PR #21, opened 2018-08-21
adapt HugeGraph to tinkerpop tests for mysql backend

### LINKED ISSUE #20: adapt to tinkerpop tests for mysql backend
1. multi-source vertex-step problem also exists, filter it.
2. mysql not supporting update vertex/edge property leads to using add vertex/edge to update property, which result in bug tx having 'added vertices' when update properties. fixed.
3. multi-thread tx commit not work. Reset mode to auto-commit after batch commit to fix

zhoney:
resolve #20

Change-Id: Ic6441af1115ecaa2f3783d1a5087bb503d5ab9fa


zhoney:
This pull request modifications will be merged with #19

PR #38, opened 2018-08-28
Fixed bug that start hugegraph will tip "The operation time out..." but curl is ok

Linary:
Change-Id: I2af8cfbea9840f236f3ca2180b5716f4b27e2c67

### DIFF
--- a/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
+++ b/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
@@ -2,6 +2,7 @@
 
 OPEN_MONITOR="false"
 VERBOSE=""
+SERVER_STARTUP_TIMEOUT_S=30
 
 while getopts "m:v" arg; do
     case ${arg} in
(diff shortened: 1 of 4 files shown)


hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -32,10 +33,11 @@ PID_FILE=$BIN/pid
 
 . $BIN/util.sh
 
-SERVER_URL=`read_property "$TOP/conf/rest-server.properties" "restserver.url"`
-SERVER_STARTUP_TIMEOUT_S=30
+GREMLIN_SERVER_URL=`read_property "$TOP/conf/rest-server.properties" "gremlinserver.url"`
```
javeme: $CONF/rest-server.properties

hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -32,10 +33,11 @@ PID_FILE=$BIN/pid
...
+REST_SERVER_URL=`read_property "$TOP/conf/rest-server.properties" "restserver.url"`
```
javeme: ditto

Linary:
done

PR #105, opened 2018-10-11
improve tx-cache and index&ids query

javeme:
* ensure return duplicate records when querying by ids
 * tx cache override queryVerticesFromBackend() method

Change-Id: Ib0116f2faf0d0da9ec40ea335c87950e1c6e7276

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/CachedGraphTransaction.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/CachedGraphTransaction.java
@@ -20,27 +20,24 @@
 package com.baidu.hugegraph.backend.cache;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
-
-import org.apache.tinkerpop.gremlin.structure.Edge;
-import org.apache.tinkerpop.gremlin.structure.Vertex;
 
 import com.baidu.hugegraph.HugeGraph;
 import com.baidu.hugegraph.backend.BackendException;
 import com.baidu.hugegraph.backend.cache.CachedBackendStore.QueryId;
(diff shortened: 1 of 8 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/GraphIndexTransaction.java
```diff
@@ -491,6 +490,25 @@ private boolean needIndexForLabel() {
         return !this.store().features().supportsQueryByLabel();
     }
 
+    @Watched(prefix = "index")
+    private Collection<Id> intersectIndexQueries(IndexQueries queries) {
+        Collection<Id> intersectIds = null;
+
+        for (Map.Entry<IndexLabel, ConditionQuery> entry: queries.entrySet()) {
```
Linary: add a space between 'entry' and ':'

hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -72,46 +69,52 @@ private Cache cache(String prefix, int capacity, long expire) {
...
+        List<HugeVertex> vertices = new ArrayList<>(query.ids().size());
+        for (Id vertexId : query.ids()) {
+            Object vertex = this.verticesCache.get(vertexId);
+            if (vertex != null) {
+                vertices.add((HugeVertex) vertex);
+            } else {
+                newQuery.query(vertexId);
+            }
         }
+        if (vertices.isEmpty()) {
```
Linary: seems no need to set 'newQuery = query'

javeme: yes both are ok

PR #113, opened 2018-10-15
fix: job status is not updated after running

### LINKED ISSUE #112: Job status has not been updated after running
### Steps to reproduce the problem 复现步骤
1. Schedule some(>=4) gremlin jobs which may cost a long time
2. Schedule a gremlin job, we assume the task id is `task-5` and status is `queued`
3. The status of `task-5` won't be updated even if task-5 is running, the status is only changed until it's completed.

hugegraph v0.7.4

javeme:
fix #112

Change-Id: Ic21aaeab25f74a2062206b029079d65deccbee73

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/task/HugeTaskScheduler.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/task/HugeTaskScheduler.java
@@ -142,8 +142,8 @@ public class HugeTaskScheduler {
         task.cancel(false);
     }
 
-    protected void remove(Id taskId) {
-        HugeTask<?> task = this.tasks.remove(taskId);
+    protected void remove(Id id) {
+        HugeTask<?> task = this.tasks.remove(id);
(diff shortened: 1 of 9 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/task/HugeTaskScheduler.java
```diff
@@ -211,6 +211,12 @@ public boolean close() {
     }
 
     public <V> HugeTask<V> deleteTask(Id id) {
+        HugeTask<?> task = this.tasks.get(id);
+        if (task != null) {
+            E.checkState(task.completed(),
+                         "Can't delete task '%s' in status %s",
+                         id, task.status());
```
zhoney: call remove(Id) to delete task in memory

javeme: task in memory would be deleted when task is completed

PR #141, opened 2018-10-29
add the number of pending tasks to metrics

javeme:
pending tasks: tasks with status RUNNING or QUEUED.
also refactor class name about task to remove prefix 'Huge'.

Change-Id: Id21e9d91a0b83b807189fffbee0247219322f234


javeme:
rebased

Issue #284, opened 2018-12-17
Compatible with more date formats


javeme:
### Expected behavior 期望表现
The following formats are accepted for date type:
yyyy-MM-dd
yyyy-MM-dd HH:mm:ss
yyyy-MM-dd HH:mm:ss.SSS

### Actual behavior 实际表现
{
"message": "Unparseable date: \"2016-1-1\", expect format: yyyy-MM-dd HH:mm:ss.SSS",
"Exception-Class": "java.lang.IllegalArgumentException",
"exceptions":[
"java.lang.IllegalArgumentException"
],
[...]

PR #286, opened 2018-12-18
Compatible with more date formats

Linary:
Implement #284

Change-Id: Ie1b3c3496b4427e3141b34b71a5ebb518678a258

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/util/DateUtil.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/util/DateUtil.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2017 HugeGraph Authors
+ *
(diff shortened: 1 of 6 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/util/DateUtil.java
```diff
@@ -0,0 +1,89 @@
...
+                    return parse(value, entry.getValue());
+                } catch (ParseException e) {
+                    throw new IllegalArgumentException(String.format(
+                              "%s, expect format: %s",
+                              e.getMessage(), entry.getValue()));
+                }
+            }
+        }
+        throw new IllegalArgumentException(String.format(
+                  "The valid date formats are: %s, but got '%s'",
```
javeme: Expect date formats: ...

PR #290, opened 2018-12-18
Remove the explicit config items about GremlinServer

Linary:
Change-Id: I9942a3fd085b3baac03545464ff69d401940319c


hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -34,10 +34,8 @@ PID_FILE=$BIN/pid
 
 . $BIN/util.sh
 
-GREMLIN_SERVER_URL=`read_property "$CONF/rest-server.properties" "gremlinserver.url"`
```
javeme: set default value if read_property returns none

Issue #302, opened 2018-12-25
Bug: query vertices by id with hasId() filter when using number id


zhoney:
### Expected behavior 期望表现
可以查询出结果

### Actual behavior 实际表现
结果为空

### Steps to reproduce the problem 复现步骤
g.V(1, 2, 3).hasId(P.within(2, 3))

### Specifications of environment 环境信息
- hugegraph version: 0.8

PR #303, opened 2018-12-25
fix bug: query vertices by ids with hasId() filter when using number id

zhoney:
Also improve exception message for query by labels
where one of labels is not exist

fixed: #302

Change-Id: I71f299061002b680a1b9d76f2bf4b34ed0f9f3cc

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/id/IdGenerator.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/id/IdGenerator.java
@@ -184,10 +184,10 @@ public abstract class IdGenerator {
 
         @Override
         public boolean equals(Object other) {
-            if (!(other instanceof LongId)) {
+            if (!(other instanceof Number)) {
                 return false;
             }
-            return this.id == ((LongId) other).id;
+            return this.id == ((Number) other).longValue();
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -390,6 +383,17 @@ public static Condition convIn2Relation(HugeGraph graph,
                 case without:
                     return Condition.nin(pkey.id(), values);
             }
+
```
javeme: move line 377~385 to if(hugeKey != null)-else-stmt

hugegraph-core/src/main/java/com/baidu/hugegraph/backend/id/IdGenerator.java
```diff
@@ -184,10 +184,10 @@ public int hashCode() {
 
         @Override
         public boolean equals(Object other) {
-            if (!(other instanceof LongId)) {
+            if (!(other instanceof Number)) {
                 return false;
             }
-            return this.id == ((LongId) other).id;
+            return this.id == ((Number) other).longValue();
```
javeme: add test

PR #313, opened 2019-01-03
Limit the name of the gremlin request less than 256 bytes

Linary:
Change-Id: I46ccdffd92087b50676e1612226424208f66dbee

### DIFF
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/job/GremlinAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/job/GremlinAPI.java
@@ -19,6 +19,11 @@
 
 package com.baidu.hugegraph.api.job;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CodingErrorAction;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


hugegraph-api/src/main/java/com/baidu/hugegraph/api/job/GremlinAPI.java
```diff
@@ -148,7 +152,21 @@ public String gremlin() {
 
         public String name() {
             // Get the first line of script as the name
-            return this.gremlin.split("\r\n|\r|\n", 2)[0];
+            String firstLine = this.gremlin.split("\r\n|\r|\n", 2)[0];
+            try {
+                byte[] srcBytes = firstLine.getBytes(CHARSET);
+                if (srcBytes.length > MAX_NAME_LENGTH) {
+                    byte[] dstBytes = new byte[MAX_NAME_LENGTH];
+                    System.arraycopy(srcBytes, 0, dstBytes, 0, MAX_NAME_LENGTH);
```
javeme: may cause encoding exception if the end byte is truncated
see: [decoder.onMalformedInput(CodingErrorAction.IGNORE);](https://stackoverflow.com/questions/3576754/truncating-strings-by-bytes)

hugegraph-api/src/main/java/com/baidu/hugegraph/api/job/GremlinAPI.java
```diff
@@ -148,7 +157,30 @@ public String gremlin() {
...
+                    decoder.reset();
+
+                    ByteBuffer buffer = ByteBuffer.wrap(bytes, 0,
+                                                        MAX_NAME_LENGTH);
+                    CharBuffer decoded = decoder.decode(buffer);
+                    return decoded.toString();
+                } else {
+                    return firstLine;
+                }
+            } catch (CharacterCodingException e) {
```
javeme: just catch line 175

PR #414, opened 2019-03-28
Throw exception if exceeds 80w without paging

zhoney:
resolved: #360

Change-Id: I46c7a61df21d6902aba6c0c965e8f63aa3b76fe2

PR #427, opened 2019-04-02
Revert "Support paging for scan api"

javeme:
Reverts hugegraph/hugegraph#361


javeme:
The API is missing some parameter: #414

PR #476, opened 2019-04-22
Reject request when there is too little free memory

Linary:
Change-Id: Ibac88733530b55d267da88e08055d71bce9b07d7

### DIFF
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/filter/LoadDetectFilter.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/filter/LoadDetectFilter.java
@@ -48,13 +48,16 @@ public class LoadDetectFilter implements ContainerRequestFilter {
     public void filter(ContainerRequestContext context) throws IOException {
         HugeConfig config = this.configProvider.get();
         long minFreeMemory = config.get(ServerOptions.MIN_FREE_MEMORY);
-        long available = Runtime.getRuntime().freeMemory() / Bytes.MB;
-        if (available < minFreeMemory) {
+        long allocatedMem = Runtime.getRuntime().totalMemory() -
+                            Runtime.getRuntime().freeMemory();
+        long presumableFreeMem = (Runtime.getRuntime().maxMemory() -
+                                  allocatedMem) / Bytes.MB;
+        if (presumableFreeMem < minFreeMemory) {
             throw new ServiceUnavailableException(String.format(
                       "The server available memory %s(MB) is below than " +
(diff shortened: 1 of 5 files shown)


hugegraph-api/src/main/java/com/baidu/hugegraph/api/filter/LoadDetectFilter.java
```diff
@@ -46,13 +47,25 @@
     @Override
     public void filter(ContainerRequestContext context) throws IOException {
         HugeConfig config = this.configProvider.get();
+        long minFreeMemory = config.get(ServerOptions.MIN_FREE_MEMORY);
+        long available = Runtime.getRuntime().freeMemory() / Bytes.MB;
```
javeme: please refer to https://stackoverflow.com/a/18366283/11297548
```
long presumableFreeMemory = Runtime.getRuntime().maxMemory() - allocatedMemory;
// where
long allocatedMemory = (Runtime.getRuntime().totalMemory()-Runtime.getRuntime().freeMemory());
```

hugegraph-api/src/main/java/com/baidu/hugegraph/config/ServerOptions.java
```diff
@@ -51,9 +51,20 @@ public static synchronized ServerOptions instance() {
...
+            );
+
+    public static final ConfigOption<Long> MIN_FREE_MEMORY =
+            new ConfigOption<>(
+                    "restserver.min_free_memory",
+                    "The minmium free memory(MB) of rest server, requests " +
+                    "will be rejected when the available memory of system " +
+                    "is lower than this value.",
+                    positiveInt(),
+                    128L
```
javeme: is it a suitable default value?

Linary: Choosed 512M, which is also the minimum memory required to start the service.

hugegraph-api/src/main/java/com/baidu/hugegraph/config/ServerOptions.java
```diff
@@ -63,7 +63,7 @@ public static synchronized ServerOptions instance() {
                     "will be rejected when the available memory of system " +
                     "is lower than this value.",
                     positiveInt(),
-                    128L
+                    512L
```
javeme: if the program is allocated with 512M memory, in general, the program itself occupies part of the memory, so it may not be unable to serve.

Linary: seems 256M is better

PR #488, opened 2019-04-28
fix: edges cache is not cleared when clearing backend

javeme:
Change-Id: I8208caa89b7b4e2985798676b4c607180c85d38e


hugegraph-test/src/main/java/com/baidu/hugegraph/unit/FakeObjects.java
```diff
@@ -34,12 +39,25 @@
...
-public final class FakeObject {
+public final class FakeObjects {
 
     private final HugeGraph graph;
 
-    public FakeObject() {
+    public FakeObjects() {
         this.graph = Mockito.mock(HugeGraph.class);
+        Mockito.doReturn(newConfig()).when(this.graph).configuration();
+        ;
```
Linary: delete

PR #493, opened 2019-04-30
Support batch updating elements' property by multiple strategy

imbajin:
1. Hard code to get elementId now
2. Edge's Direction is always OUT now
3. Other TODOs need finish later (add more unit-test also)

### DIFF
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/API.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/API.java
@@ -155,6 +155,16 @@ public class API {
         }
     }
 
+    protected static void checkUpdatingBody(
+                          Collection<? extends Checkable> bodys) {
+        E.checkArgumentNotNull(bodys, "The request body can't be empty");
+        for (Checkable body : bodys) {
+            E.checkArgumentNotNull(body,
+                                   "The batch body can't contain null record");
+            body.checkUpdate();
+        }
+    }
+
     @SuppressWarnings("unchecked")
     protected static Map<String, Object> parseProperties(String properties) {
         if (properties == null || properties.isEmpty()) {
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/API.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/API.java
@@ -149,8 +149,8 @@ public class API {
                           Collection<? extends Checkable> bodys) {
         E.checkArgumentNotNull(bodys, "The request body can't be empty");
         for (Checkable body : bodys) {
-            E.checkArgumentNotNull(body,
-                                   "The batch body can't contain null record");
+            E.checkArgument(body != null,
+                            "The batch body can't contain null record");
             body.checkCreate(true);
         }
     }
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -100,9 +100,9 @@ public class BatchAPI extends API {
 
     protected void updateExistElement(JsonElement oldElement,
                                       JsonElement newElement,
-                                      Map<String, UpdateStrategy> strategy) {
+                                      Map<String, UpdateStrategy> strategies) {
         if (oldElement != null && newElement != null) {
-            for (Map.Entry<String, UpdateStrategy> kv : strategy.entrySet()) {
+            for (Map.Entry<String, UpdateStrategy> kv : strategies.entrySet()) {
                 String key = kv.getKey();
                 UpdateStrategy updateStrategy = kv.getValue();
                 if (oldElement.properties.get(key) != null) {
@@ -118,13 +118,14 @@ public class BatchAPI extends API {
     protected void updateExistElement(HugeGraph g,
                                       Element oldElement,
                                       JsonElement newElement,
-                                      Map<String, UpdateStrategy> strategy) {
+                                      Map<String, UpdateStrategy> strategies) {
         E.checkArgument(oldElement != null && newElement != null,
                         "Both of the elements cannot be null");
-        for (Map.Entry<String, UpdateStrategy> kv : strategy.entrySet()) {
+        for (Map.Entry<String, UpdateStrategy> kv : strategies.entrySet()) {
             String key = kv.getKey();
             UpdateStrategy updateStrategy = kv.getValue();
-            if (oldElement.property(key).isPresent()) {
+            if (oldElement.property(key).isPresent() &&
+                newElement.properties.get(key) != null ) {
                 Object value = updateStrategy.checkAndUpdateProperty(
                                oldElement.property(key).value(),
                                newElement.properties.get(key));
@@ -134,9 +135,9 @@ public class BatchAPI extends API {
         }
     }
 
-    protected static void updateProperties(JsonElement jsonElement,
-                                           boolean append,
-                                           HugeElement element) {
+    protected static void updateProperties(HugeElement element,
+                                           JsonElement jsonElement,
+                                           boolean append) {
         for (Map.Entry<String, Object> e : jsonElement.properties.entrySet()) {
             String key = e.getKey();
             Object value = e.getValue();
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -30,6 +30,7 @@ import com.baidu.hugegraph.HugeException;
 import com.baidu.hugegraph.HugeGraph;
 import com.baidu.hugegraph.api.API;
 import com.baidu.hugegraph.api.schema.Checkable;
+import com.baidu.hugegraph.backend.id.Id;
 import com.baidu.hugegraph.config.HugeConfig;
 import com.baidu.hugegraph.config.ServerOptions;
 import com.baidu.hugegraph.metrics.MetricsUtil;
@@ -82,6 +83,8 @@ public class BatchAPI extends API {
     @JsonIgnoreProperties(value = {"type"})
     protected static abstract class JsonElement implements Checkable {
 
+        @JsonProperty("id")
+        public Id id;
         @JsonProperty("label")
         public String label;
         @JsonProperty("properties")
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -106,7 +106,7 @@ public class BatchAPI extends API {
         if (oldElement == null) {
             return;
         }
-        E.checkArgument(newElement != null, "The json element cannot be null");
+        E.checkArgument(newElement != null, "The json element can't be null");
 
         for (Map.Entry<String, UpdateStrategy> kv : strategies.entrySet()) {
             String key = kv.getKey();
@@ -127,7 +127,7 @@ public class BatchAPI extends API {
         if (oldElement == null) {
             return;
         }
-        E.checkArgument(newElement != null, "The json element cannot be null");
+        E.checkArgument(newElement != null, "The json element can't be null");
 
         for (Map.Entry<String, UpdateStrategy> kv : strategies.entrySet()) {
             String key = kv.getKey();
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -104,16 +104,19 @@ public class BatchAPI extends API {
     protected void updateExistElement(JsonElement oldElement,
                                       JsonElement newElement,
                                       Map<String, UpdateStrategy> strategies) {
-        if (oldElement != null && newElement != null) {
-            for (Map.Entry<String, UpdateStrategy> kv : strategies.entrySet()) {
-                String key = kv.getKey();
-                UpdateStrategy updateStrategy = kv.getValue();
-                if (oldElement.properties.get(key) != null) {
-                    Object value = updateStrategy.checkAndUpdateProperty(
-                                   oldElement.properties.get(key),
-                                   newElement.properties.get(key));
-                    newElement.properties.put(key, value);
-                }
+        if (oldElement == null) {
+            return;
+        }
+        E.checkArgument(newElement != null, "The json element cannot be null");
+
+        for (Map.Entry<String, UpdateStrategy> kv : strategies.entrySet()) {
+            String key = kv.getKey();
+            UpdateStrategy updateStrategy = kv.getValue();
+            if (oldElement.properties.get(key) != null) {
+                Object value = updateStrategy.checkAndUpdateProperty(
+                               oldElement.properties.get(key),
+                               newElement.properties.get(key));
+                newElement.properties.put(key, value);
             }
         }
     }
@@ -122,8 +125,11 @@ public class BatchAPI extends API {
                                       Element oldElement,
                                       JsonElement newElement,
                                       Map<String, UpdateStrategy> strategies) {
-        E.checkArgument(oldElement != null && newElement != null,
-                        "Both of the elements cannot be null");
+        if (oldElement == null) {
+            return;
+        }
+        E.checkArgument(newElement != null, "The json element cannot be null");
+
         for (Map.Entry<String, UpdateStrategy> kv : strategies.entrySet()) {
             String key = kv.getKey();
             UpdateStrategy updateStrategy = kv.getValue();
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -35,6 +35,7 @@ import com.baidu.hugegraph.config.ServerOptions;
 import com.baidu.hugegraph.metrics.MetricsUtil;
 import com.baidu.hugegraph.server.RestServer;
 import com.baidu.hugegraph.structure.HugeElement;
+import com.baidu.hugegraph.util.E;
 import com.baidu.hugegraph.util.Log;
 import com.codahale.metrics.Meter;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
@@ -116,21 +117,21 @@ public class BatchAPI extends API {
     }
 
     // TODO: Combine multi update methond into one (Element & JsonElement)
-    protected void updateExistElement(Element oldElement,
+    protected void updateExistElement(HugeGraph graph,
+                                      Element oldElement,
                                       JsonElement newElement,
-                                      Map<String, UpdateStrategy> strategy,
-                                      HugeGraph g) {
-        if (oldElement != null && newElement != null) {
-            for (Map.Entry<String, UpdateStrategy> kv : strategy.entrySet()) {
-                String key = kv.getKey();
-                UpdateStrategy updateStrategy = kv.getValue();
-                if (oldElement.property(key).isPresent()) {
-                    Object value = updateStrategy.checkAndUpdateProperty(
-                                   oldElement.property(key).value(),
-                                   newElement.properties.get(key));
-                    newElement.properties.put(key, g.propertyKey(key).
-                                                   convValue(value,false));
-                }
+                                      Map<String, UpdateStrategy> strategy) {
+        E.checkArgument(oldElement != null && newElement != null,
+                        "Both of the elements cannot be null");
+        for (Map.Entry<String, UpdateStrategy> kv : strategy.entrySet()) {
+            String key = kv.getKey();
+            UpdateStrategy updateStrategy = kv.getValue();
+            if (oldElement.property(key).isPresent()) {
+                Object value = updateStrategy.checkAndUpdateProperty(
+                               oldElement.property(key).value(),
+                               newElement.properties.get(key));
+                value = graph.propertyKey(key).convValue(value, false);
+                newElement.properties.put(key, value);
             }
         }
     }
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -79,7 +79,6 @@ public class BatchAPI extends API {
         }
     }
 
-    // TODO: Design more flexible for Element?
     @JsonIgnoreProperties(value = {"type"})
     protected static abstract class JsonElement implements Checkable {
 
@@ -116,8 +115,7 @@ public class BatchAPI extends API {
         }
     }
 
-    // TODO: Combine multi update methond into one (Element & JsonElement)
-    protected void updateExistElement(HugeGraph graph,
+    protected void updateExistElement(HugeGraph g,
                                       Element oldElement,
                                       JsonElement newElement,
                                       Map<String, UpdateStrategy> strategy) {
@@ -130,7 +128,7 @@ public class BatchAPI extends API {
                 Object value = updateStrategy.checkAndUpdateProperty(
                                oldElement.property(key).value(),
                                newElement.properties.get(key));
-                value = graph.propertyKey(key).convValue(value, false);
+                value = g.propertyKey(key).convValue(value, false);
                 newElement.properties.put(key, value);
             }
         }
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -30,7 +30,6 @@ import com.baidu.hugegraph.HugeException;
 import com.baidu.hugegraph.HugeGraph;
 import com.baidu.hugegraph.api.API;
 import com.baidu.hugegraph.api.schema.Checkable;
-import com.baidu.hugegraph.backend.id.Id;
 import com.baidu.hugegraph.config.HugeConfig;
 import com.baidu.hugegraph.config.ServerOptions;
 import com.baidu.hugegraph.metrics.MetricsUtil;
@@ -84,7 +83,7 @@ public class BatchAPI extends API {
     protected static abstract class JsonElement implements Checkable {
 
         @JsonProperty("id")
-        public Id id;
+        public Object id;
         @JsonProperty("label")
         public String label;
         @JsonProperty("properties")
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -19,20 +19,26 @@
 
 package com.baidu.hugegraph.api.graph;
 
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.tinkerpop.gremlin.structure.Element;
 import org.slf4j.Logger;
 
 import com.baidu.hugegraph.HugeException;
 import com.baidu.hugegraph.HugeGraph;
 import com.baidu.hugegraph.api.API;
+import com.baidu.hugegraph.api.schema.Checkable;
 import com.baidu.hugegraph.config.HugeConfig;
 import com.baidu.hugegraph.config.ServerOptions;
 import com.baidu.hugegraph.metrics.MetricsUtil;
 import com.baidu.hugegraph.server.RestServer;
+import com.baidu.hugegraph.structure.HugeElement;
 import com.baidu.hugegraph.util.Log;
 import com.codahale.metrics.Meter;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class BatchAPI extends API {
 
@@ -69,6 +75,77 @@ public class BatchAPI extends API {
             return result;
         } finally {
             batchWriteThreads.decrementAndGet();
+        }
+    }
+
+    // TODO: Design more flexible for Element?
+    @JsonIgnoreProperties(value = {"type"})
+    protected static abstract class JsonElement implements Checkable {
+
+        @JsonProperty("label")
+        public String label;
+        @JsonProperty("properties")
+        public Map<String, Object> properties;
+        @JsonProperty("type")
+        public String type;
+
+        @Override
+        public abstract void checkCreate(boolean isBatch);
+
+        @Override
+        public abstract void checkUpdate();
+
+        protected abstract Object[] properties();
+    }
+
+    protected void updateExistElement(JsonElement oldElement,
+                                      JsonElement newElement,
+                                      Map<String, UpdateStrategy> strategy) {
+        if (oldElement != null && newElement != null) {
+            for (Map.Entry<String, UpdateStrategy> kv : strategy.entrySet()) {
+                String key = kv.getKey();
+                UpdateStrategy updateStrategy = kv.getValue();
+                if (oldElement.properties.get(key) != null) {
+                    Object value = updateStrategy.checkAndUpdateProperty(
+                                   oldElement.properties.get(key),
+                                   newElement.properties.get(key));
+                    newElement.properties.put(key, value);
+                }
+            }
+        }
+    }
+
+    // TODO: Combine multi update methond into one (Element & JsonElement)
+    protected void updateExistElement(Element oldElement,
+                                      JsonElement newElement,
+                                      Map<String, UpdateStrategy> strategy,
+                                      HugeGraph g) {
+        if (oldElement != null && newElement != null) {
+            for (Map.Entry<String, UpdateStrategy> kv : strategy.entrySet()) {
+                String key = kv.getKey();
+                UpdateStrategy updateStrategy = kv.getValue();
+                if (oldElement.property(key).isPresent()) {
+                    Object value = updateStrategy.checkAndUpdateProperty(
+                                   oldElement.property(key).value(),
+                                   newElement.properties.get(key));
+                    newElement.properties.put(key, g.propertyKey(key).
+                                                   convValue(value,false));
+                }
+            }
+        }
+    }
+
+    protected static void updateProperties(JsonElement jsonElement,
+                                           boolean append,
+                                           HugeElement element) {
+        for (Map.Entry<String, Object> e : jsonElement.properties.entrySet()) {
+            String key = e.getKey();
+            Object value = e.getValue();
+            if (append) {
+                element.property(key, value);
+            } else {
+                element.property(key).remove();
+            }
         }
     }
 }
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
@@ -125,7 +125,7 @@ public class BatchAPI extends API {
             String key = kv.getKey();
             UpdateStrategy updateStrategy = kv.getValue();
             if (oldElement.property(key).isPresent() &&
-                newElement.properties.get(key) != null ) {
+                newElement.properties.containsKey(key)) {
                 Object value = updateStrategy.checkAndUpdateProperty(
                                oldElement.property(key).value(),
                                newElement.properties.get(key));
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -419,7 +419,7 @@ public class EdgeAPI extends BatchAPI {
                 Object sortKeyValue = newEdge.properties.get(sortKey);
                 E.checkArgument(sortKeyValue != null,
                                 "The value of sort key '%s' can't be null",
-                                skId);
+                                sortKey);
                 sortKeyValues.add(sortKeyValue);
             });
             sortKeys = SplicingIdGenerator.concatValues(sortKeyValues);
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -175,34 +175,32 @@ public class EdgeAPI extends BatchAPI {
                          EdgeRequest req) {
         LOG.debug("Graph [{}] update edges: {}", graph, req.jsonEdges);
         checkUpdatingBody(req.jsonEdges);
-
-        HugeGraph g = graph(manager, graph);
         checkBatchSize(config, req.jsonEdges);
 
+        HugeGraph g = graph(manager, graph);
         Map<Id, JsonEdge> maps = new HashMap<>(req.jsonEdges.size());
         TriFunction<HugeGraph, Object, String, Vertex> getVertex =
                     req.checkVertex ? EdgeAPI::getVertex : EdgeAPI::newVertex;
 
         return this.commit(config, g, maps.size(), () -> {
 
-             // 1.Put all newEdges' properties into map (combine first)
+            // 1.Put all newEdges' properties into map (combine first)
             req.jsonEdges.forEach(newEdge -> {
-                Id labelId = g.edgeLabel(newEdge.label).id();
-                Id id = newEdge.id != null ? EdgeId.parse(newEdge.id) :
-                        this.getEdgeId(g, labelId, newEdge);
-                this.updateExistElement(newEdge, maps.get(id),
+                Id newEdgeId = newEdge.id != null ? EdgeId.parse(newEdge.id) :
+                               getEdgeId(g, newEdge);
+                JsonEdge oldEdge = maps.get(newEdgeId);
+                this.updateExistElement(oldEdge, newEdge,
                                         req.updateStrategies);
-                maps.put(id, newEdge);
+                maps.put(newEdgeId, newEdge);
             });
 
             // 2.Get all oldEdges and update with new ones
-            List<Id> ids = new ArrayList<>(maps.size());
-            maps.keySet().forEach(key -> ids.add(key));
-            Iterator<Edge> oldEdges = g.edges(ids.toArray());
+            Object[] ids = maps.keySet().toArray();
+            Iterator<Edge> oldEdges = g.edges(ids);
             oldEdges.forEachRemaining(oldEdge -> {
-                updateExistElement(g, oldEdge,
-                                   maps.get(oldEdge.id().toString()),
-                                   req.updateStrategies);
+                JsonEdge newEdge = maps.get(oldEdge.id());
+                this.updateExistElement(g, oldEdge, newEdge,
+                                        req.updateStrategies);
             });
 
             // 3.Add all finalEdges
@@ -404,8 +402,9 @@ public class EdgeAPI extends BatchAPI {
         }
     }
 
-    private Id getEdgeId(HugeGraph g, Id labelId, JsonEdge newEdge) {
+    private Id getEdgeId(HugeGraph g, JsonEdge newEdge) {
         String sortKeys = "";
+        Id labelId = g.edgeLabel(newEdge.label).id();
         List<Id> sortKeyIds = g.edgeLabel(labelId).sortKeys();
         if (!sortKeyIds.isEmpty()) {
             List<Object> sortKeyValues = new ArrayList<>(sortKeyIds.size());
@@ -428,13 +427,13 @@ public class EdgeAPI extends BatchAPI {
 
     protected static class EdgeRequest {
 
-        @JsonProperty("jsonEdges")
+        @JsonProperty("edges")
         public List<JsonEdge> jsonEdges;
-        @JsonProperty("updateStrategies")
+        @JsonProperty("update_strategies")
         public Map<String, UpdateStrategy> updateStrategies;
         @JsonProperty("check_vertex")
         public boolean checkVertex = true;
-        @JsonProperty("create")
+        @JsonProperty("create_if_not_exist")
         public boolean createIfNotExist = true;
 
         @Override
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -402,7 +402,7 @@ public class EdgeAPI extends BatchAPI {
 
     private Id getEdgeId(HugeGraph g, JsonEdge newEdge) {
         if (newEdge.id != null) {
-            return EdgeId.parse(newEdge.id);
+            return EdgeId.parse(newEdge.id.asString());
         }
 
         String sortKeys = "";
@@ -450,8 +450,6 @@ public class EdgeAPI extends BatchAPI {
 
     private static class JsonEdge extends JsonElement {
 
-        @JsonProperty("id")
-        public String id;
         @JsonProperty("outV")
         public Object source;
         @JsonProperty("outVLabel")
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -132,9 +132,9 @@ public class EdgeAPI extends BatchAPI {
                                List<JsonEdge> jsonEdges) {
         LOG.debug("Graph [{}] create edges: {}", graph, jsonEdges);
         checkCreatingBody(jsonEdges);
+        checkBatchSize(config, jsonEdges);
 
         HugeGraph g = graph(manager, graph);
-        checkBatchSize(config, jsonEdges);
 
         TriFunction<HugeGraph, Object, String, Vertex> getVertex =
                     checkVertex ? EdgeAPI::getVertex : EdgeAPI::newVertex;
@@ -172,12 +172,10 @@ public class EdgeAPI extends BatchAPI {
                          @Context GraphManager manager,
                          @PathParam("graph") String graph,
                          BatchEdgeRequest req) {
+        BatchEdgeRequest.checkUpdate(req);
         LOG.debug("Graph [{}] update edges: {}", graph, req.jsonEdges);
         checkUpdatingBody(req.jsonEdges);
         checkBatchSize(config, req.jsonEdges);
-        // Not support createIfNotExist equals false now
-        E.checkArgument(req.createIfNotExist == false,
-                        "Param 'create_if_not_exist' is not supported now");
 
         HugeGraph g = graph(manager, graph);
         Map<Id, JsonEdge> map = new HashMap<>(req.jsonEdges.size());
@@ -436,9 +434,22 @@ public class EdgeAPI extends BatchAPI {
         @JsonProperty("update_strategies")
         public Map<String, UpdateStrategy> updateStrategies;
         @JsonProperty("check_vertex")
-        public boolean checkVertex = true;
+        public boolean checkVertex = false;
         @JsonProperty("create_if_not_exist")
         public boolean createIfNotExist = true;
+
+        private static void checkUpdate(BatchEdgeRequest req) {
+            E.checkArgumentNotNull(req, "BatchEdgeRequest cannot be null");
+            E.checkArgumentNotNull(req.jsonEdges,
+                                   "Parameter 'edges' cannot be null");
+            E.checkArgument(req.updateStrategies != null &&
+                            !req.updateStrategies.isEmpty(),
+                            "Parameter 'update_strategies' cannot be empty");
+
+            // Not support createIfNotExist equals false now
+            E.checkArgument(req.createIfNotExist == false, "Parameter " +
+                            "'create_if_not_exist' is not supported now");
+        }
 
         @Override
         public String toString() {
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -189,7 +189,7 @@ public class EdgeAPI extends BatchAPI {
             req.jsonEdges.forEach(newEdge -> {
                 Id labelId = g.edgeLabel(newEdge.label).id();
                 String id = newEdge.id != null ? newEdge.id :
-                            this.getEdgeId(g, labelId, newEdge);
+                            this.getEdgeId(g, labelId, newEdge).asString();
                 this.updateExistElement(newEdge, maps.get(id),
                                         req.updateStrategies);
                 maps.put(id, newEdge);
@@ -404,7 +404,7 @@ public class EdgeAPI extends BatchAPI {
         }
     }
 
-    private String getEdgeId(HugeGraph g, Id labelId, JsonEdge newEdge) {
+    private Id getEdgeId(HugeGraph g, Id labelId, JsonEdge newEdge) {
         // Set sortKeys "" or null?
         String sortKeys = "";
         List<Id> sortKeyIds = g.edgeLabel(labelId).sortKeys();
@@ -425,7 +425,7 @@ public class EdgeAPI extends BatchAPI {
                                    Directions.OUT,
                                    g.edgeLabel(newEdge.label).id(), sortKeys,
                                    HugeVertex.getIdValue(newEdge.target));
-        return edgeId.asString();
+        return edgeId;
     }
 
     protected static class EdgeRequest {
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -445,9 +445,8 @@ public class EdgeAPI extends BatchAPI {
             E.checkArgument(req.updateStrategies != null &&
                             !req.updateStrategies.isEmpty(),
                             "Parameter 'update_strategies' cannot be empty");
-
             // Not support createIfNotExist equals false now
-            E.checkArgument(req.createIfNotExist == false, "Parameter " +
+            E.checkArgument(req.createIfNotExist == true, "Parameter " +
                             "'create_if_not_exist' is not supported now");
         }
 
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -161,7 +161,7 @@ public class EdgeAPI extends BatchAPI {
 
     /**
      * Batch update steps are same like vertices
-     **/
+     */
     @PUT
     @Timed
     @Decompress
@@ -413,8 +413,9 @@ public class EdgeAPI extends BatchAPI {
             sortKeyIds.forEach(skId -> {
                 String sortKey = g.propertyKey(skId).name();
                 Object sortKeyValue = newEdge.properties.get(sortKey);
-                E.checkState(sortKeyValue != null,
-                             "The value of sort key '%s' can't be null", skId);
+                E.checkArgument(sortKeyValue != null,
+                                "The value of sort key '%s' can't be null",
+                                skId);
                 sortKeyValues.add(sortKeyValue);
             });
             sortKeys = SplicingIdGenerator.concatValues(sortKeyValues);
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
@@ -179,7 +179,7 @@ public class EdgeAPI extends BatchAPI {
         HugeGraph g = graph(manager, graph);
         checkBatchSize(config, req.jsonEdges);
 
-        Map<String, JsonEdge> maps = new HashMap<>(req.jsonEdges.size());
+        Map<Id, JsonEdge> maps = new HashMap<>(req.jsonEdges.size());
         TriFunction<HugeGraph, Object, String, Vertex> getVertex =
                     req.checkVertex ? EdgeAPI::getVertex : EdgeAPI::newVertex;
 
@@ -188,15 +188,15 @@ public class EdgeAPI extends BatchAPI {
              // 1.Put all newEdges' properties into map (combine first)
             req.jsonEdges.forEach(newEdge -> {
                 Id labelId = g.edgeLabel(newEdge.label).id();
-                String id = newEdge.id != null ? newEdge.id :
-                            this.getEdgeId(g, labelId, newEdge).asString();
+                Id id = newEdge.id != null ? EdgeId.parse(newEdge.id) :
+                        this.getEdgeId(g, labelId, newEdge);
                 this.updateExistElement(newEdge, maps.get(id),
                                         req.updateStrategies);
                 maps.put(id, newEdge);
             });
 
             // 2.Get all oldEdges and update with new ones
-            List<String> ids = new ArrayList<>(maps.size());
+            List<Id> ids = new ArrayList<>(maps.size());
             maps.keySet().forEach(key -> ids.add(key));
             Iterator<Edge> oldEdges = g.edges(ids.toArray());
             oldEdges.forEachRemaining(oldEdge -> {
@@ -405,7 +405,6 @@ public class EdgeAPI extends BatchAPI {
     }
 
     private Id getEdgeId(HugeGraph g, Id labelId, JsonEdge newEdge) {
-        // Set sortKeys "" or null?
         String sortKeys = "";
         List<Id> sortKeyIds = g.edgeLabel(labelId).sortKeys();
         if (!sortKeyIds.isEmpty()) {
@@ -422,8 +421,7 @@ public class EdgeAPI extends BatchAPI {
 
         // TODO: How to get Direction from JsonEdge easily? or any better way?
         EdgeId edgeId = new EdgeId(HugeVertex.getIdValue(newEdge.source),
-                                   Directions.OUT,
-                                   g.edgeLabel(newEdge.label).id(), sortKeys,
+                                   Directions.OUT, labelId, sortKeys,
                                    HugeVertex.getIdValue(newEdge.target));
         return edgeId;
     }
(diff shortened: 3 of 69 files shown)


hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +82,99 @@ public BatchAPI() {
...
+
+    protected void updateExistElement(JsonElement oldElement,
+                                      JsonElement newElement,
+                                      Map<String, UpdateStrategy> strategy) {
+        if (oldElement != null && newElement != null) {
+            strategy.entrySet().forEach(kv -> {
+                String key = kv.getKey();
+                if (oldElement.properties.get(key) != null) {
+                    Object value = UpdateStrategy.checkAndUpdateProperty(
+                            oldElement.properties.get(key),
```
javeme: prefer to align with UpdateStrategy

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +82,99 @@ public BatchAPI() {
...
+                            oldElement.property(key).value(),
+                            element.properties.get(key),
+                            kv.getValue());
+                    element.properties.put(key, value);
+                }
+            });
+        }
+    }
+
+    protected void operateProperties(JsonElement jsonElement, boolean append,
```
javeme: set static  and rename to updateProperties()

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+    abstract void checkPropertyType(Object oldProperty, Object newProperty);
+
+    public static Object checkAndUpdateProperty(@NotNull Object oldProperty,
+                                                @NotNull Object newProperty,
+                                                UpdateStrategy strategy) {
+        strategy.checkPropertyType(oldProperty, newProperty);
+        return strategy.updatePropertyValue(oldProperty, newProperty);
+    }
+
+    private static String getFormat(Object oldProperty, Object newProperty,
```
javeme: prefer to rename to formatError()

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            throw new NotSupportException("UpdateStrategy : eliminate");
+        }
+    };
+
+    abstract Object updatePropertyValue(Object oldProperty, Object newProperty);
+
+    abstract void checkPropertyType(Object oldProperty, Object newProperty);
+
+    public static Object checkAndUpdateProperty(@NotNull Object oldProperty,
```
javeme: we can use member method here instead of static method

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    private static String getFormat(Object oldProperty, Object newProperty,
+                                    String className) {
+        return String.format("Property must be" + className + "type, but got " +
+                             "%s, %s", oldProperty.getClass().getSimpleName(),
+                             newProperty.getClass().getSimpleName());
+    }
+
+    public static UpdateStrategy valueOf(Object strategy) {
```
javeme: It’s more concise to let json(jersey framework) deserialize enum.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +114,82 @@ public String create(@Context GraphManager manager,
...
 
+    /**
+     * TODO: PUT method should support list in Huge-common, use POST first..
+     * TODO: Adapter for param("createIfNotExist") or delete it?
+     * batch update steps like:
+     * 1. get all newVertices' ID & combine first
+     * 2. get all oldVertices & update
+     * 3. add the final vertex together
+     **/
+    @POST
```
javeme: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +114,82 @@ public String create(@Context GraphManager manager,
...
+                         @QueryParam("updateStrategies")
+                         String updateStrategies,
+                         List<JsonVertex> jsonVertices) {
+        LOG.debug("Graph [{}] update vertices: {}", graph, jsonVertices);
+        checkUpdatingBody(jsonVertices);
+
+        HugeGraph g = graph(manager, graph);
+        checkBatchSize(config, jsonVertices);
+        // Do not support auto-generate id strategy now (check once?)
+        E.checkArgument(g.vertexLabel(jsonVertices.get(0).label).idStrategy()
```
javeme: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+import javax.validation.constraints.NotNull;
+
+import com.baidu.hugegraph.exception.NotSupportException;
+import com.baidu.hugegraph.util.E;
+import com.google.common.collect.Sets;
+
+public enum UpdateStrategy {
+
+    //Only number support accumlate
+    ACCUMULATE {
```
javeme: SUM is more readable

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+    BIGGER {
+        @Override
+        Object updatePropertyValue(Object oldProperty, Object newProperty) {
+            return getComparedDate(oldProperty, newProperty, BIGGER);
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            E.checkState((oldProperty instanceof Date ||
+                          oldProperty instanceof Long) &&
```
javeme: oldProperty instanceof Number?

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+                         (newProperty instanceof Date ||
+                          newProperty instanceof Long),
```
javeme: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+            return getComparedSet(oldProperty, newProperty, INTERSECTION);
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            UNION.checkPropertyType(oldProperty, newProperty);
+        }
+    },
+
+    // TODO: Merge from "action", how to handle List? (have same elements)
```
javeme: ELIMINATE all the same elements

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +82,99 @@ public BatchAPI() {
...
+    // TODO: Combine multi update methond into one (Element & JsonElement)
+    protected void updateExistElement(Element oldElement, JsonElement element,
+                                      Map<String, UpdateStrategy> strategy) {
+        if (oldElement != null && element != null) {
+            strategy.entrySet().forEach(kv -> {
+                String key = kv.getKey();
+                if (oldElement.property(key).isPresent()) {
+                    Object value = UpdateStrategy.checkAndUpdateProperty(
+                            oldElement.property(key).value(),
+                            element.properties.get(key),
```
javeme: let new property value be converted by property-key like this [pkey.convValue(value, false)](https://github.com/hugegraph/hugegraph/blob/master/hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/optimize/TraversalUtil.java#L569)

how to get property-key: `graph.propertyKey(name)`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
imbajin: but how to use non-static method in other class directly, like `UpdateStrategy.func()` 
unless using `UpdateStrategy.SUM.func()`? (and enum is not stable)

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
imbajin: have thought about it before, but the timestamp with milliseconds can only be Long, the unattached is also close to the upper limit of Integer, and it should only allow Integer or Long?  The range of Number seems too wide

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +82,99 @@ public BatchAPI() {
...
+
+        protected abstract Object[] properties();
+    }
+
+    /**
+     * Get vertex/edge ID from JsonVerex/JsonEdge
+     * TODO: Should support AUTOMATIC ElementID ?
+     * TODO: How to get JsonEdge's fields ?
+     */
+    protected String getElementId(HugeGraph g, String labelId,
```
Linary: seems can split into two methods directly: getVertexId(g, label, JsonVertex) and getEdgeId(g, label, JsonEdge).
can avoid the method too long and reduce the number of parameters.
If need, move getVertexId into VertexAPI, getEdgeId into EdgeAPI

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +82,99 @@ public BatchAPI() {
...
+
+        @Override
+        public abstract void checkUpdate();
+
+        protected abstract Object[] properties();
+    }
+
+    /**
+     * Get vertex/edge ID from JsonVerex/JsonEdge
+     * TODO: Should support AUTOMATIC ElementID ?
```
Linary: AUTOMATIC ElementID -> AUTOMATIC Id strategy, similar in other places

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +82,99 @@ public BatchAPI() {
...
+            // EdgeId edgeId = new EdgeId(); // need jsonEdge here
+        }
+        return null;
+    }
+
+    protected void updateExistElement(JsonElement oldElement,
+                                      JsonElement newElement,
+                                      Map<String, UpdateStrategy> strategy) {
+        if (oldElement != null && newElement != null) {
+            strategy.entrySet().forEach(kv -> {
```
Linary: This logic-critical code uses a normal for loop instead of forEach

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -157,6 +159,106 @@ public String create(@Context GraphManager manager,
...
+                         @DefaultValue("true") boolean createIfNotExist,
+                         List<JsonEdge> jsonEdges) {
+        LOG.debug("Graph [{}] update edges: {}", graph, jsonEdges);
+        checkUpdatingBody(jsonEdges);
+
+        HugeGraph g = graph(manager, graph);
+        checkBatchSize(config, jsonEdges);
+
+        Map<String, JsonEdge> maps = new HashMap<>(jsonEdges.size());
+        Map<String, Object> strategies = parseProperties(updateStrategies);
```
Linary: Define a class wrap all QueryParams and jsonEdges, then deserialize UpdateStrategies will be easy. you can refer 
`CustomizedPathsAPI.PathRequest`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
javeme: we can call its method like common class:
```
UpdateStrategy strategy = UpdateStrategy.valueOf(...)
strategy.func()
```

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
javeme: Maybe we can use method compareNumber from hugegraph-common directly: `com.baidu.hugegraph.util.NumericUtil.compareNumber(Object first, Number second)` 
and call `com.baidu.hugegraph.util.NumericUtil.convertToNumber(Object value)` to convert an object to a number.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
imbajin: > 
> 
> let new property value be converted by property-key like this [pkey.convValue(value, false)](https://github.com/hugegraph/hugegraph/blob/master/hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/optimize/TraversalUtil.java#L569)
> 
> how to get property-key: `graph.propertyKey(name)`

is it like `value = g.propertyKey(key).convValue(value, false)` ? Wonder why need to convert pkValue again? or just for safety?

PS: disappear for some days, and back now~   (will fix them soon)

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +314,49 @@ private static void checkBatchSize(HugeConfig config,
...
-    @JsonIgnoreProperties(value = {"type"})
-    private static class JsonVertex implements Checkable {
+    //TODO: Should support AUTOMATIC Id strategy ?
+    private String getVertexId(HugeGraph g, String labelId, JsonVertex vertex) {
+        List<Id> pkIds = g.vertexLabel(vertex.label).primaryKeys();
+        List<Object> pkValues = new ArrayList<>(pkIds.size());
+        for (Id pkId : pkIds) {
+            String propertyKey = g.propertyKey(pkId).name();
+            Object propertyValue = vertex.properties.get(propertyKey);
+            E.checkState(propertyValue != null, "The value of primary key " +
```
javeme: wrap line after ","

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -157,6 +159,68 @@ public String create(@Context GraphManager manager,
...
+                maps.put(id, newEdge);
+            });
+
+            // 2.Get all oldEdges and update with new ones
+            List<String> ids = new ArrayList<>(maps.size());
+            maps.keySet().forEach(key -> ids.add(key));
+            Iterator<Edge> oldEdges = g.edges(ids.toArray());
+            oldEdges.forEachRemaining(oldEdge -> {
+                updateExistElement(oldEdge,
+                                   maps.get(oldEdge.id().toString()),
```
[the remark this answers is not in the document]
imbadark: first I want to use a common method called `getElementId` to get EdgeId & VertexId together, so I use `String` for both
now the methods are separated. maybe use `Map<Object, JsonVertex>` &` Map<String, JsonEdge>` is more simple because JsonVertex & JsonEdge 's  `id ` property's type is `Object` & `String`.
if use Id for them, both of them need to convert again, but seems unnecessary?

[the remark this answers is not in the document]
imbajin: first I want to use a common method called `getElementId` to get EdgeId & VertexId together, so I use String for both
now the methods are separated. maybe use `Map<Object, JsonVertex>` & `Map<String, JsonEdge> `is more simple because JsonVertex & JsonEdge 's `id` property's type is` Object `& `String`.
if use `Id` type for them, both of them need to convert again, but seems unnecessary?

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,176 @@
...
+import com.baidu.hugegraph.util.NumericUtil;
+import com.google.common.collect.Sets;
+
+public enum UpdateStrategy {
+
+    // Only number support sum
+    SUM {
+        @Override
+        Object updatePropertyValue(Object oldProperty, Object newProperty) {
+            // TODO: How to accumulae universally (Byte, Long,...Double)
```
javeme: we can use bigdecimal: https://www.tutorialspoint.com/java/math/bigdecimal_add_augend.htm

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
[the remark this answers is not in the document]
javeme: both Edge Id and Vertex Id are instance of Id, EdgeId is a subclass of id.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +314,50 @@ private static void checkBatchSize(HugeConfig config,
...
-                      "Too many counts of vertices for one time post, " +
+                      "Too many vertices for one time post, " +
                       "the maximum number is '%s'", max));
         }
     }
 
-    @JsonIgnoreProperties(value = {"type"})
-    private static class JsonVertex implements Checkable {
+    //TODO: Should support AUTOMATIC Id strategy ?
+    private Object getVertexId(HugeGraph g, String labelId, JsonVertex vertex) {
```
javeme: If let getVertexId() return Id then it can be more simple and unified in the future.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +404,63 @@ public static Direction parseDirection(String direction) {
         }
     }
 
-    @JsonIgnoreProperties(value = {"type"})
-    private static class JsonEdge implements Checkable {
+    private String getEdgeId(HugeGraph g, Id labelId, JsonEdge newEdge) {
```
javeme: same as getVertexId()

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +404,63 @@ public static Direction parseDirection(String direction) {
...
+            });
+            sortKeys = SplicingIdGenerator.concatValues(sortKeyValues);
+        }
+
+        // TODO: How to get Direction from JsonEdge easily? or any better way?
+        EdgeId edgeId = new EdgeId(HugeVertex.getIdValue(newEdge.source),
+                                   Directions.OUT,
+                                   g.edgeLabel(newEdge.label).id(), sortKeys,
+                                   HugeVertex.getIdValue(newEdge.target));
+        return edgeId.asString();
```
javeme: return edgeId itself

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -157,6 +159,68 @@ public String create(@Context GraphManager manager,
...
+        Map<String, JsonEdge> maps = new HashMap<>(req.jsonEdges.size());
+        TriFunction<HugeGraph, Object, String, Vertex> getVertex =
+                    req.checkVertex ? EdgeAPI::getVertex : EdgeAPI::newVertex;
+
+        return this.commit(config, g, maps.size(), () -> {
+
+             // 1.Put all newEdges' properties into map (combine first)
+            req.jsonEdges.forEach(newEdge -> {
+                Id labelId = g.edgeLabel(newEdge.label).id();
+                String id = newEdge.id != null ? newEdge.id :
```
javeme: Id id = newEdge.id != null ? EdgeId.parse(newEdge.id)

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +78,75 @@ public BatchAPI() {
             batchWriteThreads.decrementAndGet();
         }
     }
+
+    // TODO: Design more flexible for Element?
```
javeme: seems unused TODO

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +78,75 @@ public BatchAPI() {
...
+                    Object value = updateStrategy.checkAndUpdateProperty(
+                                   oldElement.properties.get(key),
+                                   newElement.properties.get(key));
+                    newElement.properties.put(key, value);
+                }
+            }
+        }
+    }
+
+    // TODO: Combine multi update methond into one (Element & JsonElement)
```
javeme: remove TODO since we can't combine them

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,71 @@ public String create(@Context GraphManager manager,
...
+
+        return this.commit(config, g, maps.size(), () -> {
+            /*
+            * 1.Put all newVertices' properties into map (combine first)
+            * - Consider primary-key & user-define ID mode first
+            * */
+            req.jsonVertices.forEach(newVertex -> {
+                String labelId = g.vertexLabel(newVertex.label).id().asString();
+                Object id = newVertex.id != null ? newVertex.id :
+                            this.getVertexId(g, labelId, newVertex);
```
javeme: just let `Id id = this.getVertexId(g, labelId, newVertex);`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +314,50 @@ private static void checkBatchSize(HugeConfig config,
...
+            String propertyKey = g.propertyKey(pkId).name();
+            Object propertyValue = vertex.properties.get(propertyKey);
+            E.checkState(propertyValue != null,
+                         "The value of primary key '%s' can't be null",
+                         propertyKey);
+            pkValues.add(propertyValue);
+        }
+
+        String name = SplicingIdGenerator.concatValues(pkValues);
+        return SplicingIdGenerator.splicing(labelId, name);
```
javeme:
```
Object idValue = newVertex.id != null ? newVertex.id : SplicingIdGenerator.splicing(labelId, name)
return HugeVertex.getIdValue(idValue)
```

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +404,63 @@ public static Direction parseDirection(String direction) {
...
+                             "The value of sort key '%s' can't be null", skId);
+                sortKeyValues.add(sortKeyValue);
+            });
+            sortKeys = SplicingIdGenerator.concatValues(sortKeyValues);
+        }
+
+        // TODO: How to get Direction from JsonEdge easily? or any better way?
+        EdgeId edgeId = new EdgeId(HugeVertex.getIdValue(newEdge.source),
+                                   Directions.OUT,
+                                   g.edgeLabel(newEdge.label).id(), sortKeys,
```
javeme: use labelId instead

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
javeme: just let `Id id = this.getVertexId(g, labelId, newVertex);`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +314,50 @@ private static void checkBatchSize(HugeConfig config,
...
+                      "Too many vertices for one time post, " +
                       "the maximum number is '%s'", max));
         }
     }
 
-    @JsonIgnoreProperties(value = {"type"})
-    private static class JsonVertex implements Checkable {
+    //TODO: Should support AUTOMATIC Id strategy ?
+    private Id getVertexId(HugeGraph g, String labelId, JsonVertex vertex) {
+        List<Id> pkIds = g.vertexLabel(vertex.label).primaryKeys();
```
javeme: remove TODO and assert id strategy != AUTOMATIC

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +78,75 @@ public BatchAPI() {
...
+    protected void updateExistElement(HugeGraph graph,
```
javeme: keep consistent style `HugeGraph g`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
imbajin: I was wondering if I could make it easier to merge `JsonElement` & `Element` by extends or implements a class, will del it

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,70 @@ public String create(@Context GraphManager manager,
...
+
+        Map<Id, JsonVertex> maps = new HashMap<>(req.jsonVertices.size());
+
+        return this.commit(config, g, maps.size(), () -> {
+            /*
+            * 1.Put all newVertices' properties into map (combine first)
+            * - Consider primary-key & user-define ID mode first
+            * */
+            req.jsonVertices.forEach(newVertex -> {
+                String labelId = g.vertexLabel(newVertex.label).id().asString();
```
Linary: seems doesn't need to define labelId, and getVertexId(g, newVertex) is enough

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,70 @@ public String create(@Context GraphManager manager,
...
+                         @Context GraphManager manager,
+                         @PathParam("graph") String graph,
+                         VertexRequest req) {
+        LOG.debug("Graph [{}] update vertices: {}", graph, req.jsonVertices);
+        checkUpdatingBody(req.jsonVertices);
+
+        HugeGraph g = graph(manager, graph);
+        checkBatchSize(config, req.jsonVertices);
+        // Not support automatic Id strategy now (check once?)
+        E.checkArgument(g.vertexLabel(req.jsonVertices.get(0).label).
```
Linary: Need check all vertices id strategy

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,70 @@ public String create(@Context GraphManager manager,
...
+
+        return this.commit(config, g, maps.size(), () -> {
+            /*
+            * 1.Put all newVertices' properties into map (combine first)
+            * - Consider primary-key & user-define ID mode first
+            * */
+            req.jsonVertices.forEach(newVertex -> {
+                String labelId = g.vertexLabel(newVertex.label).id().asString();
+                Id id = this.getVertexId(g, labelId, newVertex);
+                this.updateExistElement(newVertex, maps.get(id),
```
Linary: Is the parameter reversed?  I think it should be this.updateExistElement(maps.get(id), newVertex, req.updateStrategies)

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +313,53 @@ private static void checkBatchSize(HugeConfig config,
...
                       "the maximum number is '%s'", max));
         }
     }
 
-    @JsonIgnoreProperties(value = {"type"})
-    private static class JsonVertex implements Checkable {
+    private Id getVertexId(HugeGraph g, String labelId, JsonVertex vertex) {
+        assert g.vertexLabel(labelId).idStrategy() != IdStrategy.AUTOMATIC;
+
+        List<Id> pkIds = g.vertexLabel(vertex.label).primaryKeys();
```
Linary: prefer using if-else according to the id strategy, like
```
if (CUSTOMIZED) {
    xxx
} else {
    assert PRIMARY_KEY
    xxx
}
```
Clear logic is better than code clean

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,70 @@ public String create(@Context GraphManager manager,
...
+                maps.put(id, newVertex);
+            });
+
+            // 2.Get all oldVertices and update with new vertices
+            List<Id> ids = new ArrayList<>(maps.size());
+            maps.keySet().forEach(key -> ids.add(key));
+            Iterator<Vertex> oldVertices = g.vertices(ids.toArray());
+            oldVertices.forEachRemaining(oldVertex -> {
+                updateExistElement(g, oldVertex,
+                                   maps.get(oldVertex.id()),
```
Linary: seems define a local var is more readable
```
JsonVertex newVertex = maps.get(oldVertex.id());
updateExistElement(g, oldVertex, newVertex,
                   req.updateStrategies);
```

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,70 @@ public String create(@Context GraphManager manager,
...
+            req.jsonVertices.forEach(newVertex -> {
+                String labelId = g.vertexLabel(newVertex.label).id().asString();
+                Id id = this.getVertexId(g, labelId, newVertex);
+                this.updateExistElement(newVertex, maps.get(id),
+                                        req.updateStrategies);
+                maps.put(id, newVertex);
+            });
+
+            // 2.Get all oldVertices and update with new vertices
+            List<Id> ids = new ArrayList<>(maps.size());
```
Linary: Object[] ids = maps.keySet().toArray();

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +313,53 @@ private static void checkBatchSize(HugeConfig config,
...
             throw new IllegalArgumentException(String.format(
-                      "Too many counts of vertices for one time post, " +
+                      "Too many vertices for one time post, " +
                       "the maximum number is '%s'", max));
         }
     }
 
-    @JsonIgnoreProperties(value = {"type"})
-    private static class JsonVertex implements Checkable {
+    private Id getVertexId(HugeGraph g, String labelId, JsonVertex vertex) {
```
Linary: remove param labelId

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +313,53 @@ private static void checkBatchSize(HugeConfig config,
...
+
+        String name = SplicingIdGenerator.concatValues(pkValues);
+        Object idValue = vertex.id != null ? vertex.id :
+                         SplicingIdGenerator.splicing(labelId, name);
+        return HugeVertex.getIdValue(idValue);
+    }
+
+    private static class VertexRequest {
+
+        @JsonProperty("jsonVertices")
```
Linary: RESTful API params use underlined style, EdgeAPI is an exception, in order to follow the specifications of tinkerpop
@JsonProperty("jsonVertices") -> @JsonProperty("vertices")
@JsonProperty("updateStrategies") -> @JsonProperty("update_strategies")
@JsonProperty("create") -> @JsonProperty("create_if_not_exist")

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +404,61 @@ public static Direction parseDirection(String direction) {
         }
     }
 
-    @JsonIgnoreProperties(value = {"type"})
-    private static class JsonEdge implements Checkable {
+    private Id getEdgeId(HugeGraph g, Id labelId, JsonEdge newEdge) {
```
Linary: Most of the comments in VertexAPI also applied here.

javeme:
Ugh, seems CI failed, please fix https://travis-ci.org/hugegraph/hugegraph/jobs/536220267
image

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -175,34 +175,32 @@ public String update(@Context HugeConfig config,
                          EdgeRequest req) {
         LOG.debug("Graph [{}] update edges: {}", graph, req.jsonEdges);
         checkUpdatingBody(req.jsonEdges);
-
-        HugeGraph g = graph(manager, graph);
         checkBatchSize(config, req.jsonEdges);
 
+        HugeGraph g = graph(manager, graph);
```
Linary: It's a good change

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +313,53 @@ private static void checkBatchSize(HugeConfig config,
...
+        @Override
+        public String toString() {
+            return String.format("VertexRequest{jsonVertices=%s," +
+                                 "updateStrategies=%s,createIfNotExist=%s}",
+                                 this.jsonVertices, this.updateStrategies,
+                                 this.createIfNotExist);
+        }
+    }
+
+    private class JsonVertex extends JsonElement {
```
Linary: `private static class` can pass test!!!

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,179 @@
...
+        Object updatePropertyValue(Object oldProperty, Object newProperty) {
+            // TODO: Improve preformance? (like write a method in common modle)
+            BigDecimal oldNumber = new BigDecimal(oldProperty.toString());
+            BigDecimal newNumber = new BigDecimal(newProperty.toString());
+            return oldNumber.add(newNumber);
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            E.checkState(oldProperty instanceof Number &&
```
javeme: update to E.checkArgument(), same as other checks in UpdateStrategy

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,179 @@
...
+
+    SMALLER {
+        @Override
+        Object updatePropertyValue(Object oldProperty, Object newProperty) {
+            return compareNumber(oldProperty, newProperty, SMALLER);
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            BIGGER.checkPropertyType(oldProperty, newProperty);
```
javeme: it makes sense to call BIGGER.checkPropertyType

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public Object checkAndUpdateProperty(@NotNull Object oldProperty,
+                                         @NotNull Object newProperty) {
+        checkPropertyType(oldProperty, newProperty);
+        return updatePropertyValue(oldProperty, newProperty);
+    }
+
+    private static String formatError(Object oldProperty, Object newProperty,
+                                      String className) {
+        return String.format("Property must be %s type, but got %s, %s",
```
javeme: change to `Property type must be %s`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,176 @@
...
+    private static Object compareNumber(Object oldProperty, Object newProperty,
+                                        UpdateStrategy strategy) {
+        Number oldNum = NumericUtil.convertToNumber(oldProperty);
+        Number newNum = NumericUtil.convertToNumber(newProperty);
+        int result = NumericUtil.compareNumber(oldNum, newNum);
+        return strategy == BIGGER ? (result > 0 ? oldProperty : newProperty) :
+                                    (result < 0 ? oldProperty : newProperty);
+    }
+
+    private static Set combineSet(Object oldProperty, Object newProperty,
```
javeme: add template parameter like Set<?>

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,176 @@
...
+            return ((Collection) oldProperty).addAll((Collection) newProperty);
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            UNION.checkPropertyType(oldProperty, newProperty);
+        }
+    },
+
+    // TODO: Unify "action" in Vertex/EdgeAPI later
```
javeme: unused TODO?

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +307,61 @@ private static void checkBatchSize(HugeConfig config,
...
+        @Override
+        public String toString() {
+            return String.format("VertexRequest{jsonVertices=%s," +
+                                 "updateStrategies=%s,createIfNotExist=%s}",
+                                 this.jsonVertices, this.updateStrategies,
+                                 this.createIfNotExist);
+        }
+    }
+
+    private static class JsonVertex extends JsonElement {
```
javeme: override equals() & hashCode() method.
```
equals(JsonEdge other): return this.id.equals(other.id)
hashCode(): return this.id.hashCode()
```
Note: make sure that update JsonVertex.id in getVertexId() method.

The same as JsonEdge class, maybe it's better to move id field to JsonElement class.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -125,7 +125,7 @@ protected void updateExistElement(HugeGraph g,
             String key = kv.getKey();
             UpdateStrategy updateStrategy = kv.getValue();
             if (oldElement.property(key).isPresent() &&
-                newElement.properties.get(key) != null ) {
```
Linary: remember to modify these:
1. hugegraph-api pom.xml `<Implementation-Version>0.39.0.0</Implementation-Version>` to `<Implementation-Version>0.40.0.0</Implementation-Version>`
2. hugegraph-api ApiVersion.class `0.39` to `0.40`, and add an item under `[0.39] Issue-522: Add profile RESTful API`, write something like `[0.40] Issue-493: Support batch updating elements' property by multiple strategy`

hugegraph-api/src/main/java/com/baidu/hugegraph/version/ApiVersion.java
```diff
@@ -87,10 +87,12 @@
      *
      * version 0.10:
      * [0.39] Issue-522: Add profile RESTful API
+     * [0.40] Issue-493: Support batch updating elements' property by multiple
+     *        strategy
```
javeme: keep on one line

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +400,64 @@ public static Direction parseDirection(String direction) {
...
+        }
+
+        // TODO: How to get Direction from JsonEdge easily? or any better way?
+        EdgeId edgeId = new EdgeId(HugeVertex.getIdValue(newEdge.source),
+                                   Directions.OUT, labelId, sortKeys,
+                                   HugeVertex.getIdValue(newEdge.target));
+        return edgeId;
+    }
+
+    protected static class EdgeRequest {
```
javeme: BatchEdgeRequest

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +307,58 @@ private static void checkBatchSize(HugeConfig config,
...
+            String value = SplicingIdGenerator.concatValues(pkValues);
+            return SplicingIdGenerator.splicing(labelId, value);
+        } else {
+            assert idStrategy == IdStrategy.CUSTOMIZE_NUMBER ||
+                   idStrategy == IdStrategy.CUSTOMIZE_STRING;
+            return HugeVertex.getIdValue(vertex.id);
+        }
+    }
+
+    private static class VertexRequest {
```
javeme: BatchVertexRequest

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +79,76 @@ public BatchAPI() {
...
+        @Override
+        public abstract void checkUpdate();
+
+        protected abstract Object[] properties();
+    }
+
+    protected void updateExistElement(JsonElement oldElement,
+                                      JsonElement newElement,
+                                      Map<String, UpdateStrategy> strategies) {
+        if (oldElement != null && newElement != null) {
```
javeme: return if oldElement is null
and check newElement != null

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,64 @@ public String create(@Context GraphManager manager,
         });
     }
 
+    /**
+     * TODO: Adapter for param("createIfNotExist") or delete it?
```
javeme: If we really don't use this parameter, just delete it.

imbajin: @Linary any suggestion?

Linary: delete it first, consider it when inneed

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +306,58 @@ private static void checkBatchSize(HugeConfig config,
...
-        @JsonProperty("properties")
-        public Map<String, Object> properties;
-        @JsonProperty("type")
-        public String type;
+        @JsonProperty("vertices")
+        public List<JsonVertex> jsonVertices;
+        @JsonProperty("update_strategies")
+        public Map<String, UpdateStrategy> updateStrategies;
+        @JsonProperty("create_if_not_exist")
+        public boolean createIfNotExist = true;
```
javeme: we don't seem to deal with situation the user passes parameter create_if_not_exist=false

javeme: maybe we can just throw NotsupportException if create_if_not_exist=false

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,66 @@ public String create(@Context GraphManager manager,
...
+    public String update(@Context HugeConfig config,
+                         @Context GraphManager manager,
+                         @PathParam("graph") String graph,
+                         BatchVertexRequest req) {
+        LOG.debug("Graph [{}] update vertices: {}", graph, req.jsonVertices);
+        checkUpdatingBody(req.jsonVertices);
+        checkBatchSize(config, req.jsonVertices);
+        // Not support createIfNotExist equals false now
+        E.checkArgument(req.createIfNotExist == false,
+                        "Param 'create_if_not_exist' is not supported now");
```
javeme: same as BatchEdgeRequest.createIfNotExist

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -157,6 +159,66 @@ public String create(@Context GraphManager manager,
...
+    public String update(@Context HugeConfig config,
+                         @Context GraphManager manager,
+                         @PathParam("graph") String graph,
+                         BatchEdgeRequest req) {
+        LOG.debug("Graph [{}] update edges: {}", graph, req.jsonEdges);
+        checkUpdatingBody(req.jsonEdges);
+        checkBatchSize(config, req.jsonEdges);
+        // Not support createIfNotExist equals false now
+        E.checkArgument(req.createIfNotExist == false,
+                        "Param 'create_if_not_exist' is not supported now");
```
javeme: Param => Parameter
also check req != null at the first line
and check other parameters in req.
we can move all parameters check(include `req.createIfNotExist`) of BatchEdgeRequest to BatchEdgeRequest.checkUpdate() method.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
```diff
@@ -71,4 +79,82 @@ public BatchAPI() {
             batchWriteThreads.decrementAndGet();
         }
     }
+
+    @JsonIgnoreProperties(value = {"type"})
+    protected static abstract class JsonElement implements Checkable {
+
+        @JsonProperty("id")
+        public Id id;
```
javeme: use object instead, otherwise it will throw error as follows
`Can not construct instance of com.baidu.hugegraph.backend.id.Id: abstract types either need to be mapped to concrete types`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,176 @@
...
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            UNION.checkPropertyType(oldProperty, newProperty);
+        }
+    },
+
+    // Batch update Set should use union because of higher efficiency
+    APPEND {
+        @Override
```
javeme: add
`        @SuppressWarnings({ "rawtypes", "unchecked" })`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            UNION.checkPropertyType(oldProperty, newProperty);
+        }
+    },
+
+    ELIMINATE {
+        @Override
```
javeme: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    private static Set<?> combineSet(Object oldProperty, Object newProperty,
+                                     UpdateStrategy strategy) {
+        Set oldSet = oldProperty instanceof Set ?
+                     (Set) oldProperty : new HashSet((List) oldProperty);
+        Set newSet = newProperty instanceof Set ?
+                     (Set) newProperty : new HashSet((List) newProperty);
+        return strategy == UNION ? Sets.union(oldSet, newSet) :
+                                   Sets.intersection(oldSet, newSet);
```
javeme: Set is a raw type. References to generic type Set<E> should be parameterized.

please update to:
```
        Set<?> oldSet = oldProperty instanceof Set ?
                        (Set<?>) oldProperty : 
                        new HashSet<>((List<?>) oldProperty);
        Set<?> newSet = newProperty instanceof Set ?
                        (Set<?>) newProperty : 
                        new HashSet<>((List<?>) newProperty);
        return strategy == UNION ? Sets.union(oldSet, newSet) :
               Sets.intersection(oldSet, newSet);
```

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/BatchAPI.java
Linary: replace Id with Object

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -112,6 +112,7 @@ void checkPropertyType(Object oldProperty, Object newProperty) {
     },
 
     // Batch update Set should use union because of higher efficiency
+    @SuppressWarnings({ "rawtypes", "unchecked" })
```
javeme: move to line 118

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -125,6 +126,7 @@ void checkPropertyType(Object oldProperty, Object newProperty) {
         }
     },
 
+    @SuppressWarnings({ "rawtypes", "unchecked" })
```
javeme: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,64 @@ public String create(@Context GraphManager manager,
...
+        checkBatchSize(config, req.jsonVertices);
+
+        HugeGraph g = graph(manager, graph);
+        Map<Id, JsonVertex> map = new HashMap<>(req.jsonVertices.size());
+
+        return this.commit(config, g, map.size(), () -> {
+            /*
+             * 1.Put all newVertices' properties into map (combine first)
+             * - Consider primary-key & user-define ID mode first
+             * */
```
javeme: extra *

hugegraph-api/src/main/java/com/baidu/hugegraph/api/API.java
```diff
@@ -155,6 +155,16 @@ protected static void checkCreatingBody(
         }
     }
 
+    protected static void checkUpdatingBody(
+                          Collection<? extends Checkable> bodys) {
+        E.checkArgumentNotNull(bodys, "The request body can't be empty");
+        for (Checkable body : bodys) {
+            E.checkArgumentNotNull(body,
```
javeme: minor improve to shorten the length:
`E.checkArgument(body != null,`

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -157,6 +159,64 @@ public String create(@Context GraphManager manager,
         });
     }
 
+    /**
+     * Batch update steps are same like vertices
+     **/
```
javeme: extra *

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,64 @@ public String create(@Context GraphManager manager,
         });
     }
 
+    /**
+     * Batch update steps like:
+     * 1. Get all newVertices' ID & combine first
+     * 2. Get all oldVertices & update
+     * 3. Add the final vertex together
+     **/
```
javeme: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +307,70 @@ private static void checkBatchSize(HugeConfig config,
...
+        E.checkArgument(idStrategy != IdStrategy.AUTOMATIC,
+                        "Automatic Id strategy is not supported now");
+
+        if (idStrategy == IdStrategy.PRIMARY_KEY) {
+            List<Id> pkIds = vertexLabel.primaryKeys();
+            List<Object> pkValues = new ArrayList<>(pkIds.size());
+            for (Id pkId : pkIds) {
+                String propertyKey = g.propertyKey(pkId).name();
+                Object propertyValue = vertex.properties.get(propertyKey);
+                E.checkState(propertyValue != null,
```
javeme: E.checkArgument

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +400,76 @@ public static Direction parseDirection(String direction) {
...
+
+        String sortKeys = "";
+        Id labelId = g.edgeLabel(newEdge.label).id();
+        List<Id> sortKeyIds = g.edgeLabel(labelId).sortKeys();
+        if (!sortKeyIds.isEmpty()) {
+            List<Object> sortKeyValues = new ArrayList<>(sortKeyIds.size());
+            sortKeyIds.forEach(skId -> {
+                String sortKey = g.propertyKey(skId).name();
+                Object sortKeyValue = newEdge.properties.get(sortKey);
+                E.checkState(sortKeyValue != null,
```
javeme: E.checkArgument

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +307,70 @@ private static void checkBatchSize(HugeConfig config,
...
+
+        @JsonProperty("vertices")
+        public List<JsonVertex> jsonVertices;
+        @JsonProperty("update_strategies")
+        public Map<String, UpdateStrategy> updateStrategies;
+        @JsonProperty("create_if_not_exist")
+        public boolean createIfNotExist = true;
+
+        private static void checkUpdate(BatchVertexRequest req) {
+            E.checkArgumentNotNull(req, "BatchVertexRequest cannot be null");
```
javeme: prefer to replace all "cannot" to "can't"

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +307,70 @@ private static void checkBatchSize(HugeConfig config,
         int max = config.get(ServerOptions.MAX_VERTICES_PER_BATCH);
         if (vertices.size() > max) {
             throw new IllegalArgumentException(String.format(
-                      "Too many counts of vertices for one time post, " +
+                      "Too many vertices for one time post, " +
```
javeme: add check vertices.size() == 0

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+    };
+
+    abstract Object updatePropertyValue(Object oldProperty, Object newProperty);
+
+    abstract void checkPropertyType(Object oldProperty, Object newProperty);
+
+    public Object checkAndUpdateProperty(Object oldProperty,
+                                         Object newProperty) {
+        checkPropertyType(oldProperty, newProperty);
```
javeme: tiny improve: this.checkPropertyType()

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,180 @@
...
+        return updatePropertyValue(oldProperty, newProperty);
```
javeme: this.updatePropertyValue()

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -114,6 +115,64 @@ public String create(@Context GraphManager manager,
...
+    @Decompress
+    @Path("batch")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON_WITH_CHARSET)
+    public String update(@Context HugeConfig config,
+                         @Context GraphManager manager,
+                         @PathParam("graph") String graph,
+                         BatchVertexRequest req) {
+        BatchVertexRequest.checkUpdate(req);
+        LOG.debug("Graph [{}] update vertices: {}", graph, req.jsonVertices);
```
javeme: log req instead of req.jsonVertices

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -157,6 +159,64 @@ public String create(@Context GraphManager manager,
...
+    @Decompress
+    @Path("batch")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON_WITH_CHARSET)
+    public String update(@Context HugeConfig config,
+                         @Context GraphManager manager,
+                         @PathParam("graph") String graph,
+                         BatchEdgeRequest req) {
+        BatchEdgeRequest.checkUpdate(req);
+        LOG.debug("Graph [{}] update edges: {}", graph, req.jsonEdges);
```
javeme: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +400,77 @@ public static Direction parseDirection(String direction) {
...
+        private static void checkUpdate(BatchEdgeRequest req) {
+            E.checkArgumentNotNull(req, "BatchEdgeRequest cannot be null");
+            E.checkArgumentNotNull(req.jsonEdges,
+                                   "Parameter 'edges' cannot be null");
+            E.checkArgument(req.updateStrategies != null &&
+                            !req.updateStrategies.isEmpty(),
+                            "Parameter 'update_strategies' cannot be empty");
+            E.checkArgument(req.createIfNotExist == true,
+                            "Parameter 'create_if_not_exist' " +
+                            "dose not supported false now");
```
javeme: dose not support

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/VertexAPI.java
```diff
@@ -259,22 +307,70 @@ private static void checkBatchSize(HugeConfig config,
...
+        private static void checkUpdate(BatchVertexRequest req) {
+            E.checkArgumentNotNull(req, "BatchVertexRequest cannot be null");
+            E.checkArgumentNotNull(req.jsonVertices,
+                                   "Parameter 'vertices' cannot be null");
+            E.checkArgument(req.updateStrategies != null &&
+                            !req.updateStrategies.isEmpty(),
+                            "Parameter 'update_strategies' cannot be empty");
+            E.checkArgument(req.createIfNotExist == true,
+                            "Parameter 'create_if_not_exist' " +
+                            "dose not supported false now");
```
javeme: dose not support

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -350,25 +404,77 @@ public static Direction parseDirection(String direction) {
...
+        Id labelId = g.edgeLabel(newEdge.label).id();
+        List<Id> sortKeyIds = g.edgeLabel(labelId).sortKeys();
+        if (!sortKeyIds.isEmpty()) {
+            List<Object> sortKeyValues = new ArrayList<>(sortKeyIds.size());
+            sortKeyIds.forEach(skId -> {
+                String sortKey = g.propertyKey(skId).name();
+                Object sortKeyValue = newEdge.properties.get(sortKey);
+                E.checkArgument(sortKeyValue != null,
+                                "The value of sort key '%s' can't be null",
+                                skId);
```
zhoney: skId->sortKey

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,180 @@
...
+        Object updatePropertyValue(Object oldProperty, Object newProperty) {
+            return compareNumber(oldProperty, newProperty, BIGGER);
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            E.checkArgument((oldProperty instanceof Date ||
+                             oldProperty instanceof Number) &&
+                            (newProperty instanceof Date ||
+                             newProperty instanceof Number),
```
zhoney: (oldProperty instanceof Date && newProperty instanceof Date) ||
(oldProperty instanceof Number && newProperty instanceof Number)

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        @Override
+        void checkPropertyType(Object oldProperty, Object newProperty) {
+            // JsonElements are always List-type, so allows two type now.
+            E.checkArgument((oldProperty instanceof Set ||
+                             oldProperty instanceof List) &&
+                            (newProperty instanceof Set ||
+                             newProperty instanceof List),
+                            formatError(oldProperty, newProperty,
```
zhoney: (oldProperty instanceof Set && newProperty instanceof Set) ||
(oldProperty instanceof List && newProperty instanceof List)

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -0,0 +1,180 @@
...
+
+    public Object checkAndUpdateProperty(Object oldProperty,
+                                         Object newProperty) {
+        this.checkPropertyType(oldProperty, newProperty);
+        return this.updatePropertyValue(oldProperty, newProperty);
+    }
+
+    private static String formatError(Object oldProperty, Object newProperty,
+                                      String className) {
+        return String.format("Property must be %s, but got %s, %s", className,
```
zhoney: remove static and make error message more clear, like:

Property type must be %s for strategy %s, but got %s

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
imbajin: There are at least 2 situation:
1. one is `Date` type, and the other is a `Number` type.  
2. both are `Number` type.

the `list/set` collection has the same reason.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
imbajin: `formatError()`  is referenced by a static method, if we remove static, we have to make it be non-private ?

zhoney: making it protected is ok. I think clear error message is more important.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
zhoney: ok

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
imbajin: > making it protected is ok. I think clear error message is more important.

done, has unified them with non-static methods,  and since `enum` cannot be inherited ,seems default modifier is enough.

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -94,15 +96,15 @@ void checkPropertyType(Object oldProperty, Object newProperty) {
                              oldProperty instanceof List) &&
                             (newProperty instanceof Set ||
                              newProperty instanceof List),
-                            formatError(oldProperty, newProperty,
-                                        "Set or List"));
+                            this.formatError(oldProperty, newProperty,
+                                             this, "Set or List"));
```
javeme: unnecessary to pass this

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -150,24 +152,25 @@ public Object checkAndUpdateProperty(Object oldProperty,
         return this.updatePropertyValue(oldProperty, newProperty);
     }
 
-    private static String formatError(Object oldProperty, Object newProperty,
-                                      String className) {
-        return String.format("Property must be %s, but got %s, %s", className,
+    String formatError(Object oldProperty, Object newProperty,
+                       UpdateStrategy strategy, String className) {
+        return String.format("Property type must be %s for strategy %s, " +
+                             "but got type %s, %s", className, strategy,
```
javeme: just use this instead of strategy parameter

hugegraph-api/src/main/java/com/baidu/hugegraph/api/graph/UpdateStrategy.java
```diff
@@ -150,24 +152,25 @@ public Object checkAndUpdateProperty(Object oldProperty,
...
                              oldProperty.getClass().getSimpleName(),
                              newProperty.getClass().getSimpleName());
     }
 
-    private static Object compareNumber(Object oldProperty, Object newProperty,
-                                        UpdateStrategy strategy) {
+    Object compareNumber(Object oldProperty, Object newProperty,
```
javeme: keep private static

PR #527, opened 2019-05-24
suppress error log when postgre database does not exist

javeme:
error log like this:
```
May 24, 2019 1:25:14 PM org.postgresql.Driver connect
SEVERE: Connection error:
org.postgresql.util.PSQLException: FATAL: database "hugegraph" does not exist
```

Change-Id: I48aebac67ec98f5c05b96239b497940150542dcc


hugegraph-postgresql/src/main/java/com/baidu/hugegraph/backend/store/postgresql/PostgresqlSessions.java
```diff
@@ -69,4 +70,15 @@ public void createDatabase() {
             // Ignore exception if database already exists
         }
     }
+
+    @Override
+    protected String buildCreateDatabase(String database) {
```
zhoney: call it from PostgresqlSessions.createDatabase()

PR #543, opened 2019-05-30
Fix bug chinese character show garbled in MySQL

Linary:
Change-Id: I4705b7f24cda693c36d69726b203e59ea42d8c58


hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlOptions.java
```diff
@@ -73,6 +73,14 @@ public static synchronized MysqlOptions instance() {
                     ""
             );
 
+    public static final ConfigOption<String> JDBC_CHARACTER_ENCODING =
```
javeme: hardcode is ok

PR #555, opened 2019-06-05
 add conf entry jdbc.storage_engine to change storage engine

wenzhenghu:
default innodb, could change to rocksdb, but not myisam, as hugegraph will create large index that myisam could not support ("Specified key was too long; max key length is 1000 bytes")

### DIFF
--- a/hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlSessions.java
+++ b/hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlSessions.java
@@ -135,7 +135,7 @@ public class MysqlSessions extends BackendSessionPool {
 
     @Override
     protected Session newSession() {
-        return new Session(this.config);
+        return new Session();
     }
 
     public void checkSessionConnected() {
@@ -229,20 +229,7 @@ public class MysqlSessions extends BackendSessionPool {
             this.statements = new HashMap<>();
             this.opened = false;
             this.count = 0;
-            this.config = null;
-            try {
-                this.open();
-            } catch (SQLException ignored) {
-                // Ignore
-            }
-        }
-
-        public Session(HugeConfig config) {
-            this.conn = null;
-            this.statements = new HashMap<>();
-            this.opened = false;
-            this.count = 0;
-            this.config = config;
+            this.config = MysqlSessions.this.config();
             try {
                 this.open();
             } catch (SQLException ignored) {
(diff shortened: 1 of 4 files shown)


hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlSessions.java
```diff
@@ -222,19 +222,38 @@ protected Connection openWithoutDB(int timeout) {
...
+            this.config = config;
+            try {
+                this.open();
+            } catch (SQLException ignored) {
+                // Ignore
+            }
+        }
+
+        public HugeConfig config() {
+            return this.config;
```
javeme: just call MysqlSessions.this.config() and remove Session(HugeConfig config)

hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlTable.java
```diff
@@ -116,6 +116,11 @@ protected void createTable(Session session, TableDefine tableDefine) {
         }
     }
 
+    protected String engine(Session session) {
```
javeme: also update signature of [PostgresqlTable.engine()](https://github.com/hugegraph/hugegraph/blob/master/hugegraph-postgresql/src/main/java/com/baidu/hugegraph/backend/store/postgresql/PostgresqlTable.java#L49)

hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlSessions.java
```diff
@@ -222,19 +222,25 @@ protected Connection openWithoutDB(int timeout) {
...
         private boolean opened;
         private int count;
+        private HugeConfig config;
 
         public Session() {
             this.conn = null;
             this.statements = new HashMap<>();
             this.opened = false;
             this.count = 0;
+            this.config = MysqlSessions.this.config();
```
javeme: unneeded to define this.config, it's ok to return `MysqlSessions.this.config()` in config()

hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlTable.java
```diff
@@ -116,6 +116,11 @@ protected void createTable(Session session, TableDefine tableDefine) {
...
+        String engine = session.config().get(MysqlOptions.STORAGE_ENGINE);
+        return " ENGINE=" + engine;
+    }
+
```
javeme: remove `protected String engine()`

hugegraph-postgresql/src/main/java/com/baidu/hugegraph/backend/store/postgresql/PostgresqlTable.java
```diff
@@ -50,6 +51,11 @@ protected String engine() {
         return Strings.EMPTY;
     }
 
+    @Override
+    protected String engine(Session session) {
+        return Strings.EMPTY;
+    }
+
```
javeme: remove `engine()` since added `engine(Session session)`

PR #575, opened 2019-06-18
update default config properties

imbajin:
1. update cache options
2. add default config for hbase
3. make mysql's SSL configuration effective


hugegraph-mysql/src/main/java/com/baidu/hugegraph/backend/store/mysql/MysqlOptions.java
```diff
@@ -95,7 +95,7 @@ public static synchronized MysqlOptions instance() {
                     "jdbc.ssl_mode",
                     "The SSL mode of connections with database.",
                     disallowEmpty(),
-                    "disable"
+                    "false"
```
javeme: also update SSL_MODE to JDBC_SSL_MODE

PR #596, opened 2019-07-01
improve exception type if failed while waiting for async task

javeme:
Change-Id: Ifab67a27bc294618bee769f8d22540ab01905111

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/SchemaTransaction.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/SchemaTransaction.java
@@ -476,10 +476,10 @@ public class SchemaTransaction extends IndexableTransaction {
                     throw (RuntimeException) cause;
                 }
                 throw new HugeException("Async task failed with error: %s",
-                                        cause.getMessage(), cause);
+                                        cause, cause.getMessage());
             } catch (Exception e) {
                 throw new HugeException("Async task failed with error: %s",
-                                        e.getMessage(), e);
+                                        e, e.getMessage());
             }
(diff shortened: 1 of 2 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/SchemaTransaction.java
```diff
@@ -469,8 +470,16 @@ private static Id asyncRun(HugeGraph graph, HugeType schemaType,
             try {
                 task.get();
                 assert task.completed();
+            } catch (ExecutionException e) {
+                Throwable cause = e.getCause();
+                if (cause instanceof RuntimeException) {
+                    throw (RuntimeException) cause;
+                }
+                throw new HugeException("Async task failed with error: %s",
+                                        cause.getMessage(), cause);
```
zhoney: 'cause' is exception arg, 'cause.getMessage()' is message arg, so change order of them.

```
         throw new HugeException("Async task failed with error: %s",
                                 cause, cause.getMessage());
```

hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/SchemaTransaction.java
```diff
@@ -469,8 +470,16 @@ private static Id asyncRun(HugeGraph graph, HugeType schemaType,
...
             } catch (Exception e) {
-                throw new HugeException("Async task failed", e);
+                throw new HugeException("Async task failed with error: %s",
+                                        e.getMessage(), e);
```
zhoney: ditto

PR #604, opened 2019-07-04
fix: the vertex from edges is missing properties

javeme:
fix: #595
Change-Id: I7cacb37f7d7592cedbdcc6f98db7cc8bbb1bf537


hugegraph-core/src/main/java/com/baidu/hugegraph/structure/HugeEdge.java
```diff
@@ -221,21 +221,28 @@ public void remove() {
...
-            return;
+            return true;
         }
 
         Iterator<Edge> edges = tx().queryEdges(this.id());
+        boolean exist = edges.hasNext();
+        if (!exist && !throwIfNotExist) {
+            return false;
+        }
         E.checkState(edges.hasNext(), "Edge '%s' does not exist", this.id);
```
Linary: edges.hasNext() -> exits

PR #675, opened 2019-08-31
fix: left index of system type can't be deleted

javeme:
Change-Id: I654b2e4c69f8aebfc7ebed3225b2dd09da1720a0

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/HugeGraph.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/HugeGraph.java
@@ -263,6 +263,17 @@ public class HugeGraph implements GremlinGraph {
         }
     }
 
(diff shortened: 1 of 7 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/HugeGraph.java
```diff
@@ -309,6 +320,16 @@ public SchemaTransaction schemaTransaction() {
         return this.tx.schemaTransaction();
     }
 
+    public GraphTransaction systemTransaction() {
+        this.checkGraphNotClosed();
+        /*
+         * NOTE: each system operation will be auto committed,
+         * Don't need to open tinkerpop tx by readWrite() and commit manually.
```
Linary: Does the comment and the code have some inconsistencies?
```
this.tx.readWrite();
```

Issue #632, opened 2019-08-09
hugegrap-server重启失败


0xqq:
重启ｈｇｓｅｒｖｅｒ 经常出现ｈｕｇｅｇｒａｐｈ未初始化．
2019-08-09 17:20:21 4200  [main] [ERROR] com.baidu.hugegraph.server.RestServer [] - The backend store of 'hugegraph' has not been initialized

javeme:
@0xqq 提示“not been initialized”表示还在初始化中，可能需要等待一段时间才能启动完成（数据量非常大的时候可能时间比较长）。请提供一下日志和环境信息（版本、后端类型、现有数据量等），一般多久之后启动成功？

light-jing:
我遇到了相同的问题

2019-10-11 15:00:57 4091  [ReadOnlyZKClient-172.16.51.12:2181@0x76cdafa3-SendThread(172.16.51.12:2181)] [INFO ] org.apache.zookeeper.ClientCnxn [] - Opening socket connection to server 172.16.51.12/172.16.51.12:2181. Will not attempt to authenticate using SASL (unknown error)
2019-10-11 15:00:57 4091  [ReadOnlyZKClient-172.16.51.12:2181@0x76cdafa3-SendThread(172.16.51.12:2181)] [INFO ] org.apache.zookeeper.ClientCnxn [] - Opening socket connection to server 172.16.51.12/172.16.51.12:2181. Will not attempt to authenticate using SASL (unknown error)
2019-10-11 15:00:57 4092  [ReadOnlyZKClient-172.16.51.12:2181@0x76cdafa3-SendThread(172.16.51.12:2181)] [INFO ] org.apache.zookeeper.ClientCnxn [] - Socket connection established to 172.16.51.12/172.16.51.12:2181, initiating session
2019-10-11 15:00:57 4092  [ReadOnlyZKClient-172.16.51.12:2181@0x76cdafa3-SendThread(172.16.51.12:2181)] [INFO ] org.apache.zookeeper.ClientCnxn [] - Socket connection established to 172.16.51.12/172.16.51.12:2181, initiating session
[...]

PR #708, opened 2019-10-12
fix: The backend store of 'xx' has not been initialized

javeme:
The backend has been initialized, but this error is still raised.

fix: #632
Change-Id: I9890559365c01bd73bc85c25e1cb058c2637fb0d

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/store/BackendStoreSystemInfo.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/store/BackendStoreSystemInfo.java
@@ -19,14 +19,12 @@
 
 package com.baidu.hugegraph.backend.store;
 
-import java.io.UncheckedIOException;
 import java.util.Map;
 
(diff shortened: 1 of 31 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/backend/store/BackendStoreSystemInfo.java
```diff
@@ -64,17 +62,14 @@ public void init() {
 
     private Map<String, Object> info() {
         SchemaTransaction schema = this.graph.schemaTransaction();
-        PropertyKey pkey = null;
-        try {
-            pkey = schema.getPropertyKey(PK_BACKEND_INFO);
-        } catch (BackendException | IllegalStateException |
-                 UncheckedIOException ignored) {
-            // pass
-        }
```
javeme: this BackendException | IllegalStateException | UncheckedIOException catch are for rocksdb, cassandra, hbase fetch without database&table

Issue #718, opened 2019-10-25
Bug: restart timeout when exists too much tasks


zhoney:
### Steps to reproduce the problem 复现步骤

1. `HugeGraphServer` has more than 80w tasks.
2. shutdown `HugeGraphServer`, success
3. start `HugeGraphServer`, timeout

log/hugegraph-server.log

```
2019-10-25 14:45:39 9562  [main] [INFO ] com.baidu.hugegraph.server.RestServer [] - Restoring incomplete tasks for graph 'hugegraphProperty'...
Oct 25, 2019 2:45:42 PM org.glassfish.grizzly.http.server.NetworkListener shutdownNow
INFO: Stopped listener bound to [yq01-sw-scylladb01.yq01.baidu.com:8367]
2019-10-25 14:45:42 13219 [main] [ERROR] com.baidu.hugegraph.dist.HugeGraphServer [] - HugeRestServer start error:
com.baidu.hugegraph.HugeException: Failed to update/query TaskStore
        at com.baidu.hugegraph.task.TaskScheduler.call(TaskScheduler.java:397) ~[hugegraph-core-0.10.4.jar:0.10.2.0]
        at com.baidu.hugegraph.task.TaskScheduler.queryTask(TaskScheduler.java:357) ~[hugegraph-core-0.10.4.jar:0.10.2.0]
        at com.baidu.hugegraph.task.TaskScheduler.queryTask(TaskScheduler.java:352) ~[hugegraph-core-0.10.4.jar:0.10.2.0]
        at com.baidu.hugegraph.task.TaskScheduler.findTask(TaskScheduler.java:270) ~[hugegraph-core-0.10.4.jar:0.10.2.0]
        at com.baidu.hugegraph.task.TaskScheduler.restoreTasks(TaskScheduler.java:143) ~[hugegraph-core-0.10.4.jar:0.10.2.0]
        at com.baidu.hugegraph.core.GraphManager.restoreUncompletedTasks(GraphManager.java:219) ~[hugegraph-api-0.10.4.jar:0.48.0.0]
        at com.baidu.hugegraph.core.GraphManager.<init>(GraphManager.java:77) ~[hugegraph-api-0.10.4.jar:0.48.0.0]
        at com.baidu.hugegraph.server.ApplicationConfig$GraphManagerFactory$1.onEvent(ApplicationConfig.java:108) ~[hugegraph-api-0.10.4.jar:0.48.0.0]
[...]

PR #720, opened 2019-10-25
support paging for tasks API

javeme:
Change-Id: Ie036d433ef74d7c6724bf869425c3fd528f08941
related: #718


javeme:
Do not upgrade the api version because all client components need to be updated, and it's compatible to add a query param.

PR #723, opened 2019-10-30
fix mistake multiple backend info as no backend info

zhoney:
Change-Id: I4f017d7763698c7ccd01eb3b98f0fe897d8448c4


hugegraph-core/src/main/java/com/baidu/hugegraph/backend/store/BackendStoreSystemInfo.java
```diff
@@ -62,7 +66,21 @@ public void init() {
...
+            String message = String.format(
+                             "Should not exist schema with same name '%s'",
+                             PK_BACKEND_INFO);
+            if (message.equals(e.getMessage())) {
+                throw new HugeException("There exists multiple backend info " +
+                                        "of graph '%s' in backend '%s'",
+                                        this.graph.name(),
+                                        this.graph.backend());
+            }
+            // ignore
```
javeme: expect throw e here, since we assume the backend is inited before calling info()

PR #761, opened 2019-11-26
Fix start.hugegraph.sh timeout when config authentication

Linary:
Change-Id: Ic22531cead13f3698791070a044d66eacb1c26ee

### DIFF
--- a/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
+++ b/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
@@ -54,6 +54,38 @@ REST_SERVER_URL=`read_property "$CONF/rest-server.properties" "restserver.url"`
 check_port "$GREMLIN_SERVER_URL"
 check_port "$REST_SERVER_URL"
 
+
+USERNAME="admin"
+PASSWORD=`read_property "$CONF/rest-server.properties" "auth.admin_token"`
+if [[ -z "${PASSWORD}" ]]; then
+    USER_TOKENS=`read_property "$CONF/rest-server.properties" "auth.user_tokens"`
+    if [[ -z "${USER_TOKENS}" ]]; then
+        # no username and password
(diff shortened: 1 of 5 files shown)


hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -54,6 +54,38 @@ REST_SERVER_URL=`read_property "$CONF/rest-server.properties" "restserver.url"`
...
+        if [[ -z "${USERNAME}" || -z "${PASSWORD}" ]]; then
+            echo "Invalid config for auth.user_tokens"
+            exit 1
+        fi
+        # trim space
+        USERNAME=`echo ${USERNAME} | xargs`
+        PASSWORD=`echo ${PASSWORD} | xargs`
+    fi
+fi
+
```
zhoney: prefer to move it into util.sh as a function

hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -54,6 +54,38 @@ REST_SERVER_URL=`read_property "$CONF/rest-server.properties" "restserver.url"`
 check_port "$GREMLIN_SERVER_URL"
 check_port "$REST_SERVER_URL"
 
+
```
zhoney: delete one empty line

javeme:
If username and password is stored in db, we can't get it from shell. Prefer to add a specific user or a specific api.

hugegraph-dist/src/assembly/static/conf/hugegraph.properties
```diff
@@ -21,6 +21,9 @@ serializer=binary
 
 store=hugegraph
 
+search.text_analyzer=jieba
+search.text_analyzer_mode=INDEX
```
javeme: also update example and test

PR #771, opened 2019-11-30
fix TableNotFoundException when truncate HBase store

zhoney:
fixed: #770

Change-Id: Ibb74acc04018378a11b2d3304ea9e1437fc5879b

### DIFF
--- a/hugegraph-hbase/src/main/java/com/baidu/hugegraph/backend/store/hbase/HbaseStore.java
+++ b/hugegraph-hbase/src/main/java/com/baidu/hugegraph/backend/store/hbase/HbaseStore.java
@@ -24,6 +24,10 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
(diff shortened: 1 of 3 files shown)


hugegraph-hbase/src/main/java/com/baidu/hugegraph/backend/store/hbase/HbaseStore.java
```diff
@@ -310,14 +314,25 @@ public void truncate() {
...
+        try {
+            for (String table : tables) {
+                currentTable = table;
+                futures.put(table, this.sessions.truncateTable(table));
             }
+            long timeout = this.sessions.config()
+                                        .get(HbaseOptions.TRUNCATE_TIMEOUT);
+            for (Map.Entry<String, Future<Void>> entry : futures.entrySet()) {
+                currentTable = entry.getKey();
+                entry.getValue().get(timeout, TimeUnit.SECONDS);
```
javeme: call invokeAll

javeme: the current timeout will accumulate for each table

PR #827, opened 2020-01-13
fix task-cancel error with saving failure task

javeme:
This bug is caused by failure task status has not been saved to backend store(
the root cause is result size exceeded 64k, and failed to save), so its status is
still RUNNING even the task is finished, if call cancel() then the task will be loaded 
from backend store and its scheduler is null.

fix https://github.com/hugegraph/hugegraph-tools/issues/50
Change-Id: If70a2233f73c578f38842b6b015241b5ba6b86a1

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/job/schema/RebuildIndexCallable.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/job/schema/RebuildIndexCallable.java
@@ -50,7 +50,11 @@ public class RebuildIndexCallable extends SchemaCallable {
 
     @Override
     public Object execute() {
-        this.rebuildIndex(this.schemaElement());
+        SchemaElement schema = this.schemaElement();
+        // If the schema does not exist, ignore it
+        if (schema != null) {
+            this.rebuildIndex(schema);
+        }
         return null;
     }
 
@@ -65,6 +69,7 @@ public class RebuildIndexCallable extends SchemaCallable {
                     assert indexLabel.baseType() == HugeType.EDGE_LABEL;
                     label = this.graph().edgeLabel(indexLabel.baseValue());
                 }
+                assert label != null;
                 this.rebuildIndex(label, ImmutableSet.of(indexLabel.id()));
                 break;
             case VERTEX_LABEL:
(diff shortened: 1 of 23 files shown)


hugegraph-test/src/main/java/com/baidu/hugegraph/core/TaskCoreTest.java
```diff
@@ -0,0 +1,555 @@
...
+    public void testTask() throws TimeoutException {
+        HugeGraph graph = graph();
+        TaskScheduler scheduler = graph.taskScheduler();
+
+        TaskCallable<Integer> callable =  new TaskCallable<Integer>() {
+            @Override
+            public Integer call() throws Exception {
+                Thread.sleep(SLEEP_TIME);
+                return 125;
+            }
```
Linary: add empty line

hugegraph-core/src/main/java/com/baidu/hugegraph/job/schema/RebuildIndexCallable.java
```diff
@@ -50,7 +50,11 @@ public String type() {
 
     @Override
     public Object execute() {
-        this.rebuildIndex(this.schemaElement());
+        SchemaElement schema = this.schemaElement();
```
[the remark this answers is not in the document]
javeme: Maybe the schema is removed when task running, I'm not quite sure of the exact scene.

hugegraph-test/src/main/java/com/baidu/hugegraph/core/TaskCoreTest.java
```diff
@@ -0,0 +1,556 @@
...
+        Assert.assertTrue(task.result(), task.result() == null ||
+                          task.result().endsWith("InterruptedException"));
+
+        task = scheduler.findTask(task.id());
+        Assert.assertEquals(TaskStatus.CANCELLED, task.status());
+        Assert.assertEquals("test-gremlin-job", task.name());
+        Assert.assertTrue(task.result(), task.result() == null ||
+                          task.result().endsWith("InterruptedException"));
+
+        // cancel success task
```
zhoney: // Cancel success task

PR #837, opened 2020-01-17
add option vertex.check_adjacent_vertex_exist

javeme:
when check_adjacent_vertex_exist=true: 
    throw exception: "vertex does not exist" if not exist adjacent vertex

when check_adjacent_vertex_exist=false:
    return vertex with "~undefined" label if not exist adjacent vertex

fix: #835
Change-Id: I30ce32ae358d0483888895a0017b77d55f346491

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/GraphTransaction.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/GraphTransaction.java
@@ -113,7 +113,8 @@ public class GraphTransaction extends IndexableTransaction {
 
     private LockUtil.LocksTable locksTable;
 
-    private final boolean checkVertexExist;
+    private final boolean checkCustomVertexExist;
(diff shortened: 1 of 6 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -544,20 +558,26 @@ public void removeVertex(HugeVertex vertex) {
...
-                return r;
-            }
             query.mustSortByInput(false);
             Iterator<HugeVertex> it = this.queryVerticesFromBackend(query);
             QueryResults.fillMap(it, vertices);
         }
 
         return new MapperIterator<>(ids.iterator(), id -> {
-            return vertices.get(id);
+            HugeVertex vertex = vertices.get(id);
```
javeme: TODO: deal vertex with unmatch label

Issue #840, opened 2020-01-20
[Featuer] Support build docker image from Dockerfile (Task Summary)


imbajin:
> Apache [HugeGraph](https://hub.docker.com/r/hugegraph/hugegraph) DockerHub: https://hub.docker.com/r/hugegraph/hugegraph

## Usage:
```bash
# 0. download image (latest)
docker pull hugegraph/hugegraph

# 1. start container
docker run -itd hugegraph/hugegraph  # start server in daemon mode

# (Optional configs to expose network port outside)
docker run -itd --name=graph -p 18080:8080 hugegraph/hugegraph

# 2. (Optional) enter the container & start hugegraph-hubble
docker exec -it graph bash
xxx 
```

## Feature Needs
1. Add Dockerfile for running HugeServer
2. Users can pull/run the image from dockerhub directly.

And anyone who has other suggestion for docker image can comment here

_After PR, will update the doc_ 📝 

## Update in 2023.02.22:

Due to the project will move to Apache Foundation, we'll use  `1.0.0` as the server version & bind hubble with it, reopen this issue until it meets the new need, any suggestions for `Dockerfile` are welcome.

- [ ] use [docker-slim](https://github.com/docker-slim/docker-slim) to slim the image
- [ ] keep the tags same with server (like `hugegraph/hugegraph:1.0.0`)
- [ ] use hubble instead of studio to support new features
- [ ] use a script to run server & hubble together  OR use docker-compose to manage them
- [ ] pre-load some data or graphs in container so that users can traverse the graph with one step (docker run)
- [ ] use dockerhub(autobuild), currently it's not free to use (Submit a OSS request) 
- [x] support `ARM` env (for M1 laptop, refer [doc](https://github.com/marketplace/actions/docker-setup-buildx))
- [x] consider better way (like add a option/program to run backward) to avoid modify start-script #2118
- [x] adopt v0.12 (support raft based replication)
- [x] use github actions to publish images
- [x] use jdk11 for runtime #2118
- [x] use `multi-stage` Dockerfile way to slim the image
- [x] support build from the (latest) source code  #2118

Refer:
1. [Reduce the size of container images with DockerSlim](https://developers.redhat.com/articles/2022/01/17/reduce-size-container-images-dockerslim#)
2. [One docker  container with multi process](https://www.cnblogs.com/liwenchao1995/p/16732217.html) (Not graceful:)
[...]

Issue #706, opened 2019-10-09
authenticator求助


chobits102:
配置了默认的
auth.admin_token162f7848-0b6d-4faf-b557-3a0797869c55
auth.user_tokens [hugegraph:9fd95c9c-711b-415b-b85f-d4df46ba5c31]

如何用linux的curl 和 HugeGraph-Studio 进行访问

chobits102:
who can help me . thx

javeme:
@chobits102 请参考 https://github.com/hugegraph/hugegraph/issues/125#issuecomment-432902640

chobits102:
@javeme 感谢你的回复，我已经看过了#125 (comment)
但是还是不行，这是我的配置
```
# authentication
auth.require_authentication=com.baidu.hugegraph.auth.StandardAuthenticator
admin=162f7848-0b6d-4faf-b557-3a0797869c55
user=[hugegraph:9fd95c9c-711b-415b-b85f-d4df46ba5c31]
```
```
authentication: {
  authenticator: com.baidu.hugegraph.auth.StandardAuthenticator,
  config: {
    tokens: /opt/hugegraph-0.9.2/conf/rest-server.properties
  }
}
```
`gremlin.graph=com.baidu.hugegraph.auth.HugeFactoryAuthProxy
`
请求语句：
curl -v -X GET http://167.179.83.104:8080/graphs/hugegraph -u admin:162f7848-0b6d-4faf-b557-3a0797869c55
返回信息：
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying 167.179.83.104...
* TCP_NODELAY set
* Connected to 167.179.83.104 (167.179.83.104) port 8080 (#0)
* Server auth using Basic with user 'admin'
> GET /graphs/hugegraph HTTP/1.1
> Host: 167.179.83.104:8080
> Authorization: Basic YWRtaW46MTYyZjc4NDgtMGI2ZC00ZmFmLWI1NTctM2EwNzk3ODY5YzU1
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: application/json
< Date: Thu, 10 Oct 2019 01:45:28 GMT
< Connection: close
< Content-Length: 154
< 
* Closing connection 0
{"exception":"class java.lang.IllegalStateException","message":"Missing authentication context when accessing a Graph with permission control","cause":""}

我已经尝试几天了，请帮帮忙谢谢

[...]

PR #845, opened 2020-02-01
Add Dockerfile for server & studio

### LINKED ISSUE #184: Hugegraph 能否支持在docker中运行
### Expected behavior 期望表现
{type something here...}

### Actual behavior 实际表现
{type something here...}

### Steps to reproduce the problem 复现步骤
1. {step 1}
2. {step 2}
3. {step 3}

#### Vertex/Edge summary 数据量
- loaded vertices amount: {like 10 million}
- loaded edges amount: {like 20 million}
- loaded time: {like 200s}

#### Vertex/Edge example 数据示例
{type something here...}

#### Schema(VertexLabel, EdgeLabel, IndexLabel) 元数据结构
{type something here...}

### Specifications of environment 环境信息
- hugegraph version: {like v0.7.4}
- operating system: {like centos 7.4, 32 CPUs, 64G RAM}
- hugegraph backend: {like cassandra 3.10, cluster with 20 nodes, 3 x 1TB HDD disk each node}

imbajin:
Fix #184 #840  #989 :

0. Use rocksdb as the default storage engine
1. Add hosts to speed up image build (comment by default)
2. Modify some configs to enable the server to run better in container
3. Because Dockerfile doesn't allow multiple startup commands, currently only the server will be automatically started after the container is started. If you want to start studio, enter the container to start it.

Here are some simple using commands:
```bash
# 0. download image
docker pull imbajin:hugegraph

# 1. start container
docker run -itd imbajin:hugegraph  # quick start

# (Optional configs)
docker run -itd --name=graph -p 18080:8080 -p 18088:8088 imbajin:hugegraph

# 2. (Optional) enter the container
docker exec -it graph /bin/bash
```

**TODOs**:
1. Build from the latest code instead of tar package.
2. Support [hubble](https://github.com/hugegraph/hugegraph-hubble) studio
3. Add & split studio/loader into separate images, use `docker-compose` to manage them.
4. Maybe have better way to modify the configs to satisfied the container (like add a option for running in the foreground)

After the above completed, it can provides a one-click **pure** test environment for graph


Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    lsof \
+    g++ \
+    gcc \
+    openjdk-8-jdk \ 
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* 
+
+# 2. Init HugeGraph Sever (speed up)
+RUN set -e \
+    && echo "192.30.253.112 github.com\n151.101.44.249 github.global.ssl.fastly.net" >> /etc/hosts \
```
javeme: is the ip stable?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    && mkdir -p /root/server \
+    && curl -L -S ${PKG_URL}/hugegraph/releases/download/v0.10.4/hugegraph-0.10.4.tar.gz -o /root/server.tar.gz \
+    && tar xzf /root/server.tar.gz --strip-components 1 -C /root/server \
+    && rm /root/server.tar.gz \
+    && cd /root/server/ \
+    && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties \
+    && sed -n '63p' ./bin/start-hugegraph.sh | grep "&" > /dev/null && sed -i 63{s/\&$/#/g} ./bin/start-hugegraph.sh \
+    && sed -n '74p' ./bin/start-hugegraph.sh | grep "exit" > /dev/null && sed -i 74{s/^/#/g} ./bin/start-hugegraph.sh \
+    && ./bin/init-store.sh
```
javeme: add a shell install-server.sh?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# 3. Prepare for HugeGraph Studio
+RUN set -e \
+    && echo "192.30.253.112 github.com\n151.101.44.249 github.global.ssl.fastly.net" >> /etc/hosts \
+    && mkdir -p /root/studio \
+    && curl -L -S ${PKG_URL}/hugegraph-studio/releases/download/v0.10.0/hugegraph-studio-0.10.0.tar.gz -o /root/studio.tar.gz \
```
javeme: seems no studio tar when building server

Dockerfile
imbajin: Need test it:
if there is a file named `start.sh`
```
echo "old msg"
```
1. `RUN start.sh` 
2. `RUN echo "old msg" && echo "new msg"
If we use a shell file to change the code, would dockerfile think this is a repeated step and skip the rebuild?

Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM ubuntu:xenial
```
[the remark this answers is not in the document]
imbajin: We can only write a basically similar Dockerfile for CentOS. (Generally use `Debian/Ubuntu`)
Generally, I haven't seen the official image provide two types of Dockerfile, 
Maybe we can see the opinions of other community users.

And here are some refers:
1. [Cassandra_Dockerfile](https://github.com/docker-library/cassandra/blob/e5b7ed497ff74587560682dd4c6a2e9b01f818d9/3.11/Dockerfile)
2. [Nginx_Dockerfile](https://github.com/nginxinc/docker-nginx/blob/5971de30c487356d5d2a2e1a79e02b2612f9a72f/mainline/buster/Dockerfile)
3. [Other Official Dockerfile](https://hub.docker.com/search?q=&type=image&image_filter=official)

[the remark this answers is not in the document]
javeme: OK

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+# 1. Install needed dependencies of GraphServer & RocksDB
+RUN set -x \
+    && apt-get -q update \
+    && apt-get -q install -y --no-install-recommends --no-install-suggests \
+    curl \
+    lsof \
+    g++ \
+    gcc \
+    openjdk-8-jdk \
```
javeme: merge the 11-15 lines, and align with apt-get

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    && cd /root/hugegraph-server/ \
+    && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties \
+    && sed -n '63p' ./bin/start-hugegraph.sh | grep "&" > /dev/null && sed -i 63{s/\&$/#/g} ./bin/start-hugegraph.sh \
+    && sed -n '74p' ./bin/start-hugegraph.sh | grep "exit" > /dev/null && sed -i 74{s/^/#/g} ./bin/start-hugegraph.sh \
+    && ./bin/init-store.sh
+
+# 3. Prepare for HugeGraph Studio
+ENV STUDIO_VERSION 0.10.0
+RUN set -e \
+    && echo "192.30.253.112 github.com\n151.101.44.249 github.global.ssl.fastly.net" >> /etc/hosts \
```
javeme: remove it since added when building server

imbajin: Each layer of docker image use a separate network layer.
So when leave current layer(step), other step cannot reuse it

And other ways are troublesome to solve it, like [this](https://stackoverflow.com/questions/38302867/how-to-update-etc-hosts-file-in-docker-image-during-docker-build/38920219#38920219)

javeme: Get it

Dockerfile
javeme: we may need to also trigger docker image building when studio release to resolve it

imbajin:
Also update default value for auth for #706

javeme:
> Also update default values for auth for #706

There are indeed some errors, please update default value to ConfigAuthenticator:
`auth.authenticator=com.baidu.hugegraph.auth.ConfigAuthenticator`

javeme:
@imbajin Can you split the commits not related to docker?

imbajin:
> @imbajin Can you split the commits not related to docker?

OK, maybe it's better to update all the configuration files & shells separately later

hugegraph-dist/src/assembly/static/conf/hugegraph.properties
```diff
@@ -42,7 +42,7 @@ cassandra.password=
 # hbase backend config
 #hbase.hosts=localhost
 #hbase.port=2181
-#hbase.znode_parent=/hbsae
+#hbase.znode_parent=/hbase
```
javeme, line 45: also move it to independent commit

javeme (approved):

Linary (approved):

zhoney (approved):

PR #847, opened 2020-02-03
Not allowed empty primary values

Linary:
Change-Id: I6b7502892383d08348734502d0e59f5d94128fc7


hugegraph-core/src/main/java/com/baidu/hugegraph/structure/HugeVertex.java
```diff
@@ -115,6 +115,8 @@ public String name() {
                              "Primary values must not be empty " +
                              "(has properties %s)", hasProperties());
                 this.name = SplicingIdGenerator.concatValues(propValues);
+                E.checkArgument(!this.name.isEmpty(),
+                                "The primary values can't be empty");
```
javeme: The value of primary key can't be empty

PR #866, opened 2020-02-25
fix bug of missing offset with index query

javeme:
fix: #864
Change-Id: Ie83e15f35eb606b8280f8d3e35994a688ee29ca5


javeme:
image

PR #871, opened 2020-02-27
fix leak of absolute path in gremlin

zhoney:
Change-Id: I2e066014a3993862f3bea830dd4a683d2c7b6f63

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/io/HugeGraphSONModule.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/io/HugeGraphSONModule.java
@@ -19,6 +19,7 @@
 
 package com.baidu.hugegraph.io;
 
+import java.io.File;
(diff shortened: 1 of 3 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/io/HugeGraphSONModule.java
```diff
@@ -459,4 +461,20 @@ public void serialize(Shard shard, JsonGenerator jsonGenerator,
...
+        public FileSerializer() {
+            super(File.class);
+        }
+
+        @Override
+        public void serialize(File file, JsonGenerator jsonGenerator,
+                              SerializerProvider provider)
+                              throws IOException {
+            jsonGenerator.writeStartObject();
+            jsonGenerator.writeStringField("fileName", file.getName());
```
javeme: fileName => file_name or file

PR #939, opened 2020-04-10
fix ring assert error and handle two A->B not ring

zhoney:
Change-Id: Iaff8c4f16e34eb12f0fed2b33487735f783ae0ec

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/SubGraphTraverser.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/SubGraphTraverser.java
@@ -82,15 +82,20 @@ public class SubGraphTraverser extends HugeTraverser {
     }
 
     private static boolean hasMultiEdges(List<Edge> edges, Id target) {
-        int count = 0;
+        boolean hasOutEdge = false;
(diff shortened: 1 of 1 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/SubGraphTraverser.java
```diff
@@ -82,15 +82,20 @@ private PathSet subGraphPaths(Id sourceV, Directions dir, String label,
...
+        boolean hasInEdge = false;
         for (Edge edge : edges) {
             if (((HugeEdge) edge).id().otherVertexId().equals(target)) {
-                if (++count > 1) {
+                if (((HugeEdge) edge).direction() == Directions.OUT) {
+                    hasOutEdge = true;
+                } else {
+                    hasInEdge = true;
+                }
+                if (hasOutEdge && hasInEdge) {
```
javeme, line 94: still returned 3 vertices

javeme: oh seems the graph itself exist both direction edges between tow vertices

Issue #209, opened 2018-11-21
Support for distinguishing permission of actions: read, write, delete


javeme:

javeme:
Expected by #187

javeme:
Permissions control model:
`user --belong-> group --access-> target`

As vertex label: user, group, target
As edge label: belong, access

One target can be a hugegraph instance.
One access contains one permission like VERTEX_READ. All permissions defined below:
 ```java
    // general
    ANY(0),
    SCHEMA_ANY(1),
    VERTEX_ANY(2),
    EDGE_ANY(3),

    READ(15),
    WRITE(16),
    DELETE(17),

    // global
    GREMLIN(20),
    JOB_GREMLIN(21),

    TASK_READ(31),
    TASK_WRITE(32),
    TASK_DELETE(33),

    META_READ(34),
    META_WRITE(35),

    // schema
    SCHEMA_READ(50),
    SCHEMA_WRITE(51),
    SCHEMA_DELETE(52),

    // vertex
    VERTEX_READ(100),
    VERTEX_WRITE(101),
    VERTEX_DELETE(102),

    // edge
    EDGE_READ(150),
    EDGE_WRITE(151),
    EDGE_DELETE(152),

    // path
    PATH_READ(200),

    // variables
    VAR_READ(210),
    VAR_WRITE(211),
    VAR_DELETE(212),

    // index
    INDEX_REBUILD(220);
```

TODO:
Refine permissions to limit vertex|edge read/write with schema label.

PR #956, opened 2020-04-17
fix permission overflow from return results

javeme:
implement: #209
Change-Id: Iccfbd47c213e1e13122190d1242778031e8a957c

rebase v0.11 and closed #705


javeme:
image

PR #1093, opened 2020-07-13
Print heap dump error when oom

Linary:
Change-Id: I422a062530e931c8e5db54aeeae8592c0ab4d38c


hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -41,6 +41,7 @@ function abs_path() {
 BIN=`abs_path`
 TOP="$(cd $BIN/../ && pwd)"
 CONF="$TOP/conf"
+LOG="$TOP/logs"
```
javeme, line 44: rename to LOGS

PR #1134, opened 2020-08-06
fix delete/read without READ permission throw 'no such id' 

javeme:
* fix delete/read without READ permission throw 'no such id' 
  this patch define delete/update requires read permission, an ForbiddenException
  is thrown when there is no read permission instead 'No such vertex with id: xx'.
* fix rest api depend on gremlin permission, updated to not depend on gremlin.
* fix expired time is inaccurate due to call now() before tx open()
* refactor edgesOfVertex() to unified entrance with EdgeStep

Change-Id: I0d8cfd3d4b7d8ccaf10721249f8d9f86b53b442e
fix: hugegraph-test-811, hugegraph-test-837, hugegraph-test-846

### DIFF
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/auth/HugeGraphAuthProxy.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/auth/HugeGraphAuthProxy.java
@@ -416,6 +416,13 @@ public final class HugeGraphAuthProxy implements HugeGraph {
     }
 
     @Override
+    public Vertex vertex(Object object) {
+        Vertex vertex = this.hugegraph.vertex(object);
+        verifyElemPermission(HugePermission.READ, vertex);
+        return vertex;
+    }
+
+    @Override
     public Iterator<Vertex> adjacentVertex(Object id) {
         return verifyElemPermission(HugePermission.READ,
                                     this.hugegraph.adjacentVertex(id));
@@ -443,6 +450,13 @@ public final class HugeGraphAuthProxy implements HugeGraph {
     public Iterator<Edge> edges(Object... objects) {
         return verifyElemPermission(HugePermission.READ,
                                     this.hugegraph.edges(objects));
+    }
+
+    @Override
+    public Edge edge(Object id) {
+        Edge edge = this.hugegraph.edge(id);
+        verifyElemPermission(HugePermission.READ, edge);
+        return edge;
     }
 
     @Override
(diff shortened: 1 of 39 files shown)


hugegraph-api/src/main/java/com/baidu/hugegraph/auth/HugeGraphAuthProxy.java
```diff
@@ -1372,8 +1389,21 @@ public void applyStrategies(Admin<?, ?> traversal) {
...
+            /*
+             * Verify gremlin-execute permission for user gremlin(in gremlin-
+             * server-exec worker) and gremlin job(in task worker).
+             * But don't check permission in rest worker, because the following
+             * places need to call traversal():
+             *  1.vertices/edges rest api
+             *  2.oltp rest api (like crosspointpath/neighborrank)
+             *  3.olap rest api (like centrality/lpa/louvain/subgraph)
+             */
+            String caller = Thread.currentThread().getName();
```
zhoney, line 1401: olap task may run not in grizzly-http-server thread

hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
```diff
@@ -467,6 +462,11 @@ public void removeEdge(Edge edge) {
         this.graphTransaction().removeEdgeProperty((HugeEdgeProperty<V>) p);
     }
 
+    @Override
+    public Vertex vertex(Object objects) {
```
zhoney, line 466: object

hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
```diff
@@ -490,6 +490,11 @@ public boolean checkAdjacentVertexExist() {
         return this.graphTransaction().checkAdjacentVertexExist();
     }
 
+    @Override
+    public Edge edge(Object objects) {
```
zhoney, line 494: ditto

hugegraph-api/src/main/java/com/baidu/hugegraph/auth/HugeGraphAuthProxy.java
javeme: yes olap may run in task worker or olap thread, but currently we can't distinguish gremlin job and olap job

PR #1155, opened 2020-08-25
Implement a ThreadLocalCache to improve read perf

Linary:
Change-Id: I372fb5a56f374fc3993d8bdcbbb0ef07b5f160c7

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/ThreadLocalCache.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/ThreadLocalCache.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2017 HugeGraph Authors
+ *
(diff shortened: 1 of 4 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/ThreadLocalCache.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    protected V access(K id) {
+        return this.getOrNewCache().get(id);
+    }
+
+    @Override
+    protected boolean write(K id, V value) {
+        this.getOrNewCache();
```
javeme, line 66: check capacity and throw if ooc

javeme (changes requested):

Linary:
After the loader performance test, there is no obvious performance improvement for the two caches, so close this pr

PR #1156, opened 2020-08-26
fix OLTP algorithm not check if source/target vertex exist

zhoney:
Change-Id: I068f5b3f3bb57d47c284309023cc6e9b8ab4c140


hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/HugeTraverser.java
```diff
@@ -380,6 +386,11 @@ protected Id getEdgeLabelId(Object label) {
         return SchemaLabel.getLabelId(this.graph, HugeType.EDGE, label);
     }
 
+    protected void checkVertexExist(Id vertexId) {
+        // Throw NotFoundException if not exist vertex with id 'vertexId'
+        this.graph.vertex(vertexId);
```
javeme, line 391: throw 400 instead of 404

javeme (approved):

Linary (approved):

PR #1188, opened 2020-09-22
Define loading mode to speed up when loading data

Linary:
Change-Id: Ia9a93a3f0433c67353d89eb877a799ed5202fac1


hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
```diff
@@ -243,6 +243,7 @@ public GraphMode mode() {
 
     @Override
     public void mode(GraphMode mode) {
+        LOG.info("graph will work in {} mode", mode);
```
javeme, line 246: LOG.info("Graph {}  will work in {} mode", this, mode);

javeme (changes requested):

javeme (approved):

zhoney (approved):

PR #1264, opened 2020-11-16
Fix rocksdb instance not close properly

Linary:
* The open-pool opened session is always exist

Change-Id: I30504565155848659a818fbc1b57999547f0502f

### DIFF
--- a/hugegraph-rocksdb/src/main/java/com/baidu/hugegraph/backend/store/rocksdb/RocksDBStore.java
+++ b/hugegraph-rocksdb/src/main/java/com/baidu/hugegraph/backend/store/rocksdb/RocksDBStore.java
@@ -219,6 +219,12 @@ public abstract class RocksDBStore extends AbstractBackendStore<Session> {
             return;
         }
 
+        /*
+         * Transfer the session holder from db-open thread to main thread,
+         * otherwise once the db-open thread pool is closed, we can no longer
+         * close the session created by it, which will cause the rocksdb
+         * instance fail to close
+         */
         this.sessions.session();
         try {
             Consumers.executeOncePerThread(openPool, OPEN_POOL_THREADS,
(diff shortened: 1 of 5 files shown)


hugegraph-rocksdb/src/main/java/com/baidu/hugegraph/backend/store/rocksdb/RocksDBStore.java
```diff
@@ -217,6 +218,16 @@ private static void waitOpenFinish(List<Future<?>> futures,
         if (openPool.isShutdown()) {
             return;
         }
+
+        this.sessions.session();
```
javeme, line 222: add comments for why

hugegraph-rocksdb/src/main/java/com/baidu/hugegraph/backend/store/rocksdb/RocksDBStore.java
```diff
@@ -217,6 +218,22 @@ private static void waitOpenFinish(List<Future<?>> futures,
...
+        /*
+         * Transfer the session holder from db-open thread to main thread,
+         * otherwise once the db-open thread pool is closed, we can no longer
+         * close the session created by it, which will cause the rocksdb
+         * instance fail to close
+         */
+        this.sessions.session();
+        try {
+            Consumers.executeOncePerThread(openPool, OPEN_POOL_THREADS,
+                                           () -> this.sessions.closeSession());
```
javeme, line 231: is it ok if there are multi disks

javeme (changes requested):

hugegraph-rocksdb/src/main/java/com/baidu/hugegraph/backend/store/rocksdb/RocksDBStore.java
Linary: it's ok

javeme (approved):

PR #1314, opened 2020-12-25
Check if master node before schedule gremlin job

Linary:
Change-Id: I9ef3e1f3f4d30ae8b252a53c148b0b0759eff472


hugegraph-core/src/main/java/com/baidu/hugegraph/task/TaskScheduler.java
```diff
@@ -61,4 +61,8 @@
 
     public void waitUntilAllTasksCompleted(long seconds)
                                            throws TimeoutException;
+
+    default public void checkRequirement(String op) {
```
javeme, line 65: don't set default

javeme (changes requested):

javeme (approved):

houzhizhen (approved):

PR #1325, opened 2021-01-07
Fix add schema from follower then list all without it

Linary:
Change-Id: I3fa86e813fba61b4d77493e78317ebb451e58785

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/CachedSchemaTransaction.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/CachedSchemaTransaction.java
@@ -124,8 +124,7 @@ public final class CachedSchemaTransaction extends SchemaTransaction {
                                                  schema.name());
                     this.nameCache.invalidate(prefixedName);
                 }
-                // Set the cache all flag of the schema type to false
-                this.cachedTypes().put(type, false);
+                this.resetCachedAll(type);
                 return true;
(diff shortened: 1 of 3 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/CachedSchemaTransaction.java
```diff
@@ -124,6 +124,8 @@ private void listenChanges() {
                                                  schema.name());
                     this.nameCache.invalidate(prefixedName);
                 }
+                // Set the cache all flag of the schema type to false
+                this.cachedTypes().put(type, false);
```
javeme, line 128: add method to line 164: `private final void resetCachedAll(type)`

javeme (dismissed):

hugegraph-core/src/main/java/com/baidu/hugegraph/backend/cache/CachedSchemaTransaction.java
```diff
@@ -297,6 +296,11 @@ protected void removeSchema(SchemaElement schema) {
         }
     }
 
+    private final void resetCachedAll(HugeType type) {
+        // Set the cache all flag of the schema type to false
+        this.cachedTypes().put(type, false);
+    }
+
```
javeme, line 303: move to line 140

javeme (approved):

PR #1330, opened 2021-01-12
fix ttl not work due to job not close tx

zhoney:
Change-Id: Ic4a3bc11e67f92c88772149d5c304a2c1da0ec64

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/task/TaskCallable.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/task/TaskCallable.java
@@ -22,6 +22,7 @@ package com.baidu.hugegraph.task;
 import java.util.Date;
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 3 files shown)


hugegraph-core/src/main/java/com/baidu/hugegraph/task/TaskCallable.java
```diff
@@ -48,13 +48,19 @@ public TaskCallable() {
...
-        // Do nothing, subclasses may override this method
+        this.closeTx();
     }
 
     protected void cancelled() {
         // Do nothing, subclasses may override this method
     }
 
+    protected void closeTx() {
+        if (this.graph().tx().isOpen()) {
```
javeme, line 59: define tx first

javeme (changes requested):

javeme (approved):

Linary (approved):

PR #1383, opened 2021-03-09
Fix inappropriate shard will cause base64 decode error

Linary:


hugegraph-rocksdb/src/main/java/com/baidu/hugegraph/backend/store/rocksdb/RocksDBTable.java
```diff
@@ -327,6 +327,7 @@ public long estimateNumKeys(Session session) {
 
         @Override
         public byte[] position(String position) {
+            // The value of START AND END are same
```
javeme, line 330: AND => and
but prefer don't assume START and END are the same, prefer `START.equals(position) || END.equals(position)`

javeme (commented):
please update PR title

javeme (approved):

zhoney (approved):

PR #1434, opened 2021-04-27
Fix Fusiform similarity api return all vertices when similars is empty

zyxxoo:
Fix Fusiform similarity api return all vertices when similars is empty


hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/FusiformSimilarityAPI.java
```diff
@@ -113,7 +113,7 @@ public String post(@Context GraphManager manager,
         CloseableIterator.closeIterator(sources);
 
         Iterator<Vertex> iterator = QueryResults.emptyIterator();
-        if (request.withVertex) {
+        if (request.withVertex && !result.isEmpty()) {
```
zhoney, line 116: prefer:
```
Set<Id> vertices = result.vertices();
if (request.withVertex && !vertices.isEmpty()) {
    iterator = g.vertices(vertices.toArray());
}
```

zhoney (changes requested):

zhoney (approved):

Linary (approved):

PR #1435, opened 2021-04-27
Fix paths api return empty if start and end vertix same id

zyxxoo:
paths api  get and post method return same result


hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/NeighborRankTraverser.java
```diff
@@ -244,7 +244,7 @@ private Ranks contributeNewLayer(List<Adjacencies> adjacencies,
         public Step(HugeGraph g, Directions direction, List<String> labels,
                     long degree, long skipDegree, int top, int capacity) {
             E.checkArgument(top > 0 && top <= MAX_TOP,
-                            "The top of each layer can't exceed %s", MAX_TOP);
+                            "The top of each layer muse be in (0, %s]", MAX_TOP);
```
imbajin, line 247: must

Linary (approved):

zhoney (approved):

PR #1459, opened 2021-05-19
refact: let kout degree applied to all labels & remove source vertex from kneighbor default results 

imbajin:
Contains 3 point:
1. kout degree applied to all labels
2. remove source vertex from default kneighbor result
3. fix kneighbor result with limit may loss data because the duplicate elements of current layer are not excluded

PS: Also remove LimitIterator (moved to common) & add/adapt test for them on [client#122](https://github.com/hugegraph/hugegraph-client/pull/122)


hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/HugeTraverser.java
```diff
@@ -182,10 +183,19 @@ protected int concurrentDepth() {
...
-        return results;
+
+        if (limit == NO_LIMIT) {
+            return results;
+        }
+
+        Query query = new Query(HugeType.EDGE);
+        query.limit(limit);
+        return new LimitIterator<>(results, e -> {
+            long count = query.goOffset(1L);
```
javeme, line 196: use LongAdder

zhoney (dismissed):

zhoney (approved):

javeme (approved):

PR #1531, opened 2021-07-01
chore: refactor issue template with forum struct (yaml format)

imbajin:
After use it, submit issue will be more user friendly, main diff here:
1. the whole UI experience is much more better (a string of enhancement)
2. users don't need to update / write markdown by themselves
3. we could define the highlight code block by default language
4. description will only show before issue submitted, no need to hide them by any trick way.. (such like `<details>` or html-comment)
5. we can choose required info now (otherwise cannot submit issue)

PS: also reduce the daily stale issues amount from 25 -> 10

the UI before submit like: (click [here](https://github.com/hugegraph/hugegraph/blob/issue-forum-template/.github/ISSUE_TEMPLATE/bug_report.yml) to preview)
image

after submit like:

image


zhoney (dismissed):

Linary (dismissed):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,101 @@
...
+        {
+          "vertex": { "id": "xxx" }
+        }
+      render: javascript
+  
+  - type: textarea
+    attributes:
+      label: Schema [VertexLabel, EdgeLabel, IndexLabel] (元数据结构)
+      description: |
+        > 如果问题与具体的点 / 边 / 索引 label 相关, 请提供完整的 Schema 返回 JSON 结果"
```
javeme, line 92: 点类型 / 边类型 / 索引类型等元数据相关

javeme (approved):

PR #1532, opened 2021-07-01
fix clear api create snapshot for all backend(should rocksdb only)

zhoney:
Change-Id: I936ee13a0a4f8afc61384fe6adee89bbb74c8256


Linary (approved):

hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
```diff
@@ -508,6 +498,23 @@ private Analyzer analyzer() {
         return AnalyzerFactory.analyzer(name, mode);
     }
 
+    private void createSnapshotIfNeeded() {
+        if (!this.storeProvider.type().equals("rocksdb")) {
```
javeme, line 502: seems can move to RaftStoreProvider.initSystemInfo()

javeme (changes requested):

javeme (approved):

Linary (approved):

PR #1561, opened 2021-08-04
Refactor(HbaseOptions): use static inner class singleton pattern.

limpidezza:
1.instance is initialized only when the instance() method is called for the first time.
2. before optimization, every time instance() method is called, will be request lock. this optimization avoid request lock and improved performance.

### DIFF
--- a/hugegraph-hbase/src/main/java/com/baidu/hugegraph/backend/store/hbase/HbaseOptions.java
+++ b/hugegraph-hbase/src/main/java/com/baidu/hugegraph/backend/store/hbase/HbaseOptions.java
@@ -32,14 +32,15 @@ public class HbaseOptions extends OptionHolder {
         super();
     }
 
-    private static volatile HbaseOptions instance;
-
-    public static synchronized HbaseOptions instance() {
-        if (instance == null) {
-            instance = new HbaseOptions();
+    private static class InstanceHolder {
+        private final static HbaseOptions instance = new HbaseOptions();
+        static {
             instance.registerOptions();
         }
-        return instance;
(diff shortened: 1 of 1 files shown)


imbajin:
thanks for your contribution first 👍🏻 

Although the class `HbaseOption` will actually be only used once, we can indeed improve the `singleton` code, but why we don't change it like this (double check to void sync the whole method):

```java
    private static volatile HbaseOptions instance;

    // move sync from outside to inside
    public static HbaseOptions instance() {
        if (instance == null) {
            synchronized (HbaseOptions.class) {
                if (instance == null) {
                    instance = new HbaseOptions();
                    instance.registerOptions();
                }
            }
        }
        return instance;
    }
```

limpidezza:
> thanks for your contribution first 👍🏻
> 
> Although the class `HbaseOption` will actually only used once, we can indeed improve the `singleton` code, but why we don't change it like this (double check to void sync the whole method):
> 
> ```java
>     private static volatile HbaseOptions instance;
> 
>     // move sync from outside to inside
>     public static HbaseOptions instance() {
>         if (instance == null) {
>             synchronized (HbaseOptions.class) {
>                 if (instance == null) {
>                     instance = new HbaseOptions();
>                     instance.registerOptions();
>                 }
>             }
>         }
>         return instance;
>     }
> ```
> 
> so as other `XXOptions` classes (like RocksdDBOptions/CassandraOptions..etc)

hi, @imbajin , I think both `double check` and `static inner class` achieve the same effect. but `static inner class` more readable and easier to implement.

PR #1604, opened 2021-09-10
Using TCK Tested JDK builds of OpenJDK

carldea:
The AdoptOpenJDK has been discontinued since July 2021. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. Also, added a fixed (major) version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,12 +21,13 @@ jobs:
       fail-fast: false
       matrix:
         BACKEND: [memory, cassandra, scylladb, hbase, rocksdb]
+        java-version: ['8', '8.0.192']
     steps:
-      - name: Install JDK 8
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -21,12 +21,13 @@ jobs:
       fail-fast: false
       matrix:
         BACKEND: [memory, cassandra, scylladb, hbase, rocksdb]
+        java-version: ['8', '8.0.192']
```
javeme, line 24: since one latest version is ok and can remove the '8.0.192'.
and also keep JAVA_VERSION style?

javeme (commented):
Thanks for your contribution

.github/workflows/ci.yml
carldea: Cool, I updated the workflow. 👍🏼

PR #1618, opened 2021-10-13
Fix: quorum read if safe read

zyxxoo:


hugegraph-dist/src/assembly/static/conf/hugegraph.properties
```diff
@@ -39,6 +39,7 @@ raft.apply_batch=1
 raft.rpc_threads=80
 raft.rpc_connect_timeout=5000
 raft.rpc_timeout=60000
+raft.read_only_option=ReadOnlyLeaseBased
```
javeme, line 42: move to line 36

hugegraph-core/src/main/java/com/baidu/hugegraph/config/CoreOptions.java
```diff
@@ -195,6 +196,15 @@ public static synchronized CoreOptions instance() {
                     8
             );
 
+    public static final ConfigOption<String> RAFT_READ_ONLY_OPTION =
+            new ConfigOption<>(
+                    "raft.read_only_option",
```
javeme, line 201: raft.read_strategy

hugegraph-core/src/main/java/com/baidu/hugegraph/config/CoreOptions.java
```diff
@@ -195,6 +196,15 @@ public static synchronized CoreOptions instance() {
...
+                    "The linearizability of read strategy.",
+                    allowValues(ReadOnlyOption.ReadOnlyLeaseBased.name(),
```
javeme, line 203: prefer to use "ReadOnlyLeaseBased" string

PR #1623, opened 2021-10-26
oltp traversal: optimize id2code for number id

javeme:


hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/records/AbstractRecords.java
```diff
@@ -41,24 +42,35 @@ public AbstractRecords(RecordType type, boolean concurrent) {
     }
 
     @Watched
-    protected int code(Id id) {
-        return this.idMapping.object2Code(id);
+    protected final int code(Id id) {
+        if (id.number()) {
+            long l = id.asLong();
+            if (0 <= l && l <= Integer.MAX_VALUE) {
```
LiJie20190102, line 48: May I ask why this change was made? I recently encountered a problem where there is a large amount of data. The vertex ID is of type number, but I accidentally chose the String type when entering the graph. Therefore, when I execute olap, there will be an OOM issue; But when I switch to the Number type, no oom occurs. So I would like to ask why I am making code modifications here, and have I encountered similar problems as myself?

LiJie20190102:
image
May I ask why this change was made? I recently encountered a problem where there is a large amount of data. The vertex ID is of type number, but I accidentally chose the String type when entering the graph. Therefore, when I execute olap, there will be an OOM issue; But when I switch to the Number type, no oom occurs. So I would like to ask why I am making code modifications here, and have I encountered similar problems as myself?

LiJie20190102:
> image May I ask why this change was made? I recently encountered a problem where there is a large amount of data. The vertex ID is of type number, but I accidentally chose the String type when entering the graph. Therefore, when I execute olap, there will be an OOM issue; But when I switch to the Number type, no oom occurs. So I would like to ask why I am making code modifications here, and have I encountered similar problems as myself?

@javeme @zhoney @imbajin

javeme:
@LiJie20190102 maybe the idMapping collection is too large, then cause OOM. you can try to check the large objects this way `jmap -histo:live PID`

LiJie20190102:
image

PR #1641, opened 2021-11-10
fix: mysql & pg ci problem

imbajin:
fix 2 main problems:
 - mysql runtime image can't work (use mysql action now)
 - passwd for mysql/pg is empty so cannot connect database correctly
 - use "******" as the default passwd now


javeme (approved):

corgiboygsj (approved):

.github/workflows/ci.yml
```diff
@@ -49,6 +50,19 @@ jobs:
         run: |
           $TRAVIS_DIR/install-backend.sh ${{ matrix.BACKEND }}
 
+      - name: Init Mysql Env
```
javeme, line 53: can we init mysql only if necessary?

.github/workflows/ci.yml
```diff
@@ -49,6 +50,19 @@ jobs:
...
+          host port: 3306 # Optional, default value is 3306. The port of host
+          container port: 3306 # Optional, default value is 3306. The port of container
+          character set server: 'utf8mb4' # Optional, default value is 'utf8mb4'. The '--character-set-server' option for mysqld
+          collation server: 'utf8mb4_general_ci' # Optional, default value is 'utf8mb4_general_ci'. The '--collation-server' option for mysqld
+          mysql version: '5.7' # Optional, default value is "latest". The version of the MySQL
+          mysql database: 'test' # Optional, default value is "test". The specified database which will be create
+          mysql root password: "******" # Required if "mysql user" is empty, default is empty. The root superuser password
+          #mysql user: 'root' # Required if "mysql root password" is empty, default is empty. The superuser for the specified database. Can use secrets, too
+          #mysql password: ${{ secrets.DatabasePassword }} # Required if "mysql user" exists. The password for the "mysql user"
+
```
javeme, line 65: can remove them

.github/workflows/ci.yml
imbajin: yep, can only init it when backend is `mysql` (better init it in `install-mysql-docker.sh`, but it doesn't work well)

PR #1675, opened 2021-11-30
refactor olap property

javeme:
Change-Id: I137a42916eeee00775eafb55ee299a8a64e21671

### DIFF
--- a/hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
+++ b/hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
@@ -259,7 +259,7 @@ public class StandardHugeGraph implements HugeGraph {
         this.serverInfoManager().initServerInfo(serverId, serverRole);
 
         // TODO: check necessary?
-        LOG.info("Init olap property-keys for graph '{}'", this.name);
+        LOG.info("Check olap property-key tables for graph '{}'", this.name);
         for (PropertyKey pk : this.schemaTransaction().getPropertyKeys()) {
             if (pk.olap()) {
                 this.graphTransaction().initAndRegisterOlapTable(pk.id());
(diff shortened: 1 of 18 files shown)


zhoney (dismissed):

hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
```diff
@@ -258,8 +258,13 @@ public void serverStarted(Id serverId, NodeRole serverRole) {
                  serverId, serverRole, this.name);
         this.serverInfoManager().initServerInfo(serverId, serverRole);
 
-        LOG.info("Search olap property key for graph '{}'", this.name);
-        this.schemaTransaction().initAndRegisterOlapTables();
+        // TODO: check necessary?
```
imbajin, line 261: we will rebuild missing olap table here (in another pr)

hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
```diff
@@ -258,8 +258,13 @@ public void serverStarted(Id serverId, NodeRole serverRole) {
...
+        LOG.info("Init olap property-keys for graph '{}'", this.name);
```
imbajin, line 262: seems the info is not accurate,  "Check olap property-keys and tables for xxx"  is better

hugegraph-core/src/main/java/com/baidu/hugegraph/StandardHugeGraph.java
javeme: Ok, when will the table be missing?

imbajin: at least 2 case:
1. create property key success, create table failed (with any reason)
2. the olap table store file delete/broken by some operations

and once it happens, we can't delete or search or recovery it, this problem will fixed in that pr

imbajin (approved):

zhoney (approved):

PR #1687, opened 2021-12-13
fix log4j error

zhoney:
Change-Id: Ia8fc8e0036c53d3c4652b53de64d84d3d3a2eb93


zhoney:
log4j error already fixed in release `hugegraph-0.11.2-b4.tar.gz`.

PR #1723, opened 2022-01-04
refact: remove early cycle detection & limit max depth

imbajin:
1.  limit max depth to avoid stackoverflow
2. remove useless cycle detection due to it will remove valid path

TODO: maybe we should refactor the recursive findPath to for-loop

### DIFF
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/PathsAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/PathsAPI.java
@@ -114,9 +114,9 @@ public class PathsAPI extends TraverserAPI {
                                "The targets of request can't be null");
         E.checkArgumentNotNull(request.step,
                                "The step of request can't be null");
-        E.checkArgument(request.depth > 0,
-                        "The depth of request must be > 0, but got: %s",
-                        request.depth);
+        E.checkArgument(request.depth > 0 && request.depth <= 10000,
+                        "The depth of request must be in (0, 10000], " +
+                        "but got: %s", request.depth);
 
         LOG.debug("Graph [{}] get paths from source vertices '{}', target " +
                   "vertices '{}', with step '{}', max depth '{}', " +
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/PathsAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/PathsAPI.java
@@ -114,8 +114,8 @@ public class PathsAPI extends TraverserAPI {
                                "The targets of request can't be null");
         E.checkArgumentNotNull(request.step,
                                "The step of request can't be null");
-        E.checkArgument(request.depth > 0 && request.depth <= 10000,
-                        "The depth of request must be in (0, 10000], " +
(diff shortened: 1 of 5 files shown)


coderzc (dismissed):

hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/PathsAPI.java
```diff
@@ -114,9 +114,9 @@ public String post(@Context GraphManager manager,
                                "The targets of request can't be null");
         E.checkArgumentNotNull(request.step,
                                "The step of request can't be null");
-        E.checkArgument(request.depth > 0,
-                        "The depth of request must be > 0, but got: %s",
-                        request.depth);
+        E.checkArgument(request.depth > 0 && request.depth <= 5000,
```
javeme, line 117: can add a const MAX_DEPTH?

hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/PathsAPI.java
```diff
@@ -114,9 +114,9 @@ public String post(@Context GraphManager manager,
...
+                        "The depth of request must be in (0, 5000], " +
+                        "but got: %s", request.depth);
```
javeme, line 119: should also add depth check to other APIs?

hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/PathsTraverser.java
```diff
@@ -124,9 +124,6 @@ public void forward(Id targetV, Directions direction) {
                     PathSet results = this.record.findPath(target, null,
                                                            true, false);
                     for (Path path : results) {
-                        if (Objects.equals(target, targetV)) {
```
javeme, line 127: maybe someone don't want a ring path? can add a optional param

imbajin: it will remove the exist path due to the skip logic, a->b->c
when we first traverser from a to c, we'll bypass c then lost the only one path

hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/PathsAPI.java
imbajin: all path api use same limit?

javeme: yes, it's better to keep consistent

hugegraph-core/src/main/java/com/baidu/hugegraph/traversal/algorithm/PathsTraverser.java
javeme: not sure what it means, can add more details?

imbajin: the `targetV` will be the same with `target`, it means when we traverser from `a` to `c`, the first time we arrived c, we drop the path directly. (details in the context)

javeme (approved):

zhoney (approved):

PR #1820, opened 2022-04-15
CONTRIBUTING.md: add how to create a GitHub token

javeme:


CONTRIBUTING.md
```diff
@@ -103,6 +103,9 @@ Push the local commit to GitHub fork repo:
 git push origin bugfix-branch:bugfix-branch
 ```
 
+Note that since GitHub requires submitting code through `username + token` (instead of using `username + password` directly), you need to create a GitHub token from https://github.com/settings/tokens:
```
imbajin, line 106: seems it's not necessary for normal user/dev to use this?

corgiboygsj (approved):

CONTRIBUTING.md
javeme: mainly for new contributors

imbajin (approved):
todo:
- [ ] we should let dev use `ssh` to push code later (not recommend for http way)

PR #1834, opened 2022-04-19
chore: replace mysql actions by official docker image

imbajin:
also could used for PG later


javeme:
It is recommended to install the codecov app first. We installed it for hugegraph org before. Maybe the apache org is not installed codecov app. We need to install codecov for each repo separately.

corgiboygsj (approved):

javeme (approved):

Issue #1882, opened 2022-05-16
[Bug] The api of getProfile can not be accessed 


wangyao2016:
### Bug Type (问题类型)

rest-api (结果不合预期)

### Before submit

- [X] 我已经确认现有的 [Issues](https://github.com/hugegraph/hugegraph/issues) 与 [FAQ](https://hugegraph.github.io/hugegraph-doc/guides/faq.html) 中没有相同 / 重复问题

### Environment (环境信息)

- Server Version: v0.13.0
- Backend: RocksDB x 1
- OS: 8 CPUs, 32 G RAM, Centos 7.6
- Data Size:  xx

### Expected & Actual behavior (期望与实际表现)

Expected
`
{
    "service": "hugegraph",
    "version": "0.13.0.0",
    "doc": "https://hugegraph.github.io/hugegraph-doc/",
    "api_doc": "https://hugegraph.github.io/hugegraph-doc/clients/hugegraph-api.html",
    "apis": [
        "auth",
        "filter",
        "graph",
        "gremlin",
        "job",
        "metrics",
        "profile",
        "raft",
        "schema",
        "traversers",
        "variables"
    ]
}
`

 Actual behavior 
`
{
    "exception": "class java.lang.IllegalArgumentException",
    "message": "Invalid request uri ''",
    "cause": ""
}
`

### Vertex/Edge example (问题点 / 边数据举例)

```javascript
I think the reason for the problem is: There is a problem accessing the route definition
```

wangyao2016:
The pic of Actual behavior :
![image](https://user-images.githubusercontent.com/16890042/168511908-40ded10a-547a-4f1f-8a5f-4fe27bb9a92d.png)

And I have a pr for this. [fix bud: get profile api cannot be accessed](https://github.com/apache/incubator-hugegraph/pull/1883)

javeme:
fixed by #1883 & #1893

PR #1500, opened 2021-06-15
add LoginAPI

corgiboygsj:
Use HS256 generator token, token payload have `user_name` and `user_id`.

PR #1883, opened 2022-05-16
fix bug: get profile api cannot be accessed #1882

wangyao2016:
To solve this problem [#1882](https://github.com/apache/incubator-hugegraph/issues/1882),  I modified the route of getprofile API as `/profiles`

image

### DIFF
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/api/profile/ProfileAPI.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/api/profile/ProfileAPI.java
@@ -59,6 +59,7 @@ public class ProfileAPI {
     private static String API_PROFILES = null;
 
     @GET
+    @Path("profiles")
     @Timed
     @Produces(MediaType.APPLICATION_JSON)
     public String getProfile(@Context Application application) {


hugegraph-api/src/main/java/com/baidu/hugegraph/api/profile/ProfileAPI.java
```diff
@@ -59,6 +59,7 @@ public class ProfileAPI {
     private static String API_PROFILES = null;
 
     @GET
+    @Path("profiles")
```
javeme, line 62: can we fix this issue and keep the origin url path?

javeme:
@zyxxoo Do you think this issue may be caused by upgrading jersey?

hugegraph-api/src/main/java/com/baidu/hugegraph/api/profile/ProfileAPI.java
wangyao2016: It cause by this PR #1500 

[this line](https://github.com/apache/incubator-hugegraph/commit/9ee9fe12f3349a54bbceb161d002200d3689863c#diff-76994bedad9c0d52982ee900de571f9db327acfd5a8ab902c606cc20dc62a19aR284)

wangyao2016: Modify the check rule or just change routing to "/profiles"?

javeme: Thanks for pointing out the root cause.
IMO we cam add url "/" to WHITE_API_LIST to keep the continuity.

corgiboygsj: I didn't notice that ProfileApi would use an blank url. So in order to keep the original url, I chose to delete the blank url check. #1893

javeme: OK, let's keep the original url

wangyao2016:
https://github.com/apache/incubator-hugegraph/pull/1893

PR #1888, opened 2022-05-17
chore: replace `APPLICATION_JSON` with `TEXT_PLAIN` in dynamic create graph api

simon824:
The format of the graph instance config is `text` instead of `json`
image


javeme (approved):

seagle-yuan (approved):

imbajin:
> > @wangyao2016 could u take a look for it?
> 
> Sure. IMO both Json and text are fine , `PropertiesConfiguration` supports parsing two formats. But text is better.
> 
> Or we can support both, just remove the limit of `@Consumes(APPLICATION_JSON_WITH_CHARSET)`.

could enhance it together next time

imbajin (approved):

PR #1904, opened 2022-06-15
Minor: add editorconfig

jackwener:
It's a small but useful change.

It's will enable `new_line in end` which is good for `code-style`.

set `end_of_line = lf` is useful to the developers using Windows.


.editorconfig
```diff
@@ -0,0 +1,27 @@
...
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+
+[*.{java,xml,py}]
+indent_style = space
+indent_size = 4
```
imbajin, line 27: add a new line in the end of file, and When and where will this file take effect? (only in win or some IDE?)

jackwener: https://editorconfig.org/

`editorconfig` is just work in almost IDE(vsc, IDEA.....) .

imbajin: ok, seems fine

should we add more options? @seagle-yuan

javeme, line 27: expect a blank line at end of ".editorconfig" file

javeme (dismissed):
LGTM! Thanks for your contribution

.editorconfig
jackwener: fixed

javeme (approved):

PR #1932, opened 2022-07-15
close LimitIterator when reach limit (#410)

zhoney:


coderzc (approved):

zyxxoo:
The close have been called

PR #1937, opened 2022-07-22
GraphPlatform-430: contribute hugegraph code to open source

GioPhantus:
Change-Id: I7ee92b28470e235ee28dc84d31c5487ff78c23d4

### DIFF
--- a/hugegraph-api/src/main/java/com/baidu/hugegraph/define/UpdateStrategy.java
+++ b/hugegraph-api/src/main/java/com/baidu/hugegraph/define/UpdateStrategy.java
@@ -26,6 +26,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


hugegraph-api/src/main/java/com/baidu/hugegraph/define/UpdateStrategy.java
```diff
@@ -60,15 +65,20 @@ Object updatePropertyValue(Object oldProperty, Object newProperty) {
 
         @Override
         void checkPropertyType(Object oldProperty, Object newProperty) {
-            E.checkArgument((oldProperty instanceof Date ||
-                             oldProperty instanceof Number) &&
-                            (newProperty instanceof Date ||
+            E.checkArgument((oldProperty instanceof Date &&
+                             newProperty instanceof String) ||
```
imbajin, line 69: why we support string here? could u add some case for it?

PR #1953, opened 2022-08-23
chore: fix GraphOfTheMoviesExample exception

simon824:
This pr will fix these exception:
```
Exception in thread "main" java.lang.IllegalArgumentException: No need to build index on properties [name], because they contains all primary keys [name] for vertex label 'person'
```

```
Exception in thread "main" java.lang.IllegalStateException: Stop the current transaction before opening another
```

### DIFF
--- a/hugegraph-example/src/main/java/com/baidu/hugegraph/example/GraphOfTheMoviesExample.java
+++ b/hugegraph-example/src/main/java/com/baidu/hugegraph/example/GraphOfTheMoviesExample.java
@@ -105,12 +105,8 @@ public class GraphOfTheMoviesExample {
               .sourceLabel("person").targetLabel("movie")
               .create();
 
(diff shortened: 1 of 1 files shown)


hugegraph-example/src/main/java/com/baidu/hugegraph/example/GraphOfTheMoviesExample.java
```diff
@@ -105,13 +105,9 @@ public static void load(final HugeGraph graph) {
               .sourceLabel("person").targetLabel("movie")
               .create();
 
-        schema.indexLabel("personByName").onV("person").by("name")
```
imbajin, line 108: THX, could add & run it in test/CI?

javeme (approved):

imbajin (approved):

Issue #1987, opened 2022-10-23
Fix code scanning alert - User-controlled data in numeric cast


coderzc:
Tracking issue for:
- [ ] https://github.com/apache/incubator-hugegraph/security/code-scanning/13

PR #1996, opened 2022-10-30
fix: User-controlled data in numeric cast [#1987]

zyxxoo:


hugegraph-core/src/protobuf/java/com/baidu/hugegraph/backend/store/raft/rpc/RaftRequests.java
```diff
@@ -1,22 +1,3 @@
-/*
```
javeme, line 1: don't need to push this file?

hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/AdamicAdarAPI.java
```diff
@@ -67,7 +67,7 @@ public String get(@Context GraphManager manager,
                       @QueryParam("max_degree")
                       @DefaultValue(DEFAULT_MAX_DEGREE) long maxDegree,
                       @QueryParam("limit")
-                      @DefaultValue(DEFAULT_ELEMENTS_LIMIT) long limit) {
+                      @DefaultValue(DEFAULT_ELEMENTS_LIMIT) int limit) {
```
javeme, line 70: do all the algorithms(like adamicAdar) need to param limit with type int?

zyxxoo: Ok, I change the other algorithms limit type to int

zyxxoo: These place long are not suitable, because the java package only support int type on collections

hugegraph-test/src/main/java/com/baidu/hugegraph/core/RoleElectionStateMachineTest.java
```diff
@@ -149,7 +153,7 @@ public void master(StateMachineContext context) {
                 if (logRecords.size() > MAX_COUNT) {
                     context.stateMachine().shutdown();
                 }
-                System.out.println("master node: " + node);
+                LOG.info("master node: " + node);
```
javeme, line 156: prefer to define a `UtilXx.print()` method

javeme (approved):

hugegraph-api/src/main/java/com/baidu/hugegraph/api/traversers/AdamicAdarAPI.java
imbajin: yes，I also found this problem~

zyxxoo: I have change all algorithms limit type long to int, and I have put a pr for hugegraph-client

imbajin (approved):

PR #2019, opened 2022-11-18
refact: adapt apache package and dependency in all modules (Breaking Change)

z7658329:


z7658329:
image

also need to adapt license

javeme (dismissed):

javeme (approved):

hugegraph-core/src/main/java/org/apache/hugegraph/util/ConfigUtil.java
```diff
@@ -112,9 +112,11 @@ public static Map<String, String> scanGraphsDir(String graphsDirPath) {
 
     public static String writeToFile(String dir, String graphName,
                                      HugeConfig config) {
-        E.checkArgument(FileUtils.getFile(dir).exists(),
+        File file = FileUtils.getFile(dir);
+        E.checkArgument(file.exists(),
                         "The directory '%s' must exist", dir);
-        String fileName = Paths.get(dir, graphName + CONF_SUFFIX).toString();
+        String tmpGraphName = graphName + CONF_SUFFIX;
```
imbajin, line 118: the security alert let us check the user params (trim it) before using: (reasonable)
image

imbajin: > in our case, fileName is "graphName.properties ", has one dot

we could assert only one `.` could exist, and another way is we just truncate before `.` then we add `.properties` as surffix later

javeme:
@z7658329  can we move the CodeQL fixes into a separate PR?

z7658329:
> @z7658329 can we move the CodeQL fixes into a separate PR?

i have fixed, but in ci progress , 
hugegraph-ci / build (rocksdb, 11) (pull_request)   failed, should rerun it

hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/BelongAPI.java
```diff
@@ -60,7 +62,7 @@ public class BelongAPI extends API {
 
     @POST
     @Timed
-    @Status(Status.CREATED)
+    @StatusFilter.Status(StatusFilter.Status.CREATED)
```
javeme, line 65: keep "Status"?

hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/GroupAPI.java
```diff
@@ -1,4 +1,5 @@
 /*
+ *
```
javeme, line 2: not sure why need to add a blank line?

hugegraph-api/src/main/java/org/apache/hugegraph/api/graph/VertexAPI.java
```diff
@@ -221,7 +223,7 @@ public String update(@Context GraphManager manager,
 
     @GET
     @Timed
-    @Compress
+    @CompressInterceptor.Compress
```
javeme, line 226: ditto

z7658329: OK , i will check

javeme (dismissed):

hugegraph-api/src/main/java/org/apache/hugegraph/api/job/AlgorithmAPI.java
```diff
@@ -55,7 +55,7 @@ public class AlgorithmAPI extends API {
     @POST
     @Timed
     @Path("/{name}")
-    @Status(Status.CREATED)
+    @StatusFilter.Status(StatusFilter.Status.CREATED)
```
javeme, line 58: can we search all the "@StatusFilter.Status" and update to "@Status"?

hugegraph-api/src/main/java/org/apache/hugegraph/api/graph/EdgeAPI.java
```diff
@@ -261,7 +261,7 @@ public String update(@Context GraphManager manager,
 
     @GET
     @Timed
-    @Compress
+    @CompressInterceptor.Compress
```
javeme, line 264: can we search all the `@CompressInterceptor.Compress` and update to `@Compress`?

hugegraph-api/src/main/java/org/apache/hugegraph/api/job/AlgorithmAPI.java
z7658329: yes i will fix it, and notify you to review

javeme (approved):

hugegraph-core/src/main/java/org/apache/hugegraph/HugeGraph.java
```diff
@@ -247,7 +247,7 @@ public interface HugeGraph extends Graph {
     <K, V> V option(TypedOption<K, V> option);
 
     void registerRpcServices(RpcServiceConfig4Server serverConfig,
-                             RpcServiceConfig4Client clientConfig);
+                                    RpcServiceConfig4Client clientConfig);
```
imbajin, line 250: misoperation?

hugegraph-core/src/main/java/org/apache/hugegraph/auth/ResourceObject.java
```diff
@@ -82,8 +83,8 @@ public static ResourceObject<HugeElement> of(String graph,
         return new ResourceObject<>(graph, resType, elem);
     }
 
-    public static ResourceObject<SchemaDefine.AuthElement> of(String graph,
-                                                              SchemaDefine.AuthElement elem) {
+    public static ResourceObject<AuthElement> of(String graph,
+                                                              AuthElement elem) {
```
imbajin, line 87: merge to one line if < 100

z7658329: > merge to one line if < 100

ok i fixed it

imbajin:
some TODOs & Warning:
1. the `APIVersion` now is 0.67, shall we keep it?
2. CommonVersion
image

imbajin:
> why check common's version?
>  : to avoid importing two different common jars with conflict versions.

What does this mean? At present, the whole project seems to have only one `huge-common` dependency

imbajin:
> > What does this mean? At present, the whole project seems to have only one `huge-common` dependency
> 
> someone may import hg as a lib, at the same time also has imported hg-common.

Seems it should be handled by users, use a fixed version or exclude the duplicate one, and currently it have 2 problems:
1. the check fails when the version is 1.0.0 but the range is 2.x~2.x
2. how to set the new version? (because the old common also use the 1.x & 2.x as their version num)

consider handle this in a separate PR

pom.xml
```diff
@@ -130,16 +82,16 @@
 
     <dependencyManagement>
         <dependencies>
-            <!-- hugegraph-commons -->
+            <!-- TODO: could we merge them to hugegraph-commons only? -->
```
javeme, line 85: no we can't merge them, because in some scenarios it is not necessary to import both the two packages at the same time, like client or loader.

imbajin (dismissed):

javeme (approved):

pom.xml
imbajin: OK, change the comment in next PR (could modify the common version together @z7658329 )

z7658329:
> > > What does this mean? At present, the whole project seems to have only one `huge-common` dependency
> > 
> > 
> > someone may import hg as a lib, at the same time also has imported hg-common.
> 
> Seems it should be handled by users(like use a fixed version or exclude the duplicate one), and currently it have 2 problems:
> 
> 1. the check fails when the version is 1.0.0 but the range is 2.x~2.x
> 2. how to set the new version? (because the old common also use the 1.x & 2.x as their version num)
> 
> consider handle this in a separate PR

z7658329:
> > > What does this mean? At present, the whole project seems to have only one `huge-common` dependency
> > 
> > 
> > someone may import hg as a lib, at the same time also has imported hg-common.
> 
> Seems it should be handled by users(like use a fixed version or exclude the duplicate one), and currently it have 2 problems:
> 
> 1. the check fails when the version is 1.0.0 but the range is 2.x~2.x
> 2. how to set the new version? (because the old common also use the 1.x & 2.x as their version num)
> 
> consider handle this in a separate PR

image

imbajin (approved):
use `rebase and merge`

javeme (approved):

PR #2024, opened 2022-11-20
[fix][cli] Fix start cli error

coderzc:
### Motivation 
Run `bin/hugegraph start` error

```bash
Starting HugeGraphServer...
bin/hugegraph: line 68: /start-hugegraph.sh: No such file or directory
Failed to start service
```

### Modifications

Add check for `Variables `

### DIFF
--- a/hugegraph-dist/src/assembly/static/bin/hugegraph
+++ b/hugegraph-dist/src/assembly/static/bin/hugegraph
@@ -41,8 +41,8 @@
 INSTALL_DIR=
 SERVER_PORT=
 
-${INSTALL_DIR:?"Please setting variables 'INSTALL_DIR'"}
-${SERVER_PORT:?"Please setting variables 'SERVER_PORT'"}
+${INSTALL_DIR:?"Please set variables 'INSTALL_DIR'"}
+${SERVER_PORT:?"Please set variables 'SERVER_PORT'"}
 
 BIN_DIR=$INSTALL_DIR/bin
 SERVER_URL="http://localhost:${SERVER_PORT}"
--- a/hugegraph-dist/src/assembly/static/bin/hugegraph
+++ b/hugegraph-dist/src/assembly/static/bin/hugegraph
@@ -37,10 +37,19 @@
 ### END INIT INFO
 
 # Variables
-INSTALL_DIR=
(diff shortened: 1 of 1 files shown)


zyxxoo (approved):

hugegraph-dist/src/assembly/static/bin/hugegraph
```diff
@@ -37,10 +37,19 @@
 ### END INIT INFO
 
 # Variables
-INSTALL_DIR=
```
imbajin, line 40: change the comment, it requires user to set a **fixed abs path** manually (should't use a relative path)

coderzc: You are right. and I think should add check for `Variables`

hugegraph-dist/src/assembly/static/bin/hugegraph
```diff
@@ -37,6 +37,7 @@
 ### END INIT INFO
 
 # Variables
+# it requires user to set a fixed abs path manually
```
javeme, line 40: we can try `${INSTALL_DIR:?"Please setting variables 'INSTALL_DIR'"}`
for more details: https://stackoverflow.com/a/16753536

hugegraph-dist/src/assembly/static/bin/hugegraph
```diff
@@ -41,13 +41,15 @@
 INSTALL_DIR=
 SERVER_PORT=
 
+${INSTALL_DIR:?"Please setting variables 'INSTALL_DIR'"}
```
imbajin, line 44: 😺

javeme (approved):

imbajin (approved):
I'll enhance this script later

javeme (approved):

hugegraph-dist/src/assembly/static/bin/hugegraph
imbajin: > https://stackoverflow.com/a/16753536
@javeme seems a little strange, could it use directly or test it in local evn? (or should in a if condition?)

image

and it alert:
image

failed all cases in my ubuntu21

PR #2028, opened 2022-11-22
[improve] Shutdown log when shutdown process

coderzc:
### Motivation 
If set `immediateFlush="false"` and log size < bufferSize log can't flush to file, we should call `LogManager.shutdown()` when shutdown process ensures log flush to file.
https://github.com/apache/incubator-hugegraph/blob/8abf3d5524de3d3d5a97bbdce460a75652792b0d/hugegraph-dist/src/main/resources/log4j2.xml#L56-L58

### Modifications

* Shutdown log when shutdown process
* Improve `RIGHT_MARGIN` and `SOFT_MARGINS`


imbajin (approved):
lgtm

javeme (approved):

PR #2033, opened 2022-11-24
fix outdated manifest version & use project.version instead

z7658329:


hugegraph-api/pom.xml
```diff
@@ -186,7 +186,7 @@
                         </manifest>
                         <manifestEntries>
                             <!-- TODO: update it -->
-                            <Implementation-Version>0.69.0.0</Implementation-Version>
+                            <Implementation-Version>${project.version}</Implementation-Version>
```
imbajin, line 189: seems we should keep it now, it's APIversion

hugegraph-api/src/main/java/org/apache/hugegraph/version/ApiVersion.java
```diff
@@ -122,11 +122,11 @@ public final class ApiVersion {
...
      * TODO: what shall we set for this version? (consider the basic compatibility)
      */
-    public static final Version VERSION = Version.of(ApiVersion.class, "0.69");
+    public static final Version VERSION = Version.of(ApiVersion.class, "1.0.0");
 
     public static void check() {
         // Check version of hugegraph-core. Firstly do check from version 0.3
         // TODO: what shall we set for this version? (consider the basic compatibility)
-        VersionUtil.check(CoreVersion.VERSION, "0.13", "0.14", CoreVersion.NAME);
+        VersionUtil.check(CoreVersion.VERSION, "1.0.0", "1.0.1", CoreVersion.NAME);
```
imbajin, line 130: too strict

hugegraph-core/pom.xml
```diff
@@ -325,7 +325,7 @@
                             </addDefaultSpecificationEntries>
                         </manifest>
                         <manifestEntries>
-                            <Implementation-Version>0.13.0.0</Implementation-Version>
+                            <Implementation-Version>${project.version}</Implementation-Version>
```
imbajin, line 328: it's right here

hugegraph-core/src/main/java/org/apache/hugegraph/version/CoreVersion.java
```diff
@@ -41,6 +41,6 @@ public class CoreVersion {
     public static void check() {
         // Check version of hugegraph-common
         // TODO: why shall we check it? Update it if need
-        VersionUtil.check(CommonVersion.VERSION, "2.1.0", "2.2.0", CommonVersion.NAME);
+        VersionUtil.check(CommonVersion.VERSION, "1.0.0", "1.0.1", CommonVersion.NAME);
```
imbajin, line 44: maybe 1.0~1.2 first

hugegraph-api/pom.xml
z7658329: @javeme  if we should keep all version to 1.0.0 ?

hugegraph-core/src/main/java/org/apache/hugegraph/version/CoreVersion.java
z7658329: > 

i will fix it, but ApiVersion should discuss

hugegraph-api/src/main/java/org/apache/hugegraph/version/ApiVersion.java
z7658329: keep 1.0~1.2 first ? @imbajin

hugegraph-core/src/main/java/org/apache/hugegraph/version/CoreVersion.java
javeme: the right version is inclusive, `[1.0, 1.1)` is ok

hugegraph-api/pom.xml
javeme: I think we need to keep the api history

hugegraph-api/src/main/java/org/apache/hugegraph/version/ApiVersion.java
```diff
@@ -122,11 +122,11 @@ public final class ApiVersion {
      * The second parameter of Version.of() is for IDE running without JAR
      * TODO: what shall we set for this version? (consider the basic compatibility)
      */
-    public static final Version VERSION = Version.of(ApiVersion.class, "0.69");
```
imbajin, line 125: here？ shoud keep origin

imbajin (approved):

imbajin (approved):

javeme (approved):

PR #2038, opened 2022-11-29
chore: update antrun plugin & fix building problem in windows

seagle-yuan:
update antrun plugin & fix windows os

### DIFF
--- a/hugegraph-dist/pom.xml
+++ b/hugegraph-dist/pom.xml
@@ -179,7 +179,8 @@
                         </goals>
                         <configuration>
                             <target>
-                                <tar destfile="${top.level.dir}\${final.name}.tar.gz" basedir="${top.level.dir}\${final.name}" compression="gzip"/>
+                                <tar
+                                    destfile="${top.level.dir}/${final.name}.tar.gz" basedir="${top.level.dir}/${final.name}" compression="gzip"/>
                             </target>
                         </configuration>
                     </execution>
--- a/hugegraph-dist/pom.xml
+++ b/hugegraph-dist/pom.xml
@@ -30,8 +30,7 @@
 
     <properties>
         <assembly.dir>${project.basedir}/src/assembly</assembly.dir>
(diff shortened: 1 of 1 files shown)


hugegraph-dist/pom.xml
```diff
@@ -178,22 +178,10 @@
...
-                                    echo
-                                </echo>
-                                <exec executable="${shell-executable}"
-                                      dir="${top.level.dir}" failonerror="true">
-                                    <arg line="./dist.sh"/>
-                                </exec>
-                            </tasks>
+                            <target>
+                                <tar
+                                    destfile="${top.level.dir}/${final.name}.tar.gz" basedir="${top.level.dir}/${final.name}" compression="gzip"/>
```
imbajin, line 183: use one line is fine, and do u test it in local env?

seagle-yuan: test on linux and windows. & compare with old and new version .

hugegraph-dist/pom.xml
```diff
@@ -178,22 +178,9 @@
...
-                                    ${top.level.dir}/${final.name}.tar.gz"
-                                    echo
-                                </echo>
-                                <exec executable="${shell-executable}"
-                                      dir="${top.level.dir}" failonerror="true">
-                                    <arg line="./dist.sh"/>
-                                </exec>
-                            </tasks>
+                            <target>
+                                <tar destfile="${top.level.dir}/${final.name}.tar.gz" basedir="${top.level.dir}"  includes="${final.name}/*" compression="gzip"/>
```
javeme, line 182: can wrap line for every property?

seagle-yuan: like this
image

javeme: yes

imbajin (approved):
is it compatible in macOS & win & linux? 

which OS env do u test already?

zyxxoo (approved):

PR #2041, opened 2022-11-29
refact(dist): avoid script exit while restarting & clean script

imbajin:
- fix "restart" not work in `hugegraph` script
- clean shell script to avoid error in high-version bash/sh

refer:
- https://github.com/koalaman/shellcheck/wiki/SC2181
- https://github.com/koalaman/shellcheck/wiki/SC2004

and we have a lot of writing way may break the shell rules, and it will lead some unexpected error in high `bash/shell` version (like the problem when start with java11)


imbajin:
and I'll fix the script problem left in  #2024

javeme (approved):

zyxxoo (approved):

zyxxoo (approved):

PR #2051, opened 2022-12-13
chore: improve check-dependencies to avoid repeated action

simon824:
improve check-dependencies to avoid repeated action
image


imbajin (approved):
good

zyxxoo (approved):

PR #2064, opened 2022-12-31
fix flattened problem when clean phase include 

z7658329:
本地运行报错：clean package install -Dmaven.test.skip=true -U -f pom.xml

image


javeme (approved):

imbajin (changes requested):
wait for reproduce it

imbajin:
> 本地运行报错：mvn clean package install -Dmaven.test.skip=true -U -f pom.xml

seems we shouldn't use `package` and `install` in same cmd (only use one)

PR #2069, opened 2023-01-04
fix: script no permission

zyxxoo:


hugegraph-dist/pom.xml
```diff
@@ -179,10 +179,11 @@
                         </goals>
                         <configuration>
                             <target>
-                                <tar destfile="${final.destfile}"
-                                     basedir="${top.level.dir}"  
-                                     includes="${final.name}/**"
-                                     compression="gzip"/>
+                                <tar destfile="${final.destfile}" compression="gzip">
+                                    <tarfileset dir="${top.level.dir}/" filemode="755">
```
imbajin, line 183: we do need a test to ensure these problems

imbajin (approved):

hugegraph-dist/pom.xml
zyxxoo: I have tested, the permission problem is fixed

javeme (approved):

PR #2108, opened 2023-02-09
chore: Just check dependencies' license with apache action

JackyYangPassion:


zyxxoo (dismissed):

javeme (dismissed):

imbajin (changes requested):
we should handle or know how to fix the problem before merge it

image

imbajin:
and also fix dependency-review by github action

image

.github/workflows/licence-checker.yml
```diff
@@ -22,6 +22,27 @@ jobs:
           token: ${{ github.token }}
           config: .licenserc.yaml
 
+  dependency-license:
```
javeme, line 25: can we rename to `check-dependency-license` and also rename `check-license` to `check-source-license`

javeme:
Is it still in progress?

JackyYangPassion:
Close the PR first and then submit it if there is progress later.

PR #2134, opened 2023-02-27
chore: cmn algorithm optimization

zyxxoo:


hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphIndexTransaction.java
```diff
@@ -1120,21 +1120,25 @@ private static <T> boolean cmn(List<T> all, int m, int n,
...
         if (m == n) {
             result.addAll(all.subList(current, all.size()));
             n = 0;
         }
         if (n == 0) {
             // All n items are selected
-            return callback.apply(result);
+            Boolean apply = callback.apply(result);
+            while (index < result.size()) {
+                result.remove(index);
```
javeme, line 1132: when to cause this behavior?

javeme (approved):

imbajin (approved):

PR #2139, opened 2023-03-01
add github token for license check comment

JackyYangPassion:
the token that license eye uses when it needs to comment on the pull request. Set to empty ("") to disable commenting on pull request. The default value is ${{ github.token }}

### DIFF
--- a/.github/workflows/licence-checker.yml
+++ b/.github/workflows/licence-checker.yml
@@ -19,4 +19,16 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           log: info
+          token: ${{ github.token }}
           config: .licenserc.yaml
+
+      - uses: actions/setup-java@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/licence-checker.yml
```diff
@@ -19,4 +19,16 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           log: info
+          token: ${{ github.token }}
```
imbajin, line 22: Good, but how does it take effect in #2130? (related to the submit person?)

imbajin (approved):

javeme (approved):

imbajin:
@JackyYangPassion seems the problem still exist after this PR merged, and found another problem (validation not right)

## 1. we shouldn't allow `Copyright` appear in the header

image

## 2. lack of the header

PR #2151, opened 2023-03-09
optimize: remove lock of globalMasterInfo

zyxxoo:


hugegraph-core/src/main/java/org/apache/hugegraph/masterelection/GlobalMasterInfo.java
```diff
@@ -19,26 +19,20 @@
...
 
-    public synchronized void set(boolean isMaster, String url) {
-        this.isMaster = isMaster;
-        this.url = url;
-    }
-
-    public synchronized boolean isMaster() {
-        return this.isMaster;
+    public void info(boolean isMaster, String url) {
+        this.info = new Info(isMaster, url);
```
imbajin, line 31: `volatile Info info = new Info(isMaster, url);` 

Does `volatile` guarantee the atomic integrity of `Info` initialization? Or shall we consider to use `final` to guarantee to complete initialization like:

`final Info info = new Info(isMaster, url);`

zyxxoo: info define with volatile, so as my think, the read  operator happend-before write operator,
we read the info must is init finished;

hugegraph-core/src/main/java/org/apache/hugegraph/masterelection/GlobalMasterInfo.java
```diff
@@ -19,26 +19,20 @@
...
     }
 
-    public synchronized void set(boolean isMaster, String url) {
-        this.isMaster = isMaster;
-        this.url = url;
-    }
-
-    public synchronized boolean isMaster() {
-        return this.isMaster;
+    public void info(boolean isMaster, String url) {
```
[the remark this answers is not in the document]
zyxxoo: maybe rename node?
we need two operation, get, and set, so name node(isMaster, url), node()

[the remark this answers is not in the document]
zyxxoo: if write operation use update, the read operation keep name info?

hugegraph-core/src/main/java/org/apache/hugegraph/masterelection/GlobalMasterInfo.java
imbajin: > info define with volatile, so as my think, the write operator happens-before read operator, so if write operator happend first, the read operator must after the write finish

```java
        public Info(boolean isMaster, String url) {
            this.isMaster = isMaster; // ①
            this.url = url; // ②
        }
```

`new Info` include ① & ② step, when ① is finished but ② is not, will we see an inconsistent assignments?

like `Info info = new Info(true, "a")`, will we get `Info(true, "")` or `Info(false, "a")`?

zyxxoo: the case don't happend, because method parameter be bound;
when call 1 step, the isMaster and url parameter never be modify by external logic, because the boolean is copy the origin data, and string is final;

this is a interesting question:
``` 
      boolean isMaster = global.isMaster(); // 1
      String url = global.url(); //2
      this.info = new Info(isMaster, url); // 3
```
if the code run in more than one thread, the timing might look like this:
thread1: run 1 step; thread2: modify global.master; thread2: modify global.url; thread1: run 2; this case will have some concurrency problem; 
but in our case, only one thread modify global, so here is no problem

zyxxoo: here is a “copy on write” thought;
single thread copy the data and modify it, then multi-thread read data don't consider lock

hugegraph-core/src/main/java/org/apache/hugegraph/masterelection/GlobalMasterInfo.java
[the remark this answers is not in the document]
javeme: "info" may be an too generic word, I think `nodeInfo(isMaster, url)` is good, and also rename class Info to NodeInfo

hugegraph-core/src/main/java/org/apache/hugegraph/masterelection/StandardStateMachineCallback.java
```diff
@@ -94,12 +94,12 @@ public void error(StateMachineContext context, Throwable e) {
     public void initGlobalMasterInfo(StateMachineContext context) {
         StateMachineContext.MasterServerInfo master = context.master();
         if (master == null) {
-            this.globalMasterInfo.set(false, null);
+            this.globalMasterInfo.info(false, null);
```
javeme, line 97: keep consistent with the GlobalMasterInfo() let url = ""?

hugegraph-core/src/main/java/org/apache/hugegraph/masterelection/StandardStateMachineCallback.java
```diff
@@ -94,12 +94,12 @@ public void error(StateMachineContext context, Throwable e) {
...
             return;
         }
 
         boolean isMaster = Objects.equals(context.node(), master.node());
         String url = master.url();
-        this.globalMasterInfo.set(isMaster, url);
+        this.globalMasterInfo.info(isMaster, url);
```
javeme, line 103: how frequently the initGlobalMasterInfo() to be called?

zyxxoo: frequently = 1 / heatbeat

javeme (approved):

hugegraph-core/src/main/java/org/apache/hugegraph/masterelection/GlobalMasterInfo.java
```diff
@@ -19,26 +19,22 @@
...
     }
 
-    public synchronized void set(boolean isMaster, String url) {
-        this.isMaster = isMaster;
-        this.url = url;
-    }
-
-    public synchronized boolean isMaster() {
-        return this.isMaster;
+    public final void nodeInfo(boolean isMaster, String url) {
```
imbajin, line 30: nice but why do we add final for the method？（for design？）

zyxxoo: > nice but why do we add final for the method？（for design？）

final method no usefull，only express dont modify

javeme: looks useful, the subclasses cannot override it

imbajin (approved):

PR #2207, opened 2023-05-08
Modify function name

KeeProMise:
When I was debugging the code, I found that the name of the function filterExpiredResultFromFromBackend in the GraphTransaction.java file is problematic. Change the name.


javeme (approved):
LGTM

PR #2260, opened 2023-07-28
feat(cassandra): adapt hugegraph-cassandra from 3.11.12 to 4.0.10

lzyxx77:
we should check the compatibility in:

- [ ] Cassandra Server V3.11
- [ ] Cassandra Server V4.0

 Expect to be able to connect the two versions properly?


imbajin (commented):
refer [here](Expect to be able to connect the two versions properly?) to fix the license check problem

lzyxx77:
Anyone who interested in Cassandra 4 can take a look at this discussion #2261

Issue #2265, opened 2023-08-03
[Feature] integrate pd and store components into hugegraph


VGalaxies:
### Feature Description (功能描述)

# Background

Currently, the architecture of the community version of HugeGraph is still in version 1.0. In the internal version 2.0 of HugeGraph, we have the following design goals:

- Support trillion-scale data storage and tens of thousands of graph storages.
- Support multi-active, high availability, dynamic scalability, and automated operation and maintenance.
- Maximize read and write performance.

Based on these goals, we have designed a distributed architecture that supports graph data partitioning and multiple replicas, and separates storage from computation for flexible scaling.

In version 2.0, in addition to the hugegraph-server, we also introduce two additional components: hugegraph-pd and hugegraph-store

![image](https://github.com/apache/incubator-hugegraph/assets/79143929/e298925a-114d-4eab-9f4d-d5ab91e7522f)

The responsibilities of these two components are as follows:

- hugegraph-pd: pd stands for place driver, which can be simply understood as a meta server responsible for service discovery, partition information storage, and node scheduling.
- hugegraph-store: as a new built-in storage backend, it uses RocksDB as the distributed backend storage foundation.

We will gradually merge the internal version 2.0 of HugeGraph into the community version. Therefore, the first step is to integrate the hugegraph-pd and hugegraph-store modules into this repository.

# Tasks

- [ ] Adjust the project structure of this repository to include three sub-modules: hugegraph-server, hugegraph-pd, hugegraph-store at the root level.
- [ ] Merge the internal version of pd.
- [ ] Merge the internal version of store.
- [ ] Configuration adjustments (including merging dependencies, unifying version numbers, Maven compilation adjustments, etc.).
- [ ] Perform code adaptation tasks (hugegraph-server needs to introduce modules to connect with hugegraph-pd and hugegraph-store, and make adjustments to other modules, etc.).
- [ ] ...

Pengzna:
please add this [sub-task](https://github.com/apache/incubator-hugegraph/issues/2482) to this issue, thx~

VGalaxies:
close by #2575, PD and Store components have been integrated into hugegraph and will continue to evolve in future versions...

PR #2275, opened 2023-08-08
refact: adjust pom structure for directly compiling from root

heiyan-2020:
## Purpose of the PR

subtask of #2265 

Adjust the pom structure in order to directly build from root.

## Main Changes

- Remove some unused dependencies in `hugegraph-pd` module.
- Modify pom files for compiling and building.
- Manually compile the internal version of `hugegraph-core` and add it to the pom of store module as system dependencies.(temporal solution)
- The dependencies of packages imported through system dependencies will not be passed down, so we need to manually add some dependencies in `hugegraph-core` to the pom file of `hg-store-core`.

## Verifying these changes

- [x]  Nope
- [ ]  Modify configurations
- [ ]  The public API
- [ ]  Other affects

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`


hugegraph-store/hg-store-dist/src/assembly/ext-lib/hugegraph-core-1.5.0.pom
```diff
@@ -0,0 +1,441 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  License for the specific language governing permissions and limitations
+  under the License.
+  -->
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
```
VGalaxies, line 21: In accordance with the hierarchical structure, the parent of `hugegraph-core` should be `hugegraph-server`. However, due to existing discrepancies between the community version of `hugegraph-server` and the internal version (such as `revision`), a temporary adjustment has been made for the purpose of building.

hugegraph-store/pom.xml
```diff
@@ -49,6 +49,7 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <log4j2.version>2.15.0</log4j2.version>
+        <top.level.dir>${project.basedir}/..</top.level.dir>
```
imbajin, line 52: WARN: ensure we don't have multi `top.level.dir` in project (if we have & need use it, add a comment for it)

hugegraph-store/hg-store-core/pom.xml
```diff
@@ -121,9 +121,48 @@
             <version>${revision}</version>
             <scope>system</scope>
             <systemPath>
-                ${project.basedir}/../hg-store-dist/src/assembly/ext-lib/hugegraph-core-1.5.0.jar
+                ${top.level.dir}/hg-store-dist/src/assembly/ext-lib/hugegraph-core-1.5.0.jar
             </systemPath>
         </dependency>
+        <!-- tinkerpop -->
+        <dependency>
```
imbajin, line 128: why we need add tinkerpop dependency here?

@VGalaxies any context?

hugegraph-store/hg-store-test/pom.xml
```diff
@@ -210,7 +210,15 @@
             <version>${revision}</version>
             <scope>compile</scope>
         </dependency>
-
+        <dependency>
```
imbajin, line 213:
```suggestion
        <!-- TODO: we need replace it with the apache version later -->
        <dependency>
```

hugegraph-store/hg-store-core/pom.xml
```diff
@@ -121,9 +121,48 @@
             <version>${revision}</version>
             <scope>system</scope>
             <systemPath>
-                ${project.basedir}/../hg-store-dist/src/assembly/ext-lib/hugegraph-core-1.5.0.jar
+                ${top.level.dir}/hg-store-dist/src/assembly/ext-lib/hugegraph-core-1.5.0.jar
```
imbajin, line 124: add a TODO comment for it  

and unify the param
image

hugegraph-store/hg-store-core/pom.xml
heiyan-2020: As mentioned in the last item of "Main change", tinkerpop is depended on by `hg-store-core`, but in internal version the dependency is imported through `hugegraph-core`. The current temporary solution is that importing `hugegraph-core` jar package through system dependency, whose dependencies cannot be passed on, so the dependencies need to be added again in `hg-store-core`

VGalaxies: > why we need add tinkerpop dependency here?
> 
> @VGalaxies any context?

As stated by @heiyan-2020, assuming there are projects A (`hg-store-core`) and B (`hugegraph-core`), where project A depends on project B, and project B is a system dependency, the dependencies within project B will not be automatically inherited by project A.

hugegraph-store/pom.xml
VGalaxies: Are you suggesting that we should only retain the `top.level.dir` property in the root pom? 

Additionally, provide all the occurrences of it:

image

hugegraph-store/hg-store-core/pom.xml
imbajin: > As mentioned in the last item of "Main change", tinkerpop is depended on by `hg-store-core`, and in internal version the dependency is passed on through `hugegraph-core` which imported as usal maven dependency. But the current temporary solution is that importing `hugegraph-core` jar package through system dependency. In this way, the dependencies in `hugegraph-core` cannot be passed on, so the dependencies need to be added again in `hg-store-core`

@heiyan-2020 get it, but we need add `comment/TODO` mark here? (maybe in next PR)

hugegraph-store/pom.xml
imbajin: THX, just for warning

hugegraph-store/hg-store-dist/src/assembly/ext-lib/hugegraph-core-1.5.0.pom
imbajin: > In accordance with the hierarchical structure, the parent of `hugegraph-core` should be `hugegraph-server`. However, due to existing discrepancies between the community version of `hugegraph-server` and the internal version (such as `revision`), a temporary adjustment has been made for the purpose of building.

Update: miss the msg before, and we choose use a new version for all modules

PR #1880, opened 2022-05-13
feat: support swagger UI to viewing API

wangyao2016:
Detail descrition for this feature please check [#1879](https://github.com/apache/incubator-hugegraph/issues/1879)

We can view swagger UI by `/swagger-ui/index.html` like:
`http://127.0.0.1:8080/swagger-ui/index.html`

image

Issue #1408, opened 2021-03-30
源码编译打包错误


xxyp:
从git下载HugeGraph master分支执行 mvn package -DskipTests 报错,请问是什么原因呢
jdk 1.8 
Maven 3.39

[INFO] hugegraph-palo ..................................... SUCCESS [  0.638 s]
[INFO] hugegraph-hbase .................................... SUCCESS [  1.312 s]
[INFO] hugegraph-postgresql ............................... SUCCESS [  0.529 s]
[INFO] hugegraph-dist ..................................... FAILURE [ 11.909 s]
[INFO] hugegraph-example .................................. SKIPPED
[INFO] hugegraph-test ..................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 32.154 s
[INFO] Finished at: 2021-03-30T21:11:58+08:00
[INFO] Final Memory: 81M/837M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.3:run (default) on project hugegraph-dist: An Ant BuildException has occured: Execute failed: java.io.IOException: Cannot run program "bash": CreateProcess error=2, 系统找不到指定的文件。 -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :hugegraph-dist

javeme:
@xxyp 缺少bash命令，是否非非linux环境中执行的？
可以搜索下问题："Cannot run program "bash"，网上有不少解决方法的。

imbajin:
暂时不支持 windows 上**直接**打包, 相关打包脚本中的 `path` 解析会报错, 当然欢迎你尝试调整后提交 PR

如果只有 linux 环境, 推荐你使用win10内置的 [WSL2(Windows Subsystem for Linux)](https://docs.microsoft.com/zh-cn/windows/wsl/install-win10) 转换到 Linux 环境使用

Issue #173, opened 2018-11-06
Windows系统mvn package -DskipTests失败


htpbackup:
### Expected behavior 期望表现
mvn package -DskipTests

### Actual behavior 实际表现
```bash
[INFO] --- maven-antrun-plugin:1.3:run (default) @ hugegraph-dist ---
[INFO] Executing tasks
     [exec] hugegraph-0.7.4/
     [exec] hugegraph-0.7.4/bin/
     [exec] hugegraph-0.7.4/bin/checksocket.sh
     [exec] hugegraph-0.7.4/bin/dump-conf.sh
     [exec] hugegraph-0.7.4/bin/dump-store.sh
     [exec] hugegraph-0.7.4/bin/gremlin-console.sh
     [exec] hugegraph-0.7.4/bin/hugegraph-server.sh
     [exec] hugegraph-0.7.4/bin/init-store.sh
     [exec] hugegraph-0.7.4/bin/monitor-hugegraph.sh
     [exec] hugegraph-0.7.4/bin/start-hugegraph.sh
     [exec] hugegraph-0.7.4/bin/start-monitor.sh
     [exec] tar (child): Cannot execute remote shell: No such file or directory
     [exec] tar (child): D\:workspacehugegraphhugegraphhugegraph-dist/../hugegraph-0.7.4.tar.gz: Cannot open: Input/output error
     [exec] hugegraph-0.7.4/bin/stop-hugegraph.sh
     [exec] tar (child): Error is not recoverable: exiting now
     [exec] hugegraph-0.7.4/bin/stop-monitor.sh
     [exec] tar: D\:workspacehugegraphhugegraphhugegraph-dist/../hugegraph-0.7.4.tar.gz: Cannot write: Broken pipe
     [exec] hugegraph-0.7.4/bin/util.sh
     [exec] hugegraph-0.7.4/conf/
     [exec] hugegraph-0.7.4/conf/gremlin-driver-settings.yaml
     [exec] tar: Child returned status 2
     [exec] tar: Error is not recoverable: exiting now
     [exec] hugegraph-0.7.4/conf/gremlin-server.yaml
     [exec] hugegraph-0.7.4/conf/hugegraph.properties
     [exec] hugegraph-0.7.4/conf/log4j2.xml
     [exec] hugegraph-0.7.4/conf/remote-objects.yaml
     [exec] hugegraph-0.7.4/conf/remote.yaml
     [exec] hugegraph-0.7.4/conf/rest-server.properties
[...]

PR #2277, opened 2023-08-12
chore(dist): replace wget to curl to download swagger-ui 

VGalaxies:
## Purpose of the PR

close https://github.com/apache/incubator-hugegraph/issues/2223

related issue or PR:

- #1880 
- https://github.com/apache/incubator-hugegraph/pull/2262#discussion_r1291153745
- https://github.com/apache/incubator-hugegraph/issues/2011
- https://github.com/apache/incubator-hugegraph/issues/2088
- #1408
- #173
- #406
- https://github.com/apache/incubator-hugegraph/issues/2114#issuecomment-1453091210

## Main Changes

1. replace `wget` by `curl` when downloading `swagger-ui`
2. silence the output of `curl` and `tar` commands
3. reuse the existing `v4.15.5.tar.gz` before downloading
4. avoid downloading `swagger-ui` in non-Linux platforms to prevent build failures (there might be a cross-platform build approach available 🤔)
5. wrapp the script content within `<![CDATA[ ... ]]>` blocks ensures that the script retains its original format when generating the `dist.sh` script (also suppresses automatic indentation)
6. remove intermediate files at the end of the script

**An alternative approach**, during the generation of the `dist.sh` script, only the `${final.name}` property from the build process is utilized. It might be possible to separately store a `dist.sh` script within hugegraph-dist, then use `sed` during the build process to replace the value of `${final.name}`, **thereby avoiding the need to embed script content within the pom file**.

## Verifying these changes

    - xxx

- [ ]  Nope
- [x]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`


hugegraph-dist/pom.xml
```diff
@@ -179,21 +179,42 @@
...
-                                    cp -r swagger-ui-4.15.5/dist ../${final.name}/swagger-ui
-                                </echo>
-                                <exec executable="${shell-executable}" dir="${project.basedir}" failonerror="true">
-                                    <arg line="./dist.sh"/>
-                                </exec>
-                            </tasks>
+                            <target>
+                                <taskdef resource="net/sf/antcontrib/antcontrib.properties"
+                                         classpathref="maven.plugin.classpath"/>
+                                <if>
```
imbajin, line 185: maybe we could also check if curl exist in os env？ skip the swagger building if network/curl broken（do we need it？）

In general，the building shouldn't be bounded with outside network？

VGalaxies: @imbajin `curl` has been checked in [`3ae4aac` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/3ae4aacd33bcd9ee7d3913fd88683201ea4272f1#diff-5cb24cae8e78777b5af60640cc24189b37905e8d9229fcdc25298ce185caca88R190)

If there is no outside network, how can maven download dependencies from remote repositories during the building process?

imbajin: - I mean check but not exit if miss it（skip instead）
- like many machine could use inner “self-host repository”（like gitlab VS github） to download deps，and next time they will reuse it from local .m2 repo

VGalaxies: @imbajin Using this snippet without having curl installed:

```bash
curl --version >/dev/null 2>&1 || (echo 'WARN: Please install `curl` first' && exit)
```

Does the following result meet your expectations?

image

Similar outcomes can be expected in the case of network abnormalities.

imbajin: > Does the following result meet your expectations?
> 
> image
> 
> Similar outcomes can be expected in the case of network abnormalities.

LGTM, the warb message could  be "ERROR: Please install `curl` first if you need `swagger-ui`"

VGalaxies: > > Does the following result meet your expectations?
> > image
> > Similar outcomes can be expected in the case of network abnormalities.
> 
> LGTM, the warning message could be "ERROR: Please install `curl` first if you need `swagger-ui`"
> 
> BTW, could also skip download if network bad like `curl xxx || exit`?)

Sure, the corresponding modifications can be found in [`0a9215f` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/0a9215f5ccc83dcacaf80142c0910a58039323be)

hugegraph-dist/pom.xml
imbajin, line 1: image

Agree to extract as a **separate file** as suggested by you

VGalaxies:
also added maven-enforcer-plugin to check java and maven version for https://github.com/apache/incubator-hugegraph/issues/2114#issuecomment-1453091210

hugegraph-dist/pom.xml
```diff
@@ -212,6 +229,19 @@
...
+                    <dependency>
+                        <groupId>ant-contrib</groupId>
+                        <artifactId>ant-contrib</artifactId>
+                        <version>1.0b3</version>
+                        <exclusions>
+                            <exclusion>
+                                <groupId>ant</groupId>
+                                <artifactId>ant</artifactId>
+                            </exclusion>
+                        </exclusions>
```
imbajin, line 242: we still need it?

VGalaxies: To avoid downloading swagger-ui on non-Linux platforms, the utilization of the `if` tag from `ant-contrib` is demonstrated here.

An alternative [approach](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#OS) could involve controlling profile activation through the `os` attribute. **However, it seems tricky to activate the same profile for both Mac and Linux simultaneously.**

**In order to reuse the configuration of the `maven-antrun-plugin`**, I'm considering setting up the `download-swagger-ui` configuration in the root pom's `pluginManagement`. In This way, it can be reused in `hugegraph-dist`, see [`1534605` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/1534605ae8ab64a2d601fd3930ec19967bf8cd27).

VGalaxies: @imbajin By the way, what does the subtask imply here? Is there a need to support downloading swagger-ui on Windows in THIS PR?

image

imbajin: no need，fine to skip download in windows

imbajin: > To avoid downloading swagger-ui on non-Linux platforms, the utilization of the `if` tag from `ant-contrib` is demonstrated here.
> 
> An alternative [approach](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#OS) could involve controlling profile activation through the `os` attribute. **However, it seems tricky to activate the same profile for both Mac and Linux simultaneously.**
> 
> **In order to reuse the configuration of the `maven-antrun-plugin`**, I'm considering setting up the `download-swagger-ui` configuration in the root pom's `pluginManagement`. In This way, it can be reused in `hugegraph-dist`, see [`1534605` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/1534605ae8ab64a2d601fd3930ec19967bf8cd27).
> 

nice refactor （try it）

VGalaxies: > > To avoid downloading swagger-ui on non-Linux platforms, the utilization of the `if` tag from `ant-contrib` is demonstrated here.
> > An alternative [approach](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#OS) could involve controlling profile activation through the `os` attribute. **However, it seems tricky to activate the same profile for both Mac and Linux simultaneously.**
> > **In order to reuse the configuration of the `maven-antrun-plugin`**, I'm considering setting up the `download-swagger-ui` configuration in the root pom's `pluginManagement`. In This way, it can be reused in `hugegraph-dist`, see [`1534605` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/1534605ae8ab64a2d601fd3930ec19967bf8cd27).
> 
> nice refactor （try it）

done in [`1534605` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/1534605ae8ab64a2d601fd3930ec19967bf8cd27)

imbajin (dismissed):

hugegraph-dist/pom.xml
```diff
@@ -215,4 +193,40 @@
...
+        <profile>
+            <id>unix-download-swagger-ui</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                    </plugin>
+                </plugins>
+            </build>
```
z7658329, line 207: can extract to root pom file?

VGalaxies: @z7658329 done in [`509a4e2` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/509a4e27f8f58807be4c36019b7eb1330b01d63a)

VGalaxies:
The order of executions related to packaging should be `assembly-hugegraph` -> `install-swagger-ui` (copy and remove) -> `tar-package`. The `download-swagger-ui` task only needs to be completed before `install-swagger-ui` (currently bound to the prepare-package phase).

To control the execution order of different goals within the same phase, consider declaring the above executions in sequence within the `pluginManagement` of the root pom. Also use conditions based on `os` and `file` to ensure that these goals are executed only within the Unix or macOS environment, specifically within the `hugegraph-dist` module.

pom.xml
```diff
@@ -453,6 +453,30 @@
...
+                                <goal>single</goal>
+                            </goals>
+                            <configuration>
+                                <attach>false</attach>
+                                <appendAssemblyId>false</appendAssemblyId>
+                                <outputDirectory>${top.level.dir}
+                                </outputDirectory>
+                                <descriptor>
+                                    ${assembly.descriptor.dir}/assembly.xml
+                                </descriptor>
```
VGalaxies, line 474: As Maven operates within the context of the current module when executing goals, the paths specified here are relative to the `hugegraph-dist` directory and can correctly resolve the `assembly.descriptor.dir` property declared within `hugegraph-dist`.

pom.xml
```diff
@@ -676,8 +747,118 @@
...
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                    </plugin>
+                </plugins>
+            </build>
+            <activation>
+                <os>
+                    <family>unix</family>
+                    <name>Linux</name>
```
imbajin, line 803: `Unix` VS `MacOS` seems not right?  (better to use `Linux` directly)

pom.xml
```diff
@@ -676,8 +747,118 @@
...
+                                    </rules>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <id>assembly-hugegraph</id>
```
imbajin, line 777: after we move the profile, what's the proper building command?

Before:
`mvn clean install -Papache-release -DskipTests`  for building `apache-release` source/binary package

Now: ?

pom.xml
```diff
@@ -453,6 +453,77 @@
...
+                    </executions>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-antrun-plugin</artifactId>
+                    <version>1.8</version>
+                    <executions>
+                        <execution>
+                            <id>download-swagger-ui</id>
+                            <phase>prepare-package</phase>
```
imbajin, line 487: seems this step will not be inherited by other modules?

pom.xml
VGalaxies: looks like the `family` field can only be `unix` or `mac` 🤔

image

hugegraph-dist/pom.xml
```diff
@@ -168,51 +144,162 @@
...
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+        </profile>
```
VGalaxies, line 303: After introducing pd and store (#2265), the relevant package goals could be considered for reuse.

pom.xml
VGalaxies: When a profile is activated through the **implicit profile activation** mechanism, the **explicit profile activation mechanism** within the same pom becomes ineffective.

So it's necessary to explicitly specify the profiles:

```bash
mvn clean install -Papache-release,assembly-hugegraph,tar-package -DskipTests
```

> ref: https://maven.apache.org/guides/introduction/introduction-to-profiles.html#explicit-profile-activation

VGalaxies: > When a profile is activated through the **implicit profile activation** mechanism, the **explicit profile activation mechanism** within the same pom becomes ineffective.
> 
> So it's necessary to explicitly specify the profiles:
> 
> ```shell
> mvn clean install -Papache-release,assembly-hugegraph,tar-package -DskipTests
> ```
> 
> > ref: [maven.apache.org/guides/introduction/introduction-to-profiles.html#explicit-profile-activation](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#explicit-profile-activation)

The relevant scripts and documentation need to be updated later.

hugegraph-dist/pom.xml
imbajin: > can extract to root pom file?

better to keep them in dist module now

hugegraph-dist/pom.xml
imbajin: > After introducing pd and store (#2265), the relevant package goals could be considered for reuse.

@msgui could also take a look for this PR, and it will be helpful when u refactor the new struct

hugegraph-dist/pom.xml
```diff
@@ -168,51 +144,162 @@
...
-                                    </tarfileset>
-                                </tar>
-                            </target>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
         </plugins>
+
+        <pluginManagement>
```
imbajin, line 149: need `pluginManagement`?

VGalaxies: > need `pluginManagement`?
> 
> BTW, all the CI failed now 🔎

1. for reuse the configuration of the `maven-antrun-plugin`
2. because of https://github.com/apache/incubator-hugegraph/pull/2277#discussion_r1302502303, fix in [`880d283` (#2277)](https://github.com/apache/incubator-hugegraph/pull/2277/commits/880d28358c7c79e71caba2424b70696606ecc9e7)

VGalaxies:
Following this [discussion](https://github.com/apache/incubator-hugegraph/pull/2277#discussion_r1302507017):

Through researching relevant information, we found that the semantics of `activeByDefault` are somewhat counterintuitive. In a particular pom, if a profile is activated through:

1. Specifying it via the command line using `-Pxxx`
2. Defining it in the Maven settings using `<activeProfiles>`
3. Implicit profile activation, such as `jdk`, `os`, `property`, `file`

Then, within the **same** pom, a profile configured with `activeByDefault` set to true will NOT be activated.

To avoid explicitly specifying profiles suppressed from activation due to the aforementioned mechanisms via `-Pxxx` in the command line, we referred to the approach outlined in the official documentation:

```xml
<profiles>
  <profile>
    <id>xxx</id>
    <activation>
      <property>
        <name>!skip-xxx</name>
      </property>
    </activation>
    ...
  </profile>
</profiles>
```

Using the property mechanism within implicit profile activation to activate profiles configured with `activeByDefault` set to true.

Additionally, it's important to minimize occurrences of identical property names across different profiles. For example:

img_v2_97f89063-b734-4fa1-932f-0bf74c69867g

In a scenario where both profiles are activated, the value of `env` is `test` (`dev` is loaded and then overridden). Explicitly specifying `-Pdev` could result in the value of `env` becoming `dev`, which might introduce potential issues.

References:

1. [Maven Introduction to Profiles - Property Activation](https://maven.apache.org/guides/introduction/introduction-to-profiles.html#Property)
2. [Stack Overflow - How to Keep Maven Profiles which are activeByDefault active](https://stackoverflow.com/questions/5309379/how-to-keep-maven-profiles-which-are-activebydefault-active-even-if-another-prof)

VGalaxies:
**TODO:** We have observed that when the `id` of the `core-test` profile under the root pom is activated using the aforementioned approach, it leads to failures while running the majority of backend `api-test` cases. For more details, please refer to: [GitHub Actions Run](https://github.com/apache/incubator-hugegraph/actions/runs/5951571368?pr=2277).

imbajin (approved):
LGTM

z7658329 (approved):

PR #2300, opened 2023-08-27
feat(cassandra): adapt cassandra from 3.11.12 to 4.0.10

lzyxx77:
we should check the compatibility in:

- [x] Cassandra Server V3.11
- [x] Cassandra Server V4.0

Expect to be able to connect the two versions properly?


javeme (approved):
LGTM

hugegraph-dist/src/assembly/travis/install-cassandra.sh
```diff
@@ -19,7 +19,7 @@ set -ev
 
 TRAVIS_DIR=`dirname $0`
 CASS_DOWNLOAD_ADDRESS="http://archive.apache.org/dist/cassandra"
-CASS_VERSION="3.10"
+CASS_VERSION="4.0.10"
```
imbajin, line 22: After this PR merged, maybe we could remove useless bundled `Java8` env in ci (HBase also need update the version)
image


Hbase Refer: (so we need upgrade to HBase 2.3+ in ci)  
img_v2_5fbd54e3-6118-4d32-bfd7-6bd1dbf0b77g

imbajin (approved):
Thanks, and could enhance Hbase version in next PR maybe (before next release V1.2)
https://github.com/apache/incubator-hugegraph/blob/e90489f3616a4c8572a199c7b7e8ae356c563811/.github/workflows/ci.yml#L53

PR #2308, opened 2023-09-09
fix: base-ref/head-ref missed in dependency-review on master

msgui:
## Purpose of the PR

fix: base-ref/head-ref miss in dependency-review on pushing to master

## Verifying these changes

    - xxx

- [x]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/.github/workflows/check-dependencies.yml
+++ b/.github/workflows/check-dependencies.yml
@@ -3,7 +3,6 @@ name: "3rd-party"
 on:
   push:
     branches:
-      - master
       - /^release-.*$/
   pull_request:
   


.github/workflows/check-dependencies.yml
```diff
@@ -3,7 +3,6 @@ name: "3rd-party"
 on:
   push:
     branches:
-      - master
       - /^release-.*$/
   pull_request:
```
msgui, line 7: Due to automatic assignment of base-ref/head-ref on pull_request, their absence is noticed when pushing to master.
[Doc link 👇](https://github.com/actions/dependency-review-action#:~:text=base%2Dref/,none)

image

.github/workflows/check-dependencies.yml
```diff
@@ -3,7 +3,6 @@ name: "3rd-party"
 on:
   push:
     branches:
-      - master
```
javeme, line 6: I'm not sure why need to remove it, since it is a triggering branch

javeme (commented):
Thanks for your contribution

imbajin (approved):
Also update the toolchain & other influenced repo, THX

.github/workflows/check-dependencies.yml
msgui: > I'm not sure why need to remove it, since it is a triggering branch

When pushing to the master base-ref/head-ref are not automatically assigned, leading to frequent errors during such pushes. The reason for removing them is that during PRs, we can directly check for the introduction of unauthorized 3rd-party libraries (exceptional libraries) through 'Dependency Review' and make the decision on whether to introduce them. There is no need to showcase this during master branch pushes.

javeme: Get it, Thanks for the detailed description

javeme (approved):
LGTM

PR #2332, opened 2023-10-22
fix(core): handle schema Cache expandCapacity concurrent problem

### LINKED ISSUE #2330: [Bug] Schema Cache expandCapacity concurrent issue
### Bug Type (问题类型)

other exception / error (其他异常报错)

### Before submit

### Environment (环境信息)

see the following error-log
```java
org.apache.hugegraph.exception.ServerException: Index 32 out of bounds for length 32
```

```java
    public void set(int key, V value) {
        if (key >= this.currentSize) {
            this.expandCapacity();
        }
        // this.array is not the one that has done expandCapacity!
        this.array[key] = value;
    }

    private synchronized void expandCapacity() {
        if (this.currentSize == this.maxSize) {
            return;
        }
        this.currentSize = Math.min(this.currentSize * 2, this.maxSize);
        Object[] newArray = new Object[this.currentSize];
        System.arraycopy(this.array, 0, newArray, 0, this.array.length);
        this.clear();
        this.array = newArray;
    }
```

_No response_

_No response_

conghuhu:
## Purpose of the PR

fix #2330 

## Verifying these changes

    - org.apache.hugegraph.unit.store.RamIntObjectMapTest

- [x]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [x]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/ram/IntObjectMap.java
+++ b/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/ram/IntObjectMap.java
@@ -29,8 +29,8 @@ public final class IntObjectMap<V> implements RamMap {
     private static final float DEFAULT_INITIAL_FACTOR = 0.25f;
 
     private final int maxSize;
-    private int currentSize;
-    private Object[] array;
+    private volatile int currentSize;
+    private volatile Object[] array;
 
     public IntObjectMap(int size) {
         this.maxSize = size;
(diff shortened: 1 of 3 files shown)


conghuhu:
@imbajin PTAL

hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/ram/IntObjectMap.java
```diff
@@ -29,8 +29,8 @@ public final class IntObjectMap<V> implements RamMap {
     private static final float DEFAULT_INITIAL_FACTOR = 0.25f;
 
     private final int maxSize;
-    private int currentSize;
-    private Object[] array;
+    private volatile int currentSize;
+    private volatile Object[] array;
```
imbajin, line 33: seems we just need mark it as `final` to ensure the init integrity? could try it to ensure
```suggestion
    private final Object[] array;
```

conghuhu: image
If it is changed to final, then the value cannot be assigned again

imbajin: > image If it is changed to final, then the value cannot be assigned again

emm, if we need to assign multiple times, it cannot be done like this, ignore it

imbajin:
Find another critical bug @conghuhu 

image

PR #2377, opened 2023-12-05
feat: add IntMapByDynamicHash

conghuhu:
## Purpose of the PR

Add IntMapByDynamicHash for high performance, lockless, concurrency secure hashmap.

### Dynamic Capcitity
image

### With Init Capcitity
image

    - xxx

- [ ]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
@@ -17,6 +17,9 @@
 
 package org.apache.hugegraph.util.collection;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
@@ -92,8 +95,11 @@ public class IntMapByDynamicHash implements IntMap {
         return (n < 0) ? 1 : (n >= MAXIMUM_CAPACITY) ? MAXIMUM_CAPACITY : n + 1;
     }
 
+    /**
+     * updated via atomic field updater
+     */
     @SuppressWarnings("UnusedDeclaration")
-    private volatile int size; // updated via atomic field updater
+    private volatile int size;
 
     public IntMapByDynamicHash() {
         this(DEFAULT_INITIAL_CAPACITY);
@@ -110,10 +116,8 @@ public class IntMapByDynamicHash implements IntMap {
         int cap = (size >= (long) MAXIMUM_CAPACITY) ?
                   MAXIMUM_CAPACITY : tableSizeFor((int) size);
         if (cap >= PARTITIONED_SIZE_THRESHOLD) {
-            /*
-            we want 7 extra slots and 64 bytes for each
-            slot. int is 4 bytes, so 64 bytes is 16 ints.
-             */
+            // we want 7 extra slots and 64 bytes for each
+            // slot. int is 4 bytes, so 64 bytes is 16 ints.
             this.partitionedSize =
                 new int[SIZE_BUCKETS * 16];
         }
@@ -293,14 +297,12 @@ public class IntMapByDynamicHash implements IntMap {
 
     @Override
     public IntIterator keys() {
-        // TODO impl
-        return null;
+        return new KeyIterator();
     }
 
     @Override
     public IntIterator values() {
-        // TODO impl
-        return null;
+        return new ValueIterator();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


conghuhu:
@imbajin PTAL

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+                    return;
+                }
+            }
+        }
+    }
+
+    private static final class ResizeContainer extends Entry {
+        private static final int QUEUE_INCREMENT =
+            Math.min(1 << 10,
+                     Integer.highestOneBit(Runtime.getRuntime().availableProcessors()) << 4);
```
javeme, line 640: we have defined a const CPUS, let's just use it

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+        public boolean isNotDone() {
+            return this.resizers.get() > 0;
+        }
+
+        public void zeroOutQueuePosition() {
+            this.queuePosition.set(0);
+        }
+    }
+
+    private static class Entry {
```
javeme, line 708: expect a blank line

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+
+        @Override
+        public String toString() {
+            return this.key + "=" + this.value;
+        }
+    }
+
+    /* ---------------- Iterator -------------- */
+
+    private static final class IteratorState {
```
javeme, line 762: ditto

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+        }
+
+        private IteratorState(Entry[] currentTable, int start, int end) {
+            this.currentTable = currentTable;
+            this.start = start;
+            this.end = end;
+        }
+    }
+
+    private abstract class HashIterator implements IntIterator {
```
javeme, line 779: ditto

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+        final int key;
+        volatile int value;
+        volatile Entry next;
+
+        /**
+         * 0 NORMAL
+         * 1 RESIZING
+         * 2 RESIZED
+         * 3 RESIZE_SENTINEL
+         * 4 ResizeContainer
```
javeme, line 718: please also keep capital style

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+    private static Object tableAt(Object[] array, int index) {
+        return UNSAFE.getObjectVolatile(array,
+                                        ((long) index << ENTRY_ARRAY_SHIFT) +
+                                        ENTRY_ARRAY_BASE);
+    }
+
+    private static boolean casTableAt(Object[] array, int index, Object expected, Object newValue) {
+        return UNSAFE.compareAndSwapObject(
+            array,
+            ((long) index << ENTRY_ARRAY_SHIFT) + ENTRY_ARRAY_BASE,
```
javeme, line 78: we can define a local var to shorten the length of the parameter

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+                    Thread.yield();
+                }
+            }
+            if (this.resizers.get() > 0) {
+                synchronized (this) {
+                    while (this.resizers.get() > 0) {
+                        try {
+                            this.wait();
+                        } catch (InterruptedException e) {
+                            //ginore
```
javeme, line 692: expect a space `// ginore`

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
```diff
@@ -0,0 +1,899 @@
...
+        volatile Entry next;
+
+        /**
+         * 0 NORMAL
+         * 1 RESIZING
+         * 2 RESIZED
+         * 3 RESIZE_SENTINEL
+         * 4 ResizeContainer
+         */
+        final int state;
```
javeme, line 720: can we use 1 8 bits var

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/collection/IntMapByDynamicHash.java
javeme, line 708: expect to serialize it into a chunk memory

imbajin:
From line L735 ~ L972, some suggestions from Copilot:
1. **Handling InterruptedException:** In the `waitForAllResizers` method of the `ResizeContainer` class, the InterruptedException is caught but completely ignored. This could potentially cause issues with thread interruption. In most cases, InterruptedException should either some operation in the catch block or propagate it further.
2. **Endless loop:** Depending on the IntMapByDynamicHash.tableAt(currentArray, index) != RESIZED condition in the helpWithResizeWhileCurrentIndex method, Java could get stuck in an endless while loop. The condition that ends the while loop (change of IntMapByDynamicHash.tableAt(currentArray, index)) does not seem to be reachable and changeable within the loop.
3. **Concurrency:** This code has complex concurrent operations and synchronizations, which are often a source of hard-to-detect bugs. Please ensure that your multithreading logic is correct and you are properly handling thread synchronization.
4. **Thread.yield() usage:** The usage of Thread.yield() is generally not recommended since its behavior can be unpredictable and could vary with different JVMs. You would also want to avoid it for real-time systems.
5. **Null Entry objects:** In the HashIterator's findNext method, null Entry objects are omitted. This would cause hasNext and next methods to not properly reflect the presence of these null entries.
6. **Mutation of shared resources:** Since the array being operated on is not final nor encapsulated, external threads could affect the same array. This could cause unexpected behaviors if not properly handled.
7. **Potential multithreading issues:** This code contains various multithreaded operations (e.g., asynchronous resizers updating in the ResizeContainer class), which can create complex interactions. Without proper synchronization controls, this may result in data inconsistencies.
8. **Method reverseTransfer is not in the code you provided:** There's method reverseTransfer is called in helpWithResize method. However, the code for reverseTransfer is not provided and thus, can't be validated.

Please note that these are potential issues that could lead to bugs based on the provided code snippet. Whether they could actually pose problems depends on the exact workflow and environment in which this code is used. It would also benefit from unit-tests to assure that functionality is as expected.

hugegraph-server/hugegraph-test/src/test/java/org/apache/hugegraph/benchmark/map/MapRandomGetPutThroughputTest.java
```diff
@@ -57,9 +58,11 @@ public class MapRandomGetPutThroughputTest {
 
     private ConcurrentHashMap<Integer, Integer> concurrentHashMap;
 
-    private IntMap.IntMapBySegments intMapBySegments;
+    private IntMap intMapBySegments;
 
-    private IntMap.IntMapByEcSegment intMapByEcSegments;
+    private IntMap intMapByDynamicHashNonCap;
```
javeme, line 63: prefer intMapByDynamicHashWithoutCap

hugegraph-server/hugegraph-test/src/test/java/org/apache/hugegraph/benchmark/map/MapRandomGetPutThroughputTest.java
```diff
@@ -57,9 +58,11 @@ public class MapRandomGetPutThroughputTest {
...
+
+    private IntMap intMapByDynamicHash;
```
javeme, line 65: intMapByDynamicHashWithCap

imbajin (approved):
We merge this V1 version first for release 1.2.0, and link it to the summary issue (enhance it in the future)

simon824 (approved):

PR #2379, opened 2023-12-05
Fix EdgeExistenceAPI path

msgui:
## Purpose of the PR

Path name should match class name

## Verifying these changes

    - xxx

- [ ]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/traversers/EdgeExistenceAPI.java
```diff
@@ -44,7 +44,7 @@
 import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.Context;
 
-@Path("graphs/{graph}/traversers/edgeexist")
+@Path("graphs/{graph}/traversers/edgeexistence")
```
javeme, line 47: prefer to rename EdgeExistence to EdgeExist

PR #2419, opened 2024-01-17
fix(server): starting error of hgServer

Pengzna:
1. When starting hugegrpah-server in idea, we will encounter an error: `java.lang.ClassNotFoundException: io.grpc.stub.AbstractBlockingStub`. We should remove the `<scope>` line in pom to avoid that.

2. Add dependency support for arm-architecture machine


pom.xml
```diff
@@ -89,6 +89,7 @@
     </prerequisites>
 
     <properties>
+        <os.detected.classifier>osx-x86_64</os.detected.classifier>
```
VGalaxies, line 92: good catch~

hugegraph-server/hugegraph-test/src/main/resources/hugegraph.properties
```diff
@@ -17,8 +17,9 @@
 
 gremlin.graph=org.apache.hugegraph.HugeFactory
 
-backend=${backend}
-serializer=${serializer}
+backend=hstore
+serializer=binary
+task.scheduler_type=local
```
VGalaxies, line 22: The ${} here is used to replace variables for testing different backends, and no modification is necessary :)

VGalaxies (changes requested):
PTAL~

PR #2431, opened 2024-01-29
fix(pd-store): fix hstore backend core tests failure

Pengzna:
related issue: https://github.com/apache/incubator-hugegraph/issues/2420
further work of https://github.com/apache/incubator-hugegraph/pull/2424

NOTE：there are still some issues related to MultiGraphTest, EdgeCoreTest, etc. I and @VGalaxies have marked these issues with TODOs in comments, which will be expected to be gradually repaired in the subsequent migration and adaptation work.


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -85,8 +85,6 @@ public void close() {
     }
 
     private Cache<Id, Object> cache(String prefix, long capacity) {
-        // TODO: uncomment later - graph space
```
imbajin, line 88: maybe better to keep the **comment** due to we will import the graph-space feat later?

PR #2432, opened 2024-01-29
fix(server): make CacheManager constructor private to satisfy the singleton pattern

Pengzna:
For the sake of regulation and safety, we should privatize the constructor of cacheManager to satisfy the singleton pattern


Pengzna:
PTAL~ @imbajin @VGalaxies

VGalaxies (approved):
good catch~

imbajin (commented):
Good & we could also list the fixed problem in summary

imbajin (approved):

PR #2462, opened 2024-02-27
fix(server):fix the metric name when promthus collect hugegraph metric,see issue

SunnyBoy-WYH:
fix the metric name when promthus collect hugegraph metric,see issue : https://github.com/apache/incubator-hugegraph/issues/2354

finally, we got it:
img_v3_027q_f143426d-b990-4f46-bf60-61aebf94da7g


imbajin (approved):
THX

javeme (approved):

PR #2466, opened 2024-03-01
fix(server): fix server slow log, support loader import & client IP

### LINKED ISSUE #2468: [Bug] server slow log, support loader import & client IP
### Bug Type (问题类型)

others (please edit later)

### Before submit

### Environment (环境信息)

latest branch

we support slow log before ,but it cause bug when loader batch import data, the feat PR see https://github.com/apache/incubator-hugegraph/pull/2327

and later we downgrade it ,see pr : https://github.com/apache/incubator-hugegraph/pull/2347

the bug due to:

we need get post body from req, and we need set it back to request, so it changed.
the loader request use the "GZIP" header, after get post body, server cant read it
so we ready to resolve it , use BufferedInputStream to cache the stream:

` BufferedInputStream bufferedStream = new BufferedInputStream(context.getEntityStream());

        bufferedStream.mark(Integer.MAX_VALUE);

        context.setProperty(REQUEST_PARAMS_JSON, IOUtils.toString(bufferedStream, Charsets.toCharset(CHARSET)));

        bufferedStream.reset();

        context.setEntityStream(bufferedStream);

_No response_

_No response_

SunnyBoy-WYH:
fix #2468

we support slow log before  ,but it cause bug when loader batch import data, the feat PR see  https://github.com/apache/incubator-hugegraph/pull/2327

and later we downgrade it ,see pr : https://github.com/apache/incubator-hugegraph/pull/2347

the bug due to:
1. we need get post body from req, and we need set it back to request, so it changed.
2. the loader request use the "GZIP" header, after get post body, server cant read it  

so we ready to resolve it , use BufferedInputStream  to cache the stream:

`            BufferedInputStream bufferedStream = new BufferedInputStream(context.getEntityStream());

            bufferedStream.mark(Integer.MAX_VALUE);

            context.setProperty(REQUEST_PARAMS_JSON,
                                IOUtils.toString(bufferedStream, Charsets.toCharset(CHARSET)));

            bufferedStream.reset();

            context.setEntityStream(bufferedStream);
`
case：
[Slow Query] ip=127.0.0.1 execTime=22ms, body={"gremlin":"hugegraph.backendStoreFeatures()                       .supportsSharedStorage();","bindings":{},"language":"gremlin-groovy","aliases":{"g":"__g_hugegraph"}}, method=POST, path=/gremlin, query=null

and then i test it
image

image


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java
```diff
@@ -35,24 +42,29 @@ public class PathFilter implements ContainerRequestFilter {
...
+        long startTime = System.currentTimeMillis();
+
+        context.setProperty(REQUEST_TIME, startTime);
+
+        recordRequestJson(context);
+    }
 
-        // TODO: temporarily comment it to fix loader bug, handle it later
-        /*// record the request json
+    private void recordRequestJson(ContainerRequestContext context) throws IOException {
```
javeme, line 52: prefer `collectRequestParams()`, seems 'record' means output to log

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java
```diff
@@ -35,24 +42,29 @@ public class PathFilter implements ContainerRequestFilter {
...
-                                                 Charsets.toCharset(CHARSET));
-            // replace input stream because we have already read it
-            InputStream in = IOUtils.toInputStream(requestParamsJson, Charsets.toCharset(CHARSET));
-            context.setEntityStream(in);
+            BufferedInputStream bufferedStream = new BufferedInputStream(context.getEntityStream());
+
+            bufferedStream.mark(Integer.MAX_VALUE);
+
+            context.setProperty(REQUEST_PARAMS_JSON,
+                                IOUtils.toString(bufferedStream, Charsets.toCharset(CHARSET)));
```
javeme, line 60: maybe it's very large for batch-write, can we cut part of the content?

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java
```diff
@@ -35,24 +42,29 @@ public class PathFilter implements ContainerRequestFilter {
...
+
+            bufferedStream.reset();
+
+            context.setEntityStream(bufferedStream);
         } else if (method.equals(HttpMethod.GET)) {
-            MultivaluedMap<String, String> pathParameters = context.getUriInfo()
-                                                                   .getPathParameters();
-            requestParamsJson = pathParameters.toString();
+            context.setProperty(REQUEST_PARAMS_JSON,
+                                context.getUriInfo().getPathParameters().toString());
```
javeme, line 67: can we also add all other branchs like PUT/DELETE

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
```diff
@@ -109,13 +113,25 @@ public void filter(ContainerRequestContext requestContext,
...
             if (timeThreshold > 0 && executeTime > timeThreshold &&
                 needRecordLog(requestContext)) {
-                // TODO: set RequestBody null, handle it later & should record "client IP"
-                LOG.info("[Slow Query] execTime={}ms, body={}, method={}, path={}, query={}",
-                         executeTime, null, method, path, uri.getQuery());
+
+                LOG.info("[Slow Query] ip={} execTime={}ms, body={}, method={}, path={}, query={}",
+                         getClientIP(requestContext), executeTime,
+                         requestContext.getProperty(REQUEST_PARAMS_JSON), method, path,
+                         uri.getQuery());
```
javeme, line 120: are they repeated when GET: uri.getQuery() and REQUEST_PARAMS_JSON

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java
SunnyBoy-WYH: maybe we can cut the special length?  like 512?

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
SunnyBoy-WYH: for GET ， REQUEST_PARAMS_JSON will extract the `id` info from `/example/{id} `;  uri.getQuery() will extract ` id=1` from  `http://example.com/resource?id=1`

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java
SunnyBoy-WYH: or we will make it can be config? like 512/1024/2048 and default will be 512? @imbajin  @liuxiaocs7 @VGalaxies

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
```diff
@@ -114,9 +118,11 @@ public void filter(ContainerRequestContext requestContext,
             // Record slow query if meet needs, watch out the perf
             if (timeThreshold > 0 && executeTime > timeThreshold &&
                 needRecordLog(requestContext)) {
-                // TODO: set RequestBody null, handle it later & should record "client IP"
-                LOG.info("[Slow Query] execTime={}ms, body={}, method={}, path={}, query={}",
-                         executeTime, null, method, path, uri.getQuery());
+
+                LOG.info("[Slow Query] ip={} execTime={}ms, body={}, method={}, path={}, query={}",
+                         getClientIP(requestContext), executeTime,
```
javeme, line 123: prefer this style: defining a local var for better readability:
for both: getClientIP and REQUEST_PARAMS_JSON.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
```diff
@@ -128,6 +134,16 @@ public void filter(ContainerRequestContext requestContext,
         }
     }
 
+    private String getClientIP(ContainerRequestContext requestContext) {
+        try {
+            UriInfo uriInfo = requestContext.getUriInfo();
+            String host = uriInfo.getRequestUri().getHost();
```
javeme, line 140: return here if host is an ip or empty?

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
```diff
@@ -128,6 +134,16 @@ public void filter(ContainerRequestContext requestContext,
...
+            return InetAddress.getByName(host).getHostAddress();
+        } catch (UnknownHostException e) {
+            return "unknown";
```
javeme, line 143: prefer "<unknown_ip>"

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java
```diff
@@ -32,27 +39,37 @@ public class PathFilter implements ContainerRequestFilter {
...
```
javeme, line 72: can we also add all other branchs like PUT/DELETE -- seems still missing

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java
```diff
@@ -32,27 +39,37 @@ public class PathFilter implements ContainerRequestFilter {
...
-                                                 Charsets.toCharset(CHARSET));
-            // replace input stream because we have already read it
-            InputStream in = IOUtils.toInputStream(requestParamsJson, Charsets.toCharset(CHARSET));
-            context.setEntityStream(in);
+        if (method.equals(HttpMethod.POST) || method.equals(HttpMethod.PUT) ||
+            method.equals(HttpMethod.DELETE)) {
+            BufferedInputStream bufferedStream = new BufferedInputStream(context.getEntityStream());
+
+            bufferedStream.mark(Integer.MAX_VALUE);
+            String body = IOUtils.toString(bufferedStream,
```
VGalaxies, line 60: **High: Slow-log capture reads full request bodies before truncating**

`hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/PathFilter.java:60`

**Evidence**
- `collectRequestParams()` wraps every POST/PUT/DELETE entity stream, calls `bufferedStream.mark(Integer.MAX_VALUE)`, then `IOUtils.toString(bufferedStream, ...)`, and only truncates the resulting `String` afterward. Batch loader paths such as `VertexAPI`/`EdgeAPI` batch imports are POST/PUT request bodies and can be large.

**Impact**
- Large loader imports are fully read and retained by the filter before the resource method runs, causing avoidable latency and possible heap exhaustion. The 512-byte limit does not limit memory usage because truncation happens after the full body is loaded.

**Requested fix**
- Read at most `MAX_SLOW_LOG_BODY_LENGTH` bytes/chars for the log preview and replay the consumed prefix plus the remaining original stream, or skip body capture for large/batch/import endpoints and non-slow-loggable paths.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
VGalaxies, line 140: **Medium: Slow logs resolve the request URI host, not the client IP**

`hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java:140`

**Evidence**
- `getClientIP()` uses `requestContext.getUriInfo().getRequestUri().getHost()`, which is the host in the server URL requested by the client. Existing surrounding code in `AuthenticationFilter` uses `Request.getRemoteAddr()` for the actual peer address.

**Impact**
- Slow-query logs will attribute requests to the HugeGraph server/listener hostname or proxy target instead of the originating client, making the new client-IP field misleading.

**Requested fix**
- Inject/use the servlet request and log `getRemoteAddr()` or a validated forwarded header only when HugeGraph is behind a trusted proxy.

VGalaxies (commented):
**Review summary**

- **Blocking:** yes
- **Summary:** The PR compiles, but it adds request-body logging in a way that can buffer large loader/batch imports in memory and logs the server host instead of the client IP.
- **Evidence:**
  - `git diff origin/master...HEAD`, `mvn -pl hugegraph-server/hugegraph-api -DskipTests compile` passed
  - no focused filter tests found

PR #2479, opened 2024-03-15
feat(pd-store): added the OpenTelemetry trace support for pd-store

VGalaxies:
follow up https://github.com/apache/incubator-hugegraph/pull/2477

usage (after running otel-collector):

```bash
./bin/start-hugegraph-pd.sh -y true
./bin/start-hugegraph-store.sh -y true
./bin/start-hugegraph.sh -y true
```


imbajin (approved):

imbajin:
Seems we could ignore the unrelated test error?

VGalaxies:
> Seems we could ignore the unrelated test error?

Are you referring to the expected failure of the hstore core tests? It can be ignored for now.

PR #2480, opened 2024-03-16
fix(pd-store): intro `Useless` annotation & refactor and mark the failed tests in `hg-pd-test`

VGalaxies:
subtask of https://github.com/apache/incubator-hugegraph/issues/2265

Some unused classes have been annotated with `@Useless`, while some temporarily failing tests have been annotated with `@Ignore` (some tests' `@Test` annotations have been commented out).


hugegraph-pd/hg-pd-test/src/main/java/org/apache/hugegraph/pd/StoreNodeServiceTest.java
```diff
@@ -420,7 +425,7 @@ public void onPartitionRemoved(Metapb.Partition partition) {
 
     }
 
-    // @Test
```
imbajin, line 423: if a test is hard to ensure the meaning or currently we can't run it well, we could add `@Ignore` for it rather than comment `@Test`  (so
 that we could search ignore in future)

maybe we could also add a annotation `@useless` for other useless classes/method?

imbajin (approved):

PR #2498, opened 2024-04-01
feat(pd): integrate `pd-grpc`&`pd-common`&`pd-client`

VGalaxies:
subtask of https://github.com/apache/incubator-hugegraph/issues/2481

merge https://github.com/apache/incubator-hugegraph/pull/2460 to master


pom.xml
```diff
@@ -86,13 +86,13 @@
     </scm>
 
     <properties>
-        <revision>1.3.0</revision>
+        <revision>1.5.0.1</revision>
```
VGalaxies, line 89: @imbajin Which version number should we use here?

pom.xml
```diff
@@ -199,7 +199,7 @@
                                     <!-- TODO: uncomment for checking dependency conflicts -->
                                     <!-- <DependencyConvergence/> -->
                                     <requireJavaVersion>
-                                        <version>[1.8,12)</version>
+                                        <version>[11,)</version>
```
VGalaxies, line 202: @imbajin Do we enforce a specific Java version here, abandoning Java 8?

imbajin: > @imbajin Do we enforce a specific Java version here, abandoning Java 8?

In my view, Yes (we could just define it in the root pom?)

pom.xml
imbajin: > @imbajin Which version number should we use here?

`1.5.0` for master (we could change it if need)

imbajin (approved):

imbajin:
Note, set version to `1.5.0` before release~

JackyYangPassion (approved):

Issue #2481, opened 2024-03-16
[Summary] integrate new module PD into hugegraph


VGalaxies:
subtask of https://github.com/apache/incubator-hugegraph/issues/2265

- license header
	- [x] https://github.com/apache/incubator-hugegraph/pull/2446
- CI
    - [x] https://github.com/apache/incubator-hugegraph/pull/2480
    - [x] https://github.com/apache/incubator-hugegraph/pull/2492
- to master
	- [x] https://github.com/apache/incubator-hugegraph/pull/2460
	- [x] https://github.com/apache/incubator-hugegraph/pull/2498
	- [x] https://github.com/apache/incubator-hugegraph/pull/2478
	- [x] https://github.com/apache/incubator-hugegraph/pull/2528
	- [x] https://github.com/apache/incubator-hugegraph/pull/2532
	- [x] https://github.com/apache/incubator-hugegraph/pull/2536

VGalaxies:
close by #2265

PR #2536, opened 2024-05-12
chore(pd): translate CJK comments to English

VGalaxies:
subtask of #2481


hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/meta/IdMetaStore.java
imbajin, line 1: the comment in this class seems useful for new comers?

VGalaxies: Recovered and translated.

imbajin (approved):
Thanks a lot for the sync

liuxiaocs7 (approved):

PR #2550, opened 2024-06-04
chore: reset license header for file declared in LICENSE

VGalaxies:
I notice that https://github.com/apache/incubator-hugegraph/blob/master/LICENSE#L210, the license header of `StructureBasicSuite.java` should be the same as `ProcessBasicSuite.java`?


imbajin (approved):
Yes, maybe we need also search some keyword like `@author *` or `mail` format to check the source code reference

VGalaxies:
> Yes, maybe we need also search some keyword like `@author *` or `mail` format to check the source code reference

checked and found no omissions

JackyYangPassion (approved):

PR #2562, opened 2024-06-23
feat(store): integrate store-dist and store-cli submodule

Pengzna:
As title. After this pr, the whole `store` module is completely transferred and integrated.

Note & Todo:
1. there are many useless class in `store-cli` and there is a relatively huge gap between this version of `store-cli` and that in version 4.0. Need more futher work on that.
2. dockerfile is not integrated in aligns with `pd-module`. 
3. countless Chinese comment and log need to be translated to English in further work.


VGalaxies:
@Pengzna update deps accordingly~

VGalaxies (approved):

Pengzna:
@imbajin PTAL

imbajin (approved):
thanks for marking the todos

PR #2585, opened 2024-07-13
chore: upgrade revision to 1.5.0

VGalaxies:
see https://repository.apache.org/content/repositories/orgapachehugegraph-1121/


imbajin (approved):

hugegraph-pd/pom.xml
```diff
@@ -119,14 +119,6 @@
...
                         </goals>
                     </execution>
-                    <!-- auto delete .flattened-pom.xml after "install" step -->
-                    <execution>
-                        <id>remove-flattened-pom</id>
-                        <phase>install</phase>
-                        <goals>
-                            <goal>clean</goal>
-                        </goals>
-                    </execution>
```
VGalaxies, line 129: enhance later: could add an activation (then enable it when not deploying, but only effective during installation).

JackyYangPassion (approved):

PR #2592, opened 2024-07-16
chore: fix lombok compiling error and error for list cast to string

zyxxoo:
fix Fatal error compiling:
`java.lang.NoSuchFieldError: Class com.sun.tools.javac.tree.JCTree$JCImport does not have member field 'com.sun.tools.javac.tree.JCTree qualid'`


hugegraph-pd/hg-pd-client/pom.xml
```diff
@@ -34,7 +34,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
-            <version>1.18.20</version>
+            <version>1.18.30</version>
```
imbajin, line 37: Shall we use a unified version property for it?

VGalaxies: +1

javeme, line 37: Can we remove the lombok dependency:
since we don't have a large places to generate/replace code, and in the few places where it's necessary, I would rather write some code manually to maintain a consistent style.

minaxolone:
@zyxxoo The version also needs to be bumped in [hg-pd-common](https://github.com/apache/incubator-hugegraph/blob/ba5b89e374f3335865cea36a50eca1e99ccaf56e/hugegraph-pd/hg-pd-common/pom.xml#L41).

hugegraph-commons/pom.xml
```diff
@@ -124,6 +124,16 @@
         <module>hugegraph-rpc</module>
     </modules>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>1.18.30</version>
```
imbajin, line 132: Maybe we could add `<optional>true</optional>` and `provide` scope for it?

hugegraph-store/pom.xml
```diff
@@ -85,6 +85,11 @@
                 <artifactId>log4j-slf4j-impl</artifactId>
                 <version>${log4j2.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>1.18.30</version>
```
imbajin, line 91: miss here?

zyxxoo: the code is place in dependencymanagner

hugegraph-commons/pom.xml
zyxxoo: the optional mean don't transfer dependency, for example, A -> B -> C, if B. use. optional true to c, the A will not reference C
so the optional in our code maybe don't importament

javeme:
TODO: 
Can we remove the lombok dependency:
since we don't have a large places to generate/replace code, and in the few places where it's necessary, I would rather write some code manually to maintain a consistent style.

javeme (dismissed):

imbajin (approved):
TODO: replace each 3rd-level-module deps by 2nd-level-module dep

VGalaxies (approved):

PR #2599, opened 2024-07-19
chore: temporarily ignore failure core tests for hstore & simplify ci name

VGalaxies:
subtask of https://github.com/apache/incubator-hugegraph/issues/2420


hugegraph-server/pom.xml
```diff
@@ -236,13 +236,11 @@
                 <groupId>io.grpc</groupId>
                 <artifactId>grpc-netty</artifactId>
                 <version>${grpc.version}</version>
-                <scope>provided</scope>
```
imbajin, line 239: any context with the modification?

.github/workflows/server-ci.yml
```diff
@@ -1,4 +1,4 @@
-name: "hugegraph-server-ci"
+name: "server CI"
```
imbajin, line 1: - GraphServer CI (better?)
- Server CI
- graph server CI

VGalaxies: `Graph Server CI` feels better, similar `Graph PD & Store & Hstore CI`

hugegraph-server/pom.xml
VGalaxies: Dependencies marked as provided scope are available during compilation and testing, but they will not be included in the final WAR or JAR package.

If you use provided scope here, an error class not found will occur when running hstore with pd-client related code.

imbajin (dismissed):

imbajin (approved):

JackyYangPassion (approved):

PR #2612, opened 2024-08-02
chore(server): temporary ignore graphspaces segment in url for forwards compatibility

VGalaxies:
As title.

### DIFF
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/GraphSpaceFilter.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/GraphSpaceFilter.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hugegraph.api.filter;
+
+import jakarta.inject.Singleton;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.container.PreMatching;
+import jakarta.ws.rs.core.UriBuilder;
+import jakarta.ws.rs.ext.Provider;
+
+import org.apache.hugegraph.util.Log;
+import org.slf4j.Logger;
+
+import java.io.IOException;
+import java.net.URI;
+
+@Provider
+@Singleton
+@PreMatching
+public class GraphSpaceFilter implements ContainerRequestFilter {
+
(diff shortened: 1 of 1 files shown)


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/GraphSpaceFilter.java
```diff
@@ -0,0 +1,69 @@
...
+        URI relativePath = baseUri.relativize(requestUri);
+
+        String relativePathStr = relativePath.getPath();
+        // TODO: remember change the logic after we support "GraphSpace"
+        if (relativePathStr.startsWith("graphspaces/")) {
+            // Step 2: Extract the next substring after "graphspaces/"
+            String[] parts = relativePathStr.split("/");
+            if (parts.length > 2) {
+                String ignoredPart = parts[1];
+                String nextPathPart = parts[2];
```
javeme, line 55: what if parts[3,4,5...], do we miss the rest parts?

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/GraphSpaceFilter.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public void filter(ContainerRequestContext context) throws IOException {
+        // Step 1: Get relativePath
+        URI baseUri = context.getUriInfo().getBaseUri();
+        URI requestUri = context.getUriInfo().getRequestUri();
+        URI relativePath = baseUri.relativize(requestUri);
+
+        String relativePathStr = relativePath.getPath();
+        // TODO: remember change the logic after we support "GraphSpace"
+        if (relativePathStr.startsWith("graphspaces/")) {
```
javeme, line 50: prefer this style here: `return if (!relativePathStr.startsWith("graphspaces/"))`

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/GraphSpaceFilter.java
```diff
@@ -0,0 +1,69 @@
...
+            // Step 2: Extract the next substring after "graphspaces/"
```
javeme, line 51: can we add a api-test case for this code branch

VGalaxies: see GraphSpaceVertexApiTest and GraphSpaceEdgeApiTest, could be more...

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/config/ServerOptions.java
```diff
@@ -47,6 +47,14 @@ public static synchronized ServerOptions instance() {
                     "http://127.0.0.1:8080"
             );
 
+    public static final ConfigOption<Boolean> REST_SERVER_ENABLE_GRAPHSPACES_FILTER =
+            new ConfigOption<>(
+                    "restserver.enable_graphspaces_filter",
```
imbajin, line 52: `Compatible` or `Transfer`?  (seems better than filter)

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/BaseApiTest.java
```diff
@@ -84,6 +84,7 @@ public static void init() {
     @AfterClass
     public static void clear() throws Exception {
         client.close();
+        client = null;
```
imbajin, line 87: what's for?

VGalaxies: No special meaning, just set the client member variable to null to indicate that the closed client should not be used anymore.

imbajin (approved):
Thanks for the update~

simon824 (approved):

PR #2620, opened 2024-08-06
fix(server) Avoid duplicate registration of StoreEventListener

### LINKED ISSUE #2618: [Bug] StoreEventListener in CachedSystemTransactionV2 and CachedGraphTransaction will be registered and monitored repeatedly
### Bug Type (问题类型)

logic (逻辑设计问题)

### Before submit

### Environment (环境信息)

- Server Version: 1.5.0 (Apache Release Version)
- Backend: HStore

StoreEventListener in CachedSystemTransactionV2 and CachedGraphTransaction will be registered and monitored repeatedly;
Because CachedSystemTransactionV2 and CachedGraphTransaction are both ThreadLocal types in TinkerPopTransaction, each thread holds a corresponding instance. When initializing each instance, it registers a storeEventListener with the storeEventHub held by AbstractBackendStoreProvider. Since each listener implements the same functionality, storeEventHub holds a large number of redundant storeEventListeners

_No response_

_No response_

haohao0103:
close #2618 

Define initialization identifiers for storeEventListeners in AbstractBackendStoreProvider to ensure that storeEventListeners are not registered repeatedly


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/AbstractBackendStoreProvider.java
```diff
@@ -67,7 +70,10 @@ protected final void checkOpened() {
 
     @Override
     public void listen(EventListener listener) {
-        this.storeEventHub.listen(EventHub.ANY_EVENT, listener);
+        if (schemaCacheClearListened.compareAndSet(false, true) ||
+            vertexEdgeCacheClearListened.compareAndSet(false, true)) {
```
imbajin, line 74: Some question:
- we define 2 automic_param boolean A & B and try to set them to `true`, when to reset it?
- why we need 2 flag to mark it? (in this context)
- since **assignment operations** (`=`) for **primitive types** are atomic by default(64bit), it seems that we only need to use a `volatile` boolean flag to achieve the desired effect?

@Pengzna @VGalaxies

haohao0103: 1、define 2 automic_param boolean A & B;this operation is to ensure that storeEventListener does not deliver to storeEventHub of storeProvider once in each thread. I understand that there is no need to re register during the entire Hugegraph lifecycle, so there is no reset operation; Will this miss other necessary scenarios?
2、schemaCacheClearListened for schemaCache clear , vertexEdgeCacheClearListened for vertex&edge cache;
3、 I have modified it to use the volatile+synchronized，thanks

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/AbstractBackendStoreProvider.java
```diff
@@ -44,6 +45,8 @@ public abstract class AbstractBackendStoreProvider
     private final EventHub storeEventHub = new EventHub("store");
 
     protected Map<String, BackendStore> stores = null;
+    private static volatile Boolean schemaCacheClearListened = Boolean.FALSE;
```
imbajin, line 48: Is `boolean` enough?

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/AbstractBackendStoreProvider.java
```diff
@@ -20,6 +20,7 @@
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
```
imbajin, line 23:
```suggestion
```

zyxxoo:
thanks for you push code for solve the problem, this is a great code, but i have some different advise for that. ohh, the transaction monitor some event happen, meawhile, the code change the state local, every instance maybe need care the event, so  we should allow repeated event register; for the problem, i think the main problem maybe is the threadlocal not correct release, that make the listen not be remove

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -133,7 +133,7 @@ private void listenChanges() {
             }
             return false;
         };
-        this.store().provider().listen(this.storeEventListener);
+        this.store().provider().listenDataCacheClear(this.storeEventListener);
```
javeme, line 136: Sorry I have read the issue, but I still don't quite understand in what scenarios.the problem is encountered.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/BackendStoreProvider.java
```diff
@@ -67,6 +67,12 @@ public interface BackendStoreProvider {
 
     void listen(EventListener listener);
 
+    default void listenSchemaCacheClear(EventListener listener) {
```
javeme, line 70: prefer listenSchemaCacheAtMostOnce

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/BackendStoreProvider.java
```diff
@@ -67,6 +67,12 @@ public interface BackendStoreProvider {
...
+    }
+
+    default void listenDataCacheClear(EventListener listener) {
```
javeme, line 73: prefer listenGraphCacheAtMostOnce

javeme (commented):
Thanks for your contribution, some tiny comments

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -105,7 +105,8 @@ private void listenChanges() {
             }
             return false;
         };
-        this.graphParams().loadGraphStore().provider().listen(this.storeEventListener);
+        this.graphParams().loadGraphStore().provider()
+            .listenSchemaCacheClear(this.storeEventListener);
 
         // Listen cache event: "cache"(invalid cache item)
         this.cacheEventListener = event -> {
```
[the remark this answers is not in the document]
VGalaxies: CachedSchemaTransactionV2 is the CachedSchemaTransaction corresponding to the SchemaTransaction of the pd-store version. We added a V2 to distinguish it from the previous version. Due to the inconsistency of the parameters required for construction, it is currently not easy to reuse, will be merged later...

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/store/BackendStoreProvider.java
haohao0103: thanks ,  you mean to update the method name to  listenSchemaCacheAtMostOnce?

haohao0103:
> thanks for you push code for solve the problem, this is a great code, but i have some different advise for that. ohh, the transaction monitor some event happen, meawhile, the code change the state local, every instance maybe need care the event, so we should allow repeated event register; for the problem, i think the main problem maybe is the threadlocal not correct release, that make the listen not be remove

thanks , i try to understand  what you mean , Indeed, there some issues with threadlocal processing,and my code did not address this problem. In addition ,I thind it is indeed valuable to pay attention to events for each transaction instance; But in the current sceario,based on my understanding,this operation seems redundant because schema and vertex/edge cache are both global sinleton, and it seems unnecessay to execute their operations on every transaction instance. of course ,if there are personalized scenarios that allow for reregistration of listener,it is indeed necessary and valuable.this is my idea, welcome to correct it.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
haohao0103: simply put,i found that registering a listener for each transaction instance is redundant because each transaction receives events and then performs a clear operation on the schema/vertex/edge cache , i think that this is redundant.

javeme: Note each provider has multiple listeners, which is our design intention, because each listener wants to get notifications.
This modification looks a bit special at the moment, and we will propose a solution after we figure out the problem. Could you post the error stack if there is an error?

PR #2692, opened 2024-11-11
chore(dist): fix licenses and remove empty files

VGalaxies:
As title.


.gitattributes
```diff
@@ -12,6 +12,5 @@ hugegraph-store/hg-store-dist/src/assembly/static/bin/libjemalloc_aarch64.so exp
 .github/ export-ignore
 .idea/ export-ignore
 install-dist/scripts/ export-ignore
-install-dist/release-docs/ export-ignore
```
VGalaxies, line 15: because of https://github.com/apache/incubator-hugegraph/blob/master/install-dist/pom.xml#L56

imbajin (approved):
Thanks~

simon824 (approved):

PR #2702, opened 2024-12-09
fix(docker): update server image desc

VGalaxies:
follow up https://github.com/apache/incubator-hugegraph-doc/pull/381


imbajin (approved):
Remember to update the version after release done (maybe we should add it to the doc/wiki)

liuxiaocs7 (approved):

PR #2706, opened 2024-12-13
fix(dist): add license for swagger-ui & reset use stage to false in ci yml

VGalaxies:
fixup #2687

cound not found licenses declared in dist, see https://github.com/swagger-api/swagger-ui/issues/8317


install-dist/release-docs/LICENSE
imbajin, line 1: maybe we should disable 'stage' param in this PR

Also mark a TODO for use common dependency directly

VGalaxies: > Also, mark a TODO for using common/rpc dependency directly

where to mark?

imbajin: > > Also, mark a TODO for using common/rpc dependency directly
> 
> where to mark?

Using the `stage` param of the CI place + usage in the `pom.xml` (common/rpc)

imbajin (approved):

JackyYangPassion (approved):

PR #2722, opened 2025-01-14
Centralize version management.

lavanyavijayk:
- Improved version management by introducing Maven's  property
- Replaced hardcoded version strings with dynamic retrieval
- Added version.properties to support runtime version access
- Ensured updated version in scripts and classes

## Purpose of the PR

- close #xxx 

## Verifying these changes

    - xxx

- [ ]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-commons/hugegraph-common/pom.xml
+++ b/hugegraph-commons/hugegraph-common/pom.xml
@@ -238,6 +238,12 @@
     </dependencyManagement>
 
     <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+            </resource>
+        </resources>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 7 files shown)


VGalaxies:
@lavanyavijayk There is an issue with the version number [here](https://github.com/apache/incubator-hugegraph/blob/master/hugegraph-pd/hg-pd-service/pom.xml#L127) for compilation error, it should be updated to `${hugegraph-commons.version}`.

lavanyavijayk:
> @lavanyavijayk There is an issue with the version number [here](https://github.com/apache/incubator-hugegraph/blob/master/hugegraph-pd/hg-pd-service/pom.xml#L127) for compilation error, it should be updated to `${hugegraph-commons.version}`.
Thank you for pointing that out. I have made the changes. 

I also want to get more details for the following.

The ticket mentions a change in /hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/version/ApiVersion.java, but it is unclear which string should be replaced with the revision value. Could you confirm this before proceeding?
I attempted to dynamically assign the revision value to CoreVersion.DEFAULT_VERSION (/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/version/CoreVersion.java), but it caused a compilation error. The issue arises because CoreVersion.DEFAULT_VERSION is also used as an annotation value in /hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/server/ApplicationConfig.java, which requires a constant expression. Should I defer this change for now?

PR #2735, opened 2025-02-19
adjust the related filters of sofa-bolt.

haohao0103:
Optimize and adjust the related filters of sofa-bolt.

### DIFF
--- a/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
+++ b/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
@@ -19,6 +19,7 @@ package org.apache.hugegraph.pd.raft;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
```diff
@@ -130,14 +136,35 @@ public boolean init(PDConfig.Raft config) {
     /**
      * Create a Raft RPC Server for communication between PDs
      */
-    private RpcServer createRaftRpcServer(String raftAddr) {
+    private RpcServer createRaftRpcServer(String raftAddr, List<PeerId> peers) {
         Endpoint endpoint = JRaftUtils.getEndPoint(raftAddr);
         RpcServer rpcServer = RaftRpcServerFactory.createRaftRpcServer(endpoint);
+        configureRaftServerIpWhitelist(peers, rpcServer);
```
zyxxoo, line 142: How does the raft group member change?

haohao0103: 1、The whitelist IPs are sourced from the peer-list configuration file. The current solution rejects all Raft-related requests from nodes not in the peer-list.
2、I understand that the members of the Raft group are a subset of the peer-list. Please correct me if there are any errors in this understanding.
thanks

imbajin (approved):

simon824 (approved):

PR #2861, opened 2025-08-30
perf(core): optimize perf by avoid boxing long 

javeme:
- Query: E.checkArgument avoid boxing long
- BytesBuffer: E.checkArgument avoid Bytes.toHex and boxing long
- define const RANGE_TYPES to avoid construct every time

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
@@ -137,6 +137,12 @@ public abstract class Condition {
             return true;
         });
 
+        private static final Set<RelationType> RANGE_TYPES =
+                ImmutableSet.of(GT, GTE, LT, LTE);
+
+        private static final Set<RelationType> UNFLATTEN_TYPES =
+                ImmutableSet.of(IN, NOT_IN, TEXT_CONTAINS_ANY);
+
         private final String operator;
         private final BiFunction<Object, Object, Boolean> tester;
         private final Class<?> v1Class;
@@ -261,7 +267,7 @@ public abstract class Condition {
         }
 
         public boolean isRangeType() {
-            return ImmutableSet.of(GT, GTE, LT, LTE).contains(this);
+            return RANGE_TYPES.contains(this);
         }
 
         public boolean isSearchType() {
@@ -628,10 +634,6 @@ public abstract class Condition {
         // The value serialized(code/string) by backend store.
         protected Object serialValue;
 
-        protected static final Set<RelationType> UNFLATTEN_RELATION_TYPES =
-                ImmutableSet.of(RelationType.IN, RelationType.NOT_IN,
-                                RelationType.TEXT_CONTAINS_ANY);
-
         @Override
         public ConditionType type() {
             return ConditionType.RELATION;
@@ -672,7 +674,7 @@ public abstract class Condition {
 
         @Override
         public boolean isFlattened() {
-            return !UNFLATTEN_RELATION_TYPES.contains(this.relation);
+            return !RelationType.UNFLATTEN_TYPES.contains(this.relation);
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Query.java
```diff
@@ -180,7 +180,9 @@ public long offset() {
     }
 
     public void offset(long offset) {
-        E.checkArgument(offset >= 0L, "Invalid offset %s", offset);
+        if (offset < 0L) {
```
imbajin, line 183: **Performance optimization concern**: While avoiding autoboxing is good, this pattern is harder to read and maintain. Consider using a static helper method instead:

```java
private static void checkNonNegative(long value, String message, Object... args) {
    if (value < 0L) {
        E.checkArgument(false, message, args);
    }
}
```

Then call: `checkNonNegative(offset, "Invalid offset %s", offset);`

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
```diff
@@ -162,16 +165,21 @@ public int remaining() {
 
     private void require(int size) {
         // Does need to resize?
-        if (this.buffer.limit() - this.buffer.position() >= size) {
+        if (this.buffer.remaining() >= size) {
```
imbajin, line 168: **Bug fix**: Good catch! Using `buffer.remaining()` is more efficient and semantically correct than manually calculating `buffer.limit() - buffer.position()`.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
```diff
@@ -249,7 +257,7 @@ public byte peek() {
     }
 
     public byte peekLast() {
-        return this.buffer.get(this.buffer.capacity() - 1);
+        return this.buffer.get(this.buffer.limit() - 1);
```
imbajin, line 260: **Critical bug fix**: This change from `buffer.capacity() - 1` to `buffer.limit() - 1` is correct and important. Using capacity would access potentially uninitialized data beyond the actual buffer content.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
```diff
@@ -137,6 +137,12 @@ public enum RelationType implements BiPredicate<Object, Object> {
             return true;
         });
 
+        private static final Set<RelationType> RANGE_TYPES =
```
imbajin, line 140: **Performance optimization**: Excellent refactoring! Moving these constant sets to static final fields avoids recreating the same ImmutableSet instances on every method call. This is a significant performance improvement for frequently called methods.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Query.java
```diff
@@ -287,17 +293,27 @@ public long total() {
 
     public long limit() {
         if (this.capacity != NO_CAPACITY) {
```
imbajin, line 295: **Code style inconsistency**: While the performance optimization is good, this pattern creates inconsistent code style throughout the file. Consider applying the same refactoring pattern to all similar validation methods for consistency, or use a helper method approach for better maintainability.

javeme: We only modify the high-frequency codes on the main path and keep the other low-frequency codes in their original style.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Query.java
javeme: due to there is no a universal check condition, but adding methods for various check conditions is not a good idea.

VGalaxies (approved):

PR #2862, opened 2025-08-30
perf(core): StringId hold bytes to avoid decode/encode

javeme:
Change-Id: I1ff93a2ee31f142fb7ab5cfe0e038ef2da2f982d

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
@@ -23,6 +23,7 @@ import java.util.UUID;
 import org.apache.hugegraph.backend.id.Id.IdType;
 import org.apache.hugegraph.backend.serializer.BytesBuffer;
 import org.apache.hugegraph.structure.HugeVertex;
+import org.apache.hugegraph.util.Bytes;
 import org.apache.hugegraph.util.E;
 import org.apache.hugegraph.util.LongEncoding;
 import org.apache.hugegraph.util.NumericUtil;
@@ -128,14 +129,19 @@ public abstract class IdGenerator {
     public static class StringId implements Id {
 
         protected String id;
+        protected byte[] bytes;
 
         public StringId(String id) {
             E.checkArgument(!id.isEmpty(), "The id can't be empty");
             this.id = id;
+            this.bytes = null;
         }
 
         public StringId(byte[] bytes) {
-            this.id = StringEncoding.decode(bytes);
+            E.checkArgument(bytes != null && bytes.length > 0,
+                            "The id bytes can't be null or empty");
+            this.bytes = bytes;
+            this.id = null;
         }
 
         @Override
@@ -145,27 +151,33 @@ public abstract class IdGenerator {
 
         @Override
         public Object asObject() {
-            return this.id;
+            return this.asString();
         }
 
         @Override
         public String asString() {
+            if (this.id == null) {
+                this.id = StringEncoding.decode(this.bytes);
+            }
             return this.id;
         }
 
         @Override
         public long asLong() {
-            return Long.parseLong(this.id);
+            return Long.parseLong(this.asString());
         }
 
         @Override
         public byte[] asBytes() {
-            return StringEncoding.encode(this.id);
+            if (this.bytes == null) {
+                this.bytes = StringEncoding.encode(this.id);
+            }
+            return this.bytes;
         }
 
         @Override
         public int length() {
-            return this.id.length();
+            return this.asString().length();
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
```diff
@@ -128,14 +129,19 @@ public static int compareType(Id id1, Id id2) {
     public static class StringId implements Id {
 
         protected String id;
+        protected byte[] bytes;
```
imbajin, line 132: **Thread Safety Concern**: The lazy initialization pattern used here is not thread-safe. If multiple threads access the same StringId instance simultaneously, race conditions could occur during the lazy initialization of `id` and `bytes` fields. Consider using volatile fields or synchronization to ensure thread safety.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
```diff
@@ -128,14 +129,19 @@ public static int compareType(Id id1, Id id2) {
...
 
         public StringId(String id) {
```
imbajin, line 134: **Inconsistent State Risk**: The constructor initializes `bytes = null` explicitly, but the state where both `id` and `bytes` are null is not handled. Consider adding validation to ensure at least one field is non-null, or document the invariant that exactly one field should be null at any time.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
```diff
@@ -174,25 +186,34 @@ public int compareTo(Id other) {
             if (cmp != 0) {
                 return cmp;
             }
-            return this.id.compareTo(other.asString());
+            if (this.id != null) {
+                return this.id.compareTo(other.asString());
+            } else {
+                return Bytes.compare(this.bytes, other.asBytes());
```
imbajin, line 195: **Performance Optimization**: Consider using `Arrays.equals()` instead of `Bytes.equals()` for byte array comparison if `Bytes.equals()` is just a wrapper. Also, cache the result of `other.asBytes()` to avoid multiple conversions in the comparison.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
```diff
@@ -174,25 +186,34 @@ public int compareTo(Id other) {
...
-            if (!(other instanceof StringId)) {
+        public boolean equals(Object obj) {
+            if (!(obj instanceof StringId)) {
                 return false;
             }
-            return this.id.equals(((StringId) other).id);
+            StringId other = (StringId) obj;
+            if (this.id != null) {
+                return this.id.equals(other.asString());
+            } else {
```
imbajin, line 214: **Potential Memory Leak**: In the equals method, calling `other.asBytes()` when `this.id != null` may trigger unnecessary byte array creation and caching in the other object. Consider checking if the other object has the same representation type first to avoid forced conversions.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
```diff
@@ -128,14 +129,19 @@ public static int compareType(Id id1, Id id2) {
     public static class StringId implements Id {
```
imbajin, line 129: **Missing Test Coverage**: This performance optimization changes core behavior of StringId. Please ensure comprehensive test coverage including:\n- Thread safety tests with concurrent access\n- Edge cases with empty/null inputs\n- Performance regression tests\n- Memory usage validation tests

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/memory/consumer/impl/id/StringIdOffHeap.java
```diff
@@ -70,6 +70,7 @@ public void serializeSelfToByteBuf(MemoryPool memoryPool) {
     @Override
     public void releaseOriginalVarsOnHeap() {
         this.id = null;
+        this.bytes = null;
```
javeme, line 78: in serializeSelfToByteBuf we can set stringBytes = this.bytes if this.bytes != null

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
javeme: At present, I haven't found any scenarios where this would cause problems. It's safe even if multiple assignments occur at the same time.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
javeme: id is null is not allowed in this constructor

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/id/IdGenerator.java
javeme: done

VGalaxies (approved):

PR #2863, opened 2025-08-30
perf(rocksdb): RocksDBStore remove redundant checkOpened() call

javeme:
Change-Id: I4c4ae6368fecb994708ae1da89c256beefda4511

### DIFF
--- a/hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/RocksDBStore.java
+++ b/hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/RocksDBStore.java
@@ -425,7 +425,6 @@ public abstract class RocksDBStore extends AbstractBackendStore<RocksDBSessions.
         Lock readLock = this.storeLock.readLock();
         readLock.lock();
         try {
-            this.checkOpened();
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Store {} mutation: {}", this.store, mutation);
             }
@@ -490,7 +489,6 @@ public abstract class RocksDBStore extends AbstractBackendStore<RocksDBSessions.
         readLock.lock();
 
         try {
-            this.checkOpened();
             HugeType tableType = RocksDBTable.tableType(query);
             RocksDBTable table;
             RocksDBSessions.Session session;
@@ -525,7 +523,6 @@ public abstract class RocksDBStore extends AbstractBackendStore<RocksDBSessions.
         Lock readLock = this.storeLock.readLock();
         readLock.lock();
         try {
-            this.checkOpened();
             HugeType tableType = RocksDBTable.tableType(query);
             RocksDBTable table = this.table(tableType);
             return table.queryNumber(this.session(tableType), query);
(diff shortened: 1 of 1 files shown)


hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/RocksDBStore.java
```diff
@@ -813,12 +776,56 @@ private void closeSessions() {
...
+
+        // Collect session of standard disk
+        RocksDBSessions.Session session = this.sessions.session();
+        if (!session.opened()) {
+            this.checkOpened();
+        }
+
+        if (this.tableDiskMapping.isEmpty()) {
+            return ImmutableList.of(session);
+        }
```
imbajin, line 795: Performance optimization looks good, but consider session state consistency. The conditional checkOpened() call based on session.opened() could introduce race conditions between the check and actual session usage.

hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/RocksDBStore.java
```diff
@@ -813,12 +776,56 @@ private void closeSessions() {
...
 
+        // Collect session of each table with optimized disk
+        List<RocksDBSessions.Session> list = new ArrayList<>(this.tableDiskMapping.size() + 1);
+        list.add(session);
+        for (String disk : this.tableDiskMapping.values()) {
+            RocksDBSessions.Session optimizedSession = this.db(disk).session();
+            assert optimizedSession.opened();
```
imbajin, line 801: Inconsistent session state handling: optimized sessions use assert optimizedSession.opened() while standard sessions get conditional checks. Consider applying the same defensive pattern to both session types for consistency.

hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/RocksDBStore.java
```diff
@@ -813,12 +776,56 @@ private void closeSessions() {
         }
     }
 
-    private Collection<RocksDBSessions> sessions() {
+    private final Collection<RocksDBSessions> sessions() {
         return this.dbs.values();
     }
 
-    private void parseTableDiskMapping(Map<String, String> disks, String dataPath) {
+    private final List<RocksDBSessions.Session> session() {
```
imbajin, line 783: Consider making session() method final like sessions() for consistency and to prevent accidental overrides that could break the optimized session management logic.

hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/RocksDBStore.java
```diff
@@ -425,7 +425,6 @@ public void mutate(BackendMutation mutation) {
         Lock readLock = this.storeLock.readLock();
```
imbajin, line 425: Good performance optimization removing redundant checkOpened() calls. Ensure comprehensive testing covers store state consistency under concurrent access and that the performance gains are measurable without introducing subtle bugs.

hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/RocksDBStore.java
javeme: It is believed here that optimizedSession must be the opened state, otherwise it is a bug in logic. We consider this optimizedSession state to be consistent with the main session state.

VGalaxies (approved):

PR #2864, opened 2025-08-30
perf(core): add AdjacentEdgesQuery

javeme:
Change-Id: Ie7169fdb707495ff0a298ea97fa4d105117a132d

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/AdjacentEdgesQuery.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/AdjacentEdgesQuery.java
@@ -1,0 +1,396 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hugegraph.backend.query;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+import org.apache.hugegraph.backend.id.Id;
+import org.apache.hugegraph.backend.query.Condition.Relation;
+import org.apache.hugegraph.backend.query.Condition.RelationType;
+import org.apache.hugegraph.structure.HugeEdge;
+import org.apache.hugegraph.structure.HugeElement;
+import org.apache.hugegraph.type.HugeType;
+import org.apache.hugegraph.type.define.Directions;
+import org.apache.hugegraph.type.define.HugeKeys;
+import org.apache.hugegraph.util.CollectionUtil;
+import org.apache.hugegraph.util.E;
+import org.apache.hugegraph.util.InsertionOrderUtil;
+
+import com.google.common.collect.ImmutableList;
+
+public class AdjacentEdgesQuery extends ConditionQuery {
+
+    private Id ownerVertex;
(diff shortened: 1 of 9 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/AdjacentEdgesQuery.java
```diff
@@ -0,0 +1,396 @@
...
+
+    private static final Id[] EMPTY = new Id[0];
+
+    public AdjacentEdgesQuery(Id ownerVertex, Directions direction) {
+        this(ownerVertex, direction, EMPTY);
+    }
+
+    public AdjacentEdgesQuery(Id ownerVertex, Directions direction,
+                              Id[] edgeLabels) {
+        super(HugeType.EDGE);
```
imbajin, line 55: Consider adding defensive null check for 'ownerVertex' before using it in toString() and other methods to prevent potential NPE.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/AdjacentEdgesQuery.java
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> T selfCondition(Object key) {
+        if (key == HugeKeys.OWNER_VERTEX) {
+            return (T) this.ownerVertex;
+        }
+        if (key == HugeKeys.DIRECTION) {
+            return (T) this.direction;
+        }
```
imbajin, line 186: The logic in selfCondition() at line 186 should return an array for consistency when edgeLabels.length > 1, or throw the exception immediately instead of having an unreachable return statement.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/AdjacentEdgesQuery.java
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    @Override
+    public boolean isFlattened() {
+        if (this.direction == Directions.BOTH) {
+            return false;
+        }
+        if (this.edgeLabels.length > 1) {
+            return false;
+        }
```
imbajin, line 320: Consider using Collections.emptyList() instead of creating a new ArrayList when no flattening is needed for better memory efficiency.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
```diff
@@ -345,31 +345,31 @@ public static Relation neq(HugeKeys key, Object value) {
         return new SyspropRelation(key, RelationType.NEQ, value);
     }
 
-    public static Condition in(HugeKeys key, List<?> value) {
+    public static Relation in(HugeKeys key, List<?> value) {
```
imbajin, line 348: The return type changes from Condition to Relation break API compatibility. Ensure all calling code has been updated to handle Relation instead of Condition.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -1291,62 +1292,24 @@ public <V> void removeEdgeProperty(HugeEdgeProperty<V> prop) {
...
-            } else {
-                query.eq(HugeKeys.LABEL, edgeLabel.id());
-            }
-        } else if (edgeLabels.length >= 1) {
-            query.query(
-                    Condition.in(HugeKeys.LABEL,
-                                 Arrays.stream(edgeLabels)
-                                       .map(SchemaElement::id)
-                                       .collect(Collectors.toList())));
+        if (true) {
```
imbajin, line 1309: The hardcoded 'if (true)' should be replaced with a proper condition or configuration flag. This looks like debug/test code that should not be merged.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -1291,62 +1292,24 @@ public <V> void removeEdgeProperty(HugeEdgeProperty<V> prop) {
...
+            return new AdjacentEdgesQuery(sourceVertex, direction, edgeLabels);
         }
```
imbajin, line 1312: Unreachable code after 'if (true)' block should be removed to avoid confusion and maintain code cleanliness.

PR #2871, opened 2025-09-16
improve: raft state machine should report error

zyxxoo:
## Purpose of the PR

- close #xxx 

## Verifying these changes

    - xxx

- [ ]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftStateMachine.java
+++ b/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftStateMachine.java
@@ -94,6 +94,7 @@ public class RaftStateMachine extends StateMachineAdapter {
                 if (done != null) {
                     done.run(new Status(RaftError.EINTERNAL, t.getMessage()));
                 }
+                iter.setErrorAndRollback(1, new Status(RaftError.ESTATEMACHINE, t.getMessage()));
             }
             iter.next();
         }
--- a/hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/raft/HgStoreStateMachine.java
+++ b/hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/raft/HgStoreStateMachine.java
@@ -101,6 +101,7 @@ public class HgStoreStateMachine extends StateMachineAdapter {
                               done.op.getReq());
                     //    done.run(new Status(RaftError.EINTERNAL, t.getMessage()));
                 }
+                inter.setErrorAndRollback(1, new Status(RaftError.ESTATEMACHINE, t.getMessage()));
             }
             committedIndex = inter.getIndex();
 


imbajin (commented):
Score: 6.0/10; validation is insufficient for approval. Three independent reviews found no verified material runtime regression, but this change moves PD and Store apply failures into jRaft terminal-error handling without adding failure-path regression tests. Please verify leader apply failure and follower replay failure, that later entries are not applied, closure completion, and the applied-index/snapshot boundary. In PD, the existing done.run(EINTERNAL) and jRaft error propagation both complete the current closure; the present CompletableFuture consumer ignores the second completion, so this is not a claim of duplicate RPC effects or data loss. This review is based on code and jRaft 1.3.13 control-flow inspection, not an executed fault test.

imbajin (commented):
Score: 8/10 (previously 6/10). The benefit is confirmed: the same fault tests fail before the fix because entry 2 throws but lastApplied advances to 3; after the fix it remains at 1 and later entries are not applied. PD error closures are completed once, and Store no longer publishes the failed entry as applied.

Updated to master dbb6663a8 and resolved conflicts without rewriting history. Store now targets the active PartitionStateMachine, rather than the older implementation. Seven apply scenarios cover success, leader/follower failure, closure completion and the applied/snapshot boundary; existing snapshot tests remain intact. Independent review also caught and corrected a snapshot assertion that could pass for an unrelated reason.

Validation: formatting, full clean compile and the integrated core profiles passed: 179 tests, 177 passed, 2 skipped, no failures. One earlier compaction-lock test failed on a timing-sensitive state assertion; it passed on clean master and in the final integrated run. New CI is running; no approval or merge is implied.

Remaining boundaries: mocked Node/LogManager, no multi-node restart validation, and no rollback of partial business writes. Store's existing automatic restart can repeat a deterministic failure; PD's existing leader-stop callback can overwrite the ERROR probe label while readiness remains false. These recovery/probe issues remain outside this PR (#3222). Doc - No Need: internal error handling, no API/configuration change.

imbajin:
Centralized the existing jRaft version declarations in the root POM without upgrading any production component: Server remains on 1.3.11; PD and Store remain on 1.3.13. Store tests reference the same property as their production component. All 18 resolved jRaft dependency entries are identical before and after this declaration refactor.

A single-version upgrade is explicitly deferred to #3239, with a TODO in the root POM. Compatibility must be established separately; the current tests are not evidence that upgrading Server is safe. Full clean compile and the 10 relevant PD/Store tests passed after this change.

imbajin:
Fixed the Store core CI failure in HgSnapshotHandlerTest. The test treated compactionDone as a terminal state, but the subsequent successful snapshot resets it to compactionCanStart. In the failed CI log, compaction finished at 16:14:54.001 and the successful snapshot released the lock at 16:14:54.024; the test polled every 50 ms and missed that transient state.

The test now directly waits for and asserts path-lock release after interruption, while confirming the snapshot range lock is still owned. The redundant second compaction and transient-state assertion are removed; production code is unchanged. A negative control that removes unlock(path) fails at the retained lock-release assertion, demonstrating that the original regression is still detected.

I had observed the same failure locally earlier but accepted passing reruns without resolving this race. That validation gap is now addressed.

Validation: full clean compile, formatting and the Store core profile passed (37 Store tests, zero failures). The targeted negative control failed as expected.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/raft/PartitionStateMachine.java
```diff
@@ -105,6 +105,9 @@ public void onApply(Iterator iter) {
                               done.getOperation().getOp(),
                               done.getOperation().getReq());
                 }
+                iter.setErrorAndRollback(1, new Status(RaftError.ESTATEMACHINE, t.getMessage()));
```
imbajin: ⚠️ This catch also covers `done.run(Status.OK())` in the leader path. The `Change_Shard` handler updates its task and starts follow-up work before invoking the response closure, so a thrown callback is treated as an apply failure after side effects have started and can fail-stop this FSM. Please keep callback exceptions out of the rollback path.

imbajin: ⚠️ This records a terminal state-machine error for the committed entry, but `PartitionEngine.onError()` immediately restarts the engine; `shutdown()` keeps the same log/meta directories that `init()` reopens. A deterministic failure will replay the entry and restart again, leaving the partition unavailable. Please stop automatic restart for this error or add bounded recovery that does not replay the poison entry.

imbajin (commented):
Blocking: yes. Summary: A deterministic apply failure can restart and replay the same committed entry, and a completion callback exception can also fail-stop the FSM after work has started. Evidence: exact-head state-machine and PartitionEngine source plus JRaft 1.3.13 rollback control flow.

Issue #2209, opened 2023-05-13
[Bug] The backend store version is inconsistent


lanerbean:
### Bug Type (问题类型)

server status (启动/运行异常)

### Before submit

- [X] 我已经确认现有的 [Issues](https://github.com/apache/hugegraph/issues) 与 [FAQ](https://hugegraph.apache.org/docs/guides/faq/) 中没有相同 / 重复问题 (I have confirmed and searched that there are no similar problems in the historical issue and documents)

### Environment (环境信息)

- Server Version: 1.0.0 (Apache Release Version)
- Backend: MySQL
- OS: CentOS 7.x 

### Expected & Actual behavior (期望与实际表现)

When the HugeGraph Server and Hadoop cluster are shut down and restarted, an error is encountered, causing the startup to fail.
![error1](https://github.com/apache/incubator-hugegraph/assets/18582787/0e2af46d-5a07-4df4-b657-5b8f1252f725)

imbajin:
U could add & print the log (what's the differ version) and how to reproduce it?

imbajin:
> > U could add & print the log (what's the differ version) and how to reproduce it?
> 
> 您好，我是在3台虚拟机上部署的Hadoop集群，并在其中一台机器上部署了HugeGraph，每次关闭HugeGraph，Hadoop集群，并且关闭虚拟机之后，第二天重启Hadoop集群，再开启HugeGraph Server时，就会出现图数据版本不一致的问题。
> 
> 我之前采用的解决方法非常暴力，直接删除了MySQL（后台图数据库）的hugegraph，然后在MySQL重新创建了hugegraph数据库，重新init和启动了HugeGraph Server。
> 
> 这种方法，会导致之前导入的图数据都会丢失，需要重新导入，很不方便。不知道该如何回避这种问题的出现。

我理解 `hadoop` 环境在这里应该和图是基本没有关系的吧, mysql 来说, 一台单独的虚拟机甚至是容器环境应该也都足矣了, 能否保持一个 clean 的环境试试呢? 或者你甚至可以在  server 的 docker 镜像中配置一下 mysql, 然后看看是否会有类似问题

lanerbean:
是不是因为HugeGraph和MySQL不在同一台机器上导致的呢？并且MySQL也是部署在虚拟机上的，应该也是关闭过服务又重启的

robinhood-jim:
[...]

PR #2888, opened 2025-10-24
fix(server): prevent graph clear API from clearing meta table in MySQL backend storage

LYD031106:
## Purpose of the PR

- close [[Bug] The backend store version is inconsistent #2209](https://github.com/apache/incubator-hugegraph/issues/2209)

## Main Changes

The PR fixes the issue where the graph clear API mistakenly clears the meta table in MySQL backend storage. Since version 1.7.0, HugeGraph only supports "hstore, rocksdb, hbase, and memory" as backend storage options. Future checks will be performed to ensure that similar issues do not exist in these supported backends.

## Verifying these changes

    - xxx

- [ ]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/mysql/BaseMysqlUnitTest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/mysql/BaseMysqlUnitTest.java
@@ -24,6 +24,7 @@ import org.apache.hugegraph.config.HugeConfig;
 import org.apache.hugegraph.testutil.Utils;
 import org.apache.hugegraph.unit.BaseUnitTest;
 import org.junit.After;
+import org.junit.Assume;
 
 public class BaseMysqlUnitTest extends BaseUnitTest {
 
@@ -35,6 +36,10 @@ public class BaseMysqlUnitTest extends BaseUnitTest {
 
     public void setup() throws Exception {
         Configuration conf = Utils.getConf();
+        String backend = conf.getString("backend", "memory");
+        // Only run mysql related tests when backend is mysql
+        Assume.assumeTrue("Skip MySQL tests when backend is not mysql",
+                          "mysql".equalsIgnoreCase(backend));
         this.config = new HugeConfig(conf);
         this.provider = new MysqlStoreProvider();
         this.provider.open(GRAPH_NAME);
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/UnitTestSuite.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/UnitTestSuite.java
@@ -132,7 +132,6 @@ import org.junit.runners.Suite;
         /* mysql */
         MysqlUtilTest.class,
         WhereBuilderTest.class,
-        BaseMysqlUnitTest.class,
         MysqlTest.class,
 
         /* rocksdb */


hugegraph-server/hugegraph-mysql/src/main/java/org/apache/hugegraph/backend/store/mysql/MysqlStore.java
```diff
@@ -223,6 +223,7 @@ public void truncate() {
         this.checkOpened();
 
         this.truncateTables();
+        this.init();
```
imbajin: ‼️ **Critical: Missing test coverage for the fix**

This fix addresses a critical bug where the meta table gets cleared during graph truncation, but there's no test to verify the fix works correctly or prevent regression.

**Suggested action:**
Add a test that:
1. Verifies the meta table is properly initialized after truncate()
2. Checks that the backend version is still readable after truncation
3. Ensures other metadata persists correctly

Example test structure:
```java
@Test
public void testTruncatePreservesMetaTable() {
    // Write version to meta table
    store.init();
    String versionBefore = sessions.systemStore().storedVersion();
    
    // Truncate should not lose meta table
    store.truncate();
    String versionAfter = sessions.systemStore().storedVersion();
    
    Assert.assertEquals(versionBefore, versionAfter);
}
```

imbajin: ⚠️ **Important: Potential inefficiency in fix approach**

While calling `init()` after `truncateTables()` fixes the meta table issue, this approach re-initializes ALL tables, not just the meta table. This may be inefficient.

**Analysis:**
- `init()` calls `initTables()` which iterates through all tables and calls `table.init(session)` for each
- After truncation, only the meta table needs to be re-initialized
- Other backend stores (RocksDB, HBase) don't seem to require this step

**Suggested improvement:**
Consider a more targeted fix that only re-initializes the meta table:
```java
public void truncate() {
    this.checkOpened();
    
    this.truncateTables();
    // Only re-init the meta table instead of all tables
    this.initMetaTable();
    LOG.debug("Store truncated: {}", this.store);
}

private void initMetaTable() {
    Session session = this.sessions.session();
    this.sessions.systemStore().meta.init(session);
}
```

This would be more efficient and clearer in intent.

imbajin: ⚠️ **Important: Check consistency with other backend stores**

The PR description mentions that "HugeGraph only supports hstore, rocksdb, hbase, and memory" but this fix is only applied to MySQL backend.

**Questions to verify:**
1. Do RocksDB, HBase, and Memory backends have the same issue? 
2. From my quick check, HBase's `truncate()` method doesn't call `init()` after truncating tables - why is MySQL different?
3. Is there a fundamental difference in how MySQL handles the meta table vs other backends?

**Recommendation:**
- Document WHY MySQL requires this init() call while other backends don't
- Verify other supported backends (RocksDB, HBase) don't have similar issues
- Consider adding a comment explaining the MySQL-specific requirement:
```java
this.truncateTables();
// MySQL-specific: Re-initialize tables to restore meta table
// Other backends handle meta table separately during truncation
this.init();
```

imbajin: 🧹 **Minor: Consider side effects of calling init()**

The `init()` method does more than just `initTables()`:
- It calls `checkClusterConnected()`
- Creates database if not exists (`sessions.createDatabase()`)
- Opens a new session
- Then calls `checkOpened()` again

**Observation:**
Most of these operations are already done and may be redundant when called from `truncate()` (which already calls `checkOpened()` first).

**Potential issue:**
If there are any idempotency concerns with these operations being called multiple times, this could introduce subtle bugs.

**Suggestion:**
Review whether all operations in `init()` are safe to call multiple times during the store lifecycle, or extract just the table initialization logic into a separate method.

LYD031106: **Root Cause**
The issue stems from the fact that backend stores such as MySQL, HBase, and RocksDB persist metadata (e.g., `driverVersion`) to the backend storage. When the `truncate()` operation is executed to clear all table data, this metadata is also deleted. As a result, subsequent operations fail due to missing metadata or state conflicts.

**Analysis by Backend**
- **RocksDB**:  
  Although affected by the same problem, its `truncate()` method explicitly calls `this.init()` after clearing data to reinitialize the metadata, thereby avoiding errors.
```
    @Override
    public synchronized void truncate() {
        Lock writeLock = this.storeLock.writeLock();
        writeLock.lock();
        try {
            this.checkOpened();

            this.clear(false);
            this.init();
            // Clear write-batch
            this.dbs.values().forEach(BackendSessionPool::forceResetSessions);
            LOG.debug("Store truncated: {}", this.store);
        } finally {
            writeLock.unlock();
        }
    }
```
- **HBase**:  
  Confirmed to have the same issue as MySQL.
Action Required: A follow-up PR is needed to fix the HBase backend accordingly.
- **Memory / HStore**:  
  These backends are not affected because they do not persist critical version information or similar metadata to the backend storage.

imbajin: > **Root Cause** The issue stems from the fact that backend stores such as MySQL, HBase, and RocksDB persist metadata (e.g., `driverVersion`) to the backend storage. When the `truncate()` operation is executed to clear all table data, this metadata is also deleted. As a result, subsequent operations fail due to missing metadata or state conflicts.
> 
> **Analysis by Backend**
> 
> * **RocksDB**:
>   Although affected by the same problem, its `truncate()` method explicitly calls `this.init()` after clearing data to reinitialize the metadata, thereby avoiding errors.
> 
> ```
>     @Override
>     public synchronized void truncate() {
>         Lock writeLock = this.storeLock.writeLock();
>         writeLock.lock();
>         try {
>             this.checkOpened();
> 
>             this.clear(false);
>             this.init();
>             // Clear write-batch
>             this.dbs.values().forEach(BackendSessionPool::forceResetSessions);
>             LOG.debug("Store truncated: {}", this.store);
>         } finally {
>             writeLock.unlock();
>         }
>     }
> ```
> 
> * **HBase**:
>   Confirmed to have the same issue as MySQL.
>   Action Required: A follow-up PR is needed to fix the HBase backend accordingly.
> * **Memory / HStore**:
>   These backends are not affected because they do not persist critical version information or similar metadata to the backend storage.

@LYD031106 LGTM, we could add a basic test & comment for the above issue (and we could merge this PR soon)

BTW, this week we prepare to release the new version (1.7.0)

LYD031106: I have submitted a MySQL compatibility fix to the `release-1.5.0` branch. Since this is a legacy compatibility patch, I will submit a separate PR to the master branch to address the HBase storage issues.
Regarding the addition of test cases: I’ve noticed that `hugegraph-test` currently lacks backend-specific tests for `HBase` and `MySQL`. I plan to add a minimal set of regression tests. However, I’m not very familiar with how to build such backend-specific test cases. Could you please provide some guidance or reference examples?

imbajin: > I have submitted a MySQL compatibility fix to the `release-1.5.0` branch. Since this is a legacy compatibility patch, I will submit a separate PR to the master branch to address the HBase storage issues. Regarding the addition of test cases: I’ve noticed that `hugegraph-test` currently lacks backend-specific tests for `HBase` and `MySQL`. I plan to add a minimal set of regression tests. However, I’m not very familiar with how to build such backend-specific test cases. Could you please provide some guidance or reference examples?

refer ci steps: https://github.com/apache/incubator-hugegraph/actions/runs/18783126996/workflow?pr=2888#L49
image

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/mysql/BaseMysqlUnitTest.java
imbajin, line 1: why we need this class? (seems a little redundant)

hugegraph-server/hugegraph-mysql/src/main/java/org/apache/hugegraph/backend/store/mysql/MysqlStore.java
```diff
@@ -337,6 +337,9 @@ protected void clearTables() {
     protected void truncateTables() {
         Session session = this.sessions.session();
         for (MysqlTable table : this.tables()) {
+            if (table instanceof MysqlTables.Meta || table instanceof MysqlTables.Counters){
+                continue;
```
imbajin: ‼️ **Critical: Fragile implementation with instanceof checks**

The current fix uses `instanceof` checks to skip specific table types during truncation. This approach has several issues:

1. **Tight coupling**: Creates dependency on concrete implementation classes rather than using a more abstract marker (e.g., an interface like `MetaTable` or a table property)
2. **Not extensible**: If new meta tables are added in the future, developers must remember to update this condition
3. **Violates Open/Closed Principle**: The store layer shouldn't need to know about specific table implementations

```suggestion
// Consider adding a method to MysqlTable interface like:
// public boolean isSystemTable() { return false; }
// Then override in Meta and Counters tables
if (table.isSystemTable()) {
    continue;
}
```

Alternatively, you could maintain a set of system tables or use table metadata to mark which tables should be preserved.

hugegraph-server/hugegraph-mysql/src/main/java/org/apache/hugegraph/backend/store/mysql/MysqlStore.java
```diff
@@ -337,6 +337,9 @@ protected void clearTables() {
     protected void truncateTables() {
         Session session = this.sessions.session();
         for (MysqlTable table : this.tables()) {
+            if (table instanceof MysqlTables.Meta || table instanceof MysqlTables.Counters){
```
imbajin: ‼️ **Critical: Unclear why Counters table is preserved**

The PR description mentions fixing Meta table truncation but doesn't explain why the Counters table should also be skipped. This needs clarification:

1. Is this fixing an existing bug with Counters table? (not mentioned in issue #2209)
2. What data does the Counters table hold that should survive truncation?
3. Should Counters be reset during truncate but preserved during clear?

Please add a comment explaining the reasoning, or if this is a separate fix, consider splitting it into a separate PR for clarity.

hugegraph-server/hugegraph-mysql/src/main/java/org/apache/hugegraph/backend/store/mysql/MysqlStore.java
```diff
@@ -223,6 +222,7 @@ public void truncate() {
         this.checkOpened();
 
         this.truncateTables();
+        this.init();
```
imbajin: ⚠️ **Important: Verify init() is necessary after truncate**

Adding `this.init()` after truncation appears to reinitialize the meta table. However:

1. Does `init()` have any side effects beyond restoring meta data?
2. Is this call idempotent and safe to call multiple times?
3. Should this also be added to `clearTables()` method for consistency?

Please verify this is the correct approach and consider adding a comment explaining why re-initialization is needed after truncate.

hugegraph-server/hugegraph-mysql/src/main/java/org/apache/hugegraph/backend/store/mysql/MysqlStore.java
```diff
@@ -54,7 +54,6 @@ public abstract class MysqlStore extends AbstractBackendStore<Session> {
     private final BackendStoreProvider provider;
 
     private final Map<HugeType, MysqlTable> tables;
```
imbajin, line 57: 🧹 **Minor: Unnecessary whitespace change**

This blank line removal is an unrelated formatting change. Please keep formatting changes separate from functional fixes to make code reviews easier.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/mysql/MysqlTest.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.hugegraph.unit.mysql;
+
+import org.apache.hugegraph.backend.store.BackendStore;
+import org.apache.hugegraph.testutil.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class MysqlTest extends BaseMysqlUnitTest{
+    @Before
+    public void setUp(){
```
imbajin: ⚠️ **Important: Inconsistent test lifecycle annotations**

The `setUp()` method in `MysqlTest` calls `super.setup()`, but the base class method `setup()` doesn't have `@Before` annotation. This means:

1. The base class setup runs for every test in MysqlTest (due to @Before here)
2. But may not run correctly for other test classes extending BaseMysqlUnitTest
3. The naming inconsistency (`setup` vs `setUp`) is confusing

```suggestion
// Either add @Before to the base class setup() method, or rename for consistency:
@Before
public void setup() {
    super.setup();
}
```

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/mysql/MysqlTest.java
```diff
@@ -0,0 +1,43 @@
...
+        super.setup();
+    }
+
+    @Test
```
imbajin, line 36: ⚠️ **Important: Incomplete test coverage**

The test only verifies that the version is preserved, but doesn't verify:

1. **That data in non-meta tables IS actually cleared** - The bug could still exist if truncate silently fails
2. **That the Counters table content is preserved** (if that's intentional)
3. **That schema data is cleared** - The fix might have side effects on other stores

Suggested additions:
```java
@Test
public void testMysqlTruncateClearsDataButPreservesMeta() {
    // 1. Insert some test data into graph store
    // 2. Record meta version before truncate
    // 3. Call truncate
    // 4. Verify meta version unchanged
    // 5. Verify test data was actually cleared
    // 6. Verify Counters table state (if needed)
}
```

hugegraph-server/hugegraph-mysql/src/main/java/org/apache/hugegraph/backend/store/mysql/MysqlStore.java
LYD031106: Sorry, this was my misjudgment: I mistakenly thought the Counters table was a system table similar to the Meta table, so I included it in the truncation-skip logic.

imbajin (approved):
Merge it first (maybe we need handle the ci/code problem in another PR) @LYD031106 ?

PR #2916, opened 2025-12-05
chore: update the status of distributed modules

imbajin:
update for release 1.7


hugegraph-server/hugegraph-dist/src/assembly/static/conf/rest-server.properties
VGalaxies, line 1: @imbajin seems that modifying this file caused the required CI to fail 🤔

imbajin: > @imbajin seems that modifying this file caused the required CI to fail 🤔

Yes, due to some hard-code`sed` usage in legacy code.... I'll fix it soon 😢

VGalaxies (approved):

PR #2918, opened 2025-12-15
fix: optimize code and update risky deps

kenssa4eedfd:
## Purpose of the PR

- optimize code and update dependencies

## Verifying these changes

- [ ]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/StringEncoding.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/StringEncoding.java
@@ -36,6 +36,7 @@ public final class StringEncoding {
     private static final byte[] BYTES_EMPTY = new byte[0];
     private static final String STRING_EMPTY = "";
     private static final int BLOCK_SIZE = 4096;
+    private static final int BCRYPT_WORK_FACTOR = 10;
 
     static {
         final String ALG = "SHA-256";
@@ -142,7 +143,7 @@ public final class StringEncoding {
     public static String hashPassword(String password) {
         // OWASP suggests 10 as a minimum and 12–14 for production;
         // workFactor 12 is not used by default due to its 200+ ms cost.
-        return BCrypt.hashpw(password, BCrypt.gensalt(10));
+        return BCrypt.hashpw(password, BCrypt.gensalt(BCRYPT_WORK_FACTOR));
     }
 
     public static boolean checkPassword(String candidatePassword, String dbPassword) {
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/StringEncodingTest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/StringEncodingTest.java
@@ -198,6 +198,24 @@ public class StringEncodingTest {
 
         // Test that hashPassword uses the new cost factor
         String hashedPassword = StringEncoding.hashPassword(testPassword);
-        Assert.assertTrue(StringEncoding.checkPassword(testPassword, hashedPassword));
+        Assert.assertTrue("Hash should contain work factor 10",
+                          hashedPassword.matches("^\\$2[aby]\\$10\\$.*")
+        );
+
+        // Compare computational cost between work factor 4 and 10
+        long start4 = System.nanoTime();
+        StringEncoding.checkPassword(testPassword, oldPassword);
+        long elapsed4 = System.nanoTime() - start4;
+
+        long start10 = System.nanoTime();
+        StringEncoding.checkPassword(testPassword, hashedPassword);
+        long elapsed10 = System.nanoTime() - start10;
+
+        // BCrypt cost difference: (10-4) = 6 => theoretical ~2^6 = 64x
+        Assert.assertTrue(
+                "Work factor 10 should be significantly slower than work factor 4 " +
+                "(expected exponential cost increase)",
+                elapsed10 >= elapsed4 * 32
+        );
     }
 }
--- a/hugegraph-struct/src/main/java/org/apache/hugegraph/util/StringEncoding.java
+++ b/hugegraph-struct/src/main/java/org/apache/hugegraph/util/StringEncoding.java
@@ -56,6 +56,8 @@ public final class StringEncoding {
     private static final byte[] BYTES_EMPTY = new byte[0];
     private static final int BLOCK_SIZE = 4096;
 
+    private static final int BCRYPT_WORK_FACTOR = 10;
+
     static {
         final String ALG = "SHA-256";
         try {
@@ -167,7 +169,7 @@ public final class StringEncoding {
     public static String hashPassword(String password) {
         // OWASP suggests 10 as a minimum and 12–14 for production;
         // workFactor 12 is not used by default due to its 200+ ms cost.
-        return BCrypt.hashpw(password, BCrypt.gensalt(10));
+        return BCrypt.hashpw(password, BCrypt.gensalt(BCRYPT_WORK_FACTOR));
     }
 
     public static boolean checkPassword(String candidatePassword,
--- a/install-dist/release-docs/LICENSE
+++ b/install-dist/release-docs/LICENSE
@@ -649,8 +649,6 @@ The text of each license is also included in licenses/LICENSE-[project].txt.
     https://central.sonatype.com/artifact/org.lionsoul/jcseg-core/2.2.0 -> Apache 2.0
     https://central.sonatype.com/artifact/org.lionsoul/jcseg-core/2.6.2 -> Apache 2.0
     https://central.sonatype.com/artifact/org.lz4/lz4-java/1.4.0 -> Apache 2.0
-    https://central.sonatype.com/artifact/org.lz4/lz4-java/1.7.1 -> Apache 2.0
-    https://central.sonatype.com/artifact/org.lz4/lz4-java/1.8.0 -> Apache 2.0
     https://central.sonatype.com/artifact/org.lz4/lz4-java/1.8.1 -> Apache 2.0
     https://central.sonatype.com/artifact/org.nlpcn/nlp-lang/1.7.7 -> Apache 2.0
     https://central.sonatype.com/artifact/org.objenesis/objenesis/2.6 -> Apache 2.0


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
```diff
@@ -173,6 +173,21 @@ private static Path zipSlipProtect(ArchiveEntry entry, Path targetDir)
...
+            throws IOException {
+        Path targetDirResolved = targetDir.resolve(entry.getName());
+        /*
+         * Make sure normalized file still has targetDir as its prefix,
+         * else throws exception
+         */
+        Path normalizePath = targetDirResolved.normalize();
+        if (!normalizePath.startsWith(targetDir.normalize())) {
+            throw new IOException(String.format("Bad entry: %s",
+                                                entry.getName()));
```
imbajin: maybe merge into one line?

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/StringEncoding.java
```diff
@@ -140,7 +140,7 @@ public static String decompress(byte[] value, float bufferRatio) {
     }
 
     public static String hashPassword(String password) {
-        return BCrypt.hashpw(password, BCrypt.gensalt(4));
+        return BCrypt.hashpw(password, BCrypt.gensalt(12));
```
imbajin: maybe use the minimum factor in the proper range (to reduce perf influence, like 10?)

hugegraph-server/hugegraph-core/pom.xml
```diff
@@ -198,7 +198,7 @@
             <version>${commons-compress.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.lz4</groupId>
```
imbajin, line 201: expected change? (org changed?)

hugegraph-server/hugegraph-core/pom.xml
```diff
@@ -32,7 +32,7 @@
         <jraft.version>1.3.11</jraft.version>
         <ohc.version>0.7.4</ohc.version>
         <jna.version>5.12.1</jna.version>
-        <lz4.version>1.8.0</lz4.version>
+        <lz4.version>1.9.0</lz4.version>
```
imbajin: in ASF rules, we need to update the dependencies info

refer: https://hugegraph.apache.org/docs/contribution-guidelines/contribute/#321-check-licenses

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
```diff
@@ -173,6 +173,21 @@ private static Path zipSlipProtect(ArchiveEntry entry, Path targetDir)
         return normalizePath;
     }
 
+    private static Path zipSlipProtect(ZipEntry entry, Path targetDir)
+            throws IOException {
```
imbajin, line 167: one line?

hugegraph-server/hugegraph-core/pom.xml
```diff
@@ -198,7 +198,7 @@
             <version>${commons-compress.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.lz4</groupId>
+            <groupId>at.yawk.lz4</groupId>
             <artifactId>lz4-java</artifactId>
```
imbajin, line 204: ‼️ **Critical: Incorrect GroupId for lz4-java**

The dependency change switches from `org.lz4:lz4-java` to `at.yawk.lz4:lz4-java`, but this appears to be incorrect. The official lz4-java artifact is maintained under the `org.lz4` groupId.

The `at.yawk.lz4` groupId points to an unofficial fork that may not be maintained and could introduce compatibility or security issues.

**Recommendation:** Keep the original `org.lz4` groupId and only update the version to 1.9.0:

```suggestion
        <dependency>
            <groupId>org.lz4</groupId>
            <artifactId>lz4-java</artifactId>
            <version>${lz4.version}</version>
        </dependency>
```

hugegraph-struct/pom.xml
```diff
@@ -109,9 +109,9 @@
             <version>8.1.0</version>
         </dependency>
         <dependency>
-            <groupId>org.lz4</groupId>
+            <groupId>at.yawk.lz4</groupId>
             <artifactId>lz4-java</artifactId>
```
imbajin, line 115: ‼️ **Critical: Incorrect GroupId for lz4-java (hugegraph-struct)**

Same issue as in hugegraph-core - using an unofficial fork `at.yawk.lz4` instead of the official `org.lz4` groupId.

```suggestion
        <dependency>
            <groupId>org.lz4</groupId>
            <artifactId>lz4-java</artifactId>
            <version>1.9.0</version>
        </dependency>
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/StringEncoding.java
imbajin: ⚠️ **Important: BCrypt Work Factor Change Needs Migration Strategy**

Increasing BCrypt rounds from 4 to 12 is a good security improvement, but this will break authentication for all existing users whose passwords were hashed with the old work factor (4 rounds).

**Impact:** Existing users won't be able to log in after this change.

**Recommendation:** Implement a gradual migration strategy:
1. Keep verifying old hashes (4 rounds) in `checkPassword()`
2. Re-hash passwords with 12 rounds on successful login
3. Consider adding a configuration option for the work factor

Example approach:
```java
public static String hashPassword(String password) {
    // New hashes use 12 rounds
    return BCrypt.hashpw(password, BCrypt.gensalt(12));
}

public static boolean checkPassword(String candidatePassword, String dbPassword) {
    boolean valid = BCrypt.checkpw(candidatePassword, dbPassword);
    // TODO: Consider re-hashing with new work factor on successful login
    return valid;
}
```

hugegraph-struct/src/main/java/org/apache/hugegraph/util/StringEncoding.java
```diff
@@ -165,7 +165,7 @@ public static String decompress(byte[] value, float bufferRatio) {
     }
 
     public static String hashPassword(String password) {
-        return BCrypt.hashpw(password, BCrypt.gensalt(4));
+        return BCrypt.hashpw(password, BCrypt.gensalt(12));
```
imbajin: ⚠️ **Important: BCrypt Work Factor Change (hugegraph-struct)**

Same BCrypt work factor change as in hugegraph-core. This will affect password compatibility between modules.

Ensure both modules handle password migration consistently. See the comment on hugegraph-core/StringEncoding.java for migration strategy details.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
```diff
@@ -173,6 +173,21 @@ private static Path zipSlipProtect(ArchiveEntry entry, Path targetDir)
         return normalizePath;
     }
 
+    private static Path zipSlipProtect(ZipEntry entry, Path targetDir)
```
imbajin, line 166: ⚠️ **Code Quality: Redundant zipSlipProtect Method**

The new `zipSlipProtect(ZipEntry, Path)` method duplicates the existing `zipSlipProtect(ArchiveEntry, Path)` method's logic. Both perform identical path traversal validation.

**Recommendation:** Extract common logic to avoid duplication:

```suggestion
    private static Path zipSlipProtect(String entryName, Path targetDir)
            throws IOException {
        Path targetDirResolved = targetDir.resolve(entryName);
        Path normalizePath = targetDirResolved.normalize();
        if (!normalizePath.startsWith(targetDir.normalize())) {
            throw new IOException(String.format("Bad entry: %s", entryName));
        }
        return normalizePath;
    }

    private static Path zipSlipProtect(ArchiveEntry entry, Path targetDir)
            throws IOException {
        return zipSlipProtect(entry.getName(), targetDir);
    }

    private static Path zipSlipProtect(ZipEntry entry, Path targetDir)
            throws IOException {
        return zipSlipProtect(entry.getName(), targetDir);
    }
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
```diff
@@ -220,9 +235,7 @@ public static void decompressZip(String sourceFile, String outputDir,
              ZipInputStream zis = new ZipInputStream(bis)) {
             ZipEntry entry;
             while ((entry = zis.getNextEntry()) != null) {
-                String fileName = entry.getName();
-                File entryFile = new File(Paths.get(outputDir, fileName)
-                                               .toString());
+                File entryFile = new File(zipSlipProtect(entry, Paths.get(outputDir)).toString());
```
imbajin: ✅ **Good: Zip Slip Protection Applied**

Good security improvement! Applying `zipSlipProtect()` to the ZipEntry path prevents directory traversal attacks during ZIP decompression.

Minor style suggestion for consistency with existing code pattern:
```suggestion
                Path targetPath = zipSlipProtect(entry, Paths.get(outputDir));
                File entryFile = new File(targetPath.toString());
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/StringEncoding.java
imbajin: Also, we **lack tests** here to avoid any data `breaking change` in our system

hugegraph-server/hugegraph-core/pom.xml
kenssa4eedfd: You are correct, the official latest version, 1.8.1, also includes the security implementation fix. I will switch to using version 1.8.1 from the official Group ID org.lz4.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
kenssa4eedfd: I will switch to a more elegant implementation, modifying the method to:
```
private static Path zipSlipProtect(String fileName, Path targetDir)
            throws IOException {
        Path targetDirResolved = targetDir.resolve(fileName);
        /*
         * Make sure normalized file still has targetDir as its prefix,
         * else throws exception
         */
        Path normalizePath = targetDirResolved.normalize();
        if (!normalizePath.startsWith(targetDir.normalize())) {
            throw new IOException(String.format("Bad entry: %s",
                                                fileName));
        }
        return normalizePath;
    }
```
merge the two methods.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
```diff
@@ -158,17 +158,17 @@ public static void decompressTar(String sourceFile, String outputDir,
         }
     }
 
-    private static Path zipSlipProtect(ArchiveEntry entry, Path targetDir)
+    private static Path zipSlipProtect(String fileName, Path targetDir)
```
imbajin: ‼️ **Critical: Zip slip protection weakened**

The refactoring changes `zipSlipProtect` to accept `String fileName` instead of `ArchiveEntry entry`, but this removes access to other entry properties that might be needed for security validation. While the current implementation only uses `entry.getName()`, accepting the full entry object is a better practice for:

1. Future security enhancements that might need to validate other entry properties
2. Consistency with security-focused APIs that typically pass full objects

Consider keeping the `ArchiveEntry` parameter:

```suggestion
    private static Path zipSlipProtect(ArchiveEntry entry, Path targetDir)
            throws IOException {
        Path targetDirResolved = targetDir.resolve(entry.getName());
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
```diff
@@ -220,9 +220,7 @@ public static void decompressZip(String sourceFile, String outputDir,
              ZipInputStream zis = new ZipInputStream(bis)) {
             ZipEntry entry;
             while ((entry = zis.getNextEntry()) != null) {
-                String fileName = entry.getName();
-                File entryFile = new File(Paths.get(outputDir, fileName)
-                                               .toString());
+                File entryFile = new File(zipSlipProtect(entry.getName(), Paths.get(outputDir)).toString());
```
imbajin: ⚠️ **Important: Code formatting inconsistency**

This line is too long and hard to read. The original multi-line format was more maintainable:

```suggestion
                String fileName = entry.getName();
                Path entryPath = zipSlipProtect(fileName, Paths.get(outputDir));
                File entryFile = new File(entryPath.toString());
```

This improves:
- Readability by keeping lines under 100 characters
- Debuggability by allowing breakpoints on each step
- Maintainability by separating concerns

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/StringEncoding.java
```diff
@@ -140,7 +140,8 @@ public static String decompress(byte[] value, float bufferRatio) {
     }
 
     public static String hashPassword(String password) {
-        return BCrypt.hashpw(password, BCrypt.gensalt(4));
+        // OWASP suggest 10 as minimum and 12-14 as production default
+        return BCrypt.hashpw(password, BCrypt.gensalt(12));
```
imbajin: ‼️ **Critical: Breaking change for existing users**

Increasing BCrypt cost factor from 4 to 12 is a **256x increase** in computation time (2^8). While 12 is the recommended value, this change will:

1. **Break existing password hashes**: Old passwords hashed with cost=4 won't match if the application re-hashes on login
2. **Impact performance**: Password operations will be significantly slower
3. **Require migration strategy**: Need a plan to rehash existing passwords

Recommendation:
```suggestion
    public static String hashPassword(String password) {
        // OWASP recommends 12-14 for production, but use 10 as default to balance security and performance
        // TODO: Implement gradual migration from legacy cost=4 hashes
        return BCrypt.hashpw(password, BCrypt.gensalt(10));
    }
```

Add a migration path:
1. Check hash cost on login using `BCrypt.getRounds()`
2. If cost < 10, rehash with new cost
3. Document the breaking change in release notes

hugegraph-struct/src/main/java/org/apache/hugegraph/util/StringEncoding.java
```diff
@@ -165,7 +165,8 @@ public static String decompress(byte[] value, float bufferRatio) {
     }
 
     public static String hashPassword(String password) {
-        return BCrypt.hashpw(password, BCrypt.gensalt(4));
+        // OWASP suggest 10 as minimum and 12-14 as production default
+        return BCrypt.hashpw(password, BCrypt.gensalt(12));
```
imbajin: ⚠️ **Important: Duplicate code across modules**

The same BCrypt cost change is duplicated in `hugegraph-struct/src/main/java/org/apache/hugegraph/util/StringEncoding.java`. This creates maintenance burden.

Consider:
1. Consolidating the implementation in one place
2. Making the cost factor configurable
3. Adding a constant: `private static final int BCRYPT_COST_FACTOR = 12;`

This would centralize the security policy and make future updates easier.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/StringEncodingTest.java
```diff
@@ -180,4 +181,18 @@ public void testReadAsciiString() {
...
+
+    @Test
+    public void testCheckPasswordSupportsOldAndNewCost() {
+        // oldWorkFactor
+        String oldPassword = BCrypt.hashpw("123456", BCrypt.gensalt(4));
+        // newWorkFactor
+        String newPassword = BCrypt.hashpw("123456", BCrypt.gensalt(12));
+
+        Assert.assertTrue(StringEncoding.checkPassword("123456", oldPassword));
+        Assert.assertTrue(StringEncoding.checkPassword("123456", newPassword));
```
imbajin: 🧹 **Minor: Test coverage could be improved**

The test validates backward compatibility but could be more comprehensive:

```suggestion
    @Test
    public void testCheckPasswordSupportsOldAndNewCost() {
        String testPassword = "test123!@#";
        
        // Test old work factor (4)
        String oldPassword = BCrypt.hashpw(testPassword, BCrypt.gensalt(4));
        Assert.assertTrue(StringEncoding.checkPassword(testPassword, oldPassword));
        Assert.assertFalse(StringEncoding.checkPassword("wrong", oldPassword));
        
        // Test new work factor (12)
        String newPassword = BCrypt.hashpw(testPassword, BCrypt.gensalt(12));
        Assert.assertTrue(StringEncoding.checkPassword(testPassword, newPassword));
        Assert.assertFalse(StringEncoding.checkPassword("wrong", newPassword));
        
        // Test that hashPassword uses the new cost factor
        String hashedPassword = StringEncoding.hashPassword(testPassword);
        Assert.assertTrue(StringEncoding.checkPassword(testPassword, hashedPassword));
        // Verify new hashes use cost=12 (though this is implementation-dependent)
    }
```

install-dist/release-docs/LICENSE
```diff
@@ -651,6 +651,7 @@ The text of each license is also included in licenses/LICENSE-[project].txt.
     https://central.sonatype.com/artifact/org.lz4/lz4-java/1.4.0 -> Apache 2.0
     https://central.sonatype.com/artifact/org.lz4/lz4-java/1.7.1 -> Apache 2.0
     https://central.sonatype.com/artifact/org.lz4/lz4-java/1.8.0 -> Apache 2.0
```
imbajin: remove 1.8.0?

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManager.java
```diff
@@ -580,6 +580,7 @@ public HugeUser matchUser(String name, String password) {
         }
 
         if (StringEncoding.checkPassword(password, user.password())) {
+            // TODO: rehash password if bcrypt work factor is lower than expected
```
imbajin, line 583: **‼️ Critical: Missing security upgrade for existing users**

The bcrypt work factor is being increased from 4 to 10, but there's no migration strategy for existing password hashes. The TODO comment suggests awareness of this issue but provides no implementation.

**Problem:**
- Existing users with old hashes (work factor 4) remain vulnerable
- No automatic rehashing on successful login
- No admin tool to force password rehash

**Suggestion:**
Implement automatic password upgrade on successful authentication:

```suggestion
        if (StringEncoding.checkPassword(password, user.password())) {
            // Rehash password if bcrypt work factor is lower than expected
            if (requiresRehash(user.password())) {
                String newHash = StringEncoding.hashPassword(password);
                user.password(newHash);
                this.save(user);
            }
            this.pwdCache.update(user.id(), password);
            return user;
        }
```

Add helper method:
```java
private boolean requiresRehash(String hashedPassword) {
    // Extract work factor from bcrypt hash (format: a$<rounds>$...)
    String[] parts = hashedPassword.split("\\$");
    if (parts.length >= 3) {
        int workFactor = Integer.parseInt(parts[2]);
        return workFactor < 10;
    }
    return false;
}
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManagerV2.java
```diff
@@ -947,6 +947,7 @@ public HugeUser matchUser(String name, String password) {
         }
 
         if (StringEncoding.checkPassword(password, user.password())) {
+            // TODO: rehash password if bcrypt work factor is lower than expected
```
imbajin, line 950: **‼️ Critical: Duplicate code - missing upgrade logic here too**

This is identical to StandardAuthManager.java but also needs the automatic password rehashing implementation. See the comment on StandardAuthManager.java for the suggested fix.

Apply the same solution here to ensure both auth managers handle password upgrades consistently.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/StringEncodingTest.java
```diff
@@ -180,4 +181,23 @@ public void testReadAsciiString() {
...
+        Assert.assertFalse(StringEncoding.checkPassword("wrong", oldPassword));
+
+        // Test new work factor (10)
+        String newPassword = BCrypt.hashpw(testPassword, BCrypt.gensalt(10));
+        Assert.assertTrue(StringEncoding.checkPassword(testPassword, newPassword));
+        Assert.assertFalse(StringEncoding.checkPassword("wrong", newPassword));
+
+        // Test that hashPassword uses the new cost factor
+        String hashedPassword = StringEncoding.hashPassword(testPassword);
+        Assert.assertTrue(StringEncoding.checkPassword(testPassword, hashedPassword));
```
imbajin: **⚠️ Important: Missing test coverage for actual security requirement**

The test validates backward compatibility but doesn't verify the security improvement. Missing test cases:

1. **Performance test**: Verify work factor 10 takes significantly longer than work factor 4 (should be ~64x slower, around 100-200ms)
2. **Brute force resistance**: Document the computational cost difference
3. **Hash format validation**: Verify new hashes contain `$2a$10$` prefix

**Suggested addition:**
```suggestion
    @Test
    public void testHashPasswordUsesSecureWorkFactor() {
        String testPassword = "test123!@#";
        String hashedPassword = StringEncoding.hashPassword(testPassword);
        
        // Verify work factor 10 is used (hash format: a$...)
        Assert.assertTrue("Hash should use work factor 10", 
                         hashedPassword.startsWith("a$"));
        
        // Verify computational cost (work factor 10 should take 50-300ms)
        long startTime = System.currentTimeMillis();
        StringEncoding.checkPassword(testPassword, hashedPassword);
        long elapsed = System.currentTimeMillis() - startTime;
        Assert.assertTrue("Work factor 10 should take at least 50ms", elapsed >= 50);
    }
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/StringEncoding.java
```diff
@@ -140,7 +140,9 @@ public static String decompress(byte[] value, float bufferRatio) {
     }
 
     public static String hashPassword(String password) {
-        return BCrypt.hashpw(password, BCrypt.gensalt(4));
+        // OWASP suggests 10 as a minimum and 12–14 for production;
+        // workFactor 12 is not used by default due to its 200+ ms cost.
+        return BCrypt.hashpw(password, BCrypt.gensalt(10));
```
imbajin: **🧹 Minor: Consider using constant for work factor**

The magic number `10` appears in multiple places. Consider defining a constant for maintainability:

```suggestion
    private static final int BCRYPT_WORK_FACTOR = 10;
    
    public static String hashPassword(String password) {
        // OWASP suggests 10 as a minimum and 12–14 for production;
        // workFactor 12 is not used by default due to its 200+ ms cost.
        return BCrypt.hashpw(password, BCrypt.gensalt(BCRYPT_WORK_FACTOR));
    }
```

This makes it easier to adjust the work factor later and ensures consistency across the codebase.

hugegraph-struct/src/main/java/org/apache/hugegraph/util/StringEncoding.java
```diff
@@ -165,7 +165,9 @@ public static String decompress(byte[] value, float bufferRatio) {
     }
 
     public static String hashPassword(String password) {
-        return BCrypt.hashpw(password, BCrypt.gensalt(4));
+        // OWASP suggests 10 as a minimum and 12–14 for production;
+        // workFactor 12 is not used by default due to its 200+ ms cost.
```
imbajin, line 171: **🧹 Minor: Same constant suggestion for consistency**

This file should use the same constant approach as suggested for hugegraph-core's StringEncoding.java:

```suggestion
    private static final int BCRYPT_WORK_FACTOR = 10;
    
    public static String hashPassword(String password) {
        // OWASP suggests 10 as a minimum and 12–14 for production;
        // workFactor 12 is not used by default due to its 200+ ms cost.
        return BCrypt.hashpw(password, BCrypt.gensalt(BCRYPT_WORK_FACTOR));
    }
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/CompressUtil.java
```diff
@@ -160,15 +160,26 @@ public static void decompressTar(String sourceFile, String outputDir,
...
+        return zipSlipProtect(entry.getName(), targetDir);
+    }
+
+    private static Path zipSlipProtect(ZipEntry entry, Path targetDir)
+            throws IOException {
+        return zipSlipProtect(entry.getName(), targetDir);
+    }
+
+    private static Path zipSlipProtect(String entryName, Path targetDir)
+            throws IOException {
```
imbajin, line 172: **⚠️ Important: Zip slip protection can be simplified**

Good addition of zip slip protection, but the implementation has some redundancy:

1. Both `ArchiveEntry` and `ZipEntry` overloads now delegate to the same `String` version
2. The `decompressZip` method uses the protected path correctly

**Suggestion for better consistency:**
Ensure `outputDir` is normalized at the start of `decompressZip` to use the same base path for all operations. This prevents potential issues if `outputDir` itself contains path traversal characters.

PR #2941, opened 2026-01-20
fix(consumers): prevent await deadlock on ContextCallable failure

### LINKED ISSUE #2939: [Bug] Potential silent thread failure in Consumers.java due to unchecked Future
### Bug Type (问题类型)

logic (逻辑设计问题)

### Before submit

### Environment (环境信息)

Server Version: N/A (Found in Source Code Analysis / Master Branch)

Backend: N/A

OS: N/A

Data Size: N/A

Expected Behavior: 
When Consumers threads encounter a critical error (e.g., inside ContextCallable initialization or an Error that escapes the internal try-catch), the exception should be logged or propagated to the UncaughtExceptionHandler to alert the system.

Actual Behavior: 
In Consumers.java, tasks are submitted using executor.submit(...) (Line 110), but the returned Future is never checked (no future.get() is called in await() or elsewhere). This causes a "swallowed exception" scenario:

1. Exceptions occurring outside the runAndDone() internal try-catch block (e.g., in ContextCallable wrapper or strictly internal JVM errors) are captured by the Future.

2. Since the Future is ignored, these exceptions remain invisible. The system believes the consumer is running, but the thread may have terminated silently.

Code / Logs：
// Consumers.java
public void start(String name) {
    // ...
    for (int i = 0; i < this.workers; i++) {
        // Problem: 'submit' is used but the returned Future is never checked via .get()
        // If an exception happens outside runAndDone's try-catch, it is swallowed.
        this.runningFutures.add(
                this.executor.submit(new ContextCallable<>(this::runAndDone)));
    }
}

Recommendation:
Change submit() to execute() if the return value is not needed, which allows the JVM's UncaughtExceptionHandler to catch unhandled exceptions. Or ensure future.get() is checked during the lifecycle management.

```javascript
N/A
```

```javascript
N/A
```

bitflicker64:
## Purpose of the PR

- close #2939  
- Fix possible await() hang when ContextCallable fails before entering runAndDone(). In that case latch.countDown() was never called, so await() could block forever and the failure could remain hidden.

## Main Changes
 
Submit safeRun() instead of directly submitting ContextCallable

Add safeRun() wrapper to always call latch.countDown() in finally

Remove latch.countDown() from runAndDone() to avoid double-counting

## Verifying these changes

    - mvn -pl hugegraph-server/hugegraph-core -am test PASS; full build fails in hugegraph-clustertest-minicluster due to toolchain/JDK issue (unrelated).

- [x]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/ConsumersTest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/ConsumersTest.java
@@ -58,21 +58,23 @@ public class ConsumersTest {
      */
     @Test(timeout = 1000)
     public void testAwaitDoesNotHangWhenContextCallableFails() throws Throwable {
-        ExecutorService executor = Executors.newSingleThreadExecutor();
+        ExecutorService executor = Executors.newFixedThreadPool(1);
         try {
+            // Use AssertionError to bypass the inner catch(Exception) loop in runAndDone()
+            // This simulates a scenario where an exception escapes the task logic
+            // (similar to how a ContextCallable failure would behave from safeRun's perspective)
             Consumers<Integer> consumers = new Consumers<>(executor, v -> {
-                // Never reached
+                throw new AssertionError("Simulated fatal error (OOM/StackOverflow/etc)");
             });
-
-            /*
-             * start() creates ContextCallable internally.
-             * If ContextCallable.call() fails before runAndDone(),
-             * safeRun().finally MUST count down the latch.
-             */
-            consumers.start("test");
-
-            // If the fix is missing, this call hangs forever.
+            consumers.start("test-fatal-error");
+            consumers.provide(1);
+            // Verification:
+            // Without the fix, the latch would never be decremented (because runAndDone crashes), causing await() to hang.
+            // With the fix (safeRun wrapper), the finally block ensures latch.countDown() is called.
             consumers.await();
+
+            // Note: consumer.exception will be null because safeRun only catches Exception, not Error.
+            // This is acceptable behavior for fatal errors, as long as it doesn't deadlock.
         } finally {
             executor.shutdownNow();
         }


imbajin:
Hi, thanks for your PR, maybe we need at least **2 tests** to validate the modification

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/Consumers.java
```diff
@@ -104,9 +104,23 @@ public void start(String name) {
...
         for (int i = 0; i < this.workers; i++) {
-            this.runningFutures.add(
-                    this.executor.submit(new ContextCallable<>(this::runAndDone)));
+            this.runningFutures.add(this.executor.submit(this::safeRun));
         }
+
+    }
+    private Void safeRun() {
+        try {
+            new ContextCallable<>(this::runAndDone).call();
```
imbajin: **‼️ Critical:** The `safeRun()` method calls `new ContextCallable<>(this::runAndDone).call()` directly, which bypasses the context propagation mechanism that `ContextCallable` is designed to provide when submitted via `ExecutorService.submit()`.

```suggestion
    private Void safeRun() {
        try {
            this.runAndDone();
```

Since `safeRun()` is already submitted to the executor, the context should be captured at submission time, not inside the worker. The current approach may lose thread-local context that `ContextCallable` is meant to preserve.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/Consumers.java
```diff
@@ -104,9 +104,23 @@ public void start(String name) {
...
+        } catch (Throwable e) {
+            LOG.error("Worker failed before runAndDone()", e);
+            if (this.exception == null) {
+                this.exception = e;
+            }
+            exceptionHandle(e);
```
imbajin: **⚠️ Important:** Duplicate exception logging - both `safeRun()` and `runAndDone()` call `exceptionHandle(e)`, which may result in the same exception being logged/handled twice.

```suggestion
        } catch (Throwable e) {
            LOG.error("Worker failed before runAndDone()", e);
            if (this.exception == null) {
                this.exception = e;
            }
            // exceptionHandle(e) will be called inside runAndDone()
```

Consider removing one of the `exceptionHandle()` calls to avoid redundant processing.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/Consumers.java
```diff
@@ -104,9 +104,23 @@ public void start(String name) {
         LOG.info("Starting {} workers[{}] with queue size {}...",
                  this.workers, name, this.queueSize);
         for (int i = 0; i < this.workers; i++) {
-            this.runningFutures.add(
-                    this.executor.submit(new ContextCallable<>(this::runAndDone)));
+            this.runningFutures.add(this.executor.submit(this::safeRun));
         }
+
+    }
+    private Void safeRun() {
```
imbajin: **⚠️ Important:** Missing blank line before method declaration violates Java code style conventions.

```suggestion

    private Void safeRun() {
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/Consumers.java
```diff
@@ -104,9 +104,23 @@ public void start(String name) {
...
+        try {
+            new ContextCallable<>(this::runAndDone).call();
+        } catch (Throwable e) {
```
imbajin: **🧹 Minor:** The catch block catches `Throwable` which includes `Error` subclasses. In the original code via `ContextCallable.call()`, only `Exception` was caught. This change broadens exception handling - is this intentional?

Consider whether catching `Error` (e.g., `OutOfMemoryError`, `StackOverflowError`) is appropriate here, as these typically indicate unrecoverable JVM states.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/ConsumersTest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.hugegraph.testutil.Assert;
+import org.apache.hugegraph.util.Consumers;
+import org.junit.Test;
+
+public class ConsumersTest {
+
```
imbajin, line 29: **⚠️ Important:** Test timeout of 5 seconds may be too long for a unit test. The actual execution should complete in milliseconds if there's no deadlock.

```suggestion
    @Test(timeout = 1000)
```

Consider reducing to 1 second to fail faster if there's a regression.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/ConsumersTest.java
```diff
@@ -0,0 +1,81 @@
...
+            consumers.await();
+
+            Assert.assertEquals("Should process all provided elements",
+                                50, processed.get());
+        } finally {
+            executor.shutdownNow();
+        }
+    }
+
+    @Test(timeout = 5000)
```
imbajin: **🧹 Minor:** The test doesn't verify the specific deadlock scenario mentioned in the PR description - where `ContextCallable` fails *before* entering `runAndDone()`.

Consider adding a test case that explicitly simulates the failure scenario, for example:
```java
@Test(timeout = 1000)
public void testAwaitDoesNotHangWhenContextCallableFails() throws Throwable {
    // Test that simulates ContextCallable constructor/call() failure
    // before runAndDone() is entered
}
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/Consumers.java
bitflicker64: would it be alright to just implement ContextCallable wraps safeRun in start() , so the context is captured at submission time and safeRun() simply calls runAndDone() and always decrements the latch in a finally block?

imbajin: > would it be alright to just implement ContextCallable wraps safeRun in start() , so the context is captured at submission time and safeRun() simply calls runAndDone() and always decrements the latch in a finally block?

Sure, the execution Flow may like:
```
  Submission Thread: start()
    → new ContextCallable<>(this::safeRun)  // ✅ Captures context here
    → executor.submit(...)
      → Worker Thread: ContextCallable.call()
        → setContext(captured_context)  // ✅ Sets submission thread's context
        → safeRun()
          → runAndDone()
          → latch.countDown()  // ✅ Always executes
        → resetContext()
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/Consumers.java
bitflicker64: Catching Throwable wasn’t intentional. I’ve updated the code to remove that catch and keep `safeRun()` as `runAndDone()` + `latch.countDown()` in a `finally` block.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/util/ConsumersTest.java
```diff
@@ -50,7 +50,35 @@ public void testStartProvideAwaitNormal() throws Throwable {
...
+             * safeRun().finally MUST count down the latch.
+             */
+            consumers.start("test");
+
+            // If the fix is missing, this call hangs forever.
+            consumers.await();
+        } finally {
+            executor.shutdownNow();
+        }
+    }
```
imbajin, line 81: Seems current test cannot validate the failure(case)? 

A possible solution may like:

```java
public void testAwaitDoesNotHangWhenFatalErrorOccurs() throws Throwable {
        ExecutorService executor = Executors.newFixedThreadPool(1);
        try {
            // Use AssertionError to bypass the inner catch(Exception) loop in runAndDone()
            // This simulates a scenario where an exception escapes the task logic 
            // (similar to how a ContextCallable failure would behave from safeRun's perspective)
            Consumers<Integer> consumers = new Consumers<>(executor, v -> {
                throw new AssertionError("Simulated fatal error (OOM/StackOverflow/etc)");
            });
            consumers.start("test-fatal-error");
            consumers.provide(1);
            // Verification:
            // Without the fix, the latch would never be decremented (because runAndDone crashes), causing await() to hang.
            // With the fix (safeRun wrapper), the finally block ensures latch.countDown() is called.
            consumers.await();
            
            // Note: consumer.exception will be null because safeRun only catches Exception, not Error.
            // This is acceptable behavior for fatal errors, as long as it doesn't deadlock.
        } finally {
            executor.shutdownNow();
        }
    }
```

PR #2944, opened 2026-01-23
refactor(server): unify URL configs when scheme is missing

### LINKED ISSUE #2942: [Improvement] unify endpoint URL format
# Feature Description

## Background

Currently, restserver.url and some other(such as gremlinserver.url, and others in ServerOptions.java) in rest-server.properties requires an explicit scheme prefix such as http://, which can be slightly inconvenient and inconsistent with other modules.

In some components (e.g. Graph AI), URLs without a scheme are supported by defaulting to http, while allowing users to override the scheme explicitly. This provides a more user-friendly configuration experience.

## Current Behavior

 restserver.url property must include a full scheme, for example:

restserver.url=http://127.0.0.1:8080

If the scheme is omitted, the URL may fail to be parsed or behave unexpectedly.

## Expected Behavior

Support a unified and more flexible configuration rule for restserver.url:

Scheme is optional

If no scheme is provided, default to http://

restserver.url=127.0.0.1:8080

Explicit scheme should override the default

restserver.url=http://127.0.0.1:8080
restserver.url=https://127.0.0.1:8080

Proposed Change

During configuration parsing:

If restserver.url does not contain a scheme (e.g. http://, https://), automatically prepend http://

If a scheme is present, preserve the original value

bitflicker64:
## Purpose of the PR

- close #2942 
- Some server URL configs currently require an explicit scheme (`http://` / `https://`). If a user sets something like `127.0.0.1:8080`, it can fail to parse or behave inconsistently compared to other modules that already default the scheme.

This PR fixes that by normalizing **only the relevant server URL options**: if the scheme is missing, we auto-prefix a sensible default, while leaving explicitly provided schemes untouched.

Also covered the review corner case: `server.k8s_url` should default to **`https://`** (so we don’t accidentally downgrade it to `http://`).

## Main Changes

`HugeConfig.get(...)` now applies URL normalization for a small allowlist of keys:

* `restserver.url` → default `http://`
* `gremlinserver.url` → default `http://`
* `server.urls_to_pd` → default `http://`
* `server.k8s_url` → default `https://`

Normalization only triggers when:

* the config key is in the allowlist, **and**
* the value is a `String`, **and**
* the value has **no scheme**

If the value already starts with `http://` or `https://`, it’s left as-is.
Non-string config values are untouched.

## Verifying these changes

Tests are included and you can verify the behavior like this:

* Added unit tests in `HugeConfigTest` covering:

  * missing scheme → default prefix is added
  * existing scheme → preserved as-is
  * `server.k8s_url` missing scheme → becomes `https://...` (no accidental downgrade)

- [x]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status

- [x]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
+++ b/hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
@@ -43,10 +43,8 @@ public class HugeConfig extends PropertiesConfiguration {
 
     private static final Logger LOG = Log.logger(HugeConfig.class);
 
-    // Cache for url normalization metadata
-    // Populated lazily on first use to ensure OptionSpace is already registered
+    // Cache for URL normalization metadata (populated lazily per key)
     private static final Map<String, String> URL_NORMALIZATIONS = new HashMap<>();
-    private static volatile boolean cacheInitialized = false;
 
     private String configPath;
 
@@ -256,28 +254,32 @@ public class HugeConfig extends PropertiesConfiguration {
     }
 
     private static String defaultSchemeFor(String key) {
-        ensureCacheInitialized();
-        return URL_NORMALIZATIONS.get(key);
-    }
+        // Check if we already cached this key's scheme
+        if (URL_NORMALIZATIONS.containsKey(key)) {
+            return URL_NORMALIZATIONS.get(key);
+        }
 
-    private static void ensureCacheInitialized() {
-        if (!cacheInitialized) {
-            synchronized (URL_NORMALIZATIONS) {
-                if (!cacheInitialized) {
-                    // Populate cache from OptionSpace
-                    for (String optionKey : OptionSpace.keys()) {
-                        TypedOption<?, ?> option = OptionSpace.get(optionKey);
-                        if (option instanceof ConfigOption) {
-                            ConfigOption<?> configOption = (ConfigOption<?>) option;
-                            if (configOption.needsUrlNormalization()) {
-                                URL_NORMALIZATIONS.put(optionKey,
-                                                       configOption.getDefaultScheme());
-                            }
-                        }
-                    }
-                    cacheInitialized = true;
+        // We don't know yet - look it up NOW from OptionSpace
+        synchronized (URL_NORMALIZATIONS) {
+            // Double-check after acquiring lock
+            if (URL_NORMALIZATIONS.containsKey(key)) {
+                return URL_NORMALIZATIONS.get(key);
+            }
+
+            // Look up the option from OptionSpace
+            TypedOption<?, ?> option = OptionSpace.get(key);
+            String scheme = null;
+
+            if (option instanceof ConfigOption) {
+                ConfigOption<?> configOption = (ConfigOption<?>) option;
+                if (configOption.needsUrlNormalization()) {
+                    scheme = configOption.getDefaultScheme();
                 }
             }
+
+            // Cache it for next time (even if null)
+            URL_NORMALIZATIONS.put(key, scheme);
+            return scheme;
         }
     }
 


hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -88,8 +88,12 @@ private void setLayoutIfNeeded(Configuration conf) {
     public <T, R> R get(TypedOption<T, R> option) {
         Object value = this.getProperty(option.name());
         if (value == null) {
-            return option.defaultValue();
+            value = option.defaultValue();
         }
+
+        // Normalize URL options if needed (add scheme like http://)
```
imbajin: **‼️ Critical: Logic Flow Issue**

The normalization logic modifies the default value before returning it, but this happens **after** the value is already set from `defaultValue()`. This means the normalization will apply to **both** user-provided values AND default values.

However, if the `ConfigOption` already has a proper default like `"http://127.0.0.1:8080"`, the normalization is redundant. More importantly, this design couples the config retrieval logic with URL-specific assumptions.

Consider: Does this break if someone later adds a URL option that isn't in the allowlist? The normalization silently won't apply, creating inconsistent behavior.

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +217,55 @@ private static Configuration loadConfigFile(File configFile) {
...
+    private static String prefixSchemeIfMissing(String raw, String scheme) {
+        if (raw == null) {
+            return null;
+        }
+        String s = raw.trim();
+        if (s.isEmpty()) {
+            return s;
+        }
+
+        // Keep original string if scheme already exists
```
imbajin: **‼️ Critical: Case Sensitivity Bug**

Using `.toLowerCase()` for comparison but returning the original string creates a vulnerability:

```java
String lower = s.toLowerCase();
if (lower.startsWith("http://") || lower.startsWith("https://")) {
    return s;  // Returns original - might be "HTTP://" or "HtTp://"
}
```

This means `"HTTP://127.0.0.1:8080"` would pass through unchanged. While URLs are case-insensitive for the scheme part per RFC 3986, this creates inconsistent output where some schemes are lowercase and others preserve their original casing.

**Suggestion:**
```suggestion
        // Keep original string if scheme already exists (case-insensitive check)
        if (s.startsWith("http://") || s.startsWith("https://") ||
            s.startsWith("HTTP://") || s.startsWith("HTTPS://")) {
            return s;
        }
```

Or normalize consistently:
```java
// Check if scheme exists (case-insensitive)
if (lower.startsWith("http://") || lower.startsWith("https://")) {
    // Normalize scheme to lowercase per RFC 3986
    return lower;
}
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +217,55 @@ private static Configuration loadConfigFile(File configFile) {
...
+        if (value instanceof String) {
+            return prefixSchemeIfMissing((String) value, scheme);
+        }
+
+        // If it ever hits here, it means config storage returned a non-string type;
+        // leave it unchanged (safer than forcing toString()).
+        return value;
+    }
+
+    private static String defaultSchemeFor(String key) {
```
imbajin, line 256: **‼️ Critical: Hardcoded Config Keys Break Maintainability**

The switch statement hardcodes config key strings:

```java
case "restserver.url":
case "gremlinserver.url":
case "server.urls_to_pd":
case "server.k8s_url":
```

This creates tight coupling between `HugeConfig` (commons module) and specific config keys from various modules. If a key is renamed or a new URL option is added elsewhere, developers must remember to update this switch statement.

**Better approach:** Let `ConfigOption` declare whether it needs URL normalization via metadata:

```java
public class ConfigOption<T> {
    private boolean urlNormalize = false;
    private String defaultScheme = "http://";
    
    public ConfigOption<T> withUrlNormalization(String scheme) {
        this.urlNormalize = true;
        this.defaultScheme = scheme;
        return this;
    }
}
```

Then in `HugeConfig.get()`:
```java
if (option.needsUrlNormalization()) {
    value = prefixSchemeIfMissing((String) value, option.getDefaultScheme());
}
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +217,55 @@ private static Configuration loadConfigFile(File configFile) {
...
+        String s = raw.trim();
+        if (s.isEmpty()) {
+            return s;
+        }
+
+        // Keep original string if scheme already exists
+        String lower = s.toLowerCase();
+        if (lower.startsWith("http://") || lower.startsWith("https://")) {
+            return s;
+        }
```
imbajin, line 306: **⚠️ Important: Missing Edge Case Handling**

The current implementation doesn't handle URLs with credentials or userinfo:

```java
// These would incorrectly get prefixed:
"user:password@127.0.0.1:8080" → "http://user:password@127.0.0.1:8080"
"admin@localhost:8080" → "http://admin@localhost:8080"
```

While valid according to RFC 3986 (`scheme://[userinfo@]host[:port]`), detecting these requires checking for `@` before any `/`:

```suggestion
        // Keep original string if scheme already exists
        String lower = s.toLowerCase();
        if (lower.startsWith("http://") || lower.startsWith("https://")) {
            return s;
        }

        // Don't add scheme if userinfo is present without scheme
        // (e.g., "user:pass@host:port" - likely malformed or needs manual fixing)
        int slashPos = s.indexOf('/');
        int atPos = s.indexOf('@');
        if (atPos != -1 && (slashPos == -1 || atPos < slashPos)) {
            // Has userinfo component - preserve as-is to avoid masking config errors
            return s;
        }

        return scheme + s;
```

hugegraph-commons/hugegraph-common/src/test/java/org/apache/hugegraph/unit/config/HugeConfigTest.java
```diff
@@ -429,6 +429,45 @@ public void testFromMapConfigurationWithList() {
...
+
+        HugeConfig config = new HugeConfig(map);
+
+        Assert.assertEquals("https://127.0.0.1:8080",
+                            config.get(UrlOptions.restUrl));
+        Assert.assertEquals("http://127.0.0.1:8182",
+                            config.get(UrlOptions.gremlinUrl));
+        Assert.assertEquals("http://127.0.0.1:8888",
+                            config.get(UrlOptions.k8sUrl));
+    }
```
imbajin: **⚠️ Important: Test Coverage Gap**

Tests verify the happy path but miss important edge cases:

1. **Empty/whitespace-only strings** - Returns empty after trim, but should this be valid?
2. **Malformed URLs** - `"://127.0.0.1:8080"` or `"http:127.0.0.1"`
3. **Mixed case schemes** - `"HtTp://"` or `"HTTPS://"`
4. **Trailing/leading whitespace** - `" http://127.0.0.1:8080 "`
5. **IPv6 addresses** - `"[::1]:8080"` should become `"http://[::1]:8080"`

Add test cases:
```java
@Test
public void testUrlNormalizationEdgeCases() {
    Map<String, Object> map = new HashMap<>();
    
    // Whitespace handling
    map.put("restserver.url", "  127.0.0.1:8080  ");
    HugeConfig config = new HugeConfig(map);
    Assert.assertEquals("http://127.0.0.1:8080", config.get(UrlOptions.restUrl));
    
    // Mixed case scheme preservation
    map.put("restserver.url", "HTTP://127.0.0.1:8080");
    config = new HugeConfig(map);
    Assert.assertEquals("HTTP://127.0.0.1:8080", config.get(UrlOptions.restUrl));
    
    // IPv6
    map.put("restserver.url", "[::1]:8080");
    config = new HugeConfig(map);
    Assert.assertEquals("http://[::1]:8080", config.get(UrlOptions.restUrl));
}
```

hugegraph-commons/hugegraph-common/src/test/java/org/apache/hugegraph/unit/config/HugeConfigTest.java
```diff
@@ -32,6 +32,15 @@
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.collections.IteratorUtils;
```
imbajin: **🧹 Minor: Import Order Inconsistency**

The test file has imports reordered (Apache Commons imports moved before HugeGraph imports). While functionally correct, this breaks the existing import ordering convention:

```diff
-import org.apache.hugegraph.config.ConfigConvOption;
-...
 import org.apache.commons.collections.IteratorUtils;
 import org.apache.commons.configuration2.AbstractConfiguration;
+...
+import org.apache.hugegraph.config.ConfigConvOption;
```

Consider keeping the original import order to reduce diff noise and maintain consistency.

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +217,55 @@ private static Configuration loadConfigFile(File configFile) {
...
+        if (value == null) {
+            return null;
+        }
+
+        String scheme = defaultSchemeFor(key);
+        if (scheme == null) {
+            return value;
+        }
+
+        // URL options are defined as ConfigOption<String> and normalized here.
```
imbajin: **🧹 Minor: Documentation Could Be Clearer**

The comment mentions "URL options are defined as ConfigOption<String>", but this is an implementation detail that might confuse readers. Consider clarifying the intent:

```suggestion
        // Normalize URL config values by adding default scheme if missing.
        // Only applies to allowlisted URL options (see defaultSchemeFor()).
        if (value instanceof String) {
            return prefixSchemeIfMissing((String) value, scheme);
        }
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
bitflicker64:
```java
// Introduces a fromDefault flag to separate system defaults from user input,
// avoiding redundant normalization of valid defaults while preserving safety for user values.
```
```java
public <T, R> R get(TypedOption<T, R> option) {
    Object value = this.getProperty(option.name());
    boolean fromDefault = false;

    if (value == null) {
        value = option.defaultValue();
        fromDefault = true;
    }

    if (!fromDefault) {
        value = normalizeUrlOptionIfNeeded(option.name(), value);
    }

    return (R) value;
}
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
bitflicker64: The prefixSchemeIfMissing() method currently just normalizes URLs and doesn’t try to validate things like userinfo (user@host). That seems reasonable since handling credentials in URLs is discouraged and @ can also appear in valid paths. Adding special checks here would blur the line between normalization and validation.It’s probably better to handle stricter validation at the client/usage layer if needed.I’m happy to add a separate check or open an additional PR if you think stricter handling belongs here. Please let me know your preference.

imbajin:
> Hi, I tried refactoring HugeConfigTest to use ServerOptions directly as suggested. However, this test is in the hugegraph-common module, while ServerOptions is in hugegraph-server. When I add the dependency, it creates a circular dependency and the build fails with this error:
> 
> ```
> The projects in the reactor contain a cyclic reference:
> org.apache.hugegraph:hugegraph-api
>  -> org.apache.hugegraph:hugegraph-core
>  -> org.apache.hugegraph:hugegraph-common
>  -> org.apache.hugegraph:hugegraph-api
> ```
> 
> Because of this, ServerOptions cannot be used in hugegraph-common tests without breaking the module structure.

Fine, then we could keep the original logic to test it

bitflicker64:
@imbajin  Thanks for the discussion and guidance. I’ve moved the ServerOptions related tests into ServerOptionsTest and confirmed that all tests and CI are passing now.

imbajin (commented):
We should also update the config files' default value (like https://github.com/apache/incubator-hugegraph/blob/9babe493919c01f012a56e7b5fb4d8b9faf64cf5/hugegraph-server/hugegraph-dist/src/assembly/static/conf/rest-server.properties#L3C2-L3C2)

Remove the unnecessary prefix of URL scheme & Also need to update it in https://github.com/apache/incubator-hugegraph-doc (website page), thx

image

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +221,53 @@ private static Configuration loadConfigFile(File configFile) {
...
+        if (s.isEmpty()) {
+            return s;
+        }
+
+        // Keep original string if scheme already exists
+        String lower = s.toLowerCase();
+        if (lower.startsWith("http://") || lower.startsWith("https://")) {
+            return lower;  // Return LOWERCASE version
+        }
+        return scheme + lower;  // Return scheme + LOWERCASE input
```
imbajin: ‼️ **Critical Logic Error: Returns lowercase for existing schemes**

This normalization converts **all** existing URLs to lowercase, including those explicitly configured with uppercase (e.g., `HTTP://127.0.0.1:8080` → `http://127.0.0.1:8080`). While HTTP schemes are case-insensitive per RFC 3986, this creates **inconsistent behavior**:

- Missing scheme → adds default + lowercases input
- Existing scheme → lowercases entire URL

**Problem:** If a user explicitly sets `HTTPS://MyHost:8080`, the hostname gets lowercased to `myhost`, which could break case-sensitive server configurations.

**Suggested fix:** Only lowercase the scheme portion, preserve original casing for the rest:

```suggestion
private static String prefixSchemeIfMissing(String raw, String scheme) {
    if (raw == null) {
        return null;
    }
    String s = raw.trim();
    if (s.isEmpty()) {
        return s;
    }

    // Extract scheme if present
    int schemeIdx = s.indexOf("://");
    if (schemeIdx > 0) {
        String existingScheme = s.substring(0, schemeIdx).toLowerCase();
        String rest = s.substring(schemeIdx);
        return existingScheme + rest;
    }
    return scheme + s;  // Preserve original case when adding scheme
}
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +221,53 @@ private static Configuration loadConfigFile(File configFile) {
                                       e, configFile);
         }
     }
+
+    private static Object normalizeUrlOptionIfNeeded(String key, Object value) {
+        if (value == null) {
+            return null;
+        }
```
imbajin, line 231: ‼️ **Critical Design Issue: Silently modifies user-provided values**

The current implementation normalizes values **even when explicitly set by users** in config files. This violates the principle of least surprise:

```properties
# User's config file
restserver.url = 127.0.0.1:8080
```

Gets silently transformed to `http://127.0.0.1:8080` without any indication. Users might expect errors for malformed URLs rather than silent corrections.

**Suggested approach:**
1. Log a warning when auto-prefixing scheme: `LOG.warn("Config '{}' missing scheme, defaulting to '{}'", key, normalizedValue);`
2. Or validate and reject invalid URLs at startup with clear error messages

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +221,53 @@ private static Configuration loadConfigFile(File configFile) {
...
+        if (value instanceof String) {
+            return prefixSchemeIfMissing((String) value, scheme);
+        }
+
+        // If it ever hits here, it means config storage returned a non-string type;
+        // leave it unchanged (safer than forcing toString()).
+        return value;
+    }
+
+    private static String defaultSchemeFor(String key) {
```
imbajin, line 256: ⚠️ **Performance: Redundant OptionSpace lookup on every config read**

`defaultSchemeFor()` calls `OptionSpace.get(key)` on **every** config value retrieval, even for non-URL options. This adds overhead to frequently accessed config reads.

**Suggested optimization:** Cache the URL normalization metadata during `ConfigOption` registration or HugeConfig initialization:

```suggestion
// In HugeConfig constructor or static initializer
private static final Map<String, String> URL_NORMALIZATIONS = new HashMap<>();
static {
    // Populate from OptionSpace during initialization
    OptionSpace.getAll().forEach((key, option) -> {
        if (option instanceof ConfigOption) {
            ConfigOption<?> co = (ConfigOption<?>) option;
            if (co.needsUrlNormalization()) {
                URL_NORMALIZATIONS.put(key, co.getDefaultScheme());
            }
        }
    });
}

private static String defaultSchemeFor(String key) {
    return URL_NORMALIZATIONS.get(key);
}
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -87,9 +87,17 @@ private void setLayoutIfNeeded(Configuration conf) {
...
         Object value = this.getProperty(option.name());
+        boolean fromDefault = false;
+
         if (value == null) {
-            return option.defaultValue();
+            value = option.defaultValue();
+            fromDefault = true;
+        }
+
+        if (!fromDefault) {
```
imbajin, line 100: ⚠️ **Code Quality: Inconsistent behavior between default and explicit values**

Line 94-96 shows normalization **only applies to explicitly set values** (`fromDefault = false`), but **not** to default values from `option.defaultValue()`.

This creates inconsistent behavior:
- Default value `http://127.0.0.1:8080` → used as-is
- User sets `127.0.0.1:8080` → normalized to `http://127.0.0.1:8080`

**Question:** Is this intentional? If defaults already have schemes, why do we need `.withUrlNormalization()` on the options?

hugegraph-server/hugegraph-test/src/test/java/org/apache/hugegraph/unit/config/ServerOptionsTest.java
```diff
@@ -0,0 +1,76 @@
...
+        conf.setProperty("restserver.url", "HTTP://127.0.0.1:8080");
+        config = new HugeConfig(conf);
+        Assert.assertEquals("http://127.0.0.1:8080",
+                            config.get(ServerOptions.REST_SERVER_URL));
+
+        conf = new PropertiesConfiguration();
+        conf.setProperty("restserver.url", "[::1]:8080");
+        config = new HugeConfig(conf);
+        Assert.assertEquals("http://[::1]:8080",
+                            config.get(ServerOptions.REST_SERVER_URL));
```
imbajin, line 77: ⚠️ **Missing Test Coverage: IPv6 URLs with schemes**

The test at line 70-74 covers IPv6 addresses **without** schemes (`[::1]:8080`), but doesn't test the preservation behavior when schemes are present:

**Add test case:**
```java
@Test
public void testIpv6WithExistingScheme() {
    PropertiesConfiguration conf = new PropertiesConfiguration();
    conf.setProperty("restserver.url", "http://[::1]:8080");
    HugeConfig config = new HugeConfig(conf);
    Assert.assertEquals("http://[::1]:8080",
                        config.get(ServerOptions.REST_SERVER_URL));
}
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
```diff
@@ -213,4 +221,53 @@ private static Configuration loadConfigFile(File configFile) {
...
+        if (value == null) {
+            return null;
+        }
+
+        String scheme = defaultSchemeFor(key);
+        if (scheme == null) {
+            return value;
+        }
+
+        // Normalize URL config values by adding default scheme if missing.
```
imbajin: 🧹 **Minor: Unnecessary comment duplicates code logic**

The comment on line 235 restates what the code already expresses clearly through method name `prefixSchemeIfMissing()` and the allowlist check.

**Suggested simplification:**
```suggestion
        // Normalize allowlisted URL options
        if (value instanceof String) {
            return prefixSchemeIfMissing((String) value, scheme);
        }
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/ConfigOption.java
```diff
@@ -20,6 +20,22 @@
 import com.google.common.base.Predicate;
 
 public class ConfigOption<T> extends TypedOption<T, T> {
+    private boolean urlNormalize = false;
+    private String defaultScheme = null;
+
+    public ConfigOption<T> withUrlNormalization(String scheme) {
```
imbajin, line 26: ⚠️ **API Design: Builder pattern requires mutation**

The `withUrlNormalization()` method mutates the `ConfigOption` instance and returns `this`, but it's called during option declaration:

```java
new ConfigOption<>(...).withUrlNormalization("http://");
```

**Issue:** If `ConfigOption` instances are ever shared or reused, this mutation could cause race conditions.

**Suggested improvement:** Make `withUrlNormalization()` immutable or ensure `ConfigOption` is only constructed once during static initialization (add final modifier or documentation).

hugegraph-server/hugegraph-test/src/test/java/org/apache/hugegraph/unit/config/ServerOptionsTest.java
bitflicker64: added tests for preservation behaviour

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
bitflicker64: Yes, this is intentional.
The default values already have http://, so they are already correct. There is nothing to fix, so normalization is skipped. User values may not have the scheme (for example: 127.0.0.1:8080), so only those need normalization.
 .withUrlNormalization() is just metadata. It tells the system: “this option is a URL, and use this scheme when fixing user input.”

```
So:
Default → already correct → no change
User value → may be incomplete → normalize
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/HugeConfig.java
bitflicker64: fixed and added tests -
```java
@Test
public void testUrlNormalizationPreservesHostnameCase() {
    // Uppercase scheme + mixed-case hostname
    PropertiesConfiguration conf = new PropertiesConfiguration();
    conf.setProperty("restserver.url", "HTTP://MyServer:8080");
    HugeConfig config = new HugeConfig(conf);
    Assert.assertEquals("http://MyServer:8080",
                        config.get(ServerOptions.REST_SERVER_URL));

    // HTTPS with mixed-case hostname
    conf = new PropertiesConfiguration();
    conf.setProperty("restserver.url", "HTTPS://MyHost:8080");
    config = new HugeConfig(conf);
    Assert.assertEquals("https://MyHost:8080",
                        config.get(ServerOptions.REST_SERVER_URL));
}

@Test
public void testUrlNormalizationPreservesPathCase() {
    PropertiesConfiguration conf = new PropertiesConfiguration();
    conf.setProperty("restserver.url", "http://127.0.0.1:8080/SomePath/CaseSensitive");
    HugeConfig config = new HugeConfig(conf);
    Assert.assertEquals("http://127.0.0.1:8080/SomePath/CaseSensitive",
                        config.get(ServerOptions.REST_SERVER_URL));
}

@Test
public void testHttpsSchemeIsNotDowngraded() {
    PropertiesConfiguration conf = new PropertiesConfiguration();
    conf.setProperty("restserver.url", "https://127.0.0.1:8080");
    HugeConfig config = new HugeConfig(conf);
    Assert.assertEquals("https://127.0.0.1:8080",
                        config.get(ServerOptions.REST_SERVER_URL));
}
```

hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/config/ConfigOption.java
bitflicker64: this method is only used when `ConfigOption` constants are first created (for example, in `ServerOptions`) It runs once when the class is loaded, sets the metadata, and is never changed after that. Because of this, there is no real risk of concurrency issues in the current design. Ill add documentation to make this clearer

bitflicker64:
All review feedback implemented  . There's one failing test `KoutApiTest.testGet` (graph traversal) that I can't trace back to my URL changes  would really appreciate your help debugging this .also after this ill update https://github.com/apache/incubator-hugegraph-doc

imbajin:
> @imbajin hey I'm interested in contributing more to the project and was wondering if I could get an invitation to the Slack workspace?

Thank you for your interest! It’s great to see you’re looking to get more involved with the project.

Regarding the [Slack](https://app.slack.com/client/T4S1WH2J3/C059UU2FJ23) and Discord channels, they are actually public and you are welcome to join anytime without a specific invitation. However, please note that we haven't officially started active operations or management for those channels yet. If you're interested, we’d be happy to grant you the necessary permissions to help lead or moderate the community there!

Currently, our primary communication still happens through GitHub Issues/Discussions and Email.

We are always open to any suggestions or feedback you might have, whether you'd like to share them publicly or privately. Looking forward to your further contributions!

bitflicker64:
> > Hi, I tried joining via the Slack link, but it says I don’t have an account. I also couldn’t find the Discord. Could you please share active invite links? My email: [himanshuverma151006@gmail.com](mailto:himanshuverma151006@gmail.com) Thanks
> 
> Seems I don't have the permission to invite u to the ASF's Slack:)
> 
> When I add your account, it said:
> 
> > Your administrator has prohibited the use of Slack Connect in this workspace, so only ASF members and guests can join this channel.
> 
> Discord: https://discord.gg/uQqtb9U5vu (Indeed Not active)
> 
> BTW, our [docker-compose](https://github.com/apache/incubator-hugegraph/blob/master/docker/docker-compose.yml) images have some issues with the latest version, u could submit a new issue & test/fix it


> > Hi, I tried joining via the Slack link, but it says I don’t have an account. I also couldn’t find the Discord. Could you please share active invite links? My email: [himanshuverma151006@gmail.com](mailto:himanshuverma151006@gmail.com) Thanks
> 
> Seems I don't have the permission to invite u to the ASF's Slack:)
> 
> When I add your account, it said:
> 
> > Your administrator has prohibited the use of Slack Connect in this workspace, so only ASF members and guests can join this channel.
> 
> Discord: https://discord.gg/uQqtb9U5vu (Indeed Not active)
> 
> BTW, our [docker-compose](https://github.com/apache/incubator-hugegraph/blob/master/docker/docker-compose.yml) images have some issues with the latest version, u could submit a new issue & test/fix it

Thanks for checking the slack invite and i've joined the discord . I’ll look into the docker issue and update accordingly.

PR #2952, opened 2026-02-15
refactor(docker): migrate single-node compose from host to bridge networking

### LINKED ISSUE #2951: [Bug] Single node Docker setup does not work on macOS and Windows because of host networking
### Bug Type (问题类型)

others (please edit later)

### Before submit

### Environment (环境信息)

HugeGraph version 1.5.0
Docker Desktop on macOS and Windows
Deployment method docker compose single node configuration
File docker-compose.yml

Expected Behavior

Single node cluster starts normally

Actual Behavior

Server stuck printing Hugegraph server are waiting for storage backend
Server logs show UNAVAILABLE io exception
Store reports {"leaderCount":0,"partitionCount":0}
Cluster never becomes usable even though containers run

Description

The single node Docker setup uses network_mode host. This works only on Linux.
On macOS and Windows, Docker Desktop does not support host networking the same way.
Containers fall back to bridge networking but service configuration still assumes localhost style addresses.

PD and Store advertise gRPC endpoints as 0.0.0.0 or 127.0.0.1.
These are bind addresses, not reachable service addresses inside Docker networking.
Other containers cannot connect, causing PD client failures and preventing partition creation.

Root Cause

network_mode host is Linux only
Docker Desktop ignores or emulates it differently
Services advertise non routable addresses
Server fails to communicate with PD

Fix

Switch to bridge networking and container hostnames.

docker-compose.yml

Remove network_mode host
Use default bridge network
Add environment override hugegraph.pd.peers=pd:8686

docker/pd-conf/application.yml

grpc.host set to pd
raft.address set to pd:8610
raft.peers-list set to pd:8610
initial-store-list set to store:8500

docker/store-conf/application.yml

grpc.host set to store
raft.address set to store:8510
pdserver.address=pd:8686

Why This Works

Bridge networking is cross platform
Docker resolves container names automatically
Services advertise reachable addresses

Result After Fix

PD reachable via pd:8686
Store registers as store:8500
Partitions created normally
Server reachable on localhost 8080

bitflicker64:
## Purpose of the PR

Fix Docker deployment failing on macOS due to Linux-only host networking. Originally scoped to single-node, expanded during review to cover the full 3-node distributed cluster as well.

close #2951

## Main Changes

**Docker Compose:**
- Remove `network_mode: host` from both single-node and 3-node compose files
- Switch to Docker bridge network (`hg-net`) with container hostnames
- Replace config file volume mounts with environment variable injection (`HG_*` prefix)
- Add proper `depends_on: condition: service_healthy`, restart policies, and healthchecks

**Entrypoint Scripts (complete rewrite):**
- `hugegraph-pd/hg-pd-dist/docker/docker-entrypoint.sh` — `SPRING_APPLICATION_JSON` injection via `HG_PD_*` env vars, deprecated alias migration
- `hugegraph-store/hg-store-dist/docker/docker-entrypoint.sh` — `HG_STORE_*` env vars
- `hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh` — `HG_SERVER_*` env vars, auth support

**Helper Scripts:**
- `wait-storage.sh` — configurable PD auth
- `wait-partition.sh` — 120s timeout, configurable

**Documentation:**
- Created `docker/README.md` — full setup guide, env var reference, port table, troubleshooting
- Fixed 7 files pointing to dead `example/` directory
- Fixed `hugegraph-store/docs/deployment-guide.md` wrong env var names
- Updated K8s manifest env var names
- Added bridge network notes to PD configuration and README docs

## Problem

The original Docker configuration uses `network_mode: host` which only works on native Linux. Docker Desktop on macOS does not implement host networking the same way. Containers start but HugeGraph services advertise incorrect addresses (`127.0.0.1`, `0.0.0.0`).

Resulting failures:
- Server stuck in loop waiting for storage backend
- PD client UNAVAILABLE io exception errors
- Store reports zero partitions
- Cluster never becomes usable even though containers are running

## Root Cause

- `network_mode: host` is Linux-specific
- Docker Desktop falls back to bridge networking silently
- HugeGraph components advertise localhost-style addresses
- Other containers cannot route to those addresses

## Solution

Switch to bridge networking and advertise container-resolvable hostnames. Docker DNS resolves service names automatically. Configuration injected via `SPRING_APPLICATION_JSON` through `HG_*` env vars instead of mounted config files.

## Verification

Tested on macOS (Apple M4, Docker Desktop):

**Single-node cluster:**
```bash
docker ps --format "table {{.Names}}\t{{.Status}}\t{{.Ports}}"
hg-server   Up healthy   0.0.0.0:8080->8080
hg-store    Up healthy   0.0.0.0:8520->8520
hg-pd       Up healthy   0.0.0.0:8620->8620
```

**3-node cluster:**
```bash
curl -u store:admin -s http://localhost:8620/v1/stores | grep -o '"partitionCount":[0-9]*'
"partitionCount":12
"partitionCount":12
"partitionCount":12
```

All 3 stores show `partitionCount:12`. All 9 containers healthy.

**Also verified on Ubuntu 24.04 (native Docker):**
- All 9 containers healthy
- partitionCount:12 on all 3 stores
- Cluster works regardless of which PD node wins leader election

- [x] Modify configurations
- [ ] Dependencies
- [ ] The public API
- [ ] Other affects
- [ ] Nope

## Documentation Status

- [x] `Doc - Done`

---

Related fixes discovered during this work:

| Bug | Issue | Fix |
|-----|-------|-----|
| `getLeaderGrpcAddress()` NPE in bridge mode | #2959 | #2961 |
| `IpAuthHandler` hostname vs IP mismatch | #2960 | #2962 |
| 'Docker Logs' Not Working | #2979 | #2980 |
| Java process not supervised in Docker (container stays up after crash) | #3043 | #3044, #3047, #3051, #3052 |
---

## Changes Checklist
- [x] Updated Docker networking configuration
- [x] Added / verified required port mappings
- [x] Adjusted service communication to use container hostnames
- [x] Validated environment-based configuration
- [x] Verified PD, Store, and Server containers start correctly
- [x] Confirmed single-node cluster reaches healthy state
- [x] Confirmed partition assignment and leader election
- [x] Validated multi-node (3-node) cluster deployment
- [x] Tested on macOS (Apple M4)
- [x] Fixed server startup timeout to allow partition assignment to complete
- [x] Verified PD, Store and Server healthchecks pass on both single and 3-node
- [x] Test on Linux
- [x] Fixed `IpAuthHandler` hostname vs IP resolution (#2962)
- [x] Fixed `getLeaderGrpcAddress()` NPE (#2961)
- [x] Removed static IP workaround after `IpAuthHandler` fix
- [x] Update Docker images to reflect entrypoint changes #2976 
- [x] Update in-repo documentation — #2963
- [x] Fix docker Logs - #2980 
- [x] Update documentation (follow-up PR to hugegraph-doc) — apache/hugegraph-doc#455
- [x] Fix standalone RocksDB mode stuck at partition wait — #3000
- [x] improve all images build cache efficiency - #3025 
- [x] Fix foreground mode exit code propagation in startup scripts — #3044, #3047  
- [x] Fix Java process supervision in Docker entrypoints (replace `tail -f /dev/null`) — #3051  
- [x] Add `HEALTHCHECK` to all four Dockerfiles and remove unused `cron` — #3052


hugegraph-server/hugegraph-dist/src/assembly/static/bin/wait-storage.sh
```diff
@@ -70,7 +70,28 @@ done < <(env | sort -r | awk -F= '{ st = index($0, "="); print $1 " " substr($0,
...
+                    echo 'Hugegraph server are waiting for storage backend...'
+                    sleep 5
+                done
+                echo 'PD is reachable, waiting extra 10s for store registration...'
+                sleep 10
+                echo 'Storage backend is ready!'
+            " || echo "Warning: Timeout waiting for storage, proceeding anyway..."
+        else
+            echo "No pd.peers configured, skipping storage wait..."
+        fi
```
imbajin, line 129: ‼️ **Critical: wait-storage check downgraded from actual graph connectivity test to PD health ping only**

The PR deletes `detect-storage.groovy` and replaces the Gremlin-based storage readiness check with a simple `curl http://${PD_REST}/v1/health` + a fixed 10s sleep. This is a significant regression in correctness:

- PD being healthy **does not** mean: Store has registered, partitions are assigned, or the server can actually read/write the backend
- The old check used the Gremlin console to open the actual graph — a true end-to-end proof that storage is ready
- On timeout the new code silently proceeds: `echo "Warning: Timeout waiting for storage, proceeding anyway..."` — the server can start in a broken state with no hard failure

Suggested approach: restore a functional storage readiness probe. At minimum, poll the store's partition/leader endpoint (e.g. `/v1/health` or `/v1/partitions`) and verify non-zero partition/leader count before proceeding. Fail hard on timeout rather than warning-and-continue.

```suggestion
        timeout "${WAIT_STORAGE_TIMEOUT_S}s" bash -c "
            until curl -fsS http://${PD_REST}/v1/health >/dev/null 2>&1 &&
                  [ \"$(curl -fsS http://${STORE_REST}/v1/partitions 2>/dev/null | python3 -c 'import sys,json; d=json.load(sys.stdin); print(d.get(\"partitionCount\",0))' 2>/dev/null)\" -gt 0 ]; do
                echo 'Waiting for storage backend (PD + Store partitions)...'
                sleep 5
            done
            echo 'Storage backend is ready\!'
        " || { echo 'ERROR: Timeout waiting for storage backend'; exit 1; }
```

docker/docker-compose.yml
```diff
@@ -15,44 +15,98 @@
...
+    environment:
+      HG_PD_GRPC_HOST: pd
+      HG_PD_GRPC_PORT: "8686"
+      HG_PD_REST_PORT: "8620"
+      HG_PD_RAFT_ADDRESS: pd:8610
+      HG_PD_RAFT_PEERS_LIST: pd:8610
+      HG_PD_INITIAL_STORE_LIST: store:8500
+      HG_PD_DATA_PATH: /hugegraph-pd/pd_data
+    ports:
+      - "8620:8620"
```
imbajin, line 50: ‼️ **Internal cluster ports (gRPC/Raft) should not be published to the host**

The compose file now publishes several internal ports to the host that should only be accessible between containers:

- `8686:8686` — PD gRPC (internal)
- `8500:8500` — Store gRPC (internal)  
- `8510:8510` — Store Raft (internal)

These are cluster-internal communication ports. Exposing them on the host:
1. Unnecessarily expands the attack surface
2. Can conflict with other services on the host
3. Creates confusion for users about what's "public" vs "internal"

Only the REST/management ports need to be reachable from the host (`8620`, `8520`, `8080`). The gRPC/Raft ports should be removed from the `ports` section — they're already reachable between containers via the `hg-net` bridge network.

```suggestion
    ports:
      - "8620:8620"
```

docker/docker-compose.yml
```diff
@@ -15,44 +15,98 @@
 # limitations under the License.
 #
 
-version: "3"
+name: hugegraph-single
+
+networks:
+  hg-net:
+    driver: bridge
+
```
imbajin, line 23: ⚠️ **`build:` in quickstart compose can unexpectedly trigger local source builds — prefer pull-only defaults**

Using `build:` together with `image:` does **not** always force a local build. In Compose, the usual behavior (depending on `pull_policy`) is to try pulling first, then fall back to building if pull is unavailable.

That said, this is still risky for a quickstart file:

1. If the image/tag cannot be pulled, users unexpectedly need the full source tree and Docker build context.
2. Startup becomes much slower because services may be built locally.
3. Locally built images can differ from official release artifacts, reducing reproducibility.

Since `docker/docker-compose.yml` is intended for quickstart usage, I recommend keeping it pull-only by default and moving `build:` blocks to a dev-specific override (for example, `docker-compose.dev.yml`).

```suggestion
  pd:
    image: hugegraph/pd:${HUGEGRAPH_VERSION:-1.7.0}
    container_name: hg-pd
    hostname: pd
```

If local builds are needed for development, users can combine files explicitly, e.g.:
`docker compose -f docker/docker-compose.yml -f docker/docker-compose.dev.yml up`

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
```diff
@@ -15,32 +15,66 @@
...
+fail_on_deprecated() {
+    local old_name="$1" new_name="$2"
+    if [[ -n "${!old_name:-}" ]]; then
+        echo "ERROR: deprecated env '${old_name}' detected. Use '${new_name}' instead." >&2
+        exit 2
+    fi
+}
+
+set_prop() {
+    local key="$1" val="$2" file="$3"
```
imbajin, line 29: 🧹 (Optional)
**`set_prop` currently injects raw values into `sed` — escape replacement text to avoid malformed properties**

`set_prop` writes `${val}` directly into a `sed` replacement expression. If the value contains characters such as `#`, `&`, `|`, or `\`, the generated line can be corrupted.

In this script, key-regex injection is lower risk because keys are fixed constants (`backend`, `pd.peers`), but value escaping is still important for correctness and future-proofing.

```suggestion
set_prop() {
    local key="$1" val="$2" file="$3"
    local esc_key esc_val

    # Escape key for regex matching and value for sed replacement.
    esc_key=$(printf '%s' "$key" | sed -e 's/[][(){}.^$*+?|\\/]/\\&/g')
    esc_val=$(printf '%s' "$val" | sed -e 's/[&|\\]/\\&/g')

    if grep -qE "^[[:space:]]*${esc_key}[[:space:]]*=" "${file}"; then
        sed -ri "s|^([[:space:]]*${esc_key}[[:space:]]*=).*|\\1${esc_val}|" "${file}"
    else
        printf '%s=%s\n' "$key" "$val" >> "${file}"
    fi
}
```

This keeps the current behavior while making updates safe for common special characters in values.

hugegraph-pd/hg-pd-dist/docker/docker-entrypoint.sh
```diff
@@ -15,8 +15,71 @@
...
+# ── Build SPRING_APPLICATION_JSON ─────────────────────────────────────
+SPRING_APPLICATION_JSON="$(cat <<JSON
+{
+  "grpc":   { "host": "$(json_escape "${HG_PD_GRPC_HOST}")",
+              "port": "$(json_escape "${HG_PD_GRPC_PORT}")" },
+  "server": { "port": "$(json_escape "${HG_PD_REST_PORT}")" },
+  "raft":   { "address":    "$(json_escape "${HG_PD_RAFT_ADDRESS}")",
+              "peers-list": "$(json_escape "${HG_PD_RAFT_PEERS_LIST}")" },
+  "pd":     { "data-path":          "$(json_escape "${HG_PD_DATA_PATH}")",
+              "initial-store-list": "$(json_escape "${HG_PD_INITIAL_STORE_LIST}")" }
```
imbajin: ⚠️ **`HG_PD_INITIAL_STORE_COUNT` is not currently mapped into `SPRING_APPLICATION_JSON` — recommend explicit wiring**

At the moment, `HG_PD_INITIAL_STORE_COUNT` is not included in the generated `SPRING_APPLICATION_JSON`. This may still work in current packaging because `conf/application.yml` already sets `pd.initial-store-count: 1` for single-node, and JSON only overrides selected keys.

However, explicitly wiring this value is safer and clearer:

1. Avoids hidden dependency on base file defaults.
2. Keeps behavior stable if packaged defaults change later.
3. Exposes a useful tuning knob for multi-node bootstrap scenarios.

```suggestion
: "${HG_PD_INITIAL_STORE_COUNT:=1}"

SPRING_APPLICATION_JSON="$(cat <<JSON
{
  "grpc":   { "host": "$(json_escape "${HG_PD_GRPC_HOST}")",
              "port": "$(json_escape "${HG_PD_GRPC_PORT}")" },
  "server": { "port": "$(json_escape "${HG_PD_REST_PORT}")" },
  "raft":   { "address":    "$(json_escape "${HG_PD_RAFT_ADDRESS}")",
              "peers-list": "$(json_escape "${HG_PD_RAFT_PEERS_LIST}")" },
  "pd":     { "data-path":          "$(json_escape "${HG_PD_DATA_PATH}")",
              "initial-store-list": "$(json_escape "${HG_PD_INITIAL_STORE_LIST}")",
              "initial-store-count": ${HG_PD_INITIAL_STORE_COUNT} }
}
JSON
)"
```

`initial-store-count` should remain a numeric JSON value (unquoted).

hugegraph-server/hugegraph-dist/docker/scripts/detect-storage.groovy
```diff
@@ -1,26 +0,0 @@
...
-
-import org.apache.hugegraph.HugeFactory
-import org.apache.hugegraph.dist.RegisterUtil
-
-// register all the backend to avoid changes if needs to support other backend
-RegisterUtil.registerPlugins()
-RegisterUtil.registerRocksDB()
-RegisterUtil.registerHBase()
-
-graph = HugeFactory.open('./conf/graphs/hugegraph.properties')
```
imbajin, line 26: ‼️ **Critical: deleting this script while related Docker references remain can break build workflows**

This file is removed in this PR, but the server Docker build flow still references `detect-storage.groovy` in current Dockerfiles. That creates an inconsistency and can fail local build paths when Compose/build expects the file.

Please align this PR by either:
1. removing stale `COPY ... detect-storage.groovy` references from server Dockerfiles, or
2. keeping a temporary compatibility stub until those references are removed.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
```diff
@@ -15,32 +15,66 @@
...
+    fi
+}
 
-if [ ! -f "${DOCKER_FOLDER}/${INIT_FLAG_FILE}" ]; then
-    # wait for storage backend
-    ./bin/wait-storage.sh
-    if [ -z "$PASSWORD" ]; then
-        echo "init hugegraph with non-auth mode"
+# ── Guard deprecated vars ─────────────────────────────────────────────
+fail_on_deprecated "BACKEND"  "HG_SERVER_BACKEND"
```
imbajin: ⚠️ **Important: hard-failing deprecated env vars can create avoidable upgrade breakage**

Current behavior exits when legacy env names are present (`BACKEND`, `PD_PEERS`), even when values are valid. This is strict but can disrupt existing users upgrading old compose/manifests.

A smoother migration path is to map old names to new names with warnings, then enforce hard-fail in a later release.

```suggestion
migrate_env() {
    local old_name="$1" new_name="$2"
    if [[ -n "${!old_name:-}" && -z "${!new_name:-}" ]]; then
        log "WARN: deprecated env '${old_name}' detected; mapping to '${new_name}'"
        export "${new_name}=${!old_name}"
    fi
}

migrate_env "BACKEND"  "HG_SERVER_BACKEND"
migrate_env "PD_PEERS" "HG_SERVER_PD_PEERS"
```

Please consider the same compatibility behavior for PD/Store entrypoints for consistency.

docker/docker-compose.yml
```diff
@@ -15,44 +15,98 @@
...
+  hg-store-data:
 
 services:
   pd:
-    image: hugegraph/pd
-    container_name: pd
+    build:
+      context: ..
+      dockerfile: hugegraph-pd/Dockerfile
+    image: hugegraph/pd:${HUGEGRAPH_VERSION:-1.7.0}
```
imbajin: ⚠️ **Important: quickstart default tag may cause pull failures or unexpected local builds**

This compose file defaults to `${HUGEGRAPH_VERSION:-1.7.0}`. If that tag is unavailable in registry, quickstart users can hit pull failures (or unexpected local builds when `build:` is present).

For a quickstart-oriented file, prefer a reliably available default tag and let users override explicitly.

```suggestion
    image: hugegraph/pd:${HUGEGRAPH_VERSION:-latest}
```

Please keep PD/Store/Server default tag policy aligned to avoid mixed-version startup behavior.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/wait-storage.sh
```diff
@@ -70,7 +70,28 @@ done < <(env | sort -r | awk -F= '{ st = index($0, "="); print $1 " " substr($0,
...
-        "until bin/gremlin-console.sh -- -e $DETECT_STORAGE > /dev/null 2>&1; do echo \"Hugegraph server are waiting for storage backend...\"; sleep 5; done"
+        # Extract pd.peers from config or environment
+        PD_PEERS="${hugegraph_pd_peers:-}"
+        if [ -z "$PD_PEERS" ]; then
+            PD_PEERS=$(grep -E "^\s*pd\.peers\s*=" "$GRAPH_CONF" | sed 's/.*=\s*//' | tr -d ' ')
+        fi
+
+        if [ -n "$PD_PEERS" ]; then
+            # Convert gRPC address to REST address (8686 -> 8620)
+            PD_REST=$(echo "$PD_PEERS" | sed 's/:8686/:8620/g' | cut -d',' -f1)
```
imbajin: ⚠️ **Important: PD REST endpoint derivation assumes fixed ports (`8686 -> 8620`)**

Current logic derives PD REST by rewriting `:8686` to `:8620`. This works for defaults, but fails in non-default port configurations and can produce false startup timeouts.

Please support an explicit override first, and keep current derivation only as fallback.

```suggestion
: "${HG_SERVER_PD_REST_ENDPOINT:=}"
if [ -n "${HG_SERVER_PD_REST_ENDPOINT}" ]; then
    PD_REST="${HG_SERVER_PD_REST_ENDPOINT}"
else
    PD_REST=$(echo "$PD_PEERS" | sed 's/:8686/:8620/g' | cut -d',' -f1)
fi
```

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
bitflicker64: Updated the entrypoint to ensure deprecated environment variables no longer prevent startup. Legacy variables are now mapped to their modern equivalents with a warning, preserving compatibility with existing deployments while maintaining current behavior for properly configured environments.

docker/docker-compose.yml
bitflicker64: I’ve changed the compose setup so it’s pull-only and no longer builds locally. Right now, I’m bind-mounting the updated entrypoint scripts because the current published images don’t include these changes.
Once this PR is done and new images are available, I can remove the mounts. I’ll follow up with a small cleanup PR to switch completely to the official images.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/wait-storage.sh
bitflicker64: Partition-based readiness checks may be unreliable since partition assignment occurs asynchronously after wait-storage completes, and a properly registered Store can legitimately report partitionCount = 0 during normal initialization. Interpreting this as a failure condition could unintentionally block startup in otherwise healthy clusters. Would it make sense to consider separating validation into pre-startup checks (Store/PD availability) and post-startup checks (partition stabilization) instead?

imbajin: > Partition-based readiness checks may be unreliable since partition assignment occurs asynchronously after wait-storage completes, and a properly registered Store can legitimately report partitionCount = 0 during normal initialization. Interpreting this as a failure condition could unintentionally block startup in otherwise healthy clusters. Would it make sense to consider separating validation into pre-startup checks (Store/PD availability) and post-startup checks (partition stabilization) instead?

Your point is right: `partitionCount` can legitimately be `0` during early initialization, so using it as a strict pre-start gate can block healthy clusters.

The core concern I want to preserve is:
1. Current readiness is too weak (`PD /v1/health + fixed sleep` only).  
2. Timeout currently does `warning + continue`, which allows starting in a broken state (no fail-fast).

Maybe we could split validation into pre-start hard gates vs post-start stabilization checks:

```text
docker compose up
   |
   v
[Pre-start Gate]  (must pass, otherwise exit)
  - PD /v1/health == OK
  - Store /v1/health == OK
  - PD /v1/stores has at least 1 Up store
   |
   v
[Init-store with retries] (must succeed, otherwise exit)
   |
   v
[Start server]
   |
   v
[Post-start checks] (non-blocking)
  - observe partition/leader stabilization
  - warn if not converged yet, but do not treat as pre-start failure
  ```

Comment (bitflicker64) [reply hugegraph-server/hugegraph-dist/src/assembly/static/bin/wait-storage.sh:129]: Thanks for the clarification that matches what I observed during testing. I’ve updated my PR to separate pre-start readiness checks from post-start partition stabilization.

Comment (bitflicker64) [reply hugegraph-server/hugegraph-dist/src/assembly/static/bin/wait-storage.sh:129]: In follow up commit i will make sure to add hard gates to the pre start ( `wait-storage.sh`)

**Comment by bitflicker64** [comment]
@imbajin  **Update on bridge network 3-node cluster investigation:**

Hi, I've been looking into why the 3-node PD cluster fails to form raft quorum in Docker bridge network mode. To rule out any config loading issues, I tried both the env var based entrypoint approach and a fully hardcoded config, both produced the same error, so I don't think it's config related.

Both attempts show this in the logs:

```
IpAuthHandler - Blocked connection from 172.18.0.x
```

Tracing it back, I noticed this in `RaftEngine.java` around line 159:

```java
IpAuthHandler.getInstance(
    peers.stream()
        .map(PeerId::getIp)
        .collect(Collectors.toSet())
);
```

I think `PeerId::getIp` might be returning the raw hostname string (e.g. `pd1`) rather than the resolved IP. In bridge mode, incoming connections arrive with their actual container IP (`172.18.0.x`), so the allowlist check would always fail and block cross-node raft connections. This wouldn't surface in `network_mode: host` since all connections appear as `127.0.0.1`.

Could you take a look and let me know if this diagnosis makes sense? I might be missing something  still getting familiar with the codebase. If this does seem right, I'd be happy to try putting together a fix.

Thanks!

hugegraph-server/hugegraph-dist/src/assembly/static/bin/wait-storage.sh
```diff
@@ -29,11 +29,17 @@ function abs_path() {
...
+WAIT_STORAGE_TIMEOUT_S=300
 
 . "$BIN"/util.sh
 
+log() {
+  echo "[wait-storage] $1"
+}
+
+# Hardcoded PD auth
+PD_AUTH_ARGS="-u store:admin"
```
imbajin: ‼️ **Hardcoded credentials**

`PD_AUTH_ARGS` is hardcoded to `-u store:admin`. Any deployment using a non-default password will silently fail here since there is no way to override this value.

Consider reading from env vars with a fallback:
```suggestion
PD_AUTH_ARGS="-u ${PD_AUTH_USER:-store}:${PD_AUTH_PASSWORD:-admin}"
```

hugegraph-pd/hg-pd-dist/docker/docker-entrypoint.sh
```diff
@@ -15,8 +15,79 @@
...
+log "  grpc.port=${HG_PD_GRPC_PORT}"
+log "  server.port=${HG_PD_REST_PORT}"
+log "  raft.address=${HG_PD_RAFT_ADDRESS}"
+log "  raft.peers-list=${HG_PD_RAFT_PEERS_LIST}"
+log "  pd.initial-store-list=${HG_PD_INITIAL_STORE_LIST}"
+log "  pd.initial-store-count=${HG_PD_INITIAL_STORE_COUNT}"
+log "  pd.data-path=${HG_PD_DATA_PATH}"
+
+log "Patching conf/application.yml with cluster settings..."
+sed -i "s|address: 127.0.0.1:8610|address: ${HG_PD_RAFT_ADDRESS}|"                   conf/application.yml
```
imbajin: ‼️ **Dual config mechanism — `SPRING_APPLICATION_JSON` and `sed` both applied**

This entrypoint sets config via `SPRING_APPLICATION_JSON` (which Spring Boot will use at startup), and then also patches `conf/application.yml` with `sed -i`. These two approaches overlap on the same keys.

The `sed` commands silently do nothing if the default values in `application.yml` have changed (e.g. `127.0.0.1` was already replaced in a new image build). This makes the effective configuration hard to reason about.

Consider removing the `sed` block entirely and relying solely on `SPRING_APPLICATION_JSON`, which is the standard Spring Boot override mechanism and has higher priority than the yml file anyway.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
```diff
@@ -15,32 +15,78 @@
...
-    echo "Hugegraph Initialization already done. Skipping re-init..."
+    log "HugeGraph initialization already done. Skipping re-init..."
 fi
 
-# start hugegraph-server
-# remove "-g zgc" now, which is only available on ARM-Mac with java > 13 
-./bin/start-hugegraph.sh -j "$JAVA_OPTS" 
+./bin/start-hugegraph.sh -j "${JAVA_OPTS:-}" -t 120
+
+STORE_REST="${STORE_REST:-hg-store:8520}"
```
imbajin: ⚠️ **`STORE_REST` default is set after `start-hugegraph.sh`, breaking single-node deployments**

`STORE_REST` is assigned its default on line 86, but `start-hugegraph.sh` is called on line 84. More importantly, `docker-compose.yml` (single-node) does not pass `STORE_REST` as an environment variable, so when `wait-partition.sh` runs it will hit:

```
: "${STORE_REST:?STORE_REST not set}"
```

...and exit 1, crashing the container.

The fix is to move the default assignment and export to **before** `start-hugegraph.sh`:
```suggestion
STORE_REST="${STORE_REST:-store:8520}"
export STORE_REST

./bin/start-hugegraph.sh -j "${JAVA_OPTS:-}" -t 120
```

docker/docker-compose-3pd-3store-3server.yml
```diff
@@ -15,166 +15,218 @@
...
+  hg-pd0-data:
+  hg-pd1-data:
+  hg-pd2-data:
+  hg-store0-data:
+  hg-store1-data:
+  hg-store2-data:
+
+# ── Shared service templates ──────────────────────────────────────────
+
+x-pd-common: &pd-common
```
imbajin, line 34: ⚠️ **`x-pd-common` template is missing `networks: [hg-net]`**

The `x-store-common` and `x-server-common` templates both include `networks: [hg-net]`, but `x-pd-common` does not. Each PD service currently works around this by inlining its own `networks:` block, but the template is incomplete. Anyone adding a new PD node while following the existing pattern would end up with a container not connected to the bridge network.

docker/docker-compose-3pd-3store-3server.yml
```diff
@@ -15,166 +15,218 @@
...
+  image: hugegraph/pd:${HUGEGRAPH_VERSION:-latest}
+  pull_policy: always
```
imbajin: ⚠️ **`pull_policy: always` will break offline / air-gapped deployments**

With `pull_policy: always`, every `docker compose up` attempts to pull the latest image from the registry. This will fail in environments without internet access (air-gapped servers, CI with no egress, etc.).

Consider changing to `pull_policy: missing` (pull only if the image is not present locally), which is safer as a default while still ensuring a fresh pull when images are not cached.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/wait-storage.sh
imbajin, line 1: There’s one more thing that might be worth looking at: in `wait-storage.sh`, the PD REST address is derived from `pd.peers`, but in practice the script only checks the first peer.

This works fine in the current compose example, but it still feels a bit fragile in a multi-PD setup. For example, during a fairly common rolling restart, a single-node upgrade, or when one PD node simply comes up more slowly than the others, if the first PD in the list is temporarily unavailable, the script will keep treating PD as not ready even if the other two PD nodes are already up and working normally.

So I think a more reliable approach here would be either:
- explicitly passing a PD REST endpoint, or
- trying the peers from `pd.peers` in order instead of checking only the first one

This may not cause an immediate problem in the default compose setup, but from a multi-PD robustness perspective, I think it’s still worth noting.

VGalaxies (dismissed):
LGTM.

@imbajin Since #2962 is already merged and #2961 still seems relevant for the 3-node path, my suggestion would be to land #2961 first, then this PR, and follow with #2963.

docker/docker-compose.yml
```diff
@@ -14,45 +14,119 @@
...
     hostname: pd
-    network_mode: host
+    restart: unless-stopped
+    networks: [hg-net]
+
+    entrypoint: ["/hugegraph-pd/docker-entrypoint.sh"]
+
+    environment:
+      HG_PD_GRPC_HOST: pd
+      HG_PD_GRPC_PORT: "8686"
```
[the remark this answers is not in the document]
bitflicker64: The entrypoint: overrides in the compose files are a temporary workaround to bind-mount the updated scripts since the current published images predate the entrypoint changes. These will be removed in a follow-up cleanup PR once updated images are published

VGalaxies (approved):

imbajin (approved):
@VGalaxies  After this PR merged, we need to update the server/pd/store image in Dockerhub (Action repo)

PR #2961, opened 2026-03-05
fix(pd): add timeout and null-safety to getLeaderGrpcAddress()

### LINKED ISSUE #2959: [Bug] 3-node PD cluster fails when pd0 is not raft leader — getLeaderGrpcAddress() NPE in bridge network mode
### Bug Type (问题类型)

others (please edit later)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0
- Backend: hstore (3 PD + 3 Store + 3 Server)
- OS: macOS Apple M4
- related issue - #2951 
- related pr - #2952 
- Network: Docker bridge mode (static IPs via ipam)
- Docker Desktop: latest

## Summary
In a 3-node PD cluster running in Docker bridge network mode, the cluster 
only works correctly when pd0 wins the raft leader election. If pd1 or pd2 
becomes leader, store registration fails, partitions are never distributed, 
and HugeGraph servers cannot initialize.

## Root Cause
In RaftEngine.java, getLeaderGrpcAddress() makes a live bolt RPC call to 
discover the leader's gRPC address when the current node is a follower:

    return raftRpcClient.getGrpcAddress(
        raftNode.getLeaderId().getEndpoint().toString()
    ).get().getGrpcAddress();  // .get() returns null in bridge mode → NPE

This call fails in Docker bridge mode — the TCP connection establishes 
successfully but the bolt RPC response never returns, causing 
CompletableFuture.get() to return null and throw NPE.

This causes:
1. redirectToLeader() fails with NPE
2. Store registration requests landing on follower PDs are never forwarded
3. Stores register but partitions are never distributed (partitionCount:0)
4. HugeGraph servers stuck in DEADLINE_EXCEEDED loop indefinitely

## Why It Only Affects Bridge Mode
In host network mode all PD nodes communicate via 127.0.0.1 — the bolt RPC 
call succeeds instantly over loopback. In bridge mode the call traverses 
Docker's virtual network and the response never arrives properly.

## Why It's Nondeterministic
JRaft leader election is timing-based. If pd0 wins, isLeader() returns true 
and the broken code path is never reached. If pd1 or pd2 wins, pd0 becomes 
a follower and hits the NPE on every redirect attempt.

## Reproduction
1. Run the 3-node cluster in Docker bridge mode
2. Check which PD won leader election:
   docker exec hg-pd0 grep "becomes leader" /hugegraph-pd/logs/hugegraph-pd-stdout.log
   docker exec hg-pd1 grep "becomes leader" /hugegraph-pd/logs/hugegraph-pd-stdout.log
   docker exec hg-pd2 grep "becomes leader" /hugegraph-pd/logs/hugegraph-pd-stdout.log
3. If pd1 or pd2 is leader, check store partitions:
   curl -u store:admin http://localhost:8620/v1/stores | grep partitionCount
   → store1 and store2 will show partitionCount:0
4. Check pd0 logs for the NPE:
   docker exec hg-pd0 grep "getLeaderGrpcAddress" /hugegraph-pd/logs/hugegraph-pd-stdout.log

## Evidence

The NPE occurs in two separate call paths:

**Path 1 — fires immediately when node becomes follower during leader election:**
```
java.util.concurrent.ExecutionException: java.lang.NullPointerException
    at java.util.concurrent.CompletableFuture.reportGet(Unknown Source)
    at java.util.concurrent.CompletableFuture.get(Unknown Source)
    at org.apache.hugegraph.pd.raft.RaftEngine.getLeaderGrpcAddress(RaftEngine.java:242)
    at org.apache.hugegraph.pd.service.PDService.onRaftLeaderChanged(PDService.java:1345)
    at org.apache.hugegraph.pd.raft.RaftStateMachine.lambda$onStartFollowing$1(RaftStateMachine.java:141)
```
```
**Path 2 — fires on every store registration redirect attempt:**
java.util.concurrent.ExecutionException: java.lang.NullPointerException
    at java.util.concurrent.CompletableFuture.reportGet(Unknown Source)
    at java.util.concurrent.CompletableFuture.get(Unknown Source)
    at org.apache.hugegraph.pd.raft.RaftEngine.getLeaderGrpcAddress(RaftEngine.java:242)
    at org.apache.hugegraph.pd.service.PDService.redirectToLeader(PDService.java:1275)
```
**proof:** restarting the non-pd0 leader forcing pd0 to win election 
immediately resolved the issue — partitionCount:12 on all 3 stores, all 9 
containers healthy. Reproducible 100% of the time.

bitflicker64:
## Purpose of the PR
- close #2959

In a 3-node PD cluster running in Docker bridge network mode, `getLeaderGrpcAddress()` makes a bolt RPC call to discover the leader's gRPC address when the current node is a follower. This call fails in bridge mode — the TCP connection establishes but the bolt RPC response never returns properly, causing `CompletableFuture.get()` to return null and throw NPE.
This causes:
1. `redirectToLeader()` fails with NPE
2. Store registration requests landing on follower PDs are never forwarded
3. Stores register but partitions are never distributed (`partitionCount:0`)
4. HugeGraph servers stuck in `DEADLINE_EXCEEDED` loop indefinitely

The cluster only works when pd0 wins raft leader election (since `isLeader()` returns true and the broken code path is skipped). If pd1 or pd2 wins, the NPE fires on every redirect attempt.

Related PR: #2952

## Main Changes
- Cache leader `PeerId` after `waitingForLeader()` and null-check to prevent NPE when leader election times out
- Wire `config.getRpcTimeout()` into `RaftRpcClient`'s `RpcOptions` so Bolt transport timeout is consistent with `future.get()` caller timeout
- Add bounded timeout to the bolt RPC call using `config.getRpcTimeout()` instead of unbounded `.get()`
- Split `catch (TimeoutException | ExecutionException)` into separate blocks to avoid double-wrapping root cause
- Restore best-effort fallback to `leaderIp + grpcPort` with warn logging when RPC fails, times out, or returns null
- Improve `waitingForLeader()` to respect sub-second timeouts using `Math.min(1000, remaining)`
- Add unit tests covering all failure paths with mocked `RaftRpcClient`

## Verifying these changes
- [x] Done testing and can be verified as follows:
    - Deploy 3-node PD cluster in Docker bridge network mode
    - Verify cluster works regardless of which PD node wins raft leader election
    - Confirm stores show `partitionCount:12` on all 3 nodes when pd1 or pd2 is leader
    - Confirm no NPE in pd logs at `getLeaderGrpcAddress`
    - Unit tests: `RaftEngineLeaderAddressTest` covers timeout, RPC exception, null response, and null leader branches

- [ ]  Dependencies ([[add/update license](https://hugegraph.apache.org/docs/contribution-guidelines/contribute/#321-check-licenses)](https://hugegraph.apache.org/docs/contribution-guidelines/contribute/#321-check-licenses) info & [regenerate_known_dependencies.sh](../install-dist/scripts/dependency/regenerate_known_dependencies.sh))
- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status
- [ ]  `Doc - TODO`
- [ ]  `Doc - Done`
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
+++ b/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
@@ -277,7 +277,7 @@ public class RaftEngine {
         // WARNING: this may be incorrect in clusters where PD nodes use different grpc.port
         // values, a proper fix requires a cluster-wide source of truth for gRPC addresses.
         String derived = leader.getEndpoint().getIp() + ":" + config.getGrpcPort();
-        log.warn("Using derived leader gRPC address {} — may be incorrect if nodes use different ports",
+        log.info("Using derived leader gRPC address {} - may be incorrect if nodes use different ports",
                  derived);
         return derived;
     }
@@ -415,7 +415,8 @@ public class RaftEngine {
             long start = System.currentTimeMillis();
             while ((System.currentTimeMillis() - start < timeOut) && (leader == null)) {
                 try {
-                    this.wait(1000);
+                    long remaining = timeOut - (System.currentTimeMillis() - start);
+                    this.wait(Math.min(1000, Math.max(0, remaining)));
                 } catch (InterruptedException e) {
                     log.error("Raft wait for leader exception", e);
                 }
--- a/hugegraph-pd/hg-pd-test/src/main/java/org/apache/hugegraph/pd/raft/RaftEngineLeaderAddressTest.java
+++ b/hugegraph-pd/hg-pd-test/src/main/java/org/apache/hugegraph/pd/raft/RaftEngineLeaderAddressTest.java
@@ -74,7 +74,7 @@ public class RaftEngineLeaderAddressTest {
         when(mockNode.getLeaderId()).thenReturn(mockLeader);
 
         // Build mock config
-        // Use a short timeout (100ms) so the null-leader test doesn't block for seconds
+        // Use a short default timeout (100ms); specific tests may override getRpcTimeout()
         mockConfig = mock(PDConfig.Raft.class);
         when(mockConfig.getGrpcAddress()).thenReturn("127.0.0.1:" + GRPC_PORT);
         when(mockConfig.getGrpcPort()).thenReturn(GRPC_PORT);


bitflicker64:
**How I tested:**

1. Built a local Docker image from source with this fix applied
2. Brought up the 3-node cluster (3 PD + 3 Store + 3 Server) in bridge network mode
3. Confirmed cluster was healthy with pd0 as initial leader
4. Restarted pd0 to force a new leader election — pd1 won
5. Checked partition distribution and cluster health with pd1 as leader

**Results with pd1 as leader:**
```
partitionCount:12 on all 3 stores ✅
leaderCount:12 on all 3 stores ✅
{"graphs":["hugegraph"]} ✅
All 9 containers healthy ✅
```

**Confirmed fallback triggered in pd1 logs:**
```
[WARN] RaftEngine - Failed to get leader gRPC address via RPC, falling back to endpoint derivation
java.util.concurrent.ExecutionException: com.alipay.remoting.exception.RemotingException:
Create connection failed. The address is 172.20.0.10:8610
    at RaftEngine.getLeaderGrpcAddress(RaftEngine.java:247)
    at PDService.redirectToLeader(PDService.java:1275)
```

**Before this fix:** RPC returns null → NPE → follower PDs can't redirect requests to leader → cluster only worked when pd0 won leader election since it never hit the broken code path.

**After this fix:** RPC failure caught with bounded timeout → fallback to endpoint IP + gRPC port derivation → follower PDs correctly redirect to leader regardless of which PD node wins election.

Related docker bridge networking PR: #2952

hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
```diff
@@ -239,8 +241,20 @@ public String getLeaderGrpcAddress() throws ExecutionException, InterruptedExcep
...
+            if (response != null && response.getGrpcAddress() != null) {
+                return response.getGrpcAddress();
+            }
+        } catch (TimeoutException | ExecutionException e) {
+            log.warn("Failed to get leader gRPC address via RPC, falling back to endpoint derivation", e);
+        }
+
+        // Fallback: derive from raft endpoint IP + local gRPC port (best effort)
+        String leaderIp = raftNode.getLeaderId().getEndpoint().getIp();
+        return leaderIp + ":" + config.getGrpcPort();
```
imbajin: ‼️ This fallback is still incorrect for clusters where PD nodes use different `grpc.port` values. In this repo's own multi-node test configs, `application-server1.yml`, `application-server2.yml`, and `application-server3.yml` advertise `8686`, `8687`, and `8688` respectively, so a follower on `8687` will redirect to `leader-ip:8687` even when the elected leader is actually listening on `8686` or `8688`. That turns the original NPE into a silent misroute.

If we can't recover the leader's advertised gRPC endpoint here, I think it's safer to fail fast than to synthesize an address from the local port, for example:

```suggestion
        } catch (TimeoutException | ExecutionException e) {
            throw new ExecutionException(
                    String.format("Failed to resolve leader gRPC address for %s", raftNode.getLeaderId()),
                    e);
        }
```

A more complete fix would need a source of truth for the leader's actual `grpcAddress`, not the local node's port.

hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
```diff
@@ -239,8 +241,20 @@ public String getLeaderGrpcAddress() throws ExecutionException, InterruptedExcep
             waitingForLeader(10000);
         }
 
-        return raftRpcClient.getGrpcAddress(raftNode.getLeaderId().getEndpoint().toString()).get()
-                            .getGrpcAddress();
+        try {
+            RaftRpcProcessor.GetMemberResponse response = raftRpcClient
+                    .getGrpcAddress(raftNode.getLeaderId().getEndpoint().toString())
```
imbajin: ⚠️ `waitingForLeader(10000)` does not guarantee that `leaderId` is non-null; it just waits up to the timeout and returns whatever it has. That means both the RPC call and this fallback path can still dereference `raftNode.getLeaderId()` and reintroduce an NPE during leader-election gaps.

Could we cache the leader once after waiting and turn the "leader still unknown" case into a controlled exception instead?

```suggestion
        PeerId leader = raftNode.getLeaderId();
        if (leader == null) {
            throw new ExecutionException(new IllegalStateException("Leader is not ready"));
        }
```

Then reuse `leader` for both the RPC request and any follow-up handling.

hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
```diff
@@ -239,8 +241,20 @@ public String getLeaderGrpcAddress() throws ExecutionException, InterruptedExcep
...
+                    .get(config.getRpcTimeout(), TimeUnit.MILLISECONDS);
```
imbajin, line 256: ⚠️ `config.getRpcTimeout()` only bounds the callers `CompletableFuture.get(...)` here; it does **not** bound the underlying Bolt RPC. `RaftRpcClient` is still initialized with a fresh `new RpcOptions()` and `invokeAsync(...)` uses `rpcOptions.getRpcDefaultTimeout()`, so in the bridge-network failure case well return from this method quickly but still leave the actual RPC running until the client default timeout fires. On a follower that keeps redirecting traffic, those in-flight requests can accumulate.

Could we wire the configured timeout into `RaftRpcClient` itself (or cancel the future on timeout) so the transport timeout and the caller timeout stay consistent?

```text
Follower Request
       |
       v
getLeaderGrpcAddress()
       |
       +--> future.get(config.rpcTimeout)
       |          |
       |          +--> caller returns early on timeout
       |
       +--> invokeAsync(..., rpcOptions.getRpcDefaultTimeout())
                  |
                  +--> underlying Bolt RPC may still be running
                  |
                  v
        repeated follower redirects
                  |
                  v
        in-flight RPC requests can accumulate

```

hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
```diff
@@ -228,19 +229,49 @@ public PeerId getLeader() {
...
+            throw new ExecutionException(
+                    String.format("Timed out while resolving leader gRPC address for %s", leader),
+                    e);
+        } catch (ExecutionException e) {
+            // TODO: a more complete fix would need a source of truth for the leader's
+            // actual grpcAddress rather than deriving it from the local node's port config.
+            Throwable cause = e.getCause() != null ? e.getCause() : e;
+            throw new ExecutionException(
+                    String.format("Failed to resolve leader gRPC address for %s", leader), cause);
+        }
```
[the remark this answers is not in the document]
imbajin: Yep, the PR said this method should fall back to deriving the leader gRPC address from the Raft leader endpoint plus the local gRPC port when the Bolt RPC hangs / fails / returns null. 

However, in the current implementation, all of those paths still end by throwing `ExecutionException`:

- `TimeoutException` -> rethrown as `ExecutionException`
- RPC failure (`ExecutionException`) -> rethrown
- `response == null` or `response.getGrpcAddress() == null` -> throws again after the warning

That means the follower still does **not** get a usable leader gRPC address in the exact scenario this PR is trying to fix.

Contextually, the call chain still looks like this:

```text
Follower request
    |
    v
getLeaderGrpcAddress()
    |
    +-- Bolt RPC timeout / failure / null response
            |
            v
      throws ExecutionException
            |
            v
redirectToLeader()
    |
    +-- logs / catches exception
    |
    v
request is not forwarded to leader
```

```
Current:
  RPC fails/null
      -> throw
      -> no redirect

Expected:
  RPC fails/null
      -> derive leaderIp:grpcPort
      -> redirect continues
```

hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
```diff
@@ -228,19 +229,49 @@ public PeerId getLeader() {
...
-        return raftRpcClient.getGrpcAddress(raftNode.getLeaderId().getEndpoint().toString()).get()
-                            .getGrpcAddress();
+        // Cache leader to avoid repeated getLeaderId() calls and guard against
+        // waitingForLeader() returning without a leader being elected.
+        PeerId leader = raftNode.getLeaderId();
+        if (leader == null) {
+            throw new ExecutionException(new IllegalStateException("Leader is not ready"));
+        }
+
+        try {
```
imbajin, line 251: ⚠️ There is still no focused test covering the new `getLeaderGrpcAddress()` failure paths here. The current PD test additions exercise `changePeerList()` / `IpAuthHandler`, but not the timeout, RPC-exception, or null-response branches introduced in this method. 

Could add a small UIT with a mocked `RaftRpcClient` so regressions in the leader-resolution path are caught by CI.

PR #2962, opened 2026-03-05
fix(pd): resolve hostname entries in IpAuthHandler allowlist

### LINKED ISSUE #2960: [Bug] IpAuthHandler blocks all cross-node raft connections in Docker bridge mode — hostname vs IP mismatch
### Bug Type (问题类型)

None

### Before submit

## Environment
- HugeGraph PD version: 1.7.0
- Docker network mode: bridge (static IPs via ipam)
- Tested on: macOS Apple M4
- related issue - #2951 
- Related PR: #2952
- Related bug: IpAuthHandler hostname vs IP (same bridge mode root cause)

Expected: Cross-node raft connections between PD nodes work correctly in 
Docker bridge network mode.

Actual: All cross-node raft connections are silently blocked by 
IpAuthHandler. The allowlist is built using PeerId::getIp which returns 
raw hostnames (e.g. "pd0", "pd1") but incoming connections arrive with 
actual bridge IPs (e.g. "172.18.0.4") — the string comparison always 
fails so every connection is dropped with no useful error.

Error in logs:
```
WARN IpAuthHandler - Blocked connection from 172.18.0.4
```
Workaround: assign static IPs to PD containers via docker-compose ipam 
and use those IPs directly in HG_PD_RAFT_PEERS_LIST instead of hostnames.

bitflicker64:
## Purpose of the PR

- fix #2960

`IpAuthHandler` only compared the client IP with the allowlist entries directly.  
When the allowlist contains hostnames, connections from their resolved IPs could be rejected.

## Main Changes

- Resolve hostname allowlist entries using `InetAddress.getAllByName`
- Store resolved addresses in a set used for connection validation
- Keep runtime validation as a simple `Set.contains()` lookup

## Verifying these changes

- This issue was identified while testing changes in PR #2952.  

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects
- [x] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`

### DIFF
--- a/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
+++ b/hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
@@ -324,9 +324,30 @@ public class RaftEngine {
             this.raftNode.changePeers(newPeers, status -> {
                 // Use compareAndSet so a late callback does not overwrite a timeout status
                 result.compareAndSet(null, status);
+                // Refresh inside callback so it fires even if caller already timed out
+                // Note: changePeerList() uses Configuration.parse() which only supports
+                // plain comma-separated peer addresses with no learner syntax.
+                // getLearners() will always be empty here. Learner support is handled
+                // in PDService.updatePdRaft() which uses PeerUtil.parseConfig()
+                // and supports the /learner suffix.
+                if (status != null && status.isOk()) {
+                    IpAuthHandler handler = IpAuthHandler.getInstance();
+                    if (handler != null) {
+                        Set<String> newIps = newPeers.getPeers()
+                                                     .stream()
+                                                     .map(PeerId::getIp)
+                                                     .collect(Collectors.toSet());
+                        handler.refresh(newIps);
+                        log.info("IpAuthHandler refreshed after peer list change to: {}",
+                                 peerList);
+                    } else {
+                        log.warn("IpAuthHandler not initialized, skipping refresh for "
+                                 + "peer list: {}", peerList);
+                    }
+                }
                 latch.countDown();
             });
-            // Use configured RPC timeout — bare await() would block forever if
+            // Use 3x configured RPC timeout — bare await() would block forever if
             // the callback is never invoked (e.g. node not started / RPC failure)
             boolean completed = latch.await(3L * config.getRpcTimeout(),
                                             TimeUnit.MILLISECONDS);
@@ -350,22 +371,6 @@ public class RaftEngine {
         } catch (Exception e) {
             log.error("failed to changePeerList to {},{}", peerList, e);
             result.set(new Status(-1, e.getMessage()));
-        }
-
-        // Refresh IpAuthHandler so newly added peers are not blocked
-        if (result.get() != null && result.get().isOk()) {
-            IpAuthHandler handler = IpAuthHandler.getInstance();
-            if (handler != null) {
-                Set<String> newIps = newPeers.getPeers()
-                                             .stream()
-                                             .map(PeerId::getIp)
-                                             .collect(Collectors.toSet());
-                handler.refresh(newIps);
-                log.info("IpAuthHandler refreshed after peer list change to: {}", peerList);
-            } else {
-                log.warn("IpAuthHandler not initialized, skipping refresh for peer list: {}",
-                         peerList);
-            }
         }
 
         return result.get();
--- a/hugegraph-pd/hg-pd-test/src/main/java/org/apache/hugegraph/pd/raft/IpAuthHandlerTest.java
+++ b/hugegraph-pd/hg-pd-test/src/main/java/org/apache/hugegraph/pd/raft/IpAuthHandlerTest.java
@@ -76,8 +76,9 @@ public class IpAuthHandlerTest {
     @Test
     public void testUnresolvableHostnameDoesNotCrash() {
         // Should log a warning and skip — no exception thrown during construction
+        // Uses .invalid TLD which is RFC-2606 reserved and guaranteed to never resolve
         IpAuthHandler handler = IpAuthHandler.getInstance(
-                Collections.singleton("nonexistent.invalid.hostname"));
+                Collections.singleton("nonexistent.invalid"));
         // Handler was still created successfully despite bad hostname
         Assert.assertNotNull(handler);
         // Unresolvable entry is skipped so no IPs should be allowed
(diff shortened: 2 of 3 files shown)


hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/auth/IpAuthHandler.java
```diff
@@ -65,7 +71,24 @@ private static String getClientIp(ChannelHandlerContext ctx) {
     }
 
     private boolean isIpAllowed(String ip) {
-        return allowedIps.isEmpty() || allowedIps.contains(ip);
+        Set<String> resolved = this.resolvedIps;
+        return resolved.isEmpty() || resolved.contains(ip);
+    }
+
+    private static Set<String> resolveAll(Set<String> entries) {
```
imbajin, line 95: ⚠️ This change introduces new hostname-resolution behavior, but the PR still has no regression coverage around it. Please add at least one test for a hostname allowlist entry and one for a resolution failure/refresh scenario, otherwise it will be very easy to regress this path again.

hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/auth/IpAuthHandler.java
```diff
@@ -30,11 +33,14 @@
...
 public class IpAuthHandler extends ChannelDuplexHandler {
 
+    // Retained for potential refresh of resolvedIps on membership changes
     private final Set<String> allowedIps;
+    private volatile Set<String> resolvedIps;
     private static volatile IpAuthHandler instance;
 
     private IpAuthHandler(Set<String> allowedIps) {
         this.allowedIps = Collections.unmodifiableSet(allowedIps);
+        this.resolvedIps = resolveAll(allowedIps);
```
imbajin, line 40: ‼️ `resolvedIps` is only built once in the constructor, but `IpAuthHandler` is still a JVM-wide singleton. `RaftEngine#changePeerList()` can replace the peer set after startup, and hostname-based peers can also resolve to a different IP later. In both cases this handler keeps the original resolved addresses forever, so a valid PD peer can be blocked until the whole process restarts. Please add a refresh path for membership/DNS changes, or defer hostname resolution to validation time with a bounded cache/TTL.

```text
peer hostname
    |
    v
startup resolve once
    |
    v
resolvedIps (singleton, frozen)
    |
    +--> peer list changed
    |
    +--> DNS changed
            |
            v
      new peer IP not in set
            |
            v
        connection blocked

```

hugegraph-pd/hg-pd-test/src/main/java/org/apache/hugegraph/pd/raft/IpAuthHandlerTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.apache.hugegraph.pd.raft.auth.IpAuthHandler;
+import org.apache.hugegraph.testutil.Whitebox;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class IpAuthHandlerTest {
+
+    @After
```
imbajin, line 43: ⚠️ These tests are order-dependent because the singleton is only cleared in `@After`. Earlier classes in `PDCoreSuiteTest` initialize `RaftEngine`, which creates `IpAuthHandler` before this class runs; after that, `IpAuthHandler.getInstance(Set)` here just returns the pre-existing singleton and ignores the allowlist passed by the test. That makes the assertions depend on suite execution order instead of the test input. Please reset the singleton in `@Before` as well, or isolate this class from the suite so each test starts from a clean handler instance.

```text
PDCoreSuiteTest
    |
    +-- earlier test -> RaftEngine.init()
    |                   |
    |                   +-- IpAuthHandler singleton created
    |
    +-- IpAuthHandlerTest
            |
            +-- getInstance(new allowlist)
                    |
                    +-- returns old singleton
                            |
                            v
                      test input ignored
```

hugegraph-pd/hg-pd-core/src/main/java/org/apache/hugegraph/pd/raft/RaftEngine.java
```diff
@@ -326,6 +325,23 @@ public Status changePeerList(String peerList) {
                 latch.countDown();
             });
             latch.await();
+
+            // Refresh IpAuthHandler so newly added peers are not blocked
```
imbajin: ‼️ This refresh closes the `RaftEngine.changePeerList()` path, but peer membership can still be updated through `PDService.updatePdRaft()`, which calls `node.changePeers()` directly and never reaches this block. In that case the Raft config changes while `IpAuthHandler` keeps the old allowlist, so newly added hostname-based peers can still be rejected. Please centralize peer changes behind one helper that also refreshes the allowlist, or make the other update path invoke the same refresh logic on success.

```text
peer update
    |
    +-- changePeerList() --------> refresh allowlist   ✅
    |
    +-- updatePdRaft() ----------> changePeers only    ❌
                                       |
                                       v
                              old allowlist remains
                                       |
                                       v
                              new peer can be blocked
```

hugegraph-pd/hg-pd-service/src/main/java/org/apache/hugegraph/pd/service/PDService.java
```diff
@@ -1735,6 +1737,17 @@ public void updatePdRaft(Pdpb.UpdatePdRaftRequest request,
             node.changePeers(config, status -> {
                 if (status.isOk()) {
                     log.info("updatePdRaft, change peers success");
+                    // Refresh IpAuthHandler so newly added peers are not blocked
+                    IpAuthHandler handler = IpAuthHandler.getInstance();
+                    if (handler != null) {
```
imbajin, line 1742: ⚠️ The new tests cover `IpAuthHandler.refresh()` in isolation, but they do not exercise the actual regression path here: `changePeers(...) -> callback -> handler.refresh(...)`. Could we add one regression test through the peer-list update entrypoint as well? That would lock in the real behavior this PR is fixing and prevent a future refactor from accidentally dropping the callback refresh while the current tests still stay green.

```text
Current coverage
+--------------------+      +-------------------+
| IpAuthHandlerTest  | ---> | handler.refresh() |
+--------------------+      +-------------------+

Missing regression path
+------------------+      +----------------+      +-------------------+
| changePeerList() | ---> | changePeers()  | ---> | handler.refresh() |
+------------------+      +----------------+      +-------------------+
```

A small integration-style test around `RaftEngine.changePeerList()` or `PDService.updatePdRaft()` would probably be enough.

bitflicker64: Thanks for pointing that out. I’ve added a regression test (RaftEngineIpAuthIntegrationTest) that exercises the full path changePeerList → changePeers → callback → handler.refresh(). It also includes a failure-path test to ensure the allowlist is not refreshed when the peer update fails. Please let me know if you think any additional coverage would be helpful.

PR #2966, opened 2026-03-10
refactor(server): disable GraphSpaceAPI and ManagerAPI in standalone mode

### LINKED ISSUE #2964: [Bug] protect GraphSpaceAPI in non-hstore status
### Bug Type (问题类型)

rest-api (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0/master (Apache Release Version)

Currently, in non-hstore mode, the management functionality of GraphSpace is not supported.

In standalone mode, calls to GraphSpaceApi should be disabled, and a friendly message should be returned.

You can launch the standalone version and the distributed version, and request the graphspaceApi to verify this issue.

contrueCT:
## Purpose of the PR

- close #2964 

## Main Changes

The issue requires GraphSpaceAPI to return a friendly response in standalone mode. During code investigation, I found that ManagerAPI has a similar problem: in standalone mode, manager.graphSpace(name) returns a fake new GraphSpace("DEFAULT"), which causes ManagerAPI.validGraphSpace() to pass incorrectly. As a result, ManagerAPI can still enter logic that is not actually supported in standalone mode, leading to unfriendly errors or behavior that does not match the API semantics. For consistency and correctness, I applied the same standalone-mode guard to ManagerAPI as well.

- Add public isUsePD() accessor to GraphManager to expose PD status
- Add checkPdModeEnabled() helper in API base class
- Call checkPdModeEnabled() in all public methods of GraphSpaceAPI (list/get/create/manage/delete)
- Call checkPdModeEnabled() in all public methods of ManagerAPI (createManager/delete/list/checkRole/getRolesInGs)
- Returns HTTP 400 with message 'GraphSpace management is not supported in standalone mode'
- Add standalone-mode rejection tests in GraphSpaceApiTest and ManagerApiTest

## Verifying these changes

    - xxx

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/BaseApiTest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/BaseApiTest.java
@@ -74,6 +74,8 @@ public class BaseApiTest {
     private static final String GRAPH_VERTEX = "/graph/vertices";
     private static final String GRAPH_EDGE = "/graph/edges";
     private static final String BATCH = "/batch";
+    static final String STANDALONE_ERROR =
+            "GraphSpace management is not supported in standalone mode";
 
     private static final String ROCKSDB_CONFIG_TEMPLATE =
             "{ \"gremlin.graph\": \"org.apache.hugegraph.HugeFactory\"," +
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/ManagerApiStandaloneTest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/ManagerApiStandaloneTest.java
@@ -35,9 +35,6 @@ import jakarta.ws.rs.core.Response;
  */
 public class ManagerApiStandaloneTest extends BaseApiTest {
 
-    private static final String STANDALONE_ERROR =
-            "GraphSpace management is not supported in standalone mode";
-
     private static String managerPath(String graphSpace) {
         return String.format("graphspaces/%s/auth/managers", graphSpace);
     }


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -76,6 +76,7 @@ public class GraphSpaceAPI extends API {
     @Produces(APPLICATION_JSON_WITH_CHARSET)
     public Object list(@Context GraphManager manager,
                        @Context SecurityContext sc) {
+        checkPdModeEnabled(manager);
```
imbajin: ‼️ `checkPdModeEnabled()` is added to the other GraphSpace endpoints, but `/graphspaces/profile` is still unguarded. In standalone mode this method will continue to return a different result instead of the new 400, so the API contract remains inconsistent. Please add the same guard to `listProfile()` and extend the standalone coverage to hit `/graphspaces/profile` as well.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/ManagerApiTest.java
```diff
@@ -981,4 +981,45 @@ public void testUserWithDifferentRolesInMultipleSpaces() {
         Assert.assertTrue("Only graphb1 should remain in spaceb",
                           graphsList.contains("graphb1") && !graphsList.contains("graphb2"));
     }
+
+    @Test
+    public void testStandaloneModeForbidsAllEndpoints() {
+        Assume.assumeFalse("skip this test for hstore (PD mode)",
```
imbajin: ⚠️ This standalone test never actually runs. `ManagerApiTest#setUpClass()` has a class-level `Assume.assumeTrue(... hstore ...)`, so JUnit skips the whole class before the method-level `assumeFalse()` is reached here. That means this PR adds no regression coverage for the new standalone behavior. Please move these assertions into a test class that is allowed to run without PD, or relax the class-level guard so this method can execute.

contrueCT: OK, I will create a new test class for standalone test.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/GraphSpaceApiStandaloneTest.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final String PATH = "graphspaces";
+    private static final String STANDALONE_ERROR =
+            "GraphSpace management is not supported in standalone mode";
+
+    @Before
+    public void skipForPdMode() {
+        assumeStandaloneMode();
+    }
+
+    @Test
```
imbajin, line 51: ‼️ The standalone suite still doesn't cover `/graphspaces/profile`, and `GraphSpaceAPI#listProfile()` remains unguarded on the current head. That means this PR can still pass while one GraphSpace endpoint returns a different standalone behavior than the others. Please add a profile assertion here and guard `listProfile()` the same way as the rest of the GraphSpace endpoints.
```suggestion
    @Test
    public void testProfileReturnsFriendlyError() {
        Response r = this.client().get(PATH + "/profile");
        String content = assertResponseStatus(400, r);
        Assert.assertTrue(content.contains(STANDALONE_ERROR));
    }

    @Test
```

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/BaseApiTest.java
```diff
@@ -193,7 +183,8 @@ protected static void waitTaskStatus(int task, Set<String> expectedStatus) {
                 Assert.fail(String.format("Failed to wait for task %s " +
                                           "due to timeout", task));
             }
-        } while (!expectedStatus.contains(status));
+        }
+        while (!expectedStatus.contains(status));
```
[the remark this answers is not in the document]
contrueCT: This is not a part of this pr.I am not sure its logic.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/ManagerApiStandaloneTest.java
```diff
@@ -0,0 +1,90 @@
...
+    public void testDeleteManagerReturnsFriendlyError() {
+        Response r = this.client().delete(managerPath("DEFAULT"),
+                                          Map.of("user", "admin",
+                                                 "type", HugePermission.ADMIN));
+        String content = assertResponseStatus(400, r);
+        Assert.assertTrue(content.contains(STANDALONE_ERROR));
+    }
+
+    @Test
+    public void testListManagerReturnsFriendlyError() {
```
imbajin, line 65: ⚠️ The new standalone tests only exercise the `ADMIN` branch, so they don't actually lock down the regression described in the PR body. The problematic path was that `validGraphSpace()` could still succeed in standalone mode because `manager.graphSpace(name)` returns a synthetic `DEFAULT` space, and that only happens in the `SPACE` / `SPACE_MEMBER` branches.

ASCII view:
```text
checkPdModeEnabled()
        |
        +-- ADMIN         <- covered here
        +-- SPACE         <- regression path
        \-- SPACE_MEMBER <- regression path
```

Please add at least one standalone test with `type=SPACE` or `type=SPACE_MEMBER` against a non-`DEFAULT` graphspace, so this PR verifies the branch that used to fall through.

```suggestion
    @Test
    public void testListSpaceManagerReturnsFriendlyError() {
        Response r = this.client().get(managerPath("nonexistent"),
                                       Map.of("type", (Object) HugePermission.SPACE));
        String content = assertResponseStatus(400, r);
        Assert.assertTrue(content.contains(STANDALONE_ERROR));
    }
```

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/BaseApiTest.java
```diff
@@ -748,6 +739,28 @@ public static RestClient analystClient(String graphSpace, String username) {
...
+     * Skips the current test if the server is running in hstore / PD mode.
+     * Treats both {@code "hstore"} and {@code null} (i.e. the property is not
+     * set, which is the default in hstore CI runs) as PD mode.
+     * Call this from a {@code @Before} method in standalone-only test classes.
+     */
+    public static void assumeStandaloneMode() {
+        String backend = System.getProperty("backend");
+        Assume.assumeTrue(
+                "skip standalone tests: backend is '" + backend + "' (hstore/PD mode)",
+                backend != null && !backend.equals("hstore"));
```
imbajin: ⚠️ `assumeStandaloneMode()` uses a double-negative condition (`backend != null && !backend.equals("hstore")`), which reduces readability. Prefer an explicit boolean variable and `Assume.assumeFalse(...)` for clarity and null-safety:

```java
boolean isPdMode = "hstore".equals(backend);
Assume.assumeFalse("Skip in PD/distributed mode", isPdMode);
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/API.java
```diff
@@ -241,6 +241,13 @@ public static boolean checkAndParseAction(String action) {
         }
     }
 
+    public static void checkPdModeEnabled(GraphManager manager) {
```
imbajin: ⚠️ Method name `checkPdModeEnabled()` is non-idiomatic for a method that throws — Java convention uses imperative names like `ensure*` or `require*` for assertion/guard methods. Consider renaming:

```java
public static void ensurePdModeEnabled(GraphManager manager) {
```

Also missing: a null-check for `manager` and a Javadoc comment explaining what exception is thrown and when.

imbajin: ⚠️ Declaring this as `public static` on the widely-used `API` base class unnecessarily expands the public API surface. Since this guard is only called from API subclasses, consider reducing visibility to `protected`:

```java
protected static void checkPdModeEnabled(GraphManager manager) {
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/API.java
```diff
@@ -241,6 +241,13 @@ public static boolean checkAndParseAction(String action) {
...
+        if (!manager.usePD()) {
+            throw new HugeException(
+                    "GraphSpace management is not supported in standalone mode");
```
imbajin: 🧹 The error message `"GraphSpace management is not supported in standalone mode"` is duplicated across production code and tests. Extract it to a constant to avoid typos and make future changes easier:

```java
private static final String PD_MODE_REQUIRED_MSG =
        "GraphSpace management is not supported in standalone mode";
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -276,7 +276,7 @@ private static String serviceId(String graphSpace, Service.ServiceType type,
                      .replace("_", "-").toLowerCase();
     }
 
-    private boolean usePD() {
+    public boolean usePD() {
```
imbajin: ⚠️ `usePD()` is an ambiguous name for a public boolean accessor — Java convention for boolean getters is `is*` or `has*`. Consider renaming to `isPdEnabled()` or `isUsePD()` to make the intent clear:

```java
public boolean isPdEnabled() {
    return this.PDExist;
}
```

imbajin (approved):

PR #2980, opened 2026-03-28
fix(docker): enable docker logs for pd/store/server containers

### LINKED ISSUE #2979: fix(docker): enable docker logs for pd/store/server containers
## Environment
- Server Version: 1.7.0
- Backend: HStore
- Deployment: docker-compose

## Problem

`docker logs hg-pd / hg-store / hg-server` only shows a few entrypoint startup lines. All actual JVM logs go silently to files inside the container, so debugging requires manually execing in and tailing log files — not the expected Docker workflow.

## Root Cause

Two issues working together:

1. Startup scripts redirect JVM output to a file via `>> ${OUTPUT} 2>&1` before Docker can capture it
2. The `console` appender is defined in all dist `log4j2.xml` configs but never wired to the root logger

## Fix

A PR with the fix is being opened.

bitflicker64:
## Purpose of the PR
- close #2979

`docker logs` shows no application output for any HugeGraph container. All JVM logs are silently redirected to files inside the container, requiring manual exec + tail to debug. This PR fixes that.

related comment - https://github.com/apache/hugegraph/pull/2952#issuecomment-4029115619

## Main Changes

Two root causes fixed:

1. Startup scripts redirected JVM stdout/stderr to a file via `>> ${OUTPUT} 2>&1` before Docker could capture it — now gated behind a `STDOUT_MODE=true` env var so bare metal installs are unaffected
2. The console appender was defined in all dist `log4j2.xml` configs but never wired to the root logger and `org.apache.hugegraph` AsyncLogger — added `<appender-ref ref="console"/>` to both in all 3 configs

Docker compose files set `STDOUT_MODE=true` so containers stream logs via `docker logs`. File appenders are kept so non-Docker deployments are unaffected.

Also sets STDOUT_MODE=true as a default in Dockerfiles to ensure docker logs works for direct docker run usage, not just docker compose.

## Verifying these changes

- [ ]  Dependencies
- [x]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status
- [ ]  `Doc - TODO`
- [ ]  `Doc - Done`
- [x]  `Doc - No Need`

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -41,7 +41,6 @@ services:
       HG_PD_RAFT_PEERS_LIST: pd:8610
       HG_PD_INITIAL_STORE_LIST: store:8500
       HG_PD_DATA_PATH: /hugegraph-pd/pd_data
-      STDOUT_MODE: "true"
     ports:
       - "8620:8620"
     volumes:
@@ -70,7 +69,6 @@ services:
       HG_STORE_REST_PORT: "8520"
       HG_STORE_RAFT_ADDRESS: store:8510
       HG_STORE_DATA_PATH: /hugegraph-store/storage
-      STDOUT_MODE: "true"
     ports:
       - "8520:8520"
     volumes:
@@ -95,7 +93,6 @@ services:
     environment:
       HG_SERVER_BACKEND: hstore
       HG_SERVER_PD_PEERS: pd:8686
-      STDOUT_MODE: "true"
     ports:
       - "8080:8080"
     healthcheck:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -41,7 +41,7 @@ services:
       HG_PD_RAFT_PEERS_LIST: pd:8610
       HG_PD_INITIAL_STORE_LIST: store:8500
       HG_PD_DATA_PATH: /hugegraph-pd/pd_data
-      STDOUT_MODE: true
+      STDOUT_MODE: "true"
     ports:
       - "8620:8620"
     volumes:
@@ -70,7 +70,7 @@ services:
       HG_STORE_REST_PORT: "8520"
       HG_STORE_RAFT_ADDRESS: store:8510
       HG_STORE_DATA_PATH: /hugegraph-store/storage
-      STDOUT_MODE: true
+      STDOUT_MODE: "true"
     ports:
       - "8520:8520"
     volumes:
@@ -95,7 +95,7 @@ services:
     environment:
       HG_SERVER_BACKEND: hstore
       HG_SERVER_PD_PEERS: pd:8686
-      STDOUT_MODE: true
+      STDOUT_MODE: "true"
     ports:
       - "8080:8080"
     healthcheck:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -41,6 +41,7 @@ services:
       HG_PD_RAFT_PEERS_LIST: pd:8610
       HG_PD_INITIAL_STORE_LIST: store:8500
       HG_PD_DATA_PATH: /hugegraph-pd/pd_data
+      STDOUT_MODE: true
     ports:
       - "8620:8620"
     volumes:
@@ -69,6 +70,7 @@ services:
       HG_STORE_REST_PORT: "8520"
       HG_STORE_RAFT_ADDRESS: store:8510
       HG_STORE_DATA_PATH: /hugegraph-store/storage
+      STDOUT_MODE: true
     ports:
       - "8520:8520"
     volumes:
@@ -93,6 +95,7 @@ services:
     environment:
       HG_SERVER_BACKEND: hstore
       HG_SERVER_PD_PEERS: pd:8686
+      STDOUT_MODE: true
     ports:
       - "8080:8080"
     healthcheck:
--- a/hugegraph-pd/hg-pd-dist/src/assembly/static/conf/log4j2.xml
+++ b/hugegraph-pd/hg-pd-dist/src/assembly/static/conf/log4j2.xml
@@ -115,6 +115,7 @@
 
     <loggers>
         <root level="INFO">
+            <appender-ref ref="console" />
             <appender-ref ref="file" />
         </root>
         <logger name="com.alipay.sofa" level="INFO" additivity="false">
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -184,5 +184,4 @@ fi
 
 # Turn on security check
 exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \
-    org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF} \
-    >> ${OUTPUT} 2>&1
+    org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF}
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -183,5 +183,11 @@ if [ "${OPEN_TELEMETRY}" == "true" ]; then
 fi
 
 # Turn on security check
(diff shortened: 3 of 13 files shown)


bitflicker64:
@imbajin I've wired the console appender into `org.apache.hugegraph` across all three components. Should I also add it to the third-party loggers or is just the HugeGraph application logger sufficient for the Docker logging goal?

vaijosh (approved):
LGTM

hugegraph-server/hugegraph-dist/src/assembly/static/conf/log4j2.xml
```diff
@@ -104,6 +104,7 @@
 
     <loggers>
         <root level="INFO">
+            <appender-ref ref="console"/>
             <appender-ref ref="file"/>
         </root>
```
imbajin, line 109: ‼️ Adding `console` only to `root` and `org.apache.hugegraph` is not enough here: the `additivity="false"` logger blocks below still bind only `file`, so startup/runtime messages from `org.apache.cassandra`, `io.netty`, `com.datastax.driver`, `com.alipay.sofa`, etc. will still be absent from `docker logs`. Please attach `console` to the explicit logger blocks that should be visible in containers.

hugegraph-store/hg-store-dist/src/assembly/static/conf/log4j2.xml
```diff
@@ -117,6 +117,7 @@
 
     <loggers>
         <root level="INFO">
+            <appender-ref ref="console"/>
```
imbajin, line 120: ‼️ Same issue in the store config: the root console appender does not cover the `additivity="false"` logger blocks below, so `io.netty` / `org.apache.commons` / `com.alipay.sofa` output is still invisible in `docker logs`. If the goal is a complete container stream, those explicit loggers need `console` as well.

hugegraph-pd/hg-pd-dist/src/assembly/static/conf/log4j2.xml
```diff
@@ -115,6 +115,7 @@
 
     <loggers>
         <root level="INFO">
+            <appender-ref ref="console" />
             <appender-ref ref="file" />
         </root>
```
imbajin, line 120: ‼️ The same completeness gap exists here too: `root` now has `console`, but `com.alipay.sofa`, `io.netty`, and `org.apache.commons` still short-circuit to `file` only because their logger blocks keep `additivity="false"`. Container logs will still miss transport/startup output unless those explicit loggers also opt into `console`.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
```diff
@@ -184,5 +184,4 @@ fi
 
 # Turn on security check
 exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \
```
imbajin: ‼️ Removing the launcher redirection here changes the default binary-package flow too. In a plain install, JVM bootstrap stdout/stderr will no longer land in the `logs/*.log` file, so the docs that tell users to inspect that file on failure become incomplete. Please gate this on the Docker entrypoint path or preserve the file sink outside containers.

imbajin:
> @imbajin I've wired the console appender into `org.apache.hugegraph` across all three components. Should I also add it to the third-party loggers or is just the HugeGraph application logger sufficient for the Docker logging goal?

Seems not necessary for now

hugegraph-server/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -99,17 +99,14 @@ if [[ $PRELOAD == "true" ]]; then
     sed -i -e '/registerBackends/d; /serverStarted/d' "${SCRIPTS}/${EXAMPLE_SCRIPT}"
 fi
 
-# TODO: show the output message in hugegraph-server.sh when start the server
 if [[ $DAEMON == "true" ]]; then
     echo "Starting HugeGraphServer in daemon mode..."
```
imbajin, line 103: The safer, lower-risk approach is to add an explicit switch for Docker/stdout mode instead of removing `>> ${OUTPUT} 2>&1` from the default startup path. 

That keeps the binary-package flow unchanged and still lets containers stream clean logs via `docker logs` when the switch is enabled.

```text
Default package: start-*.sh -> exec java -> >> logs/*.log 2>&1

Docker package:   entrypoint -> set STDOUT_MODE=true -> exec java -> stdout + console appender
```
I would keep the file-based behavior by default, then enable stdout only in the Docker entrypoint and wire the console appender in the matching log config.

bitflicker64:
@imbajin On it fixing both the additivity loggers and the shell redirect gate. Will push shortly.

docker/docker-compose.yml
```diff
@@ -41,6 +41,7 @@ services:
       HG_PD_RAFT_PEERS_LIST: pd:8610
       HG_PD_INITIAL_STORE_LIST: store:8500
       HG_PD_DATA_PATH: /hugegraph-pd/pd_data
+      STDOUT_MODE: true
```
imbajin: 🧹 Please quote this value for consistency with the other compose file and with the shell check `[[ "${STDOUT_MODE:-false}" == "true" ]]`, so the env value is always an explicit string.

```suggestion
      STDOUT_MODE: "true"
```

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
```diff
@@ -183,6 +183,11 @@ if [ "${OPEN_TELEMETRY}" == "true" ]; then
...
-exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \
-    org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF} \
-    >> ${OUTPUT} 2>&1
+if [[ "${STDOUT_MODE:-false}" == "true" ]]; then
+    exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \
+        org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF}
+else
+    exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \
+        org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF} \
+        >> ${OUTPUT} 2>&1
```
imbajin: ⚠️ In the default non-Docker path (`STDOUT_MODE=false`), this still redirects JVM stdout/stderr to the same file used by the Log4j file appender (`logs/hugegraph-server.log`). With console now enabled, this can duplicate entries in that file and accelerate rollover.

To keep the default package flow stable while avoiding duplicate lines, please redirect this shell-level stream to a dedicated stdout log file:

```suggestion
        >> ${LOGS}/hugegraph-server-stdout.log 2>&1
```

bitflicker64:
STDOUT_MODE is currently only set via compose, so direct docker run won’t expose logs via docker logs

docker/docker-compose.yml
```diff
@@ -41,6 +41,7 @@ services:
       HG_PD_RAFT_PEERS_LIST: pd:8610
       HG_PD_INITIAL_STORE_LIST: store:8500
       HG_PD_DATA_PATH: /hugegraph-pd/pd_data
+      STDOUT_MODE: "true"
```
imbajin: ❗️This param should be removed. (same as others)

`STDOUT_MODE="true"` is already set in all three Dockerfiles, so keeping it here creates a second source of truth in compose and increases config drift risk.

Please delete this line directly in compose files (instead of keeping a commented local override).

If you want compose-level control in the future, please centralize it with a shared anchor and merge it into each service environment, for example: (Feel free to include commented-out code as a sample, if u feel it adds value for the user)

```yaml
x-stdout-env: &stdout-env
  STDOUT_MODE: "true"

services:
  pd:
    environment:
      <<: *stdout-env
      HG_PD_GRPC_HOST: pd
```

Please apply the same cleanup consistently in `docker/docker-compose.yml`, `docker/docker-compose.dev.yml`, and `docker/docker-compose-3pd-3store-3server.yml`.

bitflicker64:
Done, removed STDOUT_MODE from all three compose files. Kept the anchor comment out since the default is already in the Dockerfiles anyway.

imbajin (approved):

PR #2987, opened 2026-04-10
refactor(server): allow TinkerPop exceptions in Gremlin resp

### LINKED ISSUE #2986: [Bug] Request Gremlin exception: MissingPropertyException
### Bug Type (问题类型)

gremlin (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version: 1.0.0 (Apache Release Version)

Image

Image

Image

```javascript

```

```javascript

```

utafrali:
## Purpose of the PR
- Fixes #2986

## Main Changes
The Gremlin query API was filtering out TinkerPop exceptions like MissingPropertyException instead of passing them back to clients. Added `org.apache.tinkerpop.` to the exception allowlist so these errors bubble up properly. Also added a test to verify the behavior.

## Verifying these changes
- [ ] Trivial rework / code cleanup without any test coverage.
- [ ] Already covered by existing tests.

Added unit test in `GremlinQueryAPITest.java` covering exception handling for TinkerPop packages.

- [ ] Dependencies
- [ ] Modify configurations
- [x] The public API
- [ ] Other affects
- [ ] Nope

## Documentation Status
- [ ] Doc - TODO
- [ ] Doc - Done
- [x] Doc - No Need

### DIFF
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/gremlin/GremlinQueryAPI.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/gremlin/GremlinQueryAPI.java
@@ -43,7 +43,8 @@ public class GremlinQueryAPI extends API {
             "java.util.concurrent.TimeoutException",
             "groovy.lang.",
             "org.codehaus.",
-            "org.apache.hugegraph."
+            "org.apache.hugegraph.",
+            "org.apache.tinkerpop."
     );
 
     @Context
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/gremlin/GremlinQueryAPI.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/gremlin/GremlinQueryAPI.java
@@ -44,7 +44,7 @@ public class GremlinQueryAPI extends API {
             "groovy.lang.",
             "org.codehaus.",
             "org.apache.hugegraph.",
-            "org.apache.tinkerpop."
+            "org.apache.tinkerpop.gremlin.process.traversal.util.FastNoSuchElementException"
     );
 
     @Context
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/api/gremlin/GremlinQueryAPITest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/api/gremlin/GremlinQueryAPITest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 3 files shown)


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/gremlin/GremlinQueryAPI.java
```diff
@@ -43,7 +43,8 @@ public class GremlinQueryAPI extends API {
             "java.util.concurrent.TimeoutException",
             "groovy.lang.",
             "org.codehaus.",
-            "org.apache.hugegraph."
+            "org.apache.hugegraph.",
+            "org.apache.tinkerpop."
```
imbajin: ‼️ This package-wide allowlist is too broad. `matchBadRequestException()` would now classify classes such as `org.apache.tinkerpop.gremlin.server.auth.AuthenticationException` and `org.apache.tinkerpop.gremlin.server.authz.AuthorizationException` as 400, even though they carry 401/403 semantics. 

That changes the API contract for auth failures and hides security/server-side errors behind a generic bad request. Please whitelist only the concrete exception class(es) reproduced by #2986 instead of the whole `org.apache.tinkerpop.` namespace.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/api/gremlin/GremlinQueryAPITest.java
```diff
@@ -0,0 +1,57 @@
...
+        Method m = GremlinQueryAPI.class.getDeclaredMethod(
+                "matchBadRequestException", String.class);
+        m.setAccessible(true);
+        return (boolean) m.invoke(null, exClass);
+    }
+
+    @Test
+    public void testMatchBadRequestExceptionWithTinkerpop() throws Exception {
+        Assert.assertTrue(matchBadRequest(
+                "org.apache.tinkerpop.gremlin.structure.util.empty.EmptyProperty"));
```
imbajin: ⚠️ `EmptyProperty` is not an exception type, so this test does not pin the reported bug. In TinkerPop 3.5.1, `EmptyProperty.value()` throws `Property.Exceptions.propertyDoesNotExist()` (`IllegalStateException`), which means this assertion can pass even if the real `Exception-Class` from Gremlin is still misclassified. 

Please assert against the actual class observed in #2986, and add a **negative** case for auth/authorization exceptions so the allowlist boundary is covered.

utafrali:
Done, pushed the fix.

imbajin (approved):
THX, better to test the real query path next time(instead of only using reflection way)

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/api/gremlin/GremlinQueryAPITest.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package org.apache.hugegraph.unit.api.gremlin;
+
+import java.lang.reflect.Method;
+
+import org.apache.hugegraph.api.gremlin.GremlinQueryAPI;
+import org.apache.hugegraph.testutil.Assert;
+import org.apache.hugegraph.unit.BaseUnitTest;
+import org.junit.Test;
```
imbajin, line 25: also need format the import:)  

refer https://github.com/apache/hugegraph/wiki/The-style-config-for-HugeGraph-in-IDEA#0x01-general-configuration

PR #2991, opened 2026-04-12
fix(server): normalize bool range predicates in gremlin filters

### LINKED ISSUE #2934: [Bug] Unexpected NumberFormatException for filter-step
### Bug Type (问题类型)

gremlin (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0
- Backend: RocksDB x nodes

### Expected behavior
```gremlin
g.E().where(__.has('ep5', lt(true)))

g.E().match(__.as('start1').where(__.has('ep5', lt(true))).as('m1')).select('m1')
```
### Actual behavior
+ The first query thrown a exception: ```java.lang.NumberFormatException: Character t is neither a decimal digit number, decimal point, nor "e" notation exponential mark.```
### Additional Context
+ We moved the```where (...)``` step in first query into the ```match()```step, which should not affect the result.
+ When I executed these two queries on Janusgraph and Tinkerpop, they both returned the same results and were consistent with the results of the second query.
+ I have implemented a fuzzing tool and am using it to test HugeGraph. To reduce the burden on developers, I have simplified the test cases as much as possible and isolated irrelevant information.

```javascript
none
```

```javascript
none
```

contrueCT:
## Purpose of the PR

- fix #2934 

This PR fixes HugeGraph's inconsistent handling of boolean range predicates and unifies the internal ordering semantics as false < true. Previously, filter-step queries such as where(__.has("xxx", P.lt(true))) were pushed down through HugeGraph's condition/query pipeline and incorrectly treated as numeric-style range conditions, while equivalent match() queries could still return results normally. As a result, semantically equivalent traversals behaved differently.

## Main Changes

The fix has two parts. First, boolean support is added to backend condition comparison and range-flattening logic so gt/gte/lt/lte can be evaluated consistently. Second, runtime Gremlin boolean range predicates are normalized into equivalent eq/in/empty conditions before query planning, so they are no longer misclassified as ordinary numeric range queries that require a range index. This preserves the intended boolean ordering semantics while making has(), where(), and match() behave consistently for boolean range predicates.

## Verifying these changes

Regression coverage was also added for:

- low-level boolean lt/lte/gt/gte comparison semantics
- boolean range merge and conflicting-range flattening
- end-to-end consistency across has(), where(), and match()
- empty-result edge cases such as lt(false) and gt(true)

    - mvn -Drat.skip=true -P core-test,memory -pl hugegraph-server/hugegraph-test,hugegraph-struct -am -DfailIfNoTests=false '-Dtest=ConditionTest,ConditionQueryFlattenTest,EdgeCoreTest#testQueryEdgeByBooleanRangePredicate' clean test

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
@@ -233,16 +233,14 @@ public abstract class Condition {
         }
 
         private static int compareBoolean(Object first, Boolean second) {
-            if (first == null) {
-                first = false;
-            }
             if (first instanceof Boolean) {
                 return Boolean.compare((Boolean) first, second);
             }
 
             throw new IllegalArgumentException(String.format(
                     "Can't compare between %s(%s) and %s(%s)",
-                    first, first.getClass().getSimpleName(),
+                    first, first == null ? null :
+                           first.getClass().getSimpleName(),
                     second, second.getClass().getSimpleName()));
         }
 
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
@@ -195,7 +195,7 @@ public abstract class Condition {
          *
          * @param first  is actual value, might be Number/Date or String, It is
          *               probably that the `first` is serialized to String.
-         * @param second is value in query condition, must be Number/Date
+         * @param second is value in query condition, must be Number/Date/Boolean
          * @return the value 0 if first is numerically equal to second;
          * a value less than 0 if first is numerically less than
          * second; and a value greater than 0 if first is
@@ -208,6 +208,8 @@ public abstract class Condition {
                                                  (Number) second);
             } else if (second instanceof Date) {
                 return compareDate(first, (Date) second);
+            } else if (second instanceof Boolean) {
+                return compareBoolean(first, (Boolean) second);
             }
 
             throw new IllegalArgumentException(String.format(
@@ -222,6 +224,20 @@ public abstract class Condition {
             }
             if (first instanceof Date) {
                 return ((Date) first).compareTo(second);
+            }
+
+            throw new IllegalArgumentException(String.format(
+                    "Can't compare between %s(%s) and %s(%s)",
+                    first, first.getClass().getSimpleName(),
+                    second, second.getClass().getSimpleName()));
+        }
+
+        private static int compareBoolean(Object first, Boolean second) {
+            if (first == null) {
+                first = false;
+            }
+            if (first instanceof Boolean) {
+                return Boolean.compare((Boolean) first, second);
             }
 
             throw new IllegalArgumentException(String.format(
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
@@ -265,7 +265,7 @@ public final class ConditionQueryFlatten {
             cq.query(nonRelations);
             return ImmutableList.of(cq);
         }
-        return ImmutableList.of(query);
+        return ImmutableList.of();
     }
 
     private static Relations optimizeRelations(Relations relations) {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
@@ -427,9 +427,26 @@ public final class ConditionQueryFlatten {
         if (low == null || high == null) {
             return true;
         }
-        return compare(low, high) < 0 || compare(low, high) == 0 &&
-                                         low.relation() == Condition.RelationType.GTE &&
-                                         high.relation() == Condition.RelationType.LTE;
+        int compared = compare(low, high);
+        if (compared > 0) {
+            return false;
+        }
+        if (compared == 0) {
+            return low.relation() == Condition.RelationType.GTE &&
+                   high.relation() == Condition.RelationType.LTE;
+        }
+        return !emptyBooleanRange(low, high);
+    }
+
+    private static boolean emptyBooleanRange(Relation low, Relation high) {
+        if (!(low.value() instanceof Boolean) ||
+            !(high.value() instanceof Boolean)) {
+            return false;
+        }
+        return Boolean.FALSE.equals(low.value()) &&
+               Boolean.TRUE.equals(high.value()) &&
+               low.relation() == Condition.RelationType.GT &&
+               high.relation() == Condition.RelationType.LT;
     }
 
     private static boolean validEq(Relation eq, Relation low, Relation high) {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
@@ -507,6 +507,9 @@ public final class ConditionQueryFlatten {
             return NumericUtil.compareNumber(firstValue, (Number) secondValue);
         } else if (firstValue instanceof Date && secondValue instanceof Date) {
             return ((Date) firstValue).compareTo((Date) secondValue);
+        } else if (firstValue instanceof Boolean &&
+                   secondValue instanceof Boolean) {
+            return Boolean.compare((Boolean) firstValue, (Boolean) secondValue);
         } else {
             throw new IllegalArgumentException(String.format("Can't compare between %s and %s",
                                                              first, second));
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
@@ -265,7 +265,7 @@ public final class ConditionQueryFlatten {
             cq.query(nonRelations);
             return ImmutableList.of(cq);
         }
-        return ImmutableList.of();
+        return ImmutableList.of(query);
     }
 
     private static Relations optimizeRelations(Relations relations) {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
@@ -115,12 +115,23 @@ public final class ConditionQueryFlatten {
                 }
             case AND:
                 Condition.And and = (Condition.And) condition;
-                return new Condition.And(flattenIn(and.left(), supportIn),
-                                         flattenIn(and.right(), supportIn));
+                Condition andLeft = flattenIn(and.left(), supportIn);
+                Condition andRight = flattenIn(and.right(), supportIn);
+                if (andLeft == null || andRight == null) {
+                    return null;
+                }
+                return new Condition.And(andLeft, andRight);
             case OR:
                 Condition.Or or = (Condition.Or) condition;
-                return new Condition.Or(flattenIn(or.left(), supportIn),
-                                        flattenIn(or.right(), supportIn));
+                Condition orLeft = flattenIn(or.left(), supportIn);
+                Condition orRight = flattenIn(or.right(), supportIn);
+                if (orLeft == null) {
+                    return orRight;
+                }
+                if (orRight == null) {
+                    return orLeft;
+                }
+                return new Condition.Or(orLeft, orRight);
             default:
                 throw new AssertionError(String.format("Wrong condition type: '%s'",
                                                        condition.type()));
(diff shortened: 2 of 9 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQueryFlatten.java
```diff
@@ -265,7 +265,7 @@ private static List<ConditionQuery> flattenRelations(ConditionQuery query) {
             cq.query(nonRelations);
             return ImmutableList.of(cq);
         }
-        return ImmutableList.of(query);
+        return ImmutableList.of();
```
imbajin: ‼️ **Critical: `flattenRelations` returns empty list for ALL conflicting conditions, not just boolean**

This change from `ImmutableList.of(query)` to `ImmutableList.of()` affects **all** data types, not just booleans. Previously, when `optimizeRelations` returned `null` (meaning conditions were mutually exclusive, e.g., `age > 10 AND age == 9`), the original query was still returned — the contradiction was preserved and evaluated downstream. Now it returns an empty list, silently dropping the query.

While returning an empty list for contradictory conditions is arguably more correct (no results possible), this is a **behavioral change for numeric/date/string range contradictions** too, not just booleans. This should be:
1. Called out explicitly in the PR description since it changes existing behavior for all types
2. Covered with a test for numeric contradictory ranges to ensure no regression

Was this intentional or is the fix only meant for the boolean case?

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -488,6 +494,36 @@ private static Condition.Relation convCompare2UserpropRelation(HugeGraph graph,
...
+                       Condition.eq(key, false);
+            default:
+                throw new AssertionError(compare);
+        }
+    }
+
+    private static Condition impossibleBooleanCondition(Id key) {
+        return Condition.and(Condition.eq(key, false),
+                             Condition.eq(key, true));
+    }
```
imbajin: ⚠️ **`impossibleBooleanCondition` creates an AND node that may not be optimized away cleanly**

This creates `AND(eq(key, false), eq(key, true))` to represent an impossible condition. This is clever but has a potential issue: this compound `Condition.And` flows into the query pipeline where `ConditionQueryFlatten.flattenRelations` extracts only `Relation` conditions. An `And` node is NOT a `Relation`, so it goes into `nonRelations` and may bypass the merge/optimize logic entirely, potentially reaching the backend as a real filter condition.

Consider whether returning an empty `ConditionQuery` or a sentinel "always-false" condition would be cleaner. Alternatively, verify that the `And(eq(false), eq(true))` is properly handled in the flatten/optimize pipeline without passing contradictory conditions to the backend.

(If `Condition.in(key, ImmutableList.of())` — i.e., IN with empty list — is supported by the backend, that might be a cleaner way to express "matches nothing".)

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
```diff
@@ -230,6 +232,20 @@ private static int compareDate(Object first, Date second) {
                     second, second.getClass().getSimpleName()));
         }
 
+        private static int compareBoolean(Object first, Boolean second) {
+            if (first == null) {
+                first = false;
```
imbajin: ⚠️ **`compareBoolean`: null → false coercion is surprising**

When `first` is null, it's silently treated as `false`. This is asymmetric with `compareNumber` and `compareDate` which don't do null coercion. If a property is absent (null), treating it as `false` could cause incorrect query results — e.g., a vertex with no boolean property set would match `lte(false)`.

Consider whether null should throw or return a defined comparison result (e.g., null sorts before any non-null value), consistent with how other types handle it.

hugegraph-struct/src/main/java/org/apache/hugegraph/query/Condition.java
```diff
@@ -472,6 +474,20 @@ private static int compareDate(Object first, Date second) {
...
+            if (first == null) {
+                first = false;
+            }
+            if (first instanceof Boolean) {
+                return Boolean.compare((Boolean) first, second);
+            }
+
+            throw new IllegalArgumentException(String.format(
+                    "Can't compare between %s(%s) and %s(%s)",
+                    first, first.getClass().getSimpleName(),
```
imbajin: ⚠️ **Duplicate `compareBoolean` in two Condition classes**

The exact same `compareBoolean` method is added in both:
- `hugegraph-server/.../backend/query/Condition.java`
- `hugegraph-struct/.../query/Condition.java`

This is copy-paste duplication. If the null-handling semantics or error messages need to change later, both copies must be kept in sync. Is there a shared utility class or base class where this could live instead? If duplication is intentional (module isolation), consider at least adding a test for the `hugegraph-struct` version as well.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -488,6 +494,36 @@ private static Condition.Relation convCompare2UserpropRelation(HugeGraph graph,
...
+            case eq:
+                return Condition.eq(key, value);
+            case neq:
+                return Condition.neq(key, value);
+            case gt:
+                return value ? impossibleBooleanCondition(key) :
+                       Condition.eq(key, true);
+            case gte:
+                return value ? Condition.eq(key, true) :
+                       Condition.in(key, ImmutableList.of(false, true));
```
imbajin, line 510: 🧹 **Minor: `gte(false)` and `lte(true)` normalize to `IN(false, true)` — could simplify to "no filter"**

When `gte(false)` or `lte(true)` is applied to a boolean property, the code normalizes to `Condition.in(key, [false, true])`. Since these two values are the entire boolean domain, this is semantically equivalent to "match everything" (no filter). The `IN(false, true)` condition still gets evaluated at runtime for every candidate element.

This isn't a bug, but for a performance-conscious path, you could detect this case and skip adding the condition entirely. Low priority since boolean properties are typically small cardinality.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/Condition.java
```diff
@@ -230,6 +232,20 @@ private static int compareDate(Object first, Date second) {
                     second, second.getClass().getSimpleName()));
         }
 
+        private static int compareBoolean(Object first, Boolean second) {
+            if (first == null) {
```
imbajin: ⚠️ **null 静默转为 false，与 compareDate 行为不一致**

`compareDate` 遇到非 Date 的 `first` 时直接 throw，但 `compareBoolean` 将 `null` 静默设为 `false`。这会掩盖两类问题：
1. 属性不存在（null = 未设置）与属性值为 false 在语义上是不同的
2. 如果 `first` 是 null，后面的 `first.getClass().getSimpleName()` 本会 NPE，这里只是绕过了 NPE，没有真正解决问题

建议对齐 `compareDate` 的风格，同时修复 NPE 隐患：
```java
private static int compareBoolean(Object first, Boolean second) {
    if (first instanceof Boolean) {
        return Boolean.compare((Boolean) first, second);
    }
    throw new IllegalArgumentException(String.format(
            "Can't compare between %s(%s) and %s(%s)",
            first, first == null ? "null" : first.getClass().getSimpleName(),
            second, second.getClass().getSimpleName()));
}
```

如果业务上确实需要 null→false 的语义（某些 backend 存储缺省值为 null），请在注释中说明原因。

hugegraph-struct/src/main/java/org/apache/hugegraph/query/Condition.java
```diff
@@ -472,6 +474,20 @@ private static int compareDate(Object first, Date second) {
                     second, second.getClass().getSimpleName()));
         }
 
+        private static int compareBoolean(Object first, Boolean second) {
+            if (first == null) {
```
imbajin: ⚠️ 同上：`hugegraph-struct` 的 `compareBoolean` 有相同的 null→false 静默处理问题。两份拷贝需保持一致，建议同步修改。

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -488,6 +494,31 @@ private static Condition.Relation convCompare2UserpropRelation(HugeGraph graph,
         }
     }
 
+    private static Condition convCompare2BooleanUserpropRelation(Compare compare,
+                                                                 Id key,
+                                                                 Boolean value) {
+        switch (compare) {
+            case eq:
+                return Condition.eq(key, value);
```
imbajin, line 502: ⚠️ **`neq` 未被规范化，可能无法利用二级索引**

`neq(true)` 语义等价于 `eq(false)`，`neq(false)` 等价于 `eq(true)`。当前直接透传 `Condition.neq(key, value)` 到 backend，而 neq 通常走全扫描 + 过滤，无法命中 secondary index（例如 `strikeByArrested`）。

该 PR 的目标之一是让 boolean 谓词走正确的索引路径，`neq` 也应一并处理：
```java
case neq:
    return Condition.eq(key, !value);
```
这样 `neq(true)` → `eq(false)`，可以命中 secondary index。

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/EdgeCoreTest.java
```diff
@@ -7117,6 +7117,147 @@ public void testQueryEdgeWithNullablePropertyInCompositeIndex() {
...
+        List<Edge> hasLteTrueEdges = graph.traversal().E()
+                                           .has("arrested", P.lte(true))
+                                           .toList();
+        Assert.assertEquals(2, hasLteTrueEdges.size());
+        Set<Integer> lteTrueIds = new HashSet<>();
+        for (Edge edge : hasLteTrueEdges) {
+            lteTrueIds.add(edge.value("id"));
+        }
+        Assert.assertEquals(ImmutableSet.of(1, 2), lteTrueIds);
+
```
imbajin, line 7212: 🧹 **缺少 `neq` 谓词的端到端测试，且仅覆盖了 Edge**

- `P.neq(true)` / `P.neq(false)` 没有集成测试，而上面的 neq 实现路径与其他谓词不同（如果接受上述 neq 规范化建议，需要补充验证）
- 所有新增集成测试均在 `EdgeCoreTest`，建议同步在 `VertexCoreTest` 中添加对应的 boolean range 覆盖，确认 `TraversalUtil.convCompare2UserpropRelation` 对 `HugeType.VERTEX` 也正常工作

imbajin (commented):
整体设计合理，规范化路径（TraversalUtil 层提前将 boolean 范围谓词转换为 eq/in/empty）是正确做法，测试覆盖也较为全面。有两处中等优先级问题需要决策：compareBoolean 的 null 处理语义，以及 neq 未被规范化。

contrueCT:
@imbajin  这次提交去掉了 compareBoolean 里 null -> false 的静默转换，现在 boolean 比较遇到 null 会显式失败；hugegraph-struct 中对应实现也已同步修改，保持两边语义一致。另外，TraversalUtil 中已将 boolean neq(value) 规范化为 eq(!value)，让它可以和 boolean equality 一样走 secondary index 路径。测试也补充了 null 比较异常、Edge 上的 P.neq(true/false)，以及 Vertex 上对应的 boolean neq 路径覆盖。

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/ConditionQueryFlattenTest.java
```diff
@@ -256,5 +256,107 @@ public void testFlattenWithNotIn() {
...
+        ConditionQuery query = new ConditionQuery(HugeType.VERTEX);
+        Condition eq = Condition.eq(key, true);
+        query.query(eq.or(Condition.in(key, ImmutableList.of())));
+
+        List<ConditionQuery> queries = ConditionQueryFlatten.flatten(query);
+        Assert.assertEquals(1, queries.size());
+
+        Collection<Condition> actual = queries.iterator().next().conditions();
+        Assert.assertEquals(ImmutableList.of(eq), actual);
+    }
```
imbajin, line 342: 🧹 **Minor — consider adding a deeper-nested regression test for `IN []` null-propagation**

The new `testFlattenWithImpossibleInInsideAnd/Or/OrRight` cases all exercise **2-level** trees (`AND(IN[], eq)` / `OR(IN[], eq)`). The new `flattenIn` null-propagation is recursive and should handle arbitrary depth correctly, but there's no direct assertion for 3+ level trees such as:

```java
// AND( OR(IN[], eq(a)), eq(b) )  →  eq(a) AND eq(b)
query.query(Condition.in(key, ImmutableList.of()).or(Condition.eq(key, "a"))
                     .and(Condition.eq(key, "b")));
```

**Why this matters:** the current implementation is correct today, but if someone later refactors `flattenIn` (e.g. adds short-circuits, converts to iterative, or caches intermediate nodes), deep-nested null propagation is easy to break without a direct test signal. A single 3-level AND-over-OR case would pin this behavior cheaply.

**Impact if not added:** no correctness regression today; moderate future regression risk when `flattenIn` is touched. Happy for this to land as-is and add in a follow-up if preferred.

imbajin (commented):
LGTM overall. One minor follow-up below to harden regression coverage against future refactors of `flattenIn`.

contrueCT:
Follow-up Note:
`hugegraph-struct` currently participates in PR CI through a build/install step (`mvn -DskipTests -pl hugegraph-struct -am install`) and its test cases are not executed by default in the existing workflow.

For this PR, the new struct-side regression test was validated manually on Linux together with the related server-side verification.

A follow-up issue will track enabling `hugegraph-struct` test execution in CI.

imbajin (approved):

PR #2997, opened 2026-04-18
fix(server): auto-recover session after Cassandra restart

### LINKED ISSUE #2740: [Bug] Hugegraph isn't responding after Cassandra restarted.
### Bug Type (问题类型)

server status (启动/运行异常)

### Before submit

### Environment (环境信息)

- Server Version: Latest Code Build
- Backend: Cassandra
- OS: Ubuntu
- Data Size:  Just few `data`

I did following steps:

1. Started cassandra (V 5.0.3)
2. Started Hugegraph
```
./start-hugegraph.sh 
Starting HugeGraphServer in daemon mode...
Connecting to HugeGraphServer (http://127.0.0.1:8080/graphs)....OK
Started [pid 1252693]

```
3. Created one vertex from Swagger UI and its got created.
4. Stopped cassandra.
5. Again tried from Swagger UI and it's returning error:
```
Failed to fetch.
Possible Reasons:

CORS
Network Failure
URL scheme must be "http" or "https" for CORS request.
```
6. Restarted cassandra and tried from Swagger but same issue

```
incubator-hugegraph-master/target/apache-hugegraph-incubating-1.5.0/apache-hugegraph-server-incubating-1.5.0/bin$ ./start-hugegraph.sh 
Starting HugeGraphServer in daemon mode...
Connecting to HugeGraphServer (http://127.0.0.1:8080/graphs)....OK
Started [pid 1254569]
```

```
incubator-hugegraph-master/target/apache-hugegraph-incubating-1.5.0/apache-hugegraph-server-incubating-1.5.0/bin$ ./stop-hugegraph.sh 
no crontab for manish
The HugeGraphServer monitor has been closed
Dev/incubator-hugegraph-master/target/apache-hugegraph-incubating-1.5.0/apache-hugegraph-server-incubating-1.5.0/bin/util.sh: line 375: kill: (1254569) - No such process
Killing HugeGraphServer(pid 1254569).OK
```

IN the log i don't see anything, here is the last few line 

```
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - Configured application/vnd.gremlin-v2.0+json with org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV2d0
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - application/json already has org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV1d0 configured - it will not be replaced by org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV2d0, change order of serialization configuration if this is not desired.
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - Configured application/vnd.gremlin-v3.0+json with org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV3d0
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - application/json already has org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV1d0 configured - it will not be replaced by org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV3d0, change order of serialization configuration if this is not desired.
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - Configured application/vnd.graphbinary-v1.0 with org.apache.tinkerpop.gremlin.driver.ser.GraphBinaryMessageSerializerV1
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - Configured application/json with org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV1d0
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - Configured application/vnd.gremlin-v2.0+json with org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV2d0
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - application/json already has org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV1d0 configured - it will not be replaced by org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV2d0, change order of serialization configuration if this is not desired.
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - Configured application/vnd.gremlin-v3.0+json with org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV3d0
2025-02-28 14:43:04 [main] [INFO] o.a.t.g.s.AbstractChannelizer - application/json already has org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV1d0 configured - it will not be replaced by org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV3d0, change order of serialization configuration if this is not desired.
2025-02-28 14:43:04 [gremlin-server-boss-1] [INFO] o.a.t.g.s.GremlinServer - Gremlin Server configured with worker thread pool of 1, gremlin pool of 8 and boss thread pool of 1.
2025-02-28 14:43:04 [gremlin-server-boss-1] [INFO] o.a.t.g.s.GremlinServer - Channel started at port 8182.
2025-02-28 14:43:04 [main] [INFO] o.a.h.d.MemoryMonitor - Memory monitoring started.
```

```javascript

```

```javascript

```

dpol1:
## Purpose of the PR

closes #2740

HugeGraphServer stops responding after Cassandra is restarted and never
recovers without a full server restart.

Root cause: `CassandraSessionPool` builds the Datastax `Cluster` without a
`ReconnectionPolicy`, `CassandraSession.execute(...)` calls the driver once
with no retry, and thread-local sessions are never probed for liveness.
Once Cassandra goes down, transient `NoHostAvailableException` /
`OperationTimedOutException` errors surface to the user and the pool stays
dead even after Cassandra comes back online.

## Main Changes

- Register `ExponentialReconnectionPolicy(baseDelay, maxDelay)` on the
  `Cluster` builder so the Datastax driver keeps retrying downed nodes in
  the background.
- Wrap every `Session.execute(...)` in `executeWithRetry(Statement)` with
  exponential backoff on transient connectivity failures.
- Implement `reconnectIfNeeded()` / `reset()` on `CassandraSession` so the
  pool reopens closed sessions and issues a lightweight health-check
  (`SELECT now() FROM system.local`) before subsequent queries.
- Add four tunables in `CassandraOptions` (defaults preserve previous
  behavior for healthy clusters):

  | Option | Default | Meaning |
  |--------|---------|---------|
  | `cassandra.reconnect_base_delay`  | `1000` ms  | Initial backoff for driver reconnection policy |
  | `cassandra.reconnect_max_delay`   | `10000` ms | Cap for reconnection backoff |
  | `cassandra.query_retry_max_attempts` | `3`       | Per-query retries on transient errors (`0` disables) |
  |  `cassandra.query_retry_interval`    | `1000` ms  | Base interval for per-query exponential backoff |

- Add unit tests covering defaults, overrides, disabling retries and option keys.

## Verifying these changes

  - `mvn -pl hugegraph-server/hugegraph-test -am test -Dtest=CassandraTest` — 13/13 pass

- [x] Modify configurations

## Documentation Status

- [x] `Doc - TODO`

### DIFF
--- a/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraOptions.java
+++ b/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraOptions.java
@@ -148,7 +148,7 @@ public class CassandraOptions extends OptionHolder {
                     "exponential reconnection policy when a Cassandra host " +
                     "becomes unreachable.",
                     rangeInt(1000L, Long.MAX_VALUE),
-                    60_000L
+                    10_000L
             );
 
     public static final ConfigOption<Integer> CASSANDRA_RECONNECT_MAX_RETRIES =
@@ -159,7 +159,7 @@ public class CassandraOptions extends OptionHolder {
                     "(NoHostAvailableException / OperationTimedOutException). " +
                     "Set to 0 to disable query-time retries.",
                     rangeInt(0, Integer.MAX_VALUE),
-                    10
+                    3
             );
 
     public static final ConfigOption<Long> CASSANDRA_RECONNECT_INTERVAL =
@@ -170,6 +170,6 @@ public class CassandraOptions extends OptionHolder {
                     "actual wait grows with exponential backoff, capped at " +
                     "cassandra.reconnect_max_delay.",
                     rangeInt(100L, Long.MAX_VALUE),
-                    5000L
+                    1000L
             );
 }
--- a/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraOptions.java
+++ b/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraOptions.java
@@ -151,20 +151,21 @@ public class CassandraOptions extends OptionHolder {
                     10_000L
             );
 
-    public static final ConfigOption<Integer> CASSANDRA_RECONNECT_MAX_RETRIES =
+    public static final ConfigOption<Integer> CASSANDRA_QUERY_RETRY_MAX_ATTEMPTS =
             new ConfigOption<>(
-                    "cassandra.reconnect_max_retries",
-                    "The maximum number of retries applied at query-time when " +
-                    "a Cassandra host is temporarily unreachable " +
-                    "(NoHostAvailableException / OperationTimedOutException). " +
+                    "cassandra.query_retry_max_attempts",
+                    "The maximum number of retry attempts applied at query-time when " +
+                    "a Cassandra host is temporarily unreachable. " +
+                    "OperationTimedOutException is retried only for " +
+                    "idempotent statements. " +
                     "Set to 0 to disable query-time retries.",
                     rangeInt(0, Integer.MAX_VALUE),
                     3
             );
 
-    public static final ConfigOption<Long> CASSANDRA_RECONNECT_INTERVAL =
+    public static final ConfigOption<Long> CASSANDRA_QUERY_RETRY_INTERVAL =
             new ConfigOption<>(
-                    "cassandra.reconnect_interval",
+                    "cassandra.query_retry_interval",
                     "The interval in milliseconds between query-time retries " +
                     "when a Cassandra host is temporarily unreachable. The " +
                     "actual wait grows with exponential backoff, capped at " +
--- a/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraOptions.java
+++ b/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraOptions.java
@@ -130,4 +130,46 @@ public class CassandraOptions extends OptionHolder {
                     positiveInt(),
                     12 * 60 * 60
             );
+
+    public static final ConfigOption<Long> CASSANDRA_RECONNECT_BASE_DELAY =
+            new ConfigOption<>(
+                    "cassandra.reconnect_base_delay",
+                    "The base delay in milliseconds used by the driver's " +
+                    "exponential reconnection policy when a Cassandra host " +
+                    "becomes unreachable.",
+                    rangeInt(100L, Long.MAX_VALUE),
+                    1000L
+            );
+
+    public static final ConfigOption<Long> CASSANDRA_RECONNECT_MAX_DELAY =
+            new ConfigOption<>(
+                    "cassandra.reconnect_max_delay",
+                    "The maximum delay in milliseconds used by the driver's " +
+                    "exponential reconnection policy when a Cassandra host " +
+                    "becomes unreachable.",
+                    rangeInt(1000L, Long.MAX_VALUE),
+                    60_000L
+            );
+
+    public static final ConfigOption<Integer> CASSANDRA_RECONNECT_MAX_RETRIES =
+            new ConfigOption<>(
+                    "cassandra.reconnect_max_retries",
+                    "The maximum number of retries applied at query-time when " +
+                    "a Cassandra host is temporarily unreachable " +
+                    "(NoHostAvailableException / OperationTimedOutException). " +
+                    "Set to 0 to disable query-time retries.",
+                    rangeInt(0, Integer.MAX_VALUE),
+                    10
+            );
+
+    public static final ConfigOption<Long> CASSANDRA_RECONNECT_INTERVAL =
+            new ConfigOption<>(
+                    "cassandra.reconnect_interval",
+                    "The interval in milliseconds between query-time retries " +
+                    "when a Cassandra host is temporarily unreachable. The " +
+                    "actual wait grows with exponential backoff, capped at " +
+                    "cassandra.reconnect_max_delay.",
+                    rangeInt(100L, Long.MAX_VALUE),
+                    5000L
+            );
 }
--- a/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
+++ b/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
@@ -55,18 +55,30 @@ public class CassandraSessionPool extends BackendSessionPool {
 
     private Cluster cluster;
     private final String keyspace;
-    private int maxRetries;
-    private long retryInterval;
-    private long retryMaxDelay;
+    private final int maxRetries;
+    private final long retryInterval;
+    private final long retryMaxDelay;
 
     public CassandraSessionPool(HugeConfig config,
                                 String keyspace, String store) {
         super(config, keyspace + "/" + store);
         this.cluster = null;
         this.keyspace = keyspace;
-        this.maxRetries = 0;
-        this.retryInterval = 0L;
-        this.retryMaxDelay = 0L;
+        this.maxRetries = config.get(
+                CassandraOptions.CASSANDRA_RECONNECT_MAX_RETRIES);
+        this.retryInterval = config.get(
+                CassandraOptions.CASSANDRA_RECONNECT_INTERVAL);
+        long reconnectBase = config.get(
+                CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY);
+        long reconnectMax = config.get(
+                CassandraOptions.CASSANDRA_RECONNECT_MAX_DELAY);
+        E.checkArgument(reconnectMax >= reconnectBase,
+                        "'%s' (%s) must be >= '%s' (%s)",
+                        CassandraOptions.CASSANDRA_RECONNECT_MAX_DELAY.name(),
+                        reconnectMax,
+                        CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY.name(),
+                        reconnectBase);
+        this.retryMaxDelay = reconnectMax;
     }
 
     @Override
@@ -107,21 +119,9 @@ public class CassandraSessionPool extends BackendSessionPool {
         // with exponential backoff after they go down (see issue #2740).
         long reconnectBase = config.get(
                 CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY);
-        long reconnectMax = config.get(
-                CassandraOptions.CASSANDRA_RECONNECT_MAX_DELAY);
-        E.checkArgument(reconnectMax >= reconnectBase,
-                        "'%s' (%s) must be >= '%s' (%s)",
-                        CassandraOptions.CASSANDRA_RECONNECT_MAX_DELAY.name(),
-                        reconnectMax,
-                        CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY.name(),
-                        reconnectBase);
         builder.withReconnectionPolicy(
-                new ExponentialReconnectionPolicy(reconnectBase, reconnectMax));
-        this.retryMaxDelay = reconnectMax;
-        this.maxRetries = config.get(
-                CassandraOptions.CASSANDRA_RECONNECT_MAX_RETRIES);
-        this.retryInterval = config.get(
-                CassandraOptions.CASSANDRA_RECONNECT_INTERVAL);
+                new ExponentialReconnectionPolicy(reconnectBase,
+                                                  this.retryMaxDelay));
 
         // Credential options
         String username = config.get(CassandraOptions.CASSANDRA_USERNAME);
@@ -211,6 +211,7 @@ public class CassandraSessionPool extends BackendSessionPool {
             int processors = Math.min(statements.size(), 1023);
             List<ResultSetFuture> results = new ArrayList<>(processors + 1);
             for (Statement s : statements) {
+                // TODO: commitAsync is not retried (async retry semantics are complex)
                 ResultSetFuture future = this.session.executeAsync(s);
                 results.add(future);
 
@@ -251,7 +252,15 @@ public class CassandraSessionPool extends BackendSessionPool {
          * itself keeps retrying connections in the background via the
          * reconnection policy, so once Cassandra comes back online, a
          * subsequent attempt here will succeed without restarting the server.
-         * See issue #2740.
+         *
+         * <p><b>Blocking note:</b> retries block the calling thread via
+         * {@link Thread#sleep(long)}. Worst-case a single call blocks for
+         * {@code maxRetries * retryMaxDelay} ms. Under high-throughput
+         * workloads concurrent threads may pile up in {@code sleep()} during
+         * a Cassandra outage. For such deployments lower
+         * {@code cassandra.reconnect_max_retries} (default 10) and
+         * {@code cassandra.reconnect_max_delay} (default 60000ms) so the
+         * request fails fast and pressure is released back to the caller.
          */
         private ResultSet executeWithRetry(Statement statement) {
             int retries = CassandraSessionPool.this.maxRetries;
@@ -266,8 +275,15 @@ public class CassandraSessionPool extends BackendSessionPool {
                     if (attempt >= retries) {
                         break;
                     }
-                    long delay = Math.min(interval * (1L << Math.min(attempt, 20)),
-                                          maxDelay > 0 ? maxDelay : interval);
+                    long cap = maxDelay > 0 ? maxDelay : interval;
+                    long shift = 1L << Math.min(attempt, 20);
+                    long delay;
+                    try {
+                        // Guard against Long overflow when retryInterval is huge.
+                        delay = Math.min(Math.multiplyExact(interval, shift), cap);
+                    } catch (ArithmeticException overflow) {
+                        delay = cap;
+                    }
                     LOG.warn("Cassandra temporarily unavailable ({}), " +
                              "retry {}/{} in {} ms",
                              e.getClass().getSimpleName(), attempt + 1,
@@ -282,11 +298,14 @@ public class CassandraSessionPool extends BackendSessionPool {
                     }
                 }
             }
-            throw new BackendException("Failed to execute Cassandra query " +
-                                       "after %s retries: %s",
-                                       lastError, retries,
-                                       lastError == null ? "<null>" :
-                                               lastError.getMessage());
+            // Preserve original exception as cause (stack trace + type) by
+            // pre-formatting the message and using the (String, Throwable)
+            // constructor explicitly — avoids ambiguity with varargs overloads.
+            String msg = String.format(
+                    "Failed to execute Cassandra query after %s retries: %s",
+                    retries,
+                    lastError == null ? "<null>" : lastError.getMessage());
+            throw new BackendException(msg, lastError);
         }
 
         private void tryOpen() {
@@ -353,14 +372,20 @@ public class CassandraSessionPool extends BackendSessionPool {
                 if (this.session == null || this.session.isClosed()) {
                     this.session = null;
                     this.tryOpen();
-                    if (this.session == null) {
-                        return;
-                    }
                 }
-                this.session.execute(new SimpleStatement(HEALTH_CHECK_CQL));
+                if (this.session != null) {
+                    this.session.execute(new SimpleStatement(HEALTH_CHECK_CQL));
+                }
             } catch (DriverException e) {
                 LOG.debug("Cassandra health-check failed, " +
                           "will retry on next query: {}", e.getMessage());
+            } finally {
+                // Keep opened flag consistent with session: if tryOpen()
+                // failed to reopen, clear opened so the next execute() does
+                // not NPE before executeWithRetry() can intercept.
+                if (this.session == null) {
+                    this.opened = false;
+                }
             }
         }
 
@@ -376,7 +401,9 @@ public class CassandraSessionPool extends BackendSessionPool {
             }
             try {
                 this.session.close();
-            } catch (Throwable e) {
+            } catch (Exception e) {
+                // Do not swallow Error (OOM / StackOverflow); only log
+                // ordinary exceptions raised by the driver on close.
                 LOG.warn("Failed to reset Cassandra session", e);
             } finally {
                 this.session = null;
--- a/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
+++ b/hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
@@ -20,6 +20,7 @@ package org.apache.hugegraph.backend.store.cassandra;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.hugegraph.backend.BackendException;
 import org.apache.hugegraph.backend.store.BackendSession.AbstractBackendSession;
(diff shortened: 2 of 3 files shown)


hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -197,15 +234,59 @@ public ResultSet query(Statement statement) {
...
+        /**
+         * Execute a statement, retrying on transient connectivity failures
+         * (NoHostAvailableException / OperationTimedOutException). The driver
+         * itself keeps retrying connections in the background via the
+         * reconnection policy, so once Cassandra comes back online, a
+         * subsequent attempt here will succeed without restarting the server.
+         * See issue #2740.
+         */
+        private ResultSet executeWithRetry(Statement statement) {
+            int retries = CassandraSessionPool.this.maxRetries;
```
imbajin, line 303: ‼️ **Critical: BackendException argument order is wrong — `lastError` is passed as the first vararg instead of as the `cause`**

The constructor signature being called here is `BackendException(String message, Object... args)`, which means `lastError` is used as a format argument (`%s`), **not** as the exception cause. The original exception's stack trace is lost entirely.

The format string has two `%s` placeholders but the args are `(lastError, retries, message)` — three args. The first `%s` will be filled by `lastError.toString()` (the exception object), the second `%s` by `retries` (an int). The `message` string becomes a dangling unused arg.

To preserve the cause chain **and** format the message correctly, use the `(String, Throwable, Object...)` constructor:

```suggestion
            throw new BackendException("Failed to execute Cassandra query " +
                                       "after %s retries: %s",
                                       lastError, retries,
                                       lastError == null ? "<null>" :
                                               lastError.getMessage());
```

Should be:

```java
throw new BackendException(
    "Failed to execute Cassandra query after %s retries: %s",
    lastError, retries,
    lastError == null ? "<null>" : lastError.getMessage());
```

Wait — looking more carefully at the available constructors:
- `BackendException(String message, Throwable cause, Object... args)` — this is what you need.

The fix:
```java
throw new BackendException(
    "Failed to execute Cassandra query after %s retries",
    lastError, retries);
```

This passes `lastError` as the `cause` (Throwable) and `retries` as the format arg. The cause's message is already accessible via `getCause().getMessage()`, so repeating it in the format string is unnecessary.

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -255,6 +336,53 @@ public boolean hasChanges() {
...
         }
 
+        /**
+         * Periodic liveness probe invoked by {@link BackendSessionPool} to
+         * recover thread-local sessions after Cassandra has been restarted.
+         * Reopens the driver session if it was closed and pings the cluster
+         * with a lightweight query. Any failure here is swallowed so the
+         * caller can still issue the real query, which will drive retries via
+         * {@link #executeWithRetry(Statement)}.
+         */
```
imbajin, line 435: ⚠️ **`reconnectIfNeeded()` checks `this.opened` but never sets it to `false` on failure**

When the health check (`HEALTH_CHECK_CQL`) fails, the session stays in `opened = true` state. On the next call to `detectSession()` from `BackendSessionPool`, the idle-time check (`now - session.updated() > interval`) may skip `reconnectIfNeeded()` entirely because `session.update()` was called right after.

This means if the health check catches a failure, the session is marked as "recently updated" but is actually broken — and `reconnectIfNeeded()` won't be called again until the detect interval elapses.

Also: if `tryOpen()` fails inside `reconnectIfNeeded()`, the method silently returns. But `this.opened` was already `true` (the guard at the top), and `this.session` is now `null`. Any subsequent `execute()` call will NPE on `this.session.execute(...)` before `executeWithRetry` can even catch a `DriverException`.

Consider setting `this.opened = false` when the session is nulled out, so that `tryOpen()` is properly re-triggered on the next access.

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -197,15 +234,59 @@ public ResultSet query(Statement statement) {
...
-            return this.session.execute(statement);
+            return this.executeWithRetry(new SimpleStatement(statement));
         }
 
         public ResultSet execute(String statement, Object... args) {
-            return this.session.execute(statement, args);
+            return this.executeWithRetry(new SimpleStatement(statement, args));
+        }
+
+        /**
```
imbajin, line 278: 🧹 **Minor: `Thread.sleep()` in `executeWithRetry` blocks the calling thread**

This is acceptable for low-QPS scenarios, but for high-throughput workloads with many concurrent queries hitting a downed Cassandra, all request threads will pile up sleeping here (up to 10 retries × 60s max delay = 10 minutes worst case per thread).

Consider whether the retry count and max delay defaults might be too aggressive for production. A thread blocked for minutes can cascade into thread pool exhaustion. An alternative would be to fail faster (e.g., 3 retries with shorter delays) and let the caller/user retry at a higher level.

Not a blocker — just worth considering for the default values.

imbajin:
⚠️ **`commitAsync()` bypasses retry — still calls `this.session.executeAsync(s)` directly**

The PR wraps `execute()` and `commit()` with `executeWithRetry`, but `commitAsync()` (line 177 in the base file) still calls `this.session.executeAsync(s)` directly. If a Cassandra restart happens during an async batch commit, the same connectivity failure will surface without any retry.

Consider wrapping the async path as well, or at minimum adding a TODO/comment explaining why async commits are deliberately left un-retried (e.g., if retry semantics for async batches are too complex for this PR).

dpol1:
Thanks @imbajin for the feedback, changed!

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -86,6 +115,14 @@ public synchronized void open() {
 
         builder.withSocketOptions(socketOptions);
 
+        // Reconnection policy: let driver keep retrying nodes in background
+        // with exponential backoff after they go down (see issue #2740).
+        long reconnectBase = config.get(
+                CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY);
```
imbajin: ⚠️ **Duplicate config read for `reconnect_base_delay`**

`CASSANDRA_RECONNECT_BASE_DELAY` is read twice from config: once in the constructor (line 72) to validate the `base ≤ max` constraint, and again here in `open()` (line 120-121) to build the reconnection policy.

The constructor already stores `reconnectMax` as `this.retryMaxDelay`, but `reconnectBase` is discarded after validation. This means:
1. `config.get()` is called redundantly on every `open()`.
2. If the config object were mutable (or in tests with different config instances), the two reads could theoretically diverge.

**Suggestion:** store `reconnectBase` as a field alongside `retryMaxDelay` in the constructor, then use `this.retryBaseDelay` here:

```java
// constructor
this.retryBaseDelay = reconnectBase;   // add this field
this.retryMaxDelay  = reconnectMax;

// open()
builder.withReconnectionPolicy(
        new ExponentialReconnectionPolicy(this.retryBaseDelay,
                                          this.retryMaxDelay));
```

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -255,6 +355,61 @@ public boolean hasChanges() {
...
+                }
+            } catch (DriverException e) {
+                LOG.debug("Cassandra health-check failed, " +
+                          "will retry on next query: {}", e.getMessage());
+            } finally {
+                // Keep opened flag consistent with session: if tryOpen()
+                // failed to reopen, clear opened so the next execute() does
+                // not NPE before executeWithRetry() can intercept.
+                if (this.session == null) {
+                    this.opened = false;
```
imbajin: ‼️ **Mutating `this.opened` in `reconnectIfNeeded()` breaks parent-class lifecycle invariants**

`opened` is a field of `AbstractBackendSession` that tracks the ref-counted lifecycle managed by `BackendSessionPool`. Setting it to `false` here bypasses that contract.

The normal teardown flow is:
```
BackendSessionPool.closeSession()
  → session.detach()         // decrements ref
  → if ref == 0: session.close()
  → threadLocalSession.remove()
  → sessions.remove(threadId)
  → sessionCount.decrementAndGet()
```

If `this.opened = false` is set here instead, the session object stays registered in `sessions` and `sessionCount` is never decremented. On the next `getOrNewSession()` call the pool sees `opened == false` and creates a new session, leaking the old entry.

Additionally, `opened()` already guards against a null session:
```java
@Override
public boolean opened() {
    if (this.opened && this.session == null) {
        this.tryOpen();           // already handles the re-open
    }
    return this.opened && this.session != null;
}
```
So the NPE concern in the `finally` comment is already handled by `opened()`. The `finally` block can be removed entirely — just catch the exception and log, letting the normal `opened()` / `executeWithRetry()` path handle the rest.

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -174,6 +211,7 @@ public void commitAsync() {
             int processors = Math.min(statements.size(), 1023);
             List<ResultSetFuture> results = new ArrayList<>(processors + 1);
             for (Statement s : statements) {
+                // TODO: commitAsync is not retried (async retry semantics are complex)
```
imbajin: ‼️ **`commitAsync()` has no retry protection — inconsistent write reliability**

`commit()` (sync path) wraps all statements in `executeWithRetry`, but `commitAsync()` calls `this.session.executeAsync(s)` directly. This means write operations on the async path get no protection against transient `NoHostAvailableException` / `OperationTimedOutException` — exactly the failures this PR targets.

Timeline during a Cassandra restart:

```
Thread A (sync):   commit() → executeWithRetry() → retries → succeeds ✓
Thread B (async):  commitAsync() → executeAsync() → NoHostAvailableException ✗
```

Both paths commit the same batch type; users cannot know which one they're calling or whether it will be protected.

If implementing async retry is genuinely deferred, the TODO should be more explicit about the consequence — callers may see write failures during a Cassandra restart even with `maxRetries > 0` configured. Consider at minimum logging a warning at startup when `maxRetries > 0`.

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -197,15 +235,77 @@ public ResultSet query(Statement statement) {
...
+         * <p><b>Blocking note:</b> retries block the calling thread via
+         * {@link Thread#sleep(long)}. Worst-case a single call blocks for
+         * {@code maxRetries * retryMaxDelay} ms. Under high-throughput
+         * workloads concurrent threads may pile up in {@code sleep()} during
+         * a Cassandra outage. For such deployments lower
+         * {@code cassandra.reconnect_max_retries} (default 10) and
+         * {@code cassandra.reconnect_max_delay} (default 60000ms) so the
+         * request fails fast and pressure is released back to the caller.
+         */
+        private ResultSet executeWithRetry(Statement statement) {
```
imbajin, line 302: ⚠️ **Default settings allow a single query call to block for up to ~7 minutes**

With the defaults `maxRetries = 10`, `reconnect_interval = 5000 ms`, and `reconnect_max_delay = 60 000 ms`, the worst-case wall time for one `executeWithRetry()` call is:

```
attempt 0 → sleep(min(5000 * 1, 60000)) =  5 000 ms
attempt 1 → sleep(min(5000 * 2, 60000)) = 10 000 ms
attempt 2 → sleep(min(5000 * 4, 60000)) = 20 000 ms
attempt 3 → sleep(min(5000 * 8, 60000)) = 40 000 ms
attempts 4–9    → sleep(60 000 ms each) = 360 000 ms
                                          ──────────
total blocking time ≈ 435 seconds (~7 min)
```

During a Cassandra outage every in-flight thread sleeps through this sequence. On a busy server, hundreds of threads pile up in `Thread.sleep()`, exhausting the thread pool well before `maxRetries` is hit.

Consider lowering the defaults so the query-level retry is a short hiccup buffer, while the driver-level `ExponentialReconnectionPolicy` handles the actual node reconnection in the background:

| Option | Current default | Suggested |
|--------|----------------|-----------|
| `reconnect_max_retries` | 10 | 3 |
| `reconnect_interval` | 5 000 ms | 1 000 ms |
| `reconnect_max_delay` | 60 000 ms | 10 000 ms |

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -255,6 +355,61 @@ public boolean hasChanges() {
...
+            if (!this.opened) {
+                return;
+            }
+            try {
+                if (this.session == null || this.session.isClosed()) {
+                    this.session = null;
+                    this.tryOpen();
+                }
+                if (this.session != null) {
+                    this.session.execute(new SimpleStatement(HEALTH_CHECK_CQL));
```
imbajin, line 447: ⚠️ **Health-check failure silently leaves the dead session in place**

When `session.execute(HEALTH_CHECK_CQL)` throws a `DriverException`, the exception is caught and logged at DEBUG — but the `session` object that produced the failure is kept. The probe's purpose is to detect a dead session early; if it fails the session should be marked unhealthy immediately.

Current flow:
```
health-check → DriverException
  → LOG.debug(...)
  → session is NOT cleared
  → next executeWithRetry() tries the same dead session
    → pays the full retry cost anyway
```

Desired flow:
```
health-check → DriverException
  → session = null   ← mark unhealthy now
  → opened() will call tryOpen() before the next real query
```

Suggested fix:
```java
} catch (DriverException e) {
    LOG.debug("Cassandra health-check failed, resetting session: {}",
              e.getMessage());
    this.session = null;   // force re-open on next query via opened()
}
```

This also makes the `finally` block that sets `this.opened = false` unnecessary (see separate comment on that).

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cassandra/CassandraTest.java
```diff
@@ -192,4 +198,99 @@ public void testParseReplicaWithNetworkTopologyStrategyAndDoubleReplica() {
...
+        HugeConfig config = new HugeConfig(conf);
+        Assert.assertEquals(0, (int) config.get(
+                CassandraOptions.CASSANDRA_RECONNECT_MAX_RETRIES));
+    }
+
+    @Test
+    public void testExecuteWithRetrySucceedsAfterTransientFailures() {
+        Configuration conf = new PropertiesConfiguration();
+        HugeConfig config = new HugeConfig(conf);
+        CassandraSessionPool pool = new CassandraSessionPool(config,
```
imbajin, line 273: ⚠️ **Test constructs `CassandraSessionPool` with a blank config — fragile setup**

`new HugeConfig(new PropertiesConfiguration())` produces an empty config. The constructor calls `config.get(CASSANDRA_RECONNECT_BASE_DELAY)` etc., which falls back to option defaults only if `HugeConfig` performs that fallback correctly. If any option's `rangeInt` validator is applied to the absent value rather than the default, this test will fail with a confusing config error instead of a clear assertion.

The fields set via `Whitebox.setInternalState` immediately after construction throw away the constructed values anyway, so the constructor call is only there to get a valid object — but it's a fragile path.

Suggested fix: explicitly set the required config values so the test is self-describing and not dependent on default-fallback behaviour:

```java
Configuration conf = new PropertiesConfiguration();
conf.setProperty(CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY.name(), 100L);
conf.setProperty(CassandraOptions.CASSANDRA_RECONNECT_MAX_DELAY.name(), 1000L);
conf.setProperty(CassandraOptions.CASSANDRA_RECONNECT_MAX_RETRIES.name(), 3);
conf.setProperty(CassandraOptions.CASSANDRA_RECONNECT_INTERVAL.name(), 1L);
HugeConfig config = new HugeConfig(conf);
```

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cassandra/CassandraTest.java
```diff
@@ -192,4 +198,99 @@ public void testParseReplicaWithNetworkTopologyStrategyAndDoubleReplica() {
...
+        Whitebox.setInternalState(session, "session", driverSession);
+
+        ResultSet result = session.execute("SELECT now() FROM system.local");
+        Assert.assertSame(rs, result);
+        Mockito.verify(driverSession, Mockito.times(3))
+               .execute(Mockito.any(Statement.class));
+    }
+
+    @Test
+    public void testReconnectOptionsExposeExpectedKeys() {
```
imbajin: 🧹 **`testReconnectOptionsExposeExpectedKeys` only checks string literals — low-value test**

This test asserts that a static constant's `.name()` returns the exact string that is defined on the next line of the same class. If someone renames the key, they update both places together, so this test adds no real safety net.

What is actually worth testing — and not currently covered — is that the validators reject out-of-range values, e.g.:

```java
// base delay below minimum (100 ms) should be rejected
Configuration conf = new PropertiesConfiguration();
conf.setProperty(CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY.name(), 50L);
Assert.assertThrows(/* config validation exception */, () -> new HugeConfig(conf));

// max < base should be rejected (the E.checkArgument in the constructor)
conf.setProperty(CassandraOptions.CASSANDRA_RECONNECT_BASE_DELAY.name(), 5000L);
conf.setProperty(CassandraOptions.CASSANDRA_RECONNECT_MAX_DELAY.name(), 1000L);
Assert.assertThrows(IllegalArgumentException.class,
                    () -> new CassandraSessionPool(new HugeConfig(conf), "ks", "s"));
```

Consider replacing or augmenting this test with boundary checks like the above.

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraOptions.java
```diff
@@ -130,4 +130,46 @@ public static synchronized CassandraOptions instance() {
...
+            );
+
+    public static final ConfigOption<Long> CASSANDRA_RECONNECT_MAX_DELAY =
+            new ConfigOption<>(
+                    "cassandra.reconnect_max_delay",
+                    "The maximum delay in milliseconds used by the driver's " +
+                    "exponential reconnection policy when a Cassandra host " +
+                    "becomes unreachable.",
+                    rangeInt(1000L, Long.MAX_VALUE),
+                    10_000L
```
imbajin, line 151: 🧹 Minor: The default values in the PR description table don't match the actual code defaults:

| Option | PR Description | Code |
|--------|---------------|------|
| `reconnect_max_delay` | 60000 ms | **10000 ms** |
| `reconnect_max_retries` | 10 | **3** |
| `reconnect_interval` | 5000 ms | **1000 ms** |

Please update the PR description to match the code.

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -197,15 +256,92 @@ public ResultSet query(Statement statement) {
...
+                        // Cassandra is still unreachable; the catch below
+                        // treats that as a transient failure.
+                        this.open();
+                    }
+                    return this.session.execute(statement);
+                } catch (NoHostAvailableException | OperationTimedOutException e) {
+                    lastError = e;
+                    // Discard the (possibly broken) driver session so the
+                    // next iteration reopens cleanly.
+                    this.reset();
```
imbajin: ⚠️ Consider not calling `reset()` on each transient failure.

After `NoHostAvailableException`, the Datastax driver `Session` itself is still valid — it's the *nodes* that are unreachable, not the session object. The `ExponentialReconnectionPolicy` registered on the `Cluster` keeps retrying downed nodes in the background, so once Cassandra comes back, the **existing** session will route queries to the recovered nodes automatically.

Closing and reopening the session on every retry attempt adds the overhead of `cluster.connect(keyspace)` per attempt, which establishes a new connection pool. With the default 3 retries, that's up to 3 unnecessary reconnects during an outage.

Suggestion: simply retry `session.execute()` without `reset()`, and reserve `reset()` as a last-resort fallback only on the final attempt before giving up (or remove it from the retry loop entirely and rely on `reconnectIfNeeded()` for session-level recovery).

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -255,6 +391,56 @@ public boolean hasChanges() {
...
+            try {
+                if (this.session == null || this.session.isClosed()) {
+                    this.session = null;
+                    this.tryOpen();
+                }
+                if (this.session != null) {
+                    this.session.execute(new SimpleStatement(HEALTH_CHECK_CQL));
+                }
+            } catch (DriverException e) {
+                LOG.debug("Cassandra health-check failed, resetting session: {}",
```
imbajin, line 450: ⚠️ `DriverException` is the base class for **all** Cassandra driver exceptions, including `AuthenticationException`, `SyntaxError`, `InvalidQueryException`, etc. Catching it this broadly means non-transient errors (e.g., auth failure, bad CQL in the health check) will be silently swallowed and the session nullified.

Consider narrowing the catch to match `executeWithRetry()`:
```java
} catch (NoHostAvailableException | OperationTimedOutException e) {
```

hugegraph-server/hugegraph-cassandra/src/main/java/org/apache/hugegraph/backend/store/cassandra/CassandraSessionPool.java
```diff
@@ -174,6 +228,11 @@ public void commitAsync() {
             int processors = Math.min(statements.size(), 1023);
             List<ResultSetFuture> results = new ArrayList<>(processors + 1);
             for (Statement s : statements) {
+                // TODO(issue #2740): commitAsync() bypasses executeWithRetry().
+                // During a Cassandra restart, async writes may fail with
+                // NoHostAvailableException even when maxRetries > 0. Callers
+                // must handle CompletableFuture failures. A follow-up will
+                // wrap each future with retry semantics.
                 ResultSetFuture future = this.session.executeAsync(s);
```
imbajin: ⚠️ Worth tracking this gap explicitly. During a Cassandra restart, all async writes issued through `commitAsync()` will fail with `NoHostAvailableException`, and the failure is only observable through the `ResultSetFuture`. Callers that don't inspect futures will silently lose writes.

Consider opening a follow-up issue to track async retry support, so this TODO doesn't get forgotten.

dpol1:
Going through the feedback — NPE in `commitAsync()` when session is unavailable, TODO pointing to wrong type (`CompletableFuture` → Datastax `ResultSetFuture`), `reset()` not actually called on heatlh-check failure, catch too broad, `reset()` on every transient failure, idempotency check missing for `OperationTimedOutException` retries. Unit tests for all of it. Pushing soon.

Quick question before I do: worth renaming `cassandra.reconnect_max_retries` / `cassandra.reconnect_interval` to `cassandra.query_retry_max_attempts` / `cassandra.query_retry_interval`? The current names bleed into `ExponentialReconnectionPolicy` semantics. Keys are new in this PR so cost is low — but happy to leave it if you want to keep the diff minimal.

imbajin:
> Quick question before I do: worth renaming `cassandra.reconnect_max_retries` / `cassandra.reconnect_interval` to `cassandra.query_retry_max_attempts` / `cassandra.query_retry_interval`? The current names bleed into `ExponentialReconnectionPolicy` semantics. Keys are new in this PR so cost is low — but happy to leave it if you want to keep the diff minimal.

+1 on the rename. `query_retry_*` is much clearer — avoids confusion with the driver-level `ReconnectionPolicy`. Since these keys are new in this PR, no compatibility cost. Go for it.

imbajin (approved):
LGTM, THX

BTW, C*/*SQL will not be maintained after 1.5.0, the community focus on the RocksDB(standalone & cluster) instead

PR #3000, opened 2026-04-18
fix(docker): skip partition wait for standalone rocksdb mode

### LINKED ISSUE #2999: [Bug] Standalone Docker image stuck at "Waiting for partition assignment..." for 120s
### Bug Type

server status (startup / runtime exception)

### Before submit

### Environment

- Server Version: 1.7.0 (latest master)
- Backend: RocksDB (standalone, default)
- OS: macOS / Linux (Docker)
- Image: `hugegraph/hugegraph:latest` built from `hugegraph-server/Dockerfile`

**Expected:** The standalone image built from `hugegraph-server/Dockerfile` should start successfully and be ready to use immediately after `docker run`.

**Actual:** The server starts up (HTTP listener bound to 8080), but then keeps printing `Waiting for partition assignment...` for ~120 seconds before timing out:

```
INFO: Started listener bound to [0.0.0.0:8080]
INFO: [HttpServer] Started.
Started [pid 1017]
Waiting for partition assignment...
Waiting for partition assignment...
Waiting for partition assignment...
Waiting for partition assignment...
Waiting for partition assignment...
```

**Root Cause:**

PR #2952 (`0154c064`) refactored docker-compose and added an **unconditional** call to `wait-partition.sh` in `docker-entrypoint.sh`:

```bash
# Line 84-90 in docker-entrypoint.sh
STORE_REST="${STORE_REST:-store:8520}"
export STORE_REST
./bin/wait-partition.sh || log "WARN: partitions not assigned yet"
```

`wait-partition.sh` polls `http://store:8520` looking for `"partitionCount":[1-9]` — this is only meaningful in distributed hstore mode. In standalone rocksdb mode, there is no Store service, so it loops for 120 seconds before timing out.

In contrast, `wait-storage.sh` correctly guards against this: it skips the PD wait entirely when no hstore-related environment variables are set.

### Vertex/Edge example

_N/A — this is a Docker startup issue, not a data issue._

### Schema [VertexLabel, EdgeLabel, IndexLabel]

_N/A_

imbajin:
## Summary

Fixes #2999

- Standalone Docker image (`hugegraph-server/Dockerfile`, default `rocksdb` backend) gets stuck printing `Waiting for partition assignment...` for 120s after startup
- Root cause: `wait-partition.sh` was called unconditionally in `docker-entrypoint.sh` — it polls `http://store:8520` which only exists in distributed `hstore` mode
- Fix: read the actual `backend` value from `hugegraph.properties` and only run `wait-partition.sh` when `backend=hstore`

## Test plan

- [x] Built patched image, ran standalone container — no more "Waiting for partition" messages
- [x] Verified API responds normally (`/versions`, `/graphs`)
- [ ] Verify distributed mode (docker-compose with hstore) still works correctly with the guard

### DIFF
--- a/hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
+++ b/hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
@@ -84,7 +84,7 @@ fi
 ./bin/start-hugegraph.sh -j "${JAVA_OPTS:-}" -t 120
 
 # Post-startup cluster stabilization check (hstore only — rocksdb has no partitions)
-ACTUAL_BACKEND=$(grep -E '^\s*backend\s*=' "${GRAPH_CONF}" | sed 's/.*=\s*//' | tr -d ' ')
+ACTUAL_BACKEND=$(grep -E '^[[:space:]]*backend[[:space:]]*=' "${GRAPH_CONF}" | head -n 1 | sed 's/.*=//' | tr -d '[:space:]' || true)
 if [[ "${ACTUAL_BACKEND}" == "hstore" ]]; then
     STORE_REST="${STORE_REST:-store:8520}"
     export STORE_REST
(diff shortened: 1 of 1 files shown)


Pengzna (approved):
LGTM

PR #3008, opened 2026-04-24
feat(server): adapt Hubble 2.0 & add graph/role management

Yeaury:
## Purpose of the PR

Adapt the Server APIs needed by Hubble 2.0 so the community Server can support
Hubble's graph, GraphSpace, default graph, default role, schema template, and
manager flows.

Hubble2 will use only the canonical APIs listed below. Since Hubble2 has not
been released, legacy form-based APIs and state-changing GET aliases are not
compatibility requirements and should be removed before release.

Hubble2 reference: https://github.com/hugegraph/hugegraph-toolchain/pull/4

## Main Changes

- Graph management (`GraphsAPI.java`)
- GraphSpace and default-role APIs (`GraphSpaceAPI.java`)
- GraphSpace managers (`ManagerAPI.java`)
- Schema templates (`SchemaTemplateAPI.java`)
- Authentication and graph metadata support

## Integration Boundaries

- Hubble2 uses canonical graph create/default APIs:
  - `POST /graphspaces/{graphspace}/graphs/{name}`
  - `POST /graphspaces/{graphspace}/graphs/{name}/default`
  - `DELETE /graphspaces/{graphspace}/graphs/{name}/default`
- Server and Hubble should remove legacy form-urlencoded graph creation and
  state-changing GET aliases before Hubble2 is released.
- GraphSpace lifecycle, default role, schema template, and manager APIs are
  PD-mode APIs. Standalone/non-PD Hubble should hide or degrade these flows
  instead of requiring Server to emulate the full PD GraphSpace model.
- Hubble2 non-PD flows should continue to use `DEFAULT` GraphSpace and graph
  operations that are meaningful in standalone mode.
- Graph clear must remain user-oriented and expose two real operations:
  data-only clear that preserves schema, and schema-and-data clear. Server must
  implement and verify both semantics before Hubble wires both user actions.

## Known Integration Follow-ups

- Current PD smoke found `GET /graphspaces/DEFAULT/graphs` returning an empty
  list while direct `GET /graphspaces/DEFAULT/graphs/hugegraph` succeeds.
- Hubble Gremlin overview can still show `max` counts when Server-side Gremlin
  execution does not provide the expected `g` alias.
- Default-role checks still need a full Hubble PD workflow smoke after the API
  boundary is finalized.

## Server follow-up status

- [x] Require `memory_monitor.period` to be positive, while keeping
  `memory_monitor.threshold=1.0` as the disable switch.
- [x] Roll back MemoryMonitor, Gremlin, and REST resources if MemoryMonitor
  construction or startup fails.
- [x] Validate GraphSpace and `OBSERVER` graph existence in the manager
  default-role check.
- [x] Pass the selected Maven backend profile into the test JVM and remove the
  nonexistent singular `auth/manager/default` test route.
- [ ] Implement two explicit graph-clear operations in Server: data-only clear
  that preserves schema, and schema-and-data clear. Define canonical APIs,
  confirmation/error semantics, and verify behavior across supported backends.
- [ ] Remove the Server public GET default-graph mutation aliases before
  Hubble2 is released; retain only the canonical POST/DELETE operations.

## Hubble migration status

- [x] Use the canonical JSON graph-create resource path.
- [x] Use canonical `POST`/`DELETE` methods to set and unset the default graph.
- [x] Read the default graph through the canonical Server endpoint.
- [x] Align default-role calls and handle the empty `204` delete response.
- [x] Align schema-template and graph-profile contracts.
- [x] Use the canonical Server `PUT` contract for backend-to-Server graph
  metadata updates; the legacy Hubble GET `/update` route is removed.
- [x] Remove the form-urlencoded graph-create facade and legacy default-role
  facades; keep only canonical resource APIs.
- [x] Normalize the default-graph read to `/graphs/default` and remove Hubble
  GET default-graph mutation aliases.
- [x] Make graph clear confirmation explicit and safe: show the graph name and
  deletion scope, warn that it is irreversible, and require strong confirmation.
- [x] Stop Hubble from using per-graph reload. The shared Client overload stays
  for now, but is explicitly deprecated because Server accepts only `update`.
- [x] Hide the default-GraphSpace mutation; Hubble no longer exposes its own GET
  mutation facade.

## Hubble TODOs

- [ ] After Server implements both graph-clear modes, wire the two confirmed
  Hubble actions to non-GET operations: data-only clear that preserves schema,
  and schema-and-data clear. Keep both user-facing actions and add UI smoke
  coverage proving the invoked mode and deletion scope.
- [ ] Complete the outstanding PD integration smoke for graph list/lifecycle,
  default-role/account/role, loader peers, algorithms, and async tasks once the
  Server API boundary and test environment are available.

## Remaining design decisions

- [x] Hubble2 is unreleased and does not require legacy API compatibility.
- [ ] Decide whether default-role owner input needs an explicit user/group type
  when a user and group can share the same name.
- [ ] Decide the canonical Server contract for default GraphSpace before the
  feature is exposed by Hubble.

## Verifying these changes

- [x] Server unit coverage for the master check and default-role permission
  boundaries.
- [x] Server regression coverage for MemoryMonitor configuration/startup
  rollback, manager resource validation, and backend-profile propagation.
- [x] Hubble2 non-PD smoke for graph management, default graph APIs, and
  canonical graph create path.
- [x] Hubble2 packaged loader/auth smoke with current toolchain loader/client.
- [x] Hubble2 PD minimal API smoke for versions, GraphSpace list/get, direct
  graph get, schema template list, and Gremlin schema/data read-write.
- [ ] Server tests for data-only and schema-and-data clear across the supported
  standalone and distributed backends.
- [ ] Hubble UI smoke proving that each clear action invokes the correct Server
  mode and that its confirmation accurately describes the deletion scope.
- [ ] Hubble2 PD graph list/lifecycle smoke.
- [ ] Hubble2 PD default-role/account/role smoke.
- [ ] Hubble2 PD loader peers, algorithms, and async task smoke.


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/SchemaTemplateAPI.java
```diff
@@ -0,0 +1,199 @@
+/*
+ * Copyright 2017 HugeGraph Authors
+ *
```
imbajin: Keep consistent with the License Header

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -136,6 +223,61 @@ public Object get(@Context GraphManager manager,
         return ImmutableMap.of("name", g.name(), "backend", g.backend());
     }
 
+    @GET
```
imbajin: ‼️ **Critical: Using `GET` for state-changing operations violates REST semantics**

Both `setDefault` and `unsetDefault` modify server state (they call `authManager.setDefaultGraph` / `unsetDefaultGraph`) but use `@GET`. This is a serious REST anti-pattern that can cause:

- **Browser/proxy prefetching** triggering unintended state changes
- **HTTP caches** serving stale responses instead of forwarding the request
- **Crawlers/link scanners** accidentally modifying user default graphs

These should use `@POST` (or `@PUT`) and `@DELETE` (or `@POST`) respectively:

```suggestion
    @POST
    @Timed
    @Path("{name}/default")
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/SchemaTemplateAPI.java
```diff
@@ -0,0 +1,197 @@
...
+    @DELETE
+    @Timed
+    @Path("{name}")
+    @Produces(APPLICATION_JSON_WITH_CHARSET)
+    public void delete(@Context GraphManager manager,
+                       @PathParam("graphspace") String graphSpace,
+                       @PathParam("name") String name,
+                       @Context SecurityContext sc) {
+        LOG.debug("Remove schema template by name '{}' for graph space",
+                  name, graphSpace);
```
imbajin, line 127: ‼️ **Bug: LOG format string has one `{}` placeholder but two arguments**

The format string `"Remove schema template by name '{}' for graph space"` has only one `{}`, so `graphSpace` is silently ignored and won't appear in the log output.

```suggestion
        LOG.debug("Remove schema template by name '{}' for graph space '{}'",
                  name, graphSpace);
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/ConfigUtil.java
```diff
@@ -188,4 +192,30 @@ private static void validateGraphName(String graphName) {
...
+            if (config.file() == null) {
+                Map<String, Object> configMap = new HashMap<>();
+                Iterator<String> iterator = config.getKeys();
+                while (iterator.hasNext()) {
+                    String key = iterator.next();
+                    configMap.put(key, config.getProperty(key));
+                }
+                content = JsonUtil.toJson(configMap);
+            } else {
+                File file = config.file();
```
imbajin: ‼️ **Bug: Duplicate null check — the `else` branch already guarantees `config.file() != null`**

The outer `if (config.file() == null)` at line 198 means the `else` branch is only reached when `config.file()` is non-null. The inner `if (file == null)` check on this line is dead code and can never be true.

```suggestion
                File file = config.file();
                content = FileUtils.readFileToString(file);
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/SchemaTemplateAPI.java
```diff
@@ -0,0 +1,197 @@
...
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.SecurityContext;
+
+@Path("graphspaces/{graphspace}/schematemplates")
+@Singleton
+@Tag(name = "SchemaTemplateAPI")
+public class SchemaTemplateAPI extends API {
+
```
imbajin, line 59: ⚠️ **Security: All SchemaTemplateAPI endpoints are missing `@RolesAllowed` annotations**

Other APIs in this project (`GraphsAPI`, `GraphSpaceAPI`) consistently use `@RolesAllowed` for access control. The `delete` and `update` methods do manual permission checks via `isSpaceManager`, but `list`, `get`, and `create` have no access control at all. This means any authenticated user can create and view schema templates regardless of their role.

Consider adding appropriate `@RolesAllowed` annotations (e.g., `@RolesAllowed({"space_member", "$dynamic"})` for read operations, `@RolesAllowed({"space"})` for write operations) to be consistent with the rest of the codebase.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -120,6 +128,85 @@ public Object list(@Context GraphManager manager,
...
+                    profiles.add(profile);
+                }
+            } catch (ForbiddenException ignored) {
+                // ignore graphs the current user has no access to
+            }
+        }
+        defaultProfiles.addAll(profiles);
+        return defaultProfiles;
+    }
+
```
imbajin, line 224: ⚠️ **Code quality: `isPrefix` should be `private static`, and `nickname` can be null causing NPE**

1. This method is only used within `listProfile` — it should be `private static`, not `public`.
2. `hg.nickname()` can return `null` (graphs may not have a nickname set). When `nickname` is null, `nickname.toString().startsWith(prefix)` will throw `NullPointerException`.

```suggestion
    private static boolean isPrefix(Map<String, Object> profile, String prefix) {
        if (StringUtils.isEmpty(prefix)) {
            return true;
        }
        String name = profile.get("name").toString();
        Object nickname = profile.get("nickname");
        return name.startsWith(prefix) ||
               (nickname != null && nickname.toString().startsWith(prefix));
    }
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -155,6 +297,76 @@ public void drop(@Context GraphManager manager,
...
+                String graphName = (String) graphMap.get("name");
+                E.checkArgument(graphName != null && graphName.equals(name),
+                                "Different name in update body '%s' with path '%s'",
+                                graphName, name);
+                HugeGraph exist = graph(manager, graphSpace, name);
+                String nickname = (String) graphMap.get("nickname");
+                if (!Strings.isEmpty(nickname)) {
+                    GraphManager.checkNickname(nickname);
+                    E.checkArgument(!manager.isExistedGraphNickname(graphSpace, nickname) ||
+                                    nickname.equals(exist.nickname()),
```
imbajin: ⚠️ **Large commented-out code block should be removed**

~15 lines of commented-out code (`GRAPH_ACTION_CLEAR` and `GRAPH_ACTION_RELOAD` cases) should not be checked in. If these features are planned for a future PR, track them in an issue instead. The unused constants `CLEAR_SCHEMA` and `GRAPH_ACTION_CLEAR` defined earlier should also be removed since they're not referenced by active code.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -155,6 +296,60 @@ public void drop(@Context GraphManager manager,
...
+                                    "Nickname '%s' has already existed in graphspace '%s'",
+                                    nickname, graphSpace);
+                    exist.nickname(nickname);
```
imbajin: ‼️ **Bug: nickname 更新未持久化，且 `isExistedGraphNickname` 在非 PD 模式下会 NPE**

两个问题：

1. `exist.nickname(nickname)` 只修改了内存中的 `StandardHugeGraph.nickname` 字段，没有调用 `metaManager.updateGraphConfig()` 写回存储。创建图时有完整的持久化链路（`GraphManager:1345` 设内存 → `:1353` 写 metaManager），但这里的更新路径缺少持久化步骤，重启后 nickname 会丢失。

2. `manager.isExistedGraphNickname()` 内部调用 `metaManager.graphConfigs(graphSpace)` — 在非 PD（RocksDB 单机）模式下 MetaManager 未初始化，会抛 NPE。GraphsAPI 的端点在非 PD 模式下通过 `graphspace=DEFAULT` 正常使用，所以这里需要加 `isPDEnabled()` 分支做兼容（可参考 `GraphManager.graphs()` 的处理方式）。

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -207,11 +402,14 @@ public Object create(@Context GraphManager manager,
         if (StringUtils.isEmpty(clone)) {
             // Only check required parameters when creating new graph, not when cloning
             E.checkArgument(configs != null, "Config parameters cannot be null");
-            String[] requiredKeys = {"backend", "serializer", "store"};
-            for (String key : requiredKeys) {
-                Object value = configs.get(key);
-                E.checkArgument(value instanceof String && !StringUtils.isEmpty((String) value),
-                                "Required parameter '%s' is missing or empty", key);
+            // Auto-fill defaults for PD/HStore mode when not provided
+            configs.putIfAbsent("backend", "hstore");
```
imbajin: ‼️ **Bug: auto-fill `backend=hstore` 会导致非 PD 模式创建图失败**

GraphsAPI 的创建端点在非 PD（RocksDB 单机）模式下也会被调用（通过 `graphspace=DEFAULT`）。这里无条件填充 `backend=hstore` 会导致单机版用户创建图时使用错误的后端，应该加 PD 模式判断：

```suggestion
            // Auto-fill defaults for PD/HStore mode when not provided
            if (manager.isPDEnabled()) {
                configs.putIfAbsent("backend", "hstore");
                configs.putIfAbsent("serializer", "binary");
            }
            configs.putIfAbsent("store", name);
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/util/ConfigUtil.java
```diff
@@ -188,4 +192,18 @@ private static void validateGraphName(String graphName) {
...
+    public static String writeConfigToString(HugeConfig config) {
+        // Always serialize config to a JSON object by iterating all keys,
+        // regardless of whether the config was loaded from a file or not.
+        // This ensures callers (e.g. GraphsAPI.listProfile) can safely parse
+        // the result via JsonUtil.fromJson() without format inconsistencies.
+        Map<String, Object> configMap = new HashMap<>();
+        Iterator<String> iterator = config.getKeys();
+        while (iterator.hasNext()) {
+            String key = iterator.next();
+            configMap.put(key, config.getProperty(key));
```
imbajin, line 209: ‼️ **Security: 序列化全部配置项，可能泄露敏感信息**

`config.getKeys()` 会遍历所有配置键值对，包括可能包含密码、token 等敏感字段（如 `gremlin.server.password`、认证相关配置等）。该方法的返回值会通过 `GraphsAPI.listProfile()` 返回给前端展示。

建议做白名单过滤（只输出前端需要的字段），或者至少排除包含 `password`、`secret`、`token`、`credential` 等关键词的 key。

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/ManagerAPI.java
```diff
@@ -259,6 +262,41 @@ public String getRolesInGs(@Context GraphManager manager,
...
+        ensurePdModeEnabled(manager);
+        AuthManager authManager = manager.authManager();
+        String user = HugeGraphAuthProxy.username();
+
+        E.checkArgument(StringUtils.isNotEmpty(role) &&
+                        StringUtils.isNotEmpty(graphSpace),
+                        "Must pass graphspace and role params");
+
+        HugeDefaultRole defaultRole =
+                HugeDefaultRole.valueOf(role.toUpperCase());
```
imbajin: ‼️ **Bug: `HugeDefaultRole.valueOf()` 未做异常处理，非法 role 值返回 500**

`valueOf(role.toUpperCase())` 在 role 值非法时会抛出 `IllegalArgumentException`，变成 HTTP 500 而不是 400。同一个 PR 中 `GraphSpaceAPI.setDefaultRole()` 已经用 try-catch 做了处理，这里应该保持一致：

```suggestion
        HugeDefaultRole defaultRole;
        try {
            defaultRole = HugeDefaultRole.valueOf(role.toUpperCase());
        } catch (IllegalArgumentException e) {
            E.checkArgument(false, "Invalid role value '%s'", role);
            defaultRole = null; // unreachable
        }
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -120,6 +126,86 @@ public Object list(@Context GraphManager manager,
...
+                }
+            } catch (ForbiddenException ignored) {
+                // ignore graphs the current user has no access to
+            }
+        }
+        defaultProfiles.addAll(profiles);
+        return defaultProfiles;
+    }
+
+    private static boolean isPrefix(Map<String, Object> profile, String prefix) {
```
imbajin: ⚠️ **重复代码: `isPrefix` 在 GraphsAPI 和 GraphSpaceAPI 中有两份相同实现**

`GraphSpaceAPI` 中也定义了同样签名和逻辑的 `isPrefix` 方法。建议提取到公共基类 `API` 或工具类中，避免后续维护中两处不一致。

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -155,6 +296,60 @@ public void drop(@Context GraphManager manager,
...
+    @RolesAllowed({"space"})
+    public Map<String, String> manage(@Context GraphManager manager,
+                                      @Parameter(description = "The graph space name")
+                                      @PathParam("graphspace") String graphSpace,
+                                      @Parameter(description = "The graph name")
+                                      @PathParam("name") String name,
+                                      @Parameter(description = "Action map: {'action':'update','update':{...}}")
+                                      Map<String, Object> actionMap) {
+        LOG.debug("Manage graph '{}' with action '{}'", name, actionMap);
+        E.checkArgument(actionMap != null && actionMap.size() == 2 &&
```
imbajin: ⚠️ **校验过严: `actionMap.size() == 2` 会在前端多传字段时直接 400**

要求请求体恰好包含 2 个 key。如果前端在 JSON 中额外带了字段（很常见的兼容场景），请求会被拒绝。建议放宽为只校验必需字段：

```suggestion
        E.checkArgument(actionMap != null &&
                        actionMap.containsKey(GRAPH_ACTION),
                        "Invalid request body '%s'", actionMap);
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManager.java
```diff
@@ -813,6 +814,179 @@ public HugeGroup findGroup(String name) {
         return null;
     }
 
+    private static final String DEFAULT_GRAPH_MARKER = "~default_graph";
```
imbajin, line 850: ⚠️ **建议加注释说明 marker group 设计的上下文**

使用 `~default_graph:` 和 `~default_role:` 前缀的特殊 group 来模拟默认图/角色关系是一种基于现有机制的 workaround。当前方案可工作，但有以下已知限制：
- 这些 marker group 会出现在 `listGroups` 结果中
- Belong ID 拼接格式（`userId + "->ug->" + groupId`）依赖内部约定

建议在此处加几行注释说明设计背景和已知限制，方便后续维护者理解。未来可考虑引入独立的 default graph/role 存储机制来替代。

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -120,6 +126,86 @@ public Object list(@Context GraphManager manager,
...
+            throw new HugeException("Graphspace not exist!");
+        }
+        GraphSpace gs = manager.graphSpace(graphSpace);
+        String gsNickname = gs.nickname();
+
+        AuthManager authManager = manager.authManager();
+        String user = HugeGraphAuthProxy.username();
+        Map<String, Date> defaultGraphs = authManager.getDefaultGraph(graphSpace, user);
+
+        SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
imbajin: ⚠️ **建议使用 `DateTimeFormatter` 替代 `SimpleDateFormat`**

`SimpleDateFormat` 虽然这里是局部变量（线程安全），但 `DateTimeFormatter` 是 Java 8+ 推荐的替代方案，线程安全且可复用为 static final 常量，性能更好：

```java
private static final DateTimeFormatter DATE_FORMAT =
        DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
```

imbajin:
## 新增 API 单机版（非 PD）兼容性对照

目标：社区默认的 RocksDB 单机版通过 `graphspace=DEFAULT` 使用 GraphsAPI，应支持图的增删改查和基础权限管理。GraphSpaceAPI、ManagerAPI、SchemaTemplateAPI 属于 PD 专属功能，不需要兼容。

### 需要兼容单机版的端点（GraphsAPI）

| API 端点 | 当前状态 | 问题说明 | 需要的改动 |
|---------|:---:|---------|----------|
| `GET .../graphs/profile` | ⚠️ | `ConfigUtil.writeConfigToString` 序列化全部配置项，可能泄露 password/token 等敏感字段 | 白名单过滤或排除敏感 key |
| `PUT .../graphs/{name}` (manage/update nickname) | ❌ | 1. `isExistedGraphNickname()` 调用 `metaManager.graphConfigs()` — 非 PD 下 MetaManager 未初始化，**NPE**；2. `exist.nickname()` 只改内存，缺少持久化调用 | 1. 加 `isPDEnabled()` 分支，非 PD 下用内存图列表做去重；2. 补充持久化调用 |
| `POST .../graphs` (create JSON) | ❌ | 无条件执行 `configs.putIfAbsent("backend", "hstore")`，单机版默认 RocksDB 会被覆盖为 hstore 导致**创建失败** | 用 `if (manager.isPDEnabled())` 包裹 hstore/binary 默认值填充 |
| `POST .../graphs/{name}` (createByText) | ❌ | 委托 `create()` 方法，同上 | 同上（修复 create 即可） |
| `POST .../graphs/{name}/default` (设置默认图) | ✅ | StandardAuthManager 已实现 | 无 |
| `DELETE .../graphs/{name}/default` (取消默认图) | ✅ | StandardAuthManager 已实现 | 无 |
| `GET .../graphs/default` (查询默认图) | ✅ | StandardAuthManager 已实现 | 无 |

### PD 专属端点（无需兼容单机版）

| API 端点 | 归属 | 说明 |
|---------|------|------|
| `POST/GET/DELETE .../graphspaces/{gs}/role` | GraphSpaceAPI | 图空间级角色管理，PD-only by design |
| `GET /auth/manager/default` | ManagerAPI | 已有 `ensurePdModeEnabled()` 门禁 |
| `GET/POST/PUT/DELETE .../schematemplates[/{name}]` | SchemaTemplateAPI | 全部走 metaManager，PD-only |

**总结**：7 个 GraphsAPI 端点需要兼容单机版。其中 **3 个需要代码改动**（manage 需加 `isPDEnabled()` 分支 + 持久化修复，create/createByText 需条件化 hstore 默认值），1 个需过滤敏感配置，3 个已可正常工作。

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -120,6 +128,102 @@ public Object list(@Context GraphManager manager,
...
+    private static boolean isPrefix(Map<String, Object> profile, String prefix) {
+        if (StringUtils.isEmpty(prefix)) {
+            return true;
+        }
+        // graph name or nickname is not empty
+        String name = profile.get("name").toString();
+        Object nicknameObj = profile.get("nickname");
+        String nickname = nicknameObj != null ? nicknameObj.toString() : "";
+        return name.startsWith(prefix) || nickname.startsWith(prefix);
+    }
```
imbajin: ‼️ **Blocker: duplicate `isPrefix()` now breaks compilation**

`API` already defines a `protected static isPrefix(Map<String, Object>, String)`, so this `private static` method in `GraphsAPI` attempts to hide it with weaker access privileges and causes `hugegraph-api` compilation to fail. Since both implementations are now equivalent, please remove this duplicate method and reuse the inherited helper.

```suggestion
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManagerV2.java
```diff
@@ -1510,6 +1516,195 @@ public HugeGroup findGroup(String name) {
...
+    public void setDefaultGraph(String graphSpace, String graph, String user) {
+        try {
+            HugeBelong belong = new HugeBelong(graphSpace,
+                                               IdGenerator.of(user),
+                                               IdGenerator.of(graph +
+                                                              DEFAULT_SETTER_ROLE_KEY));
+            this.tryInitDefaultGraph(graphSpace, graph);
+            this.updateCreator(belong);
+            belong.create(belong.update());
+            this.metaManager.createBelong(graphSpace, belong);
```
imbajin, line 1536: ‼️ **Bug: setting the same default graph twice fails in PD mode**

`MetaManager.createBelong()` rejects an existing belong record, so repeated `POST /graphs/{name}/default` calls for the same user/graph will be wrapped as `HugeException` instead of behaving idempotently. `unsetDefaultGraph()` already checks `existBelong()` first; `setDefaultGraph()` should do the same before creating the belong binding.

```suggestion
            String role = graph + DEFAULT_SETTER_ROLE_KEY;
            String belongId = this.metaManager.belongId(user, role);
            if (!this.metaManager.existBelong(graphSpace, IdGenerator.of(belongId))) {
                this.metaManager.createBelong(graphSpace, belong);
            }
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManagerV2.java
```diff
@@ -1510,6 +1516,195 @@ public HugeGroup findGroup(String name) {
...
+            } else {
+                belong = new HugeBelong(graphSpace, IdGenerator.of(owner),
+                                        null, IdGenerator.of(roleName),
+                                        HugeBelong.UR);
+            }
+
+            this.tryInitDefaultRole(graphSpace, roleName, graph);
+            this.updateCreator(belong);
+            belong.create(belong.update());
+            Id result = this.metaManager.createBelong(graphSpace, belong);
```
imbajin, line 1614: ‼️ **Bug: default role creation is not idempotent in PD mode**

This has the same `createBelong()` issue as default graph: assigning the same default role twice to the same user/group will fail because the belong record already exists. Hubble-side retries or duplicate clicks should not turn an already-set role into a server error. Please check `existBelong()` before create, or update/no-op when the binding is already present.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -155,6 +329,61 @@ public void drop(@Context GraphManager manager,
...
+        String action = (String) value;
+        switch (action) {
+            case UPDATE:
+                E.checkArgument(actionMap.containsKey(UPDATE),
+                                "Please pass '%s' for graph update",
+                                UPDATE);
+                value = actionMap.get(UPDATE);
+                E.checkArgument(value instanceof Map,
+                                "The '%s' must be map, but got %s",
+                                UPDATE, value.getClass());
```
imbajin: ⚠️ **Bug: null `update` body still turns validation into NPE**

If the request body is `{"action":"update","update":null}`, `value instanceof Map` is false, but the error message still evaluates `value.getClass()`, causing a 500 instead of a clear 4xx validation error. Please make this message null-safe, similar to the `action` validation above.

```suggestion
                E.checkArgument(value instanceof Map,
                                "The %s must be map, but got %s",
                                UPDATE,
                                value == null ? "null" : value.getClass().getSimpleName());
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/ManagerAPI.java
```diff
@@ -259,6 +262,46 @@ public String getRolesInGs(@Context GraphManager manager,
                                 result));
     }
 
+    @GET
+    @Timed
+    @Path("default")
+    @Consumes(APPLICATION_JSON)
+    public String checkDefaultRole(@Context GraphManager manager,
+                                   @QueryParam("graphspace") String graphSpace,
```
imbajin: ⚠️ **Bug: this endpoint ignores the path graphspace and trusts query `graphspace` instead**

`ManagerAPI` is already mounted under `graphspaces/{graphspace}/auth/managers`, but this method reads `graphspace` from the query string. A request can therefore hit one graphspace path while checking another graphspace via query params, which is surprising and easy to misuse. Please use the existing path parameter here (or move the API to a non-graphspace route if the intended contract is global).

```suggestion
                                   @PathParam("graphspace") String graphSpace,
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -2174,6 +2199,43 @@ public Map<String, Object> graphConfig(String graphSpace,
...
+            if (configs != null) {
+                configs.put("nickname", nickname);
+                this.metaManager.updateGraphConfig(graphSpace, graphName, configs);
+                this.metaManager.notifyGraphUpdate(graphSpace, graphName);
+            }
+        } catch (Exception e) {
+            // Roll back the in-memory change so that the runtime state stays
+            // consistent with what was actually persisted.
+            if (g != null) {
+                g.nickname(null);
```
imbajin: ⚠️ **Bug: failed PD persistence rolls back the nickname to `null` instead of the previous value**

`updateGraphNickname()` mutates the in-memory graph before writing PD meta. If `updateGraphConfig()` or `notifyGraphUpdate()` fails, the catch block sets the nickname to `null`, which can leave the running server in a third state that matches neither the requested value nor the persisted value. Please capture the old nickname before mutation and restore that value on failure.

```suggestion
        String oldNickname = g != null ? g.nickname() : null;
        if (g != null) {
            g.nickname(nickname);
        }
```

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -136,6 +240,76 @@ public Object get(@Context GraphManager manager,
...
+    @RolesAllowed({"space_member", "$owner=$name"})
+    public Map<String, Object> setDefault(@Context GraphManager manager,
+                                          @Parameter(description = "The graph space name")
+                                          @PathParam("graphspace") String graphSpace,
+                                          @Parameter(description = "The graph name")
+                                          @PathParam("name") String name) {
+        LOG.debug("Set default graph '{}' in graph space '{}'", name, graphSpace);
+        E.checkArgument(manager.graph(graphSpace, name) != null,
+                        "Graph '%s/%s' does not exist", graphSpace, name);
+        // Default graph persistence requires PD meta storage.
```
imbajin: ⚠️ **Behavior regression: standalone mode silently ignores default graph changes**

The earlier standalone-compatibility goal was that `POST/DELETE/GET .../graphs/default` should work through `StandardAuthManager`. This branch now returns an empty set in non-PD mode, so Hubble can call the API without an exception but the user preference is never saved or reflected by `GET /graphs/default` / `GET /graphs/profile`. If default graph is expected in standalone RocksDB, please route this through the existing `AuthManager` implementation instead of making it a no-op; otherwise the endpoint should be documented as PD-only so the frontend can avoid treating the empty response as success.

hugegraph-server/hugegraph-dist/src/main/java/org/apache/hugegraph/dist/HugeGraphServer.java
```diff
@@ -63,18 +63,33 @@ public HugeGraphServer(String gremlinServerConf, String restServerConf)
...
+        } catch (Throwable e) {
+            LOG.error("HugeGremlinServer prepare error: ", e);
+            throw e;
+        }
+
         try {
-            // Start HugeRestServer
+            // Start HugeRestServer (loads graphs from PD; events go to
+            // ContextGremlinServer which is already listening)
             this.restServer = HugeRestServer.start(restServerConf, hub);
```
imbajin: ‼️ **Blocker: startup failures can leak process-wide state and the prepared Gremlin server**

After `System.setSecurityManager(null)`, the original security manager is only restored in the final `startPrepared()` `finally` block. If `HugeGremlinServer.prepare(...)` or `HugeRestServer.start(...)` throws, the constructor exits before that `finally`, leaving the process-wide `SecurityManager` as `null`. Also, once `prepare()` succeeds, a later `HugeRestServer.start(...)` failure rethrows without stopping the prepared Gremlin server, so the listeners/resources created during preparation are left behind.

Please wrap the whole startup sequence after disabling the security manager in an outer `try/finally` that always restores it, and stop the prepared Gremlin server when REST startup fails before `startPrepared()` is reached.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManagerV2.java
```diff
@@ -1510,6 +1516,211 @@ public HugeGroup findGroup(String name) {
...
+
+    @Override
+    public Id createDefaultRole(String graphSpace, String owner,
+                                HugeDefaultRole role, String graph) {
+        String roleName = (role.isGraphRole()) ?
+                getGraphDefaultRole(graph, role.toString()) : role.toString();
+        try {
+            HugeBelong belong;
+            String link;
+            if (HugeGroup.isGroup(owner)) {
```
imbajin: ‼️ **Blocker: group default-role assignments are stored as user bindings**

`GraphSpaceAPI.setDefaultRole()` accepts either a user or a group (`findUser(user) != null || findGroup(user) != null`), but this implementation decides whether to create a group binding by calling `HugeGroup.isGroup(owner)`. That helper only checks whether the string starts with `group-`; normal group names such as `ops` are stored and looked up by name, so they take the user-binding (`UR`) branch instead of the group-binding (`GR`) branch.

This makes `POST /graphspaces/{gs}/role` appear to succeed for a group name, but members of that group will not inherit the default role. Please determine the owner type from the actual auth entity lookup, or pass an explicit user/group distinction into the auth layer instead of inferring it from the string prefix.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -103,6 +107,161 @@ public Object get(@Context GraphManager manager,
...
+        E.checkArgument(!hasGraph || StringUtils.isNotEmpty(graph),
+                        "Must set a graph for observer");
+
+        Map <String, String> result = new HashMap<>();
+        result.put("user", user);
+        result.put("role", jsonRole.role);
+        result.put("graphSpace", name);
+
+        if (hasGraph) {
+            authManager.createDefaultRole(name, user, role, graph);
```
imbajin, line 163: ⚠️ **Validate graphspace and observer graph existence before default-role operations**

These role handlers do not verify that the path graphspace exists before calling the auth manager, and for `OBSERVER` they only check that `graph` is non-empty before creating a graph-level default role. In PD mode, a typo in the graphspace or graph name can create role/target/access/belong metadata under an arbitrary key, and that metadata may become effective later if a graphspace or graph with that name is created.

Please fail early with the same graphspace validation used by other graphspace-scoped APIs, and validate the observer graph belongs to this graphspace before create/check/delete.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -155,6 +316,62 @@ public void drop(@Context GraphManager manager,
...
+                                    nickname.equals(exist.nickname()),
+                                    "Nickname '%s' has already existed in graphspace '%s'",
+                                    nickname, graphSpace);
+                    // Delegate to GraphManager: handles both in-memory update and
+                    // PD-mode persistence (non-PD mode is in-memory only).
+                    manager.updateGraphNickname(graphSpace, name, nickname);
+                }
+                return ImmutableMap.of(name, "updated");
+            default:
+                throw new AssertionError(String.format(
```
imbajin: ⚠️ **Return 400 for malformed manage requests instead of 500**

This endpoint handles client-provided JSON, but an unknown `action` currently throws `AssertionError`, which maps to a server error even though the request is simply invalid. The same method also casts `graphMap.get("name")` and `graphMap.get("nickname")` directly to `String`; values like `123` will throw `ClassCastException` and surface as a 500.

Please use `E.checkArgument(false, ...)` for unsupported actions and validate field types before casting so malformed client requests consistently return a 4xx response.

imbajin (commented):
Found a few remaining issues that can affect startup cleanup, default-role correctness, and API error semantics. I also added +1 reactions to existing Copilot comments for duplicate findings instead of reposting them.

imbajin:
Follow-up TODOs that are not necessarily blockers for this PR, but are worth tracking separately if they are not handled here:

1. Consider using one documented marker convention for default graph/default role metadata. V1 uses a reserved `~default_*` prefix while V2 uses the `_DEFAULT_SETTER_ROLE` suffix; the suffix form can collide with user-defined role names that happen to end with that marker.
2. Consider splitting the default graph/default role storage workaround into a follow-up issue. The marker-group approach is now documented, but it still leaks implementation details into auth entities and may deserve a dedicated storage model later.
3. Consider replacing `role.equals(HugeDefaultRole.OBSERVER)` checks with `role.isGraphRole()` so future graph-scoped default roles do not need to touch all API branches again.
4. `JsonDefaultRole` implements `Checkable`, but `checkCreate()` and `checkUpdate()` are empty while validation is inline in the API method. Either move validation into the DTO or drop the interface to avoid a misleading contract.
5. Rename `@PathParam("graphspace") String name` to `graphSpace` in the default-role methods for readability and consistency with the surrounding APIs.
6. `GraphsAPI.listProfile()` still calls `manager.graphSpace(graphSpace)` twice and the diff has a few trailing-whitespace lines. These are low-risk cleanups if another revision is already planned.
7. The response ordering in `GraphsAPI.listProfile()` appears to put default profiles first, then other profiles. If the frontend depends on this, please document or explicitly sort the response.

Some broader test gaps are already covered by existing review comments, so I am not repeating them here. If these TODOs are out of scope for this PR, they can be split into separate tracking issues.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManager.java
```diff
@@ -813,6 +814,197 @@ public HugeGroup findGroup(String name) {
...
+
+    private Id createBelongBinding(String owner, Id groupId) {
+        HugeBelong existing = findBelongBinding(owner, groupId);
+        if (existing != null) {
+            return existing.id();
+        }
+        Id userId = IdGenerator.of(owner);
+        HugeBelong belong = new HugeBelong(userId, groupId);
+        belong.creator(owner);
+        return this.belong.add(belong);
```
imbajin: ‼️ **Invalidate the user cache after V1 default graph/role mutations**

`createBelongBinding()` and `removeBelongBinding()` mutate the auth relationship store, but unlike the other mutating methods in this class they never call `invalidateUserCache()`. This can leave permission/default-role checks using stale cached auth data after `setDefaultGraph()`, `unsetDefaultGraph()`, `createDefaultRole()`, or `deleteDefaultRole()` until the cache naturally expires.

Please invalidate the cache after adding or deleting the belong binding. For example:

```suggestion
        Id id = this.belong.add(belong);
        this.invalidateUserCache();
        return id;
```

And mirror the same invalidation after `this.belong.delete(existing.id())` in `removeBelongBinding()`.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/GraphsAPI.java
```diff
@@ -220,15 +490,17 @@ public Object create(@Context GraphManager manager,
         if (StringUtils.isNotEmpty(clone)) {
             // Clone from existing graph
             LOG.debug("Clone graph '{}' to '{}' in graph space '{}'", clone, name, graphSpace);
-            graph = manager.cloneGraph(graphSpace, clone, name, convConfig(configs));
+            Map<String, Object> cloneConfigs = configs != null ? configs : new HashMap<>();
+            graph = manager.cloneGraph(graphSpace, clone, name, convConfig(cloneConfigs));
         } else {
             // Create new graph
             graph = manager.createGraph(graphSpace, name, creator,
                                         convConfig(configs), true);
```
imbajin, line 517: ⚠️ **Validate config values before stringifying them**

`entry.getValue().toString()` turns malformed client input into either a 500 (`null` value) or a lossy config string (`schema: { ... }` becomes a Java map string). Since this is request input for graph creation/cloning, please reject null or non-scalar values with a clear 4xx before converting to backend config strings.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -2174,6 +2199,45 @@ public Map<String, Object> graphConfig(String graphSpace,
...
+        if (!isPDEnabled()) {
+            // Non-PD mode: in-memory only, acceptable for standalone RocksDB
+            return;
+        }
+        try {
+            Map<String, Object> configs =
+                    this.metaManager.getGraphConfig(graphSpace, graphName);
+            if (configs != null) {
+                configs.put("nickname", nickname);
+                this.metaManager.updateGraphConfig(graphSpace, graphName, configs);
```
imbajin: ⚠️ **Avoid rolling back local state after the PD config write has already succeeded**

If `updateGraphConfig()` succeeds but `notifyGraphUpdate()` throws, the catch block restores the local in-memory nickname to the old value while PD meta already contains the new nickname. That leaves this node inconsistent with the persisted state and with any node that observes the meta update.

Please distinguish persistence failure from notification failure, or only roll back the in-memory nickname when the config write itself did not succeed.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/StandardAuthManagerV2.java
```diff
@@ -1510,6 +1516,211 @@ public HugeGroup findGroup(String name) {
...
+        try {
+            String belongId;
+            if (HugeGroup.isGroup(owner)) {
+                belongId = this.metaManager.belongId(owner, role,
+                                                     HugeBelong.GR);
+            } else {
+                belongId = this.metaManager.belongId(owner, role,
+                                                     HugeBelong.UR);
+            }
+            this.metaManager.deleteBelong(graphSpace,
```
imbajin: ⚠️ **Make default-role delete idempotent like unset default graph**

`deleteBelong()` validates existence, so deleting an already-removed default role can still be wrapped as a server error. `unsetDefaultGraph()` now checks `existBelong()` and treats missing bindings as success; default-role deletion should follow the same behavior so retries or duplicate UI actions do not fail unnecessarily.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -103,6 +107,161 @@ public Object get(@Context GraphManager manager,
...
+            authManager.createSpaceDefaultRole(name, user, role);
+        }
+
+        return manager.serializer().writeMap(result);
+    }
+
+    @GET
+    @Timed
+    @Path("{graphspace}/role")
+    @Consumes(APPLICATION_JSON)
```
imbajin: 🧹 **Drop `@Consumes` from entity-less role endpoints**

`GET` and `DELETE` here do not consume a request body, so declaring `@Consumes(APPLICATION_JSON)` is misleading and can make generated clients think they need to send a JSON entity/content type. Please remove it from the entity-less role endpoints here, and from the similar `ManagerAPI.checkDefaultRole()` endpoint.

imbajin (commented):
Follow-up review after the latest update: these are remaining current-PR issues that look small enough to fix directly, while already-covered findings are intentionally not duplicated.

imbajin:
Follow-up TODO for CI coverage:

Add a real distributed-mode CI job that boots PD + Store and runs the hstore-backed graphspace API tests against that environment. The tests added in this PR compile and are gated to `backend=hstore`, but they still rely on an external PD/Store runtime to prove the full happy path.

Suggested scope for a separate issue:
1. Start a minimal PD + Store + Server stack in CI.
2. Run the hstore/PD graphspace API suite, including default graph, default role, and schema template lifecycle coverage.
3. Keep the existing standalone/RocksDB compatibility tests separate so both modes remain visible.

This is not necessarily a blocker for the current PR, but it is the right follow-up to prevent these Hubble-compatible APIs from only being validated at compile time.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/ServerInfoManager.java
```diff
@@ -122,19 +122,25 @@ public synchronized void initServerInfo(GlobalMasterInfo nodeInfo) {
...
+                        existed = servers.next();
+                        E.checkArgument(!existed.role().master() || !existed.alive(),
+                                        "Already existed master '%s' in current cluster",
+                                        existed.id());
+                    }
+                    if (page != null) {
+                        page = PageInfo.pageInfo(servers);
+                    }
+                } while (page != null);
+            } catch (Exception e) {
```
VGalaxies: **High: Existing live master detection is swallowed**

`hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/ServerInfoManager.java:139`

**Evidence**
- The uniqueness check at lines 131-133 throws when another alive master exists, but the new broad `catch (Exception e)` at line 139 logs and continues to `saveServerInfo()` at line 149.

**Impact**
- A node can register itself as master even after detecting an existing alive master, allowing multiple masters in the same cluster.

**Requested fix**
- Do not catch the invariant failure. Only handle the specific schema-mismatch/read error if needed, and rethrow when an existing alive master is found.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -103,6 +107,171 @@ public Object get(@Context GraphManager manager,
         return gsInfo;
     }
 
+    @POST
+    @Timed
+    @Status(Status.CREATED)
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON_WITH_CHARSET)
+    @Path("{graphspace}/role")
+    @RolesAllowed({"analyst"})
```
VGalaxies: **High: Analysts can grant or revoke default roles for arbitrary users**

`hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java:116`

**Evidence**
- `POST /graphspaces/{graphspace}/role` and `DELETE /graphspaces/{graphspace}/role` are gated only by `@RolesAllowed("analyst")`; after that, non-admin callers are blocked only for `SPACE`, while `ANALYST`, `SPACE_MEMBER`, and `OBSERVER` are created/deleted at lines 166-171 and 268-271. `ANALYST` grants READ/WRITE/DELETE/EXECUTE in `StandardAuthManagerV2.java:1785-1787`.

**Impact**
- Any caller satisfying the analyst role can grant destructive graphspace permissions to other users or remove their roles, bypassing the existing admin/space-manager role-management boundary.

**Requested fix**
- Require admin or space manager permission for these role mutations, matching `ManagerAPI`, and keep admin-only enforcement for `SPACE` if needed.

VGalaxies (commented):
**Review summary**

- **Blocking:** yes
- **Summary:** The PR introduces a cluster master safety regression and an authorization gap in default-role management.
- **Evidence:**
  - `git diff origin/master...HEAD`
  - `mvn -pl hugegraph-server/hugegraph-api -am -DskipTests -Dmaven.javadoc.skip=true compile` passed

imbajin:
Follow-up boundary notes after rechecking the current Hubble2 branch:

- Hubble2 currently still calls graph creation as form-urlencoded `POST /graphspaces/{graphspace}/graphs` and default graph through GET aliases such as `setdefault/getdefault`. I am not adding these extra Server compatibility aliases in this pass. Since Hubble2 is still under development, the cleaner direction is to adjust Hubble2 to the Server API contract unless we explicitly decide to expand the Server public API.
- The master heartbeat uniqueness concern is tied to the deprecated master-worker scheduling path. I am not adding more code to that path here.
- The current Server-side patch is limited to issues that remain valid independent of those boundaries: default-role auth cache invalidation, keeping GraphSpace profile PD-only, and making hstore graphspace API tests fail fast instead of hiding API failures as skips.

imbajin (commented):
Blocking: yes. Summary: Current head still has a shared-storage server-info regression and a test coverage gap; visible hbase CI is failing. Evidence: final gate saw build-server (hbase, 11) and codecov/patch failures; static review of refs/remotes/pr/3008.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/ServerInfoManager.java
```diff
@@ -198,16 +206,12 @@ public synchronized void heartbeat() {
...
         /*
-         * Missing server info on non-master node, may be caused by graph
-         * truncated on master node then synced by raft.
-         * TODO: we just patch it here currently, to be improved.
+         * Missing server info on any node (master or worker), may be caused by
+         * graph truncated or store data cleared.
+         * Re-save ServerInfo to recover automatically.
          */
+        LOG.warn("ServerInfo is missing: {}, re-saving it now", this.selfNodeId());
         serverInfo = this.saveServerInfo(this.selfNodeId(), this.selfNodeRole());
```
imbajin: `heartbeat()` now re-saves missing ServerInfo for master nodes too, and the latest hbase job is failing on that path. The log shows `ServerInfo is missing: DEFAULT-hugegraph/server-1, re-saving it now`, then `GremlinApiTest.testClearAndInit` hits `Already existed master 'DEFAULT-hugegraph/server-1' in current cluster`, followed by `decreaseLoad()` NPEs because self ServerInfo is still missing for `DEFAULT-hugegraph/server1`. Please make current-node master re-registration idempotent, or skip only the normalized current node in the uniqueness scan, and guard `decreaseLoad()` when self ServerInfo is unavailable.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/ManagerAPI.java
```diff
@@ -259,6 +262,45 @@ public String getRolesInGs(@Context GraphManager manager,
...
+            E.checkArgument(false, "Invalid role value '%s'", role);
+            defaultRole = null; // unreachable, satisfies compiler
+        }
+        boolean hasGraph = defaultRole.equals(HugeDefaultRole.OBSERVER);
+        E.checkArgument(!hasGraph || StringUtils.isNotEmpty(graph),
+                        "Must set a graph for observer");
+
+        boolean result;
+        if (hasGraph) {
+            result = authManager.isDefaultRole(graphSpace, graph, user,
```
imbajin, line 299: This endpoint checks raw default-role metadata without first validating that the path graphspace exists, and for `OBSERVER` it only checks that `graph` is non-empty. A typo or stale graph name can therefore return a metadata result instead of a clear 4xx. Please mirror `GraphSpaceAPI`: validate the graphspace before the auth lookup, and for observer roles verify that the graph exists in that graphspace before calling `isDefaultRole()`.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/api/ApiTestSuite.java
```diff
@@ -44,6 +44,9 @@
         GraphSpaceApiTest.class,
         GraphSpaceApiStandaloneTest.class,
         ManagerApiStandaloneTest.class,
+        GraphsApiStandaloneTest.class,
```
imbajin, line 47: Adding the backend-specific suites here does not make them run in CI as written. The new standalone and hstore tests gate on `System.getProperty("backend")`, but `run-api-test.sh` invokes Maven as `mvn ... -P api-test,$BACKEND`, and the surefire config does not pass the Maven `backend` profile property into the test JVM. That leaves `backend` null, so `assumeStandaloneMode()` skips the standalone suite and `GraphSpaceApiTest` skips as non-hstore. Please pass `backend` into surefire, for example with `systemPropertyVariables`, or call Maven with `-Dbackend=$BACKEND`.

imbajin (approved):
TODO: update related doc

PR #3011, opened 2026-04-28
fix(server): sync hstore schema cache clears

### LINKED ISSUE #2617: [Bug] cacheEventListener in CachedSchemaTransactionV2   Work does not meet expectations
### Bug Type (问题类型)

other exception / error (其他异常报错)

### Before submit

### Environment (环境信息)

- Server Version: 1.5.0 (Apache Release Version)
- Backend: HStore 

When using multi server mode, when a client connects to a server to create a graph schema, other servers will not perceive this and perform the action of clearing the local schema cache based on the cacheEventListener; If there is no written action in the future and only one server holds schema data, the schema will appear intermittently on the bubble page
Note: The reason why there are no errors during the writing process is that if the required schema cannot be queried in the local cache during the construction of the edge, it will be queried on the PD side and the cache will be updated.

_No response_

_No response_

dpol1:
## Purpose of the PR

- close #2617

In HStore / multi-server mode, schema changes can leave other server nodes with stale local schema caches. `CachedSchemaTransactionV2` already publishes schema-cache clear events through `MetaManager`, but it did not listen for those meta events and clear the affected local V2 caches.

## Main Changes

- Register a JVM-wide schema cache clear listener for `CachedSchemaTransactionV2`
- Clear V2 `schema-id` / `schema-name` caches and the attached array cache by graph name
- Publish schema cache clear events after schema add, update, and remove paths
- Keep the local `EventHub` cache listener registered per transaction instance
- Add unit coverage for graph-scoped V2 schema cache clearing

## Schema Cache Clear Flow

```text
Before

+----------+       schema add/remove       +---------------+
| Server A | ----------------------------> | HStore / Meta |
+----------+                               +---------------+
     |
     | local V2 cache updated
     v
+-------------------+
| Server A cache    |
| fresh             |
+-------------------+

+-------------------+
| Server B cache    |
| stale             |
+-------------------+

Problem:
Server B had no effective cross-JVM V2 schema-cache-clear listener, so it could
keep serving stale schema from local cache.
```

```text
After

+----------+       schema add/remove       +---------------+
| Server A | ----------------------------> | HStore / Meta |
| source A |                               +---------------+
+----------+                                      |
     |                                            |
     | publish schema-cache-clear                 |
     | {"graph":"DEFAULT-g","source":"A"}         |
     v                                            v
+-------------------+                    +-------------------+
| Server A listener |                    | Server B listener |
| source == A       |                    | source != B       |
| skip self echo    |                    | clear V2 caches   |
+-------------------+                    +-------------------+
                                                |
                                                v
                                      +---------------------+
                                      | schema-id cache     |
                                      | schema-name cache   |
                                      | array attachment    |
                                      +---------------------+
```

 ## Compatibility / Scope Note

  This PR does not change persisted schema data. It only changes the runtime schema-cache-clear meta-event value so new events can carry a per-JVM source id. New consumers still accept legacy plain-string graph names, so mixed-version rollouts during upgrade are safe.

  The reconnect recovery hook remains manual until MetaManager/MetaDriver exposes a reconnect callback. The `~create_time` String/Date schema userdata round-trip is tracked separately in #3013 .

## Verifying these changes

    - `mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test -Dtest=CachedSchemaTransactionTest -DfailIfNoTests=false "-Djacoco.skip=true"`

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Other affects: HStore schema cache invalidation behavior in multi-server deployments
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO`
- [ ]  `Doc - Done`
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
@@ -49,8 +49,14 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
     private static final String NAME_CACHE_PREFIX = "schema-name";
 
     // MetaDriver doesn't expose unlisten, register the meta listener once.
+    // Lifecycle: this JVM-global flag is intentionally never reset by
+    // unlistenChanges() (the underlying gRPC watch is process-wide), but IS
+    // reset on transport reconnect via resetMetaListenerForReconnect() so a
+    // new watch is installed after the old one was silently dropped.
     private static final AtomicBoolean metaEventListenerRegistered =
             new AtomicBoolean(false);
+
+    private static final Object META_LISTENER_LOCK = new Object();
 
     private final Cache<Id, Object> idCache;
     private final Cache<Id, Object> nameCache;
@@ -107,21 +113,23 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
     private static void clearSchemaCache(String spaceGraphName) {
         Map<String, Cache<Id, Object>> caches = CacheManager.instance().caches();
 
-        Cache<Id, Object> idCache = caches.get(cacheName(ID_CACHE_PREFIX,
-                                                         spaceGraphName));
-        if (idCache != null) {
-            idCache.clear();
-
-            SchemaCaches<?> arrayCaches = idCache.attachment();
-            if (arrayCaches != null) {
-                arrayCaches.clear();
-            }
-        }
-
+        // Clear name cache first so the (name -> id -> object) lookup path
+        // fails fast instead of returning a stale object backed by an
+        // already-empty id cache during the TOCTOU window.
         Cache<Id, Object> nameCache = caches.get(cacheName(NAME_CACHE_PREFIX,
                                                            spaceGraphName));
         if (nameCache != null) {
             nameCache.clear();
+        }
+
+        Cache<Id, Object> idCache = caches.get(cacheName(ID_CACHE_PREFIX,
+                                                         spaceGraphName));
+        if (idCache != null) {
+            SchemaCaches<?> arrayCaches = idCache.attachment();
+            if (arrayCaches != null) {
+                arrayCaches.clear();
+            }
+            idCache.clear();
         }
     }
 
@@ -185,34 +193,73 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
     }
 
     private static void listenSchemaCacheClear() {
-        if (!metaEventListenerRegistered.compareAndSet(false, true)) {
+        synchronized (META_LISTENER_LOCK) {
+            if (metaEventListenerRegistered.get()) {
+                return;
+            }
+            try {
+                MetaManager.instance().listenSchemaCacheClear(
+                        CachedSchemaTransactionV2::handleSchemaCacheClearEvent);
+                // Set AFTER the underlying watch is live so a concurrent
+                // caller that observes the flag is guaranteed an active
+                // subscription, and a failure leaves the flag false so the
+                // next caller retries registration.
+                metaEventListenerRegistered.set(true);
+            } catch (Exception e) {
+                throw e instanceof RuntimeException
+                      ? (RuntimeException) e
+                      : new RuntimeException(
+                              "Failed to register schema cache clear listener",
+                              e);
+            }
+        }
+    }
+
+    /**
+     * Consumer invoked by the MetaManager schema-cache-clear watch. Extracted
+     * as a package-private static method so end-to-end tests can drive the
+     * publish -> callback -> {@link #clearSchemaCache(String)} path without
+     * depending on a live etcd/PD watch.
+     */
+    static <T> void handleSchemaCacheClearEvent(T response) {
+        List<String> graphNames =
+                MetaManager.instance()
+                           .extractSchemaCacheClearGraphNamesFromResponse(
+                                   response);
+        if (graphNames == null) {
             return;
         }
+        for (String graphName : graphNames) {
+            LOG.debug("Graph {} clear schema cache on meta event", graphName);
+            clearSchemaCache(graphName);
+        }
+    }
 
-        try {
-            MetaManager.instance().listenSchemaCacheClear(response -> {
-                List<String> graphNames =
-                        MetaManager.instance().metaDriver()
-                                   .extractValuesFromResponse(response);
-                if (graphNames == null) {
-                    return;
-                }
-                for (String graphName : graphNames) {
-                    LOG.debug("Graph {} clear schema cache on meta event",
-                              graphName);
-                    clearSchemaCache(graphName);
-                }
-            });
-        } catch (RuntimeException e) {
-            metaEventListenerRegistered.set(false);
-            throw e;
+    /**
+     * Reset the JVM-global meta listener flag after a MetaManager transport
+     * reconnect. The underlying gRPC watch is silently dropped on reconnect;
+     * without this reset {@link #metaEventListenerRegistered} would stay
+     * {@code true} forever and this JVM would stop receiving cross-node
+     * schema cache clear events with no error or warning.
+     *
+     * <p>TODO: wire this into MetaManager once it exposes a transport
+     * reconnect callback (e.g. {@code listenReconnect} /
+     * {@code onTransportReconnect}). Until then it must be invoked
+     * explicitly by code that detects the reconnect.
+     */
+    public static void resetMetaListenerForReconnect() {
+        if (metaEventListenerRegistered.compareAndSet(true, false)) {
+            LOG.warn("Schema cache clear meta listener lost on reconnect — " +
+                     "will re-register on next schema operation.");
         }
     }
 
     public void clearCache(boolean notify) {
-        this.idCache.clear();
+        // Same TOCTOU ordering as clearSchemaCache(String): clear nameCache
+        // first, then the array attachment, then idCache last.
         this.nameCache.clear();
         this.arrayCaches.clear();
+        this.idCache.clear();
 
         if (notify) {
             this.maybeNotifySchemaCacheClear();
@@ -268,7 +315,8 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
         super.updateSchema(schema, updateCallback);
 
         this.updateCache(schema);
-        this.maybeNotifySchemaCacheClear();
+        // Status transitions are internal bookkeeping; notifying here causes a
+        // broadcast storm for every updateSchemaStatus() call from background jobs.
     }
 
     @Override
@@ -277,7 +325,11 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
 
         this.updateCache(schema);
 
-        this.maybeNotifySchemaCacheClear();
+        // Schema additions must always propagate to remote nodes regardless
+        // of TASK_SYNC_DELETION (which only gates removal flows).
+        MetaManager.instance()
+                   .notifySchemaCacheClear(this.graph().graphSpace(),
+                                           this.graph().name());
     }
 
     private void updateCache(SchemaElement schema) {
@@ -305,6 +357,9 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
     }
 
     private void maybeNotifySchemaCacheClear() {
+        // Only suppress notifications for removal tasks when
+        // TASK_SYNC_DELETION=true: the caller propagates cache invalidation
+        // synchronously, so the meta-event broadcast would be redundant.
         if (!this.graph().option(CoreOptions.TASK_SYNC_DELETION)) {
             MetaManager.instance()
                        .notifySchemaCacheClear(this.graph().graphSpace(),
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
@@ -48,7 +48,7 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
     private static final String ID_CACHE_PREFIX = "schema-id";
     private static final String NAME_CACHE_PREFIX = "schema-name";
 
-    // MetaDriver doesn't expose unlisten, register the PD listener once.
+    // MetaDriver doesn't expose unlisten, register the meta listener once.
     private static final AtomicBoolean metaEventListenerRegistered =
             new AtomicBoolean(false);
 
@@ -134,7 +134,8 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
             if (storeEvents.contains(event.name())) {
                 LOG.debug("Graph {} clear schema cache on event '{}'",
                           this.graph(), event.name());
-                this.clearCache(true);
+                boolean notify = !Events.STORE_INIT.equals(event.name());
+                this.clearCache(notify);
                 return true;
             }
             return false;
@@ -189,10 +190,10 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
         }
 
         try {
-            MetaDriver metaDriver = MetaManager.instance().metaDriver();
             MetaManager.instance().listenSchemaCacheClear(response -> {
                 List<String> graphNames =
-                        metaDriver.extractValuesFromResponse(response);
+                        MetaManager.instance().metaDriver()
+                                   .extractValuesFromResponse(response);
                 if (graphNames == null) {
                     return;
                 }
@@ -214,7 +215,7 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
         this.arrayCaches.clear();
 
         if (notify) {
-            this.notifySchemaCacheClear();
+            this.maybeNotifySchemaCacheClear();
         }
     }
 
@@ -267,7 +268,7 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
         super.updateSchema(schema, updateCallback);
 
         this.updateCache(schema);
-        this.notifySchemaCacheClear();
+        this.maybeNotifySchemaCacheClear();
     }
 
     @Override
@@ -276,7 +277,7 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
 
         this.updateCache(schema);
 
-        this.notifySchemaCacheClear();
+        this.maybeNotifySchemaCacheClear();
     }
 
     private void updateCache(SchemaElement schema) {
@@ -300,10 +301,10 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
 
         this.invalidateCache(schema.type(), schema.id());
 
-        this.notifySchemaCacheClear();
+        this.maybeNotifySchemaCacheClear();
     }
 
-    private void notifySchemaCacheClear() {
+    private void maybeNotifySchemaCacheClear() {
         if (!this.graph().option(CoreOptions.TASK_SYNC_DELETION)) {
             MetaManager.instance()
                        .notifySchemaCacheClear(this.graph().graphSpace(),
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
@@ -35,6 +36,7 @@ import org.apache.hugegraph.event.EventHub;
 import org.apache.hugegraph.event.EventListener;
 import org.apache.hugegraph.meta.MetaDriver;
 import org.apache.hugegraph.meta.MetaManager;
+import org.apache.hugegraph.meta.MetaManager.SchemaCacheClearEvent;
 import org.apache.hugegraph.perf.PerfUtil;
 import org.apache.hugegraph.schema.SchemaElement;
 import org.apache.hugegraph.type.HugeType;
@@ -57,6 +59,15 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
             new AtomicBoolean(false);
 
     private static final Object META_LISTENER_LOCK = new Object();
+
+    /**
+     * Per-JVM identifier emitted with every schema-cache-clear meta event so
+     * the listener can skip its own echo. Lifecycle: generated once per
+     * classloader at class init, never reused, regenerated on JVM restart.
+     * This is not a stable node identity, only a local self-echo filter.
+     */
+    private static final String SCHEMA_CACHE_CLEAR_SOURCE =
+            UUID.randomUUID().toString();
 
     private final Cache<Id, Object> idCache;
     private final Cache<Id, Object> nameCache;
@@ -222,14 +233,18 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
      * depending on a live etcd/PD watch.
      */
     static <T> void handleSchemaCacheClearEvent(T response) {
-        List<String> graphNames =
+        List<SchemaCacheClearEvent> events =
                 MetaManager.instance()
-                           .extractSchemaCacheClearGraphNamesFromResponse(
+                           .extractSchemaCacheClearEventsFromResponse(
                                    response);
-        if (graphNames == null) {
+        if (events == null) {
             return;
         }
-        for (String graphName : graphNames) {
+        for (SchemaCacheClearEvent event : events) {
+            if (SCHEMA_CACHE_CLEAR_SOURCE.equals(event.source())) {
+                continue;
+            }
+            String graphName = event.graph();
             LOG.debug("Graph {} clear schema cache on meta event", graphName);
             clearSchemaCache(graphName);
         }
@@ -249,7 +264,7 @@ public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
      */
     public static void resetMetaListenerForReconnect() {
         if (metaEventListenerRegistered.compareAndSet(true, false)) {
-            LOG.warn("Schema cache clear meta listener lost on reconnect — " +
+            LOG.warn("Schema cache clear meta listener lost on reconnect - " +
                      "will re-register on next schema operation.");
         }
     }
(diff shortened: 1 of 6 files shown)


dpol1:
Hi @imbajin - while working on this PR I spotted what might be a separate bug in the legacy `EventHub` path. Flagging before I file anything.

Producers emit past-tense actions:
- `Cache.ACTION_INVALIDED` (`"invalided"`)
- `Cache.ACTION_CLEARED` (`"cleared"`)

Listeners only match the present-tense ones:
- `Cache.ACTION_INVALID` (`"invalid"`)
- `Cache.ACTION_CLEAR` (`"clear"`)

So those events look like they get dropped silently. Wider blast radius than the HStore `CachedSchemaTransactionV2` fix here, so I kept it out of scope.

Worth a separate issue or maybe sub-issue - or is this intentional? If you're +1, I'll take care of it

imbajin:
> Producers emit past-tense actions: `Cache.ACTION_INVALIDED` / `Cache.ACTION_CLEARED`
> Listeners only match the present-tense ones: `Cache.ACTION_INVALID` / `Cache.ACTION_CLEAR`
> So those events look like they get dropped silently.

Good catch @dpol1 — I dug into the code and confirmed the mismatch is real.

Opened #3012 to track this separately — feel free to take it or I can look at it after #3011 lands.

dpol1:
> > Producers emit past-tense actions: `Cache.ACTION_INVALIDED` / `Cache.ACTION_CLEARED`
> > Listeners only match the present-tense ones: `Cache.ACTION_INVALID` / `Cache.ACTION_CLEAR`
> > So those events look like they get dropped silently.
> 
> Good catch @dpol1 — I dug into the code and confirmed the mismatch is real.
> 
> Opened #3012 to track this separately — feel free to take it or I can look at it after #3011 lands.

I'll do in a separate PR, when this lands!

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -202,6 +268,7 @@ protected void updateSchema(SchemaElement schema,
         super.updateSchema(schema, updateCallback);
 
         this.updateCache(schema);
+        this.maybeNotifySchemaCacheClear();
     }
```
imbajin, line 338: ‼️ **CRITICAL: `updateSchema()` fires a full cluster-wide schema cache clear on *every* status transition, causing notification storms**

`CachedSchemaTransactionV2.updateSchema()` is invoked by `SchemaTransactionV2.updateSchemaStatus()`, which background jobs call multiple times per operation:

- `IndexLabelRebuildJob`: 4 calls (REBUILDING → INVALID → CREATED + error rollback)
- `EdgeLabelRemoveJob` / `VertexLabelRemoveJob` / `IndexLabelRemoveJob`: 2–3 calls each

Each call now fires `MetaManager.notifySchemaCacheClear()`, broadcasting a **full clear of all PK/VL/EL/IL caches** across all cluster nodes. A single `rebuildIndex` on a 5-node cluster generates ~20 full-cache-clear broadcasts during normal background execution.

Note: `CachedSchemaTransaction` (V1) deliberately did NOT notify in `updateSchema()` — status transitions are internal bookkeeping, not topology changes. The fix should be scoped to `addSchema` and `removeSchema` only:

```suggestion
    @Override
    protected void updateSchema(SchemaElement schema,
                                Consumer<SchemaElement> updateCallback) {
        super.updateSchema(schema, updateCallback);

        this.updateCache(schema);
        // Status transitions are internal bookkeeping; notifying here causes a
        // broadcast storm for every updateSchemaStatus() call from background jobs.
    }
```

imbajin (commented):
The fix direction is correct: a JVM-global MetaManager listener with graph-scoped `clearSchemaCache` is the right architectural shape for HStore cache synchronization. Two critical behavioral issues need to be resolved before merge — `updateSchema` triggering unnecessary broadcast storms, and the static listener having no lifecycle reset path — plus a few important design concerns below.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -43,6 +45,13 @@
 
 public class CachedSchemaTransactionV2 extends SchemaTransactionV2 {
 
+    private static final String ID_CACHE_PREFIX = "schema-id";
+    private static final String NAME_CACHE_PREFIX = "schema-name";
+
+    // MetaDriver doesn't expose unlisten, register the meta listener once.
+    private static final AtomicBoolean metaEventListenerRegistered =
+            new AtomicBoolean(false);
```
imbajin, line 60: ‼️ **CRITICAL: Static `AtomicBoolean` has no reset path on MetaManager reconnect — silent listener loss**

If the MetaManager transport reconnects (etcd client reconnect after network partition), the underlying watch subscription is silently lost but `metaEventListenerRegistered` remains `true` forever, preventing re-registration. After a reconnect, this JVM stops receiving cross-node schema-cache-clear events with no error or warning.

Also: `unlistenChanges()` clears store/EventHub listeners but deliberately leaves the meta listener as a ghost (by design per the comment). After a graph is dropped, the meta listener continues firing `clearSchemaCache()` for that graph name — harmless because `CacheManager` returns null for non-existent caches, but this gap in lifecycle ownership should be explicitly documented.

Minimum fix: log a warning if MetaManager reconnects without re-registration, and consider adding a reconnect hook.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -145,12 +180,43 @@ private void listenChanges() {
         if (!schemaEventHub.containsListener(Events.CACHE)) {
             schemaEventHub.listen(Events.CACHE, this.cacheEventListener);
         }
+
+        listenSchemaCacheClear();
+    }
+
+    private static void listenSchemaCacheClear() {
+        if (!metaEventListenerRegistered.compareAndSet(false, true)) {
+            return;
```
imbajin, line 242: ⚠️ **`AtomicBoolean` is set `true` *before* the listener is fully registered — narrow race window**

The flag is flipped via `compareAndSet(false, true)` before `MetaManager.instance().listenSchemaCacheClear(...)` completes the underlying gRPC watch setup. Another thread calling `listenSchemaCacheClear()` in the same window sees `true` and returns immediately, but the subscription may not yet be live. Events published in that window are silently dropped.

The `catch` rollback handles construction failures, not the concurrent setup race. Consider wrapping the CAS-to-registration sequence in a `synchronized` block on a class-level lock.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -145,12 +180,43 @@ private void listenChanges() {
...
+                    return;
+                }
+                for (String graphName : graphNames) {
+                    LOG.debug("Graph {} clear schema cache on meta event",
+                              graphName);
+                    clearSchemaCache(graphName);
+                }
+            });
+        } catch (RuntimeException e) {
+            metaEventListenerRegistered.set(false);
```
imbajin: ⚠️ **Exception catch is too narrow — non-`RuntimeException` failures leave `AtomicBoolean` stuck at `true`**

If `MetaManager.instance().listenSchemaCacheClear()` throws a checked exception (wrapped or not) or an `Error`, the reset at line 208 is never reached and the listener is permanently skipped for this JVM.

```suggestion
        } catch (Exception e) {
            metaEventListenerRegistered.set(false);
            throw e instanceof RuntimeException ? (RuntimeException) e
                                                : new RuntimeException("Failed to register schema cache clear listener", e);
        }
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -145,12 +180,43 @@ private void listenChanges() {
...
+    private static void listenSchemaCacheClear() {
+        if (!metaEventListenerRegistered.compareAndSet(false, true)) {
+            return;
+        }
+
+        try {
+            MetaManager.instance().listenSchemaCacheClear(response -> {
+                List<String> graphNames =
+                        MetaManager.instance().metaDriver()
+                                   .extractValuesFromResponse(response);
```
imbajin: ⚠️ **Direct `metaDriver()` call breaks the MetaManager abstraction layer**

```java
MetaManager.instance().metaDriver().extractValuesFromResponse(response)
```

The cache package accessing `MetaDriver` internals directly couples `CachedSchemaTransactionV2` to the meta package's internals. `MetaManager` already exposes typed extraction helpers (`extractGraphsFromResponse`, `extractServicesFromResponse`, etc.). This PR added `metaDriver()` as a public accessor just for this one use case — the right fix is a dedicated helper on `MetaManager`:

```suggestion
                List<String> graphNames =
                        MetaManager.instance()
                                   .extractSchemaCacheClearGraphNamesFromResponse(response);
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -86,11 +95,36 @@ public void close() {
...
+    }
+
+    private static void clearSchemaCache(String spaceGraphName) {
+        Map<String, Cache<Id, Object>> caches = CacheManager.instance().caches();
+
+        Cache<Id, Object> idCache = caches.get(cacheName(ID_CACHE_PREFIX,
+                                                         spaceGraphName));
+        if (idCache != null) {
+            idCache.clear();
+
```
imbajin: ⚠️ **`clearSchemaCache` is non-atomic: TOCTOU window between `idCache.clear()` and `nameCache.clear()`**

Between these two `clear()` calls a reader thread can see `idCache` empty while `nameCache` still holds stale entries. `getSchema(type, name)` checks `nameCache` first — a hit from the still-live `nameCache` bypasses any reload logic and returns a stale object. This same window exists in the instance-method `clearCache()` and both should be addressed together.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -238,6 +301,10 @@ public void removeSchema(SchemaElement schema) {
 
         this.invalidateCache(schema.type(), schema.id());
 
+        this.maybeNotifySchemaCacheClear();
+    }
+
+    private void maybeNotifySchemaCacheClear() {
         if (!this.graph().option(CoreOptions.TASK_SYNC_DELETION)) {
             MetaManager.instance()
```
imbajin: 🧹 **`TASK_SYNC_DELETION` gate in `maybeNotifySchemaCacheClear` is semantically mismatched for `addSchema`/`updateSchema`**

`CoreOptions.TASK_SYNC_DELETION` controls whether task-based schema *deletion* runs synchronously. Using it to suppress `addSchema` and `updateSchema` notifications means: when `TASK_SYNC_DELETION=true` (typical test/single-node config), schema additions and updates are silently never propagated to remote nodes — even though they have nothing to do with deletion tasks.

As a side effect, integration tests that run with `TASK_SYNC_DELETION=true` never exercise the notification path added by this PR. Consider separating the guard: `removeSchema` only for the deletion gate, always notify for `addSchema`.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedSchemaTransactionTest.java
```diff
@@ -165,6 +172,98 @@ public void testGetSchema() throws Exception {
...
+            Assert.assertEquals(1L, otherIdCache.size());
+            Assert.assertSame(arrayCacheSchema,
+                              getV2SchemaCache(arrayCaches,
+                                               HugeType.PROPERTY_KEY,
+                                               arrayCacheId));
+
+            Whitebox.invokeStatic(CachedSchemaTransactionV2.class,
+                                  new Class<?>[]{String.class},
+                                  "clearSchemaCache", graphName);
+
```
imbajin, line 230: ⚠️ **Test only covers `clearSchemaCache()` in isolation via reflection — the end-to-end listener path is untested**

The test calls the private static method directly via `Whitebox.invokeStatic`. It does not verify:
- That `listenSchemaCacheClear()` actually registers a MetaManager watcher
- That `maybeNotifySchemaCacheClear()` → MetaManager publish → listener callback → `clearSchemaCache()` completes end-to-end
- The `AtomicBoolean` idempotency: two graph instances in the same JVM should register only one listener
- The `TASK_SYNC_DELETION=true` suppression path (which also means CI tests never exercise the notification)

The behavior this PR claims to fix — cross-node cache sync — has zero test coverage.

dpol1 (commented):
@imbajin I noticed one more issue exposed by my cache-clear change

The failing casts are here:

- [`PropertyKeyCoreTest.java#L674`](https://github.com/apache/hugegraph/blob/78f47ad7a4f2e735808707468322ae5473b3ab33/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/PropertyKeyCoreTest.java#L674)
- [`IndexLabelCoreTest.java#L1773`](https://github.com/apache/hugegraph/blob/78f47ad7a4f2e735808707468322ae5473b3ab33/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/IndexLabelCoreTest.java#L1773)

After the V2 cache clear, the schema is loaded again from backend. In that path, userdata is restored here:

[`TextSerializer.java#L917`](https://github.com/apache/hugegraph/blob/78f47ad7a4f2e735808707468322ae5473b3ab33/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/TextSerializer.java#L917)

`~create_time` comes back as `String`, but those tests expect `Date`.

I’ll fix this in `Userdata`, not in `TextSerializer`. The constructor currently just does `putAll(map)` here:

[`Userdata.java#L38-L39`](https://github.com/apache/hugegraph/blob/78f47ad7a4f2e735808707468322ae5473b3ab33/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java#L38-L39)

I’ll normalize `Userdata.CREATE_TIME` in `put()` / `putAll()`, so serialized date strings/timestamps are converted back to `Date` in one place. Then I’ll add a small regression test and rerun the two HStore create-time tests. That's my way to handle it if you have another idea on your POV, give me a feedback :)

dpol1:
@imbajin Dropped the Userdata idea — too wide for this PR.
Schema-add still propagates a cache clear, but now V2 events carry a per-JVM source id so the publisher skips its own echo. Remote HStore servers still see the event and invalidate normally.
Plain-string (legacy) events clear the same way as before, so older publishers won't break.

The ~create_time String/Date round-trip: pretty sure that's a backend serialization thing. Leaving it out to keep this PR scoped.

imbajin:
Thanks for the update. The current direction looks much better scoped than the earlier `Userdata` normalization idea.

From my reading, the current PR mechanism is:

- keep the local `EventHub` schema-cache listener for same-JVM cache events
- add one JVM-wide `MetaManager` listener for cross-server schema-cache-clear events
- publish graph-scoped schema-cache-clear events on schema add/remove
- include a per-JVM `source` id so the publisher can skip its own watch echo
- clear all V2 schema cache layers on remote JVMs: `schema-id`, `schema-name`, and the array attachment cache
- continue accepting legacy plain-string event values from older publishers

Could you please update the PR description/comment with a compact before -> after diagram? I think that will make the design much easier to review, especially since most of the added code is test coverage.

Suggested structure:

```text
Before

+----------+       schema add/remove       +---------------+
| Server A | ----------------------------> | HStore / Meta |
+----------+                               +---------------+
     |
     | local V2 cache updated
     v
+-------------------+
| Server A cache    |
| fresh             |
+-------------------+

+-------------------+
| Server B cache    |
| stale             |
+-------------------+

Problem:
Server B had no effective cross-JVM V2 schema-cache-clear listener, so it could
keep serving stale schema from local cache.
```

```text
After

+----------+       schema add/remove       +---------------+
| Server A | ----------------------------> | HStore / Meta |
| source A |                               +---------------+
+----------+                                      |
     |                                            |
     | publish schema-cache-clear                 |
     | {"graph":"DEFAULT-g","source":"A"}         |
     v                                            v
+-------------------+                    +-------------------+
| Server A listener |                    | Server B listener |
| source == A       |                    | source != B       |
| skip self echo    |                    | clear V2 caches   |
+-------------------+                    +-------------------+
                                                |
                                                v
                                      +---------------------+
                                      | schema-id cache     |
                                      | schema-name cache   |
                                      | array attachment    |
                                      +---------------------+
```

A few points I would like to confirm before merge:

1. `resetMetaListenerForReconnect()` does not seem to be wired to any actual `MetaManager` / `MetaDriver` reconnect callback.

   If there is an existing call path, could you point it out? If not, please adjust the comment so it is clear this is only a future/manual recovery hook, not a complete reconnect recovery mechanism in this PR.

2. The `~create_time` String/Date round-trip should stay out of this PR, but please mention it as a follow-up.

   The self-echo skip avoids forcing the publisher to immediately reload its freshly created schema from backend, but remote nodes can still reload schema after cache invalidation. So the backend userdata serialization issue still exists; it is just outside the current PR scope.

3. Please add a short compatibility note.

   Something like:

   > This PR does not change persisted schema data. It only changes the runtime schema-cache-clear meta-event value so new events can carry a per-JVM `source` id. New consumers still accept legacy plain-string graph names.

Overall, the design now looks reasonable to me. The main remaining thing is to make the lifecycle and compatibility boundaries explicit so future readers do not assume the reconnect and `~create_time` issues are fully solved here.

dpol1:
Thanks @imbajin. Addressed all three points.

  **1. Reconnect lifecycle**                                                                                                                                            
   
  Confirmed: no `MetaManager`/`MetaDriver` reconnect callback exists today. `resetMetaListenerForReconnect()` is a manual hook only — clarified in                      
  `CachedSchemaTransactionV2` comment + Javadoc, TODO kept for wiring once a reconnect callback is exposed. This PR is not an automatic reconnect-recovery mechanism.
                                                                                                                                                                        
  **2. `~create_time` round-trip**

  Out of scope, filed as follow-up: #3013 .
                                                                                                                                                                        
  **3. Compatibility note**
                                                                                                                                                                        
  Added to PR description and as code comment at `MetaManager.SchemaCacheClearEvent.fromValue()` (legacy plain-string branch):
                                                                                                                                                                        
  > This PR does not change persisted schema data. It only changes the runtime schema-cache-clear meta-event value so new events can carry a per-JVM `source` id. New 
  consumers still accept legacy plain-string graph names, so mixed-version rollouts are safe.                                                                           
                             
  `MetaManagerSchemaCacheClearEventTest` already covers both formats (legacy plain-string + JSON with `graph`/`source`). Before/after diagram also in PR description.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -238,13 +369,25 @@ public void removeSchema(SchemaElement schema) {
...
-                                               this.graph().name());
+            this.notifySchemaCacheClear();
         }
     }
 
+    private void notifySchemaCacheClear() {
+        MetaManager.instance()
+                   .notifySchemaCacheClear(this.graph().graphSpace(),
+                                           this.graph().name(),
+                                           SCHEMA_CACHE_CLEAR_SOURCE);
```
imbajin, line 388: 🧹 **Nit (non-blocking): consider adding a `@VisibleForTesting` or explicit visibility note on `resetMetaListenerForReconnect()`**

The method is `public static` but the Javadoc makes clear it's an escape hatch, not part of the stable API. A `@VisibleForTesting` (Guava) or at least a `@apiNote` line clarifying "internal recovery hook; do not call from application code" would prevent external callers from building on what's intended as a temporary shim until MetaManager exposes a reconnect callback.

```suggestion
    /**
     * Manually reset the JVM-global meta listener flag after detecting that
     * the MetaManager transport reconnected and dropped the underlying gRPC
     * watch. This method is not wired to a MetaManager/MetaDriver reconnect
     * callback today; callers must invoke it explicitly after detecting that
     * condition. Without such a manual reset {@link #metaEventListenerRegistered}
     * would stay {@code true} forever and this JVM would stop receiving
     * cross-node schema cache clear events with no error or warning.
     *
     * <p>TODO: wire this into MetaManager once it exposes a transport
     * reconnect callback (e.g. {@code listenReconnect} /
     * {@code onTransportReconnect}). Until then it must be invoked
     * explicitly by code that detects the reconnect.
     *
     * @apiNote Internal recovery hook. Not part of the stable API — external
     *          callers should not depend on this method; it may be replaced
     *          by an automatic reconnect callback in a future release.
     */
    public static void resetMetaListenerForReconnect() {
```

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedSchemaTransactionTest.java
```diff
@@ -165,6 +187,528 @@ public void testGetSchema() throws Exception {
...
+        // cleared and that other graphs in the same JVM are left untouched.
+        String targetGraph = "DEFAULT-meta-target-v2";
+        String otherGraph = "DEFAULT-meta-other-v2";
+
+        Cache<Id, Object> targetIdCache =
+                CacheManager.instance().cache("schema-id-" + targetGraph, 10L);
+        Cache<Id, Object> targetNameCache =
+                CacheManager.instance()
+                            .cache("schema-name-" + targetGraph, 10L);
+        Cache<Id, Object> otherIdCache =
```
imbajin, line 461: 🧹 **Nit (non-blocking): the TODO references `#2617` but that's the parent issue, not a dedicated follow-up for the IT gap**

The comment says `TODO(#2617): port these assertions into a dedicated CachedSchemaTransactionV2IT`. Since #2617 is about to close when this PR lands, the TODO will point to an issue nobody's tracking anymore. Consider either:

1. File a new follow-up issue for the integration test specifically (like you did for #3013 / #3012), and reference that here; or
2. Drop the issue number and just leave the rationale inline so future readers understand the gap without chasing a closed ticket.

Not a merge blocker — just a small housekeeping item to keep the TODO actionable after merge.

imbajin (commented):
## Merge-readiness audit

Re-reviewed against the 10 comments from my earlier pass + Copilot's 2 comments, cross-checked the three clarifications in your yesterday reply. Summary: **all blocking issues resolved**, CI green on 13/13 jobs, mergeable=true. Two non-blocking TODOs remain (documented in-code).

### Before-merge review status

| # | Prior issue | Severity | Status in head commit |
|---|---|---|---|
| 1 | `updateSchema()` broadcast storm | ‼️ Critical | ✅ Fixed — notify removed, comment explains why |
| 2 | Static `AtomicBoolean` has no reset path | ‼️ Critical | ⚠️ Mitigated — `resetMetaListenerForReconnect()` + WARN log + Javadoc + TODO; explicitly scoped as manual hook (confirmed in reply) |
| 3 | AtomicBoolean CAS-before-register race | ⚠️ Important | ✅ Fixed — `synchronized(META_LISTENER_LOCK)` + `set(true)` after `listenSchemaCacheClear` returns |
| 4 | Narrow `catch (RuntimeException)` | ⚠️ Important | ✅ Fixed — now catches `Exception`, wraps checked as RuntimeException |
| 5 | `metaDriver()` leak from MetaManager | ⚠️ Important | ✅ Fixed — new `extractSchemaCacheClearEventsFromResponse()` encapsulates extraction |
| 6 | `clearSchemaCache` non-atomic TOCTOU | ⚠️ Important | ✅ Mitigated — name-first ordering in both static+instance paths; comment spells out why |
| 7 | `TASK_SYNC_DELETION` semantic mismatch | 🧹 Minor | ✅ Fixed — `addSchema` always notifies; `maybeNotifySchemaCacheClear` only gates removal |
| 8 | End-to-end listener path untested | ⚠️ Important | ✅ Fixed — `handleSchemaCacheClearEvent` extracted package-private, 5 new tests mock MetaDriver and drive publish→consume→clear |
| 9 | Copilot: array-attachment assertion missing | ⚠️ Important | ✅ Fixed — `testClearSchemaCacheClearsArrayAttachmentMaps` verifies pks/vls/els/ils + cachedTypes emptied |
| 10 | Copilot: test cleanup not in try/finally | ⚠️ Important | ✅ Fixed — all new tests wrap assertions in try/finally |

### Your three clarifications (image reply) — code evidence

1. **Reconnect lifecycle** → confirmed in `CachedSchemaTransactionV2.java:51-56` (class-level comment) and `:253-267` (Javadoc on `resetMetaListenerForReconnect`). Both say manual-only, TODO to wire once MetaManager exposes callback. ✅
2. **`~create_time` round-trip** → out of scope, filed as #3013. ✅
3. **Compatibility note** → in PR description and `MetaManager.SchemaCacheClearEvent.fromValue()` legacy branch comment (`MetaManager.java:1355-1358`). `MetaManagerSchemaCacheClearEventTest` covers plain-string + JSON formats. ✅

### CI

All 13 checks green: dependency-check, CodeQL, commons (11), struct, server (memory/rocksdb/hbase), license, dep-review, pd, store, hstore. Last run 2026-05-05 on head SHA `beef815`.

### New test coverage added

- `testListenSchemaCacheClearIsIdempotent` — 8-thread race on the JVM-global flag, asserts short-circuit + zero exceptions
- `testClearSchemaCacheClearsArrayAttachmentMaps` — pks/vls/els/ils + cachedTypes drained
- `testHandleSchemaCacheClearEventSkipsLocalSource` — self-echo filter verified
- `testHandleSchemaCacheClearEventClearsTargetGraphOnly` — cross-graph isolation
- `testHandleSchemaCacheClearEventNullGraphsIsNoop`
- `testHandleSchemaCacheClearEventClearsMultipleGraphs`
- `MetaManagerSchemaCacheClearEventTest` (6 cases) — legacy + JSON payload parsing

### Non-blocking follow-ups (already tracked in code)

- `CachedSchemaTransactionV2.java:388` TODO — wire `resetMetaListenerForReconnect()` into MetaManager reconnect callback once exposed.
- `CachedSchemaTransactionTest.java:454-461` TODO(#2617) — add `CachedSchemaTransactionV2IT` once `mockito-inline` is on the test classpath, so `TASK_SYNC_DELETION=true` removal suppression + unconditional `addSchema` notify can be asserted at the instance level (not just the static helper). Currently exercised only by hstore integration.

### Verdict

**✅ LGTM — approve for merge.**

The PR direction was correct from the start; the three rounds of iteration closed every blocking concern. The two remaining gaps (automatic reconnect recovery, instance-level `TASK_SYNC_DELETION` unit test) are explicitly scoped out with in-code TODOs and do not regress any existing behavior. Mixed-version rollout is safe because `SchemaCacheClearEvent.fromValue()` still accepts legacy plain-string values.

The two inline comments below are minor polish suggestions, not merge blockers.

imbajin (approved):
✅ LGTM , feel free to follow up on the two in-code TODOs at your pace

PR #1024, opened 2020-06-08
add copy-schema support

javeme:
Change-Id: I1b1393c589d8a928b3cb1805a6dc264563d14dc1

PR #3017, opened 2026-05-08
fix(server): align cache event actions in legacy EventHub path

### LINKED ISSUE #3012: [Bug] EventHub action name mismatch silently drops cache invalidation events in legacy path
## Summary

While working on #3011, @dpol1 spotted a potential silent-drop bug in the legacy `EventHub` cache-notification path. This issue tracks the investigation and fix.

## Root Cause

`Cache.java` defines four string constants:

```java
// present-tense — "command" semantics
String ACTION_INVALID   = "invalid";
String ACTION_CLEAR     = "clear";

// past-tense — "event happened" semantics
String ACTION_INVALIDED = "invalided";
String ACTION_CLEARED   = "cleared";
```

**Producers emit past-tense events:**

| File | Lines | Emitted action |
|------|-------|----------------|
| `CachedGraphTransaction.java` | 400, 414, 428 | `ACTION_INVALIDED`, `ACTION_CLEARED` |
| `CachedSchemaTransaction.java` | 182, 224, 233, 286 | `ACTION_INVALIDED`, `ACTION_CLEARED` |

**Local listeners only match present-tense events:**

| File | Lines | Listened action |
|------|-------|----------------|
| `CachedGraphTransaction.java` | 147, 179 | `ACTION_INVALID`, `ACTION_CLEAR` |
| `CachedSchemaTransaction.java` | 120, 127 | `ACTION_INVALID`, `ACTION_CLEAR` |
| `CachedSchemaTransactionV2.java` | 117, 137 | `ACTION_INVALID`, `ACTION_CLEAR` |

## How the Current Bridge Works

`StandardHugeGraph.AbstractCacheNotifier` (`StandardHugeGraph.java:1397`) listens for `"invalided"`/`"cleared"` and converts them into RPC calls to remote nodes. Remote nodes then call `AbstractCacheNotifier.invalid/clear()`, which emits `"invalid"`/`"clear"` on their local hubs — which local listeners do catch.

```
Local transaction commits
  → notifyChanges("invalided") on local schemaEventHub
  → AbstractCacheNotifier.listener catches "invalided"
  → proxy.invalid(type, id)  ← RPC call to remote nodes only
  → Remote node: AbstractCacheNotifier.invalid() emits "invalid" on remote hub
  → Remote CachedSchemaTransaction.listener catches "invalid" ✓
```

## Where Events Are Silently Dropped

The bridge **only routes to remote nodes via RPC**. The local node never re-emits `"invalid"` after `"invalided"` is published, which means:

1. **Single-node / no-RPC deployments**: `proxy` is a no-op; `"invalided"` events from `CachedSchemaTransaction`/`CachedGraphTransaction` are consumed by `AbstractCacheNotifier` and go nowhere. Local consistency is currently maintained by direct `updateCache()` calls inside each transaction, but any secondary consumer that relies solely on the EventHub would miss these events.

2. **If the RPC proxy is misconfigured or unavailable**: The `"invalided"` event is silently swallowed — no error, no fallback.

3. **Concurrent local transactions**: Due to the `containsListener` guard (only one listener per hub/event), if a second transaction instance needs to be notified via EventHub it will not be — a latent hazard in scenarios where multiple transactions share the same EventHub.

## Suggested Fix

Align the action constants: either
- Have producers emit present-tense (`ACTION_INVALID` / `ACTION_CLEAR`) and remove the past-tense constants, or
- Have `AbstractCacheNotifier` also re-publish `"invalid"`/`"clear"` locally after forwarding the RPC, so local listeners are always notified.

The safer approach is option 1: unify on present-tense `ACTION_INVALID` / `ACTION_CLEAR` throughout, since RaftContext (`RaftContext.java:287,313`) already uses the present-tense constants correctly.

## Related

- PR #3011 (HStore `CachedSchemaTransactionV2` fix that surfaced this)
- Reported by @dpol1 in https://github.com/apache/hugegraph/pull/3011#issuecomment-4334740756

dpol1:
<!-- 
  Thank you very much for contributing to Apache HugeGraph, we are happy that you want to help us improve it!

  Here are some tips for you:
    1. If this is your first time, please read the [contributing guidelines](https://github.com/apache/hugegraph/blob/master/CONTRIBUTING.md)

    2. If a PR fix/close an issue, type the message "close xxx" (xxx is the link of related 
issue) in the content, GitHub will auto link it (Required)

    3. Name the PR title in "Google Commit Format", start with "feat | fix | perf | refactor | doc | chore", 
      such like: "feat(core): support the PageRank algorithm" or "fix: wrong break in the compute loop" (module is optional)
      skip it if you are unsure about which is the best component.

    4. One PR address one issue, better not to mix up multiple issues.

    5. Put an `x` in the `[ ]` to mark the item as CHECKED. `[x]` (or click it directly after 
published)
-->

## Purpose of the PR

- close #3012 <!-- or use "fix #xxx", "xxx" is the ID-link of related issue, e.g: close #1024 -->

<!--
Please explain more context in this section, clarify why the changes are needed. 
e.g:
- If you propose a new API, clarify the use case for a new API.
- If you fix a bug, you can clarify why it is a bug, and should be associated with an issue.
-->

Cache producers in the legacy `EventHub` path were emitting past-tense action                                                                                         
  names (`invalided` / `cleared`) while local cache listeners only matched the
  present-tense actions (`invalid` / `clear`). On single-node deployments, or                                                                                           
  whenever the cache RPC bridge in `StandardHugeGraph.AbstractCacheNotifier` was                                                                                        
  not active, those local cache invalidation events were silently dropped: no                                                                                           
  error, no warning, just stale cache.                                                                                                                                  
                                                                                                                                                                        
  The fix aligns producers, listeners, and the bridge on the present-tense                                                                                              
  constants, removes the obsolete past-tense ones, and wires `notifyExcept(...)`
  so the bridge does not loop back into itself once both ends speak the same                                                                                            
  action name.                                                                                                                                                          
                                                                                                                                                                        
  ## Main Changes                                                                                                                                                       
                                                            
  - Drop `Cache.ACTION_INVALIDED` / `Cache.ACTION_CLEARED`. All producers                                                                                               
    (`CachedGraphTransaction`, `CachedSchemaTransaction`) now emit
    `Cache.ACTION_INVALID` / `Cache.ACTION_CLEAR`, matching what local listeners                                                                                        
    and `RaftContext` already expect.                                                                                                                                   
  - Update `StandardHugeGraph.AbstractCacheNotifier` to listen on the                                                                                                   
    present-tense actions and to forward through `EventHub.notifyExcept(...)` so                                                                                        
    the bridge does not re-receive its own re-emit.                                                                                                                     
  - Add `EventHub.notifyExcept(event, ignoredListener, args...)` so a caller can                                                                                        
    notify everyone on the hub except the listener that originated the event.                                                                                           
    The signature is documented as a public API on `EventHub`, `notify(...)`                                                                                            
    is unchanged and now delegates internally.                                                                                                                          
  - Use `notifyExcept(...)` from the cache transactions and from                                                                                                        
    `AbstractCacheNotifier`, with the registered cache listener as the ignored                                                                                          
    one, so producers do not feed their own listener and the RPC re-emit does                                                                                           
    not loop the bridge.                                                                                                                                                
  - Track the registered cache listener per graph in
    `GraphTransaction.graphCacheEventListeners` and in a matching map inside
    `CachedSchemaTransaction`, with owner-aware teardown on close. This is the
    minimum needed for `notifyExcept(...)` to exclude the listener that is
    actually registered on the hub: with the old per-instance lambda +
    `containsListener(...)` / boolean guard pattern, only the first transaction
    per graph registered a listener, while later transactions could call
    `notifyExcept(...)` with a lambda that was never registered. In that case the
    real registered listener was not excluded and could still process the
    transaction's own local emit.                                                                                                      
                                                                                                                                                                        
  `CachedSchemaTransactionV2` is intentionally not touched here. That class was
  just adjusted in #3011 for HStore/V2 schema cache propagation through
  `MetaManager`. The remaining `containsListener(...)` guard there is a related
  listener-lifecycle concern, not the action-name mismatch fixed in this PR; it
  should be handled in a follow-up that audits the remaining V2 path and, if
  needed, moves cache listener ownership to a ref-counted holder keyed by graph
  identity and `EventHub`.                                                                                                                                                    
                                                                                                                                                                        
  ## Verifying these changes                                                                                                                                            
                                                            
      - `EventHubTest#testNotifyExcept` covers the new `notifyExcept(...)`
        semantics, including `ANY_EVENT` listeners and the ignored-listener                                                                                             
        exclusion.                                                                                                                                                      
      - `CachedSchemaTransactionTest#testLegacySchemaChangeEmitsActionInvalid`                                                                                          
        asserts the schema producer now emits `ACTION_INVALID` end-to-end.                                                                                              
      - `CachedSchemaTransactionTest#testCacheListenerSurvivesOwnerClose`
          asserts that closing the owner transaction does not drop the hub
          listener while a second transaction is still alive, and that cache
          invalidation events are still delivered to the surviving transaction.
      - `CachedSchemaTransactionTest#testLastCloseRemovesSchemaCacheListener`
        asserts that the hub listener is unregistered and the registry entry
        removed only when the last transaction for that graph closes, and that
        a subsequent graph open registers a fresh holder.
      - `CachedGraphTransactionTest#testCacheListenerSurvivesOwnerClose`
        same owner-first-close guarantee for the graph vertex/edge cache path.
      - `CachedGraphTransactionTest#testLastCloseRemovesGraphCacheListener`
        same last-close cleanup guarantee for the graph cache path.                                                                                                              
      - `CachedSchemaTransactionTest#testCacheNotifierLocalEventCallsProxyOnce`                                                                                         
        and `testCacheNotifierRpcInvalidDoesNotLoopToProxy` cover the bridge                                                                                            
        no-loop behavior.                                                                                                                                               
       - `CachedGraphTransactionTest#testClearCacheEmitsActionClear` asserts
          graph cache clears now emit `ACTION_CLEAR`.
       - `CachedGraphTransactionTest#testVertexMutationEmitsActionInvalid`
          asserts graph vertex mutations now emit `ACTION_INVALID`.
       - `CachedGraphTransactionTest#testClosingNonOwnerKeepsGraphCacheListenerRegistered`
          asserts that closing a graph transaction which reused an existing
          per-graph listener does not unregister the shared graph cache listener.
                                                                                              
                                                                                                                                                                        
  - [ ] Dependencies                                        
  - [ ] Modify configurations                                                                                                                                           
  - [x] The public API                                      
  - [ ] Other affects
  - [ ] Nope                                                                                                                                                            
   
  `EventHub.notifyExcept(...)` is a new public method; existing `notify(...)`                                                                                           
  behavior is unchanged.                                    
                                                                                                                                                                        
  ## Documentation Status                                   

  - [ ] `Doc - TODO`
  - [ ] `Doc - Done`
  - [x] `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CacheListenerHolder.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CacheListenerHolder.java
@@ -29,6 +29,8 @@ final class CacheListenerHolder {
 
     final EventListener listener;
     final EventHub hub;
+    // Must only be read or written inside ConcurrentMap.compute() for the
+    // enclosing registry; ConcurrentHashMap.compute() serialises per-key access.
     int refCount;
 
     CacheListenerHolder(EventListener listener, EventHub hub) {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CacheListenerHolder.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CacheListenerHolder.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hugegraph.backend.cache;
+
+import org.apache.hugegraph.event.EventHub;
+import org.apache.hugegraph.event.EventListener;
+
+/*
+ * Listener lifetime must cover all active transactions for the graph.
+ * The holder is removed from the registry and unregistered from EventHub
+ * only when the last transaction releases it.
+ */
+final class CacheListenerHolder {
+
+    final EventListener listener;
+    final EventHub hub;
+    int refCount;
+
+    CacheListenerHolder(EventListener listener, EventHub hub) {
+        this.listener = listener;
+        this.hub = hub;
+        this.refCount = 1;
+    }
+}
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
@@ -68,7 +68,7 @@ public final class CachedGraphTransaction extends GraphTransaction {
      * only when the last transaction releases it.
      */
     private static final ConcurrentMap<String, CacheListenerHolder>
-            graphCacheEventListeners = new ConcurrentHashMap<>();
+            GRAPH_CACHE_EVENT_LISTENERS = new ConcurrentHashMap<>();
 
     private final Cache<Id, Object> verticesCache;
     private final Cache<Id, Object> edgesCache;
@@ -197,7 +197,7 @@ public final class CachedGraphTransaction extends GraphTransaction {
         };
         EventHub graphEventHub = this.params().graphEventHub();
         String graphName = this.params().spaceGraphName();
-        CacheListenerHolder acquired = graphCacheEventListeners.compute(
+        CacheListenerHolder acquired = GRAPH_CACHE_EVENT_LISTENERS.compute(
                 graphName, (key, existing) -> {
                     if (existing == null || existing.hub != graphEventHub) {
                         // Graph close/reopen creates a new EventHub for the
@@ -221,7 +221,7 @@ public final class CachedGraphTransaction extends GraphTransaction {
         String graphName = this.params().spaceGraphName();
         CacheListenerHolder ours = this.holder;
         if (ours != null) {
-            graphCacheEventListeners.compute(graphName, (key, existing) -> {
+            GRAPH_CACHE_EVENT_LISTENERS.compute(graphName, (key, existing) -> {
                 if (existing == null || existing != ours) {
                     return existing;
                 }
@@ -236,7 +236,7 @@ public final class CachedGraphTransaction extends GraphTransaction {
             this.cacheEventListener = null;
         }
         // TODO (follow-up): storeEventListenStatus has the same owner-first
-        // close bug this PR fixes for graphCacheEventListeners. A non-owner
+        // close bug this PR fixes for GRAPH_CACHE_EVENT_LISTENERS. A non-owner
         // transaction can remove the tracking entry, unlisten its own
         // never-registered storeEventListener as a no-op, and leave the
         // original store listener registered but untracked. Apply the same
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
@@ -24,6 +24,8 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import org.apache.hugegraph.HugeGraphParams;
 import org.apache.hugegraph.backend.cache.CachedBackendStore.QueryId;
@@ -60,12 +62,20 @@ public final class CachedGraphTransaction extends GraphTransaction {
     private static final long AVG_VERTEX_ENTRY_SIZE = 40L;
     private static final long AVG_EDGE_ENTRY_SIZE = 100L;
 
+    /*
+     * Listener lifetime must cover all active transactions for the graph.
+     * The holder is removed from the registry and unregistered from EventHub
+     * only when the last transaction releases it.
+     */
+    private static final ConcurrentMap<String, CacheListenerHolder>
+            graphCacheEventListeners = new ConcurrentHashMap<>();
+
     private final Cache<Id, Object> verticesCache;
     private final Cache<Id, Object> edgesCache;
 
     private EventListener storeEventListener;
     private EventListener cacheEventListener;
-    private boolean registeredCacheEventListener;
+    private CacheListenerHolder holder;
 
     public CachedGraphTransaction(HugeGraphParams graph, BackendStore store) {
         super(graph, store);
@@ -186,29 +196,51 @@ public final class CachedGraphTransaction extends GraphTransaction {
             return false;
         };
         EventHub graphEventHub = this.params().graphEventHub();
-        EventListener previous =
-                graphCacheEventListeners.putIfAbsent(
-                        this.params().spaceGraphName(), listener);
-        if (previous == null) {
-            this.cacheEventListener = listener;
-            this.registeredCacheEventListener = true;
-            graphEventHub.listen(Events.CACHE, this.cacheEventListener);
-        } else {
-            this.cacheEventListener = previous;
-            this.registeredCacheEventListener = false;
-        }
+        String graphName = this.params().spaceGraphName();
+        CacheListenerHolder acquired = graphCacheEventListeners.compute(
+                graphName, (key, existing) -> {
+                    if (existing == null || existing.hub != graphEventHub) {
+                        // Graph close/reopen creates a new EventHub for the
+                        // same graph name; replace the stale holder. Old
+                        // transactions skip decrement via identity check.
+                        if (existing != null) {
+                            existing.hub.unlisten(Events.CACHE,
+                                                  existing.listener);
+                        }
+                        graphEventHub.listen(Events.CACHE, listener);
+                        return new CacheListenerHolder(listener, graphEventHub);
+                    }
+                    existing.refCount++;
+                    return existing;
+                });
+        this.holder = acquired;
+        this.cacheEventListener = acquired.listener;
     }
 
     private void unlistenChanges() {
         String graphName = this.params().spaceGraphName();
-        if (this.registeredCacheEventListener) {
-            EventHub graphEventHub = this.params().graphEventHub();
-            if (graphCacheEventListeners.remove(graphName,
-                                                this.cacheEventListener)) {
-                graphEventHub.unlisten(Events.CACHE, this.cacheEventListener);
-            }
-            this.registeredCacheEventListener = false;
+        CacheListenerHolder ours = this.holder;
+        if (ours != null) {
+            graphCacheEventListeners.compute(graphName, (key, existing) -> {
+                if (existing == null || existing != ours) {
+                    return existing;
+                }
+                existing.refCount--;
+                if (existing.refCount == 0) {
+                    existing.hub.unlisten(Events.CACHE, existing.listener);
+                    return null;
+                }
+                return existing;
+            });
+            this.holder = null;
+            this.cacheEventListener = null;
         }
+        // TODO (follow-up): storeEventListenStatus has the same owner-first
+        // close bug this PR fixes for graphCacheEventListeners. A non-owner
+        // transaction can remove the tracking entry, unlisten its own
+        // never-registered storeEventListener as a no-op, and leave the
+        // original store listener registered but untracked. Apply the same
+        // ref-counted holder pattern in a follow-up PR.
         if (storeEventListenStatus.remove(graphName) != null) {
             this.store().provider().unlisten(this.storeEventListener);
         }
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
@@ -65,6 +65,7 @@ public final class CachedGraphTransaction extends GraphTransaction {
 
     private EventListener storeEventListener;
     private EventListener cacheEventListener;
+    private boolean registeredCacheEventListener;
 
     public CachedGraphTransaction(HugeGraphParams graph, BackendStore store) {
         super(graph, store);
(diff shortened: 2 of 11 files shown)


dpol1:
@imbajin  I took the liberty to quickly fix a flaky test which failed a workflow twice - if you think is out of the scope with this PR I' ll revert it - Actually, I just noticed the first commit is all greens

imbajin (commented):
Fix looks correct and the new tests are solid. Two concerns worth tracking: (1) `storeEventListenStatus` has the same ownership bug this PR fixes for the cache listener, acknowledged in the test's `finally` block — worth a follow-up; (2) the owner-first-close scenario for the cache listener deserves an explicit test and/or a ref-counted ownership model.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java
```diff
@@ -138,6 +147,96 @@ public void testEventInvalid() throws Exception {
...
+
+            Assert.assertSame(registered, cacheListeners.get(graphName));
+            Assert.assertTrue(this.params.graphEventHub()
+                                         .listeners(Events.CACHE)
+                                         .contains(registered));
+        } finally {
+            // Closing the secondary transaction exercises the pre-existing
+            // store listener guard too; restore it so teardown unregisters the
+            // primary transaction's store listener.
+            storeListeners.putIfAbsent(graphName, true);
```
imbajin: ⚠️ TODO (follow-up): this `putIfAbsent` is a workaround for a pre-existing bug in `storeEventListenStatus` that mirrors exactly the one this PR fixes for `graphCacheEventListeners`. In `unlistenChanges()`:

```java
if (storeEventListenStatus.remove(graphName) != null) {
    this.store().provider().unlisten(this.storeEventListener);
}
```

when a non-owner transaction closes, it (a) pops the map entry even though it doesn't own the registration, (b) unlistens its own per-instance `storeEventListener` which was never registered on the provider (no-op), and (c) leaves the primary's real store listener registered but un-tracked — so the owner's later close silently skips the unlisten and leaks it.

Suggest a follow-up PR that applies the same ref-counted / owner-aware pattern introduced here for the cache listener (i.e. a `Map<String, EventListener>` + `registeredStoreEventListener` flag). Keeping as TODO for now since it's out of scope.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -184,17 +185,29 @@ private void listenChanges() {
...
         String graphName = this.params().spaceGraphName();
-        if (graphCacheListenStatus.remove(graphName) != null) {
+        if (this.registeredCacheEventListener) {
             EventHub graphEventHub = this.params().graphEventHub();
-            graphEventHub.unlisten(Events.CACHE, this.cacheEventListener);
+            if (graphCacheEventListeners.remove(graphName,
+                                                this.cacheEventListener)) {
+                graphEventHub.unlisten(Events.CACHE, this.cacheEventListener);
+            }
+            this.registeredCacheEventListener = false;
```
imbajin: ‼️ What happens in the reverse ordering — owner closes first while non-owners are still alive? The new tests only cover non-owner close.

Trace:
1. T1 opens (owner), registers `listener`, puts it in `graphCacheEventListeners`.
2. T2 opens (non-owner), caches `this.cacheEventListener = <T1's listener>`.
3. T1 closes → `unlistenChanges()` removes the map entry and calls `hub.unlisten(...)`.
4. T2 still alive. Its `notifyChanges(...)` calls `notifyExcept(hub, <T1's listener>, ...)`. That listener is no longer on the hub, so the exclusion is a no-op and the emit fans out to whatever else is registered (e.g. `AbstractCacheNotifier`'s bridge). Meanwhile T2 has no working local cache listener at all — the per-graph listener slot is empty until a new transaction repopulates it.

Two questions:
- Is T1-closes-before-T2 a realistic lifecycle in practice (e.g. thread-local transactions on a long-lived graph)? If yes, a ref-counted ownership holder (unregister only when the last referent closes) would be safer than "first-wins" ownership.
- Either way, worth adding a companion test to `testClosingNonOwnerKeepsGraphCacheListenerRegistered` that exercises the owner-first-close path and asserts whatever the intended semantics are, so the invariant is captured.

dpol1: Both fair, thanks.                                                                                                                                                    
                                                                       
1. `storeEventListenStatus` - same ownership bug, the `putIfAbsent` in `finally` is a workaround. Follow-up PR with whatever ownership model we land on.                 
   
2. Owner-first-close - realistic, yeah. Two options: ref-counted holder, or stop unregistering on tx close (cache is already a `CacheManager` singleton per `spaceGraphName`, so listener lifetime = graph, not tx; cleanup moves to graph dispose). I'd lean toward the second - kills the race by killing the owner concept.
                                                                                                                                                                        
  Wouldn't add a test pinning current behavior - codifies a limitation. Real test needs the fix first.                                                                  
   
  #3012 is the action-name + `notifyExcept` fix, so my preference is to merge this as-is and do ownership in a follow-up. Can fold either fix in here instead - your call.

dpol1: Address by b77fb66

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java
dpol1: Address by 80bd29c

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -60,11 +62,20 @@ public final class CachedGraphTransaction extends GraphTransaction {
     private static final long AVG_VERTEX_ENTRY_SIZE = 40L;
     private static final long AVG_EDGE_ENTRY_SIZE = 100L;
 
+    /*
+     * Listener lifetime must cover all active transactions for the graph.
+     * The holder is removed from the registry and unregistered from EventHub
+     * only when the last transaction releases it.
+     */
+    private static final ConcurrentMap<String, CacheListenerHolder>
+            graphCacheEventListeners = new ConcurrentHashMap<>();
```
imbajin: ⚠️ **Naming inconsistency with `SCHEMA_CACHE_EVENT_LISTENERS`**

Both maps are `private static final` and should follow `SCREAMING_SNAKE_CASE` per Java convention, but this one uses `lowerCamelCase` while the schema counterpart in `CachedSchemaTransaction` does not. Suggest renaming to `GRAPH_CACHE_EVENT_LISTENERS` for consistency.

```suggestion
        GRAPH_CACHE_EVENT_LISTENERS = new ConcurrentHashMap<>();
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CacheListenerHolder.java
```diff
@@ -0,0 +1,39 @@
...
+/*
+ * Listener lifetime must cover all active transactions for the graph.
+ * The holder is removed from the registry and unregistered from EventHub
+ * only when the last transaction releases it.
+ */
+final class CacheListenerHolder {
+
+    final EventListener listener;
+    final EventHub hub;
+    int refCount;
```
imbajin, line 34: ⚠️ **`refCount` is a bare `int` with no synchronization contract documented**

The field is safe today because all mutations happen inside `ConcurrentMap.compute()`, which serialises per-key — but that invariant is invisible to future maintainers. A stray read/write outside `compute()` would introduce a data race with no compiler warning.

Either document the contract:

```suggestion
    // Must only be read or written inside ConcurrentMap.compute() for the
    // enclosing registry; ConcurrentHashMap.compute() serialises per-key access.
    int refCount;
```

Or use `AtomicInteger` — consistent with `BackendSessionPool.sessionCount` and other ref-counted fields in the codebase — which makes the intent self-documenting.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java
```diff
@@ -138,6 +186,176 @@ public void testEventInvalid() throws Exception {
...
+        try {
+            second.close();
+
+            Assert.assertSame(holder, cacheListeners.get(graphName));
+            Assert.assertEquals(refCount, holderRefCount(holder));
+            Assert.assertTrue(this.params.graphEventHub()
+                                         .listeners(Events.CACHE)
+                                         .contains(registered));
+        } finally {
+            // Closing the secondary transaction exercises the pre-existing
```
imbajin: ⚠️ **Test manipulates internal production state to compensate for a known bug**

`storeListeners.putIfAbsent(graphName, true)` restores `storeEventListenStatus` so that `teardown` can unlisten the primary store listener — but only because closing `second` already consumed the map entry (the ownership-first-close bug acknowledged in the TODO inside `unlistenChanges`).

This means the test is silently accepting broken teardown behaviour rather than isolating the intended scenario. If `putIfAbsent` fails for any reason, teardown will fail with an error unrelated to the test's actual assertion.

Consider adding a comment linking this workaround to the follow-up PR that will fix `storeEventListenStatus`, or isolating it in a shared helper that makes the intent explicit.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CacheManagerTest.java
```diff
@@ -285,17 +288,23 @@ public void testCacheExpire() {
...
-        cache2.update(IdGenerator.of("fake-id"), "fake-value");
+        TimerTask task = Whitebox.invoke(CacheManager.class, "scheduleTimer",
+                                         manager, 0.1F);
+        try {
+            cache1.update(IdGenerator.of("fake-id"), "fake-value",
+                          -30 * 1000L);
+            cache2.update(IdGenerator.of("fake-id"), "fake-value");
 
-        waitTillNext(40);
+            waitTillNext(1);
```
imbajin: 🧹 **`waitTillNext(1)` may be flaky under CI load**

`scheduleTimer(0.1F)` creates a 100 ms `Timer`, and 1 second should give ~10 ticks in theory. However, `Timer` uses a single background thread with no scheduling-lag compensation, and `waitTillNext` is a spin-wait rather than a latch. Under a CPU-starved CI runner the timer thread can be delayed enough that fewer ticks fire within the window.

Consider extending the wait to 3 seconds for a reasonable safety margin, or replacing the spin-wait with a `CountDownLatch` triggered from inside the `tick()` mock.

imbajin (commented):
No blocking issues. The `CacheListenerHolder` ref-count pattern correctly solves listener lifecycle under concurrent transactions, and the action-name alignment fixes the long-standing silent cache invalidation. Four items worth cleaning up before merge: a naming inconsistency on the graph-side registry, an undocumented synchronization contract on `refCount`, a fragile test workaround for the acknowledged `storeEventListenStatus` bug, and a potential timing flakiness in `testCacheExpire`.

dpol1:
Thanks @imbajin, solved!

imbajin (approved):
THX~

PR #3024, opened 2026-05-12
feat(rocksdb): support ToplingDB provider

WaterWhisperer:
## Purpose of the PR

Add ToplingDB support as an optional RocksDB-compatible storage path.

This PR rebases and prepares the existing ToplingDB adaptation work for review on the current `master`. ToplingDB is integrated as an optional provider for RocksDB-based storage, allowing users to enable RocksDB/ToplingDB configuration and the ToplingDB Web UI while preserving the standard RocksDB path.

## Main Changes

- Add `hugegraph-rocksdb-provider` module
- Update HugeGraph Server RocksDB integration
- Add ToplingDB preload support in server dist
- Add PD / Store ToplingDB integration
- Add Maven / CI / dependency metadata updates
- Add ToplingDB docs and specs

## Verifying these changes

    - Built packaged dist artifacts for Server / PD / Store
    - Verified packaged `rocksdbjni-8.10.2-SNAPSHOT.jar` contains ToplingDB resources including `org/rocksdb/SidePluginRepo.class`
    - Ran Server with ToplingDB enabled
    - Ran Server without `rocksdb.option_path`
    - Ran PD / Store with ToplingDB config

- [x]  Dependencies ([add/update license](https://hugegraph.apache.org/docs/contribution-guidelines/contribute/#321-check-licenses) info & [regenerate_known_dependencies.sh](../install-dist/scripts/dependency/regenerate_known_dependencies.sh)) 
- [x]  Modify configurations
- [ ]  The public API
- [x]  Other affects (typed here)
  - Adds native library preload logic for ToplingDB-enabled RocksDB JNI runtime.
  - Adds optional ToplingDB Web UI support.
  - Adds RocksDB provider SPI and routes RocksDB opening through the provider loader.
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [x]  `Doc - Done` 
- [ ]  `Doc - No Need`


hugegraph-rocksdb-provider/src/main/java/org/apache/hugegraph/rocksdb/provider/AbstractRocksDBProvider.java
```diff
@@ -0,0 +1,176 @@
...
+        try {
+            // Delegate to provider-specific implementation
+            RocksDB rocksDB = doOpenRocksDB(options, dataPath);
+
+            return rocksDB;
+        } catch (Exception e) {
+            LOG.error("Failed to open RocksDB with provider: {} at path: {}",
+                      getProviderName(), dataPath, e);
+            throw e;
+        }
```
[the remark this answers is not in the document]
WaterWhisperer: wrong

pom.xml
```diff
@@ -91,6 +91,7 @@
         <hugegraph-commons.version>1.7.0</hugegraph-commons.version>
         <lombok.version>1.18.30</lombok.version>
         <release.name>hugegraph</release.name>
+        <rocksdbjni.version>8.10.2-SNAPSHOT</rocksdbjni.version>
```
imbajin: ‼️ **Default RocksDB now depends on a Topling snapshot**

The current dependency flow makes ToplingDB part of the default RocksDB path, not an opt-in provider:

```text
standard RocksDB user
  -> root pom selects rocksdbjni:8.10.2-SNAPSHOT
  -> build needs GitHub Packages
  -> Linux startup scripts source Topling preload logic
  -> native preload/runtime assumptions affect default startup
```

**Evidence**
- This line changes the global `rocksdbjni.version` to `8.10.2-SNAPSHOT`.
- `.github/configs/settings.xml` adds GitHub Packages as the source for that snapshot.
- `start-hugegraph.sh` / `init-store.sh` source `preload-topling.sh` unconditionally on the server startup path.

**Impact**
Users who did not configure `rocksdb.option_path` can still depend on a mutable snapshot artifact and Topling native preload behavior. That is risky for normal contributor builds, offline deployments, and release reproducibility.

**Suggested fix**
Keep the default dependency on a released standard `rocksdbjni` artifact. Select the Topling snapshot only through an explicit profile/configuration path, and only enter native preload when Topling is explicitly enabled.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/common-topling.sh
```diff
@@ -0,0 +1,299 @@
...
+        return 1
+    fi
+}
+
+function preload_toplingdb() {
+    local lib_dir="$1"
+    local dest_dir="$2"
+    local os_name
+
+    # NOTE: The current ToplingDB rocksdbjni snapshot bundles Linux x86_64 native libraries.
```
imbajin, line 254: ‼️ **Linux arm64 does not actually fall back to standard RocksDB**

The comment says the current ToplingDB snapshot bundles Linux x86_64 native libraries, but the implementation only checks the OS:

```text
Linux arm64/aarch64
  -> os_name == Linux
  -> extract .so from rocksdbjni snapshot
  -> preload librocksdbjni-linux64.so
  -> Java provider may still select Topling by class presence
```

**Evidence**
- This script skips only non-Linux platforms.
- It later preloads `librocksdbjni-linux64.so` without checking `uname -m`.
- `ToplingRocksDBProvider.isAvailable()` only checks for `org.rocksdb.SidePluginRepo`, not CPU architecture.
- The root POM only has a macOS profile that falls back to standard `rocksdbjni`; Linux arm64 still gets the Topling snapshot.

**Impact**
Linux arm64/aarch64 will not follow the documented fallback behavior. It can try to load x86_64 native libraries and fail during startup instead of using standard RocksDB.

**Suggested fix**
Add the same platform boundary in all three places:

```text
Maven dependency selection: Linux x86_64 -> Topling snapshot, otherwise standard rocksdbjni
Shell preload: return unless uname -s == Linux and uname -m is x86_64/amd64
Java provider availability: return false on unsupported OS/arch
```

Please also add at least one fallback check for Linux arm64 so this does not regress.

hugegraph-store/hg-store-rocksdb/src/main/java/org/apache/hugegraph/rocksdb/access/RocksDBSession.java
```diff
@@ -450,8 +451,11 @@ private void openRocksDB(String dbDataPath, long version) {
                         new ColumnFamilyDescriptor(RocksDB.DEFAULT_COLUMN_FAMILY, cfOptions));
             }
             List<ColumnFamilyHandle> columnFamilyHandleList = new ArrayList<>();
-            this.rocksDB = RocksDB.open(dbOptions, dbPath, columnFamilyDescriptorList,
-                                        columnFamilyHandleList);
+            this.rocksDB =
+                    RocksDBProviderLoader.openRocksDB(dbOptions, dbPath, columnFamilyDescriptorList,
```
imbajin, line 458: ‼️ **Store can start the Topling HTTP server once per RocksDB session**

Server already gates `open_http` to one logical DB, but Store passes the configured flag into every `RocksDBSession`:

```text
RocksDBFactory
  -> createGraphDB(db1) -> RocksDBSession -> startHttpServer()
  -> createGraphDB(db2) -> RocksDBSession -> startHttpServer()
  -> same YAML listening_ports
```

**Evidence**
- This call forwards `hugeConfig.get(RocksDBOptions.OPEN_HTTP)` directly.
- `RocksDBFactory.createGraphDB()` can create multiple sessions in the same Store process.
- `ToplingRocksDBProvider.startHttpServerIfNeeded()` calls `startHttpServer()` whenever `openHttp` is true.
- Server code has a `GRAPH_STORE` gate, but Store has no equivalent process-level gate.

**Impact**
Multiple Store RocksDB instances can compete for the same Topling HTTP port. Startup then depends on open order and may fail partway through DB initialization.

**Suggested fix**
Move Topling HTTP lifecycle to a process-level singleton, or add an explicit Store-side gating rule so only one chosen DB can start the HTTP server. Please cover the multi-session case in a regression test.

hugegraph-rocksdb-provider/src/main/java/org/apache/hugegraph/rocksdb/provider/ToplingRocksDBProvider.java
```diff
@@ -0,0 +1,614 @@
...
+                Class.forName(SIDE_PLUGIN_REPO_CLASS);
+                result = true;
+                LOG.info(
+                        "SidePluginRepo found. Will attempt to open default CF RocksDB using " +
+                        "Topling.");
+            } catch (ClassNotFoundException e) {
+                LOG.warn("SidePluginRepo not found, even though 'optionPath' was provided. " +
+                         "Falling back to the standard RocksDB default CF opening method. " +
+                         "The configuration in '{}' will be ignored.", optionPath);
+            } catch (Exception e) {
```
imbajin, line 476: ‼️ **Explicit Topling config errors should fail fast, not silently fall back**

When users configure `rocksdb.option_path`, the current behavior treats invalid Topling config as a warning and then opens standard RocksDB:

```text
rocksdb.option_path is set
  -> path/YAML validation fails
  -> warn only
  -> validateConfiguration() returns false
  -> standard RocksDB opens successfully
```

**Evidence**
- Path traversal, missing file, unreadable file, and invalid YAML are all caught here as generic exceptions.
- The method then returns `false`, which makes the caller use the standard RocksDB path.
- The provider module also hard-codes `./conf/` as the allowed deployment directory.

**Impact**
A typo or malformed YAML can make the service start successfully while not using ToplingDB at all. That is dangerous operationally because the user explicitly requested Topling behavior but gets a silent downgrade.

**Suggested fix**
Only allow fallback when Topling was not requested, or when the platform is explicitly unsupported by design. If `option_path` is explicitly configured and validation fails, startup should fail fast with a clear error. Path policy should be resolved in Server/PD/Store config layers and passed to the provider as a normalized path, rather than hard-coded in this shared provider.

hugegraph-store/hg-store-rocksdb/src/main/java/org/apache/hugegraph/rocksdb/access/RocksDBSession.java
imbajin, line 458: ‼️ **Snapshot verification still bypasses the provider path**

Normal open now goes through `RocksDBProviderLoader`, but this recovery path still uses raw RocksDB APIs:

```text
Topling DB / snapshot
  -> verifySnapshot()
  -> RocksDB.listColumnFamilies()
  -> RocksDB.openReadOnly()
  -> no option_path / SidePluginRepo initialization
```

**Evidence**
- Snapshot verification calls `RocksDB.listColumnFamilies()` directly here.
- It then calls `RocksDB.openReadOnly()` directly below.
- Store startup has a similar pre-open direct `listColumnFamilies()` path, and Server `listCFs()` also bypasses the provider.

**Impact**
If a Topling database or snapshot relies on YAML-defined factories/options, restart or snapshot restore can fail before the provider gets a chance to import the Topling configuration. This leaves lifecycle paths inconsistent with the new normal open path.

**Suggested fix**
Add provider-aware helpers for CF listing and read-only snapshot verification. At minimum, when `rocksdb.option_path` is set, initialize/import the Topling config before CF discovery and read-only verification. Please add regression coverage for Topling DB restart and snapshot verify/load.

imbajin (commented):
This review leaves 5 blocking concerns to resolve before merge: ToplingDB currently leaks into the default RocksDB build/startup path, and the platform fallback, Store HTTP lifecycle, explicit config failure semantics, and snapshot/CF lifecycle paths are not fully provider-aware. I verified the touched provider/rocksdb modules compile locally, so the remaining concerns are design/runtime coverage issues rather than simple compilation failures.

## Test coverage required before merge

The current test delta is not enough for a storage backend/provider change of this size. Please add tests in priority order so the fix remains rigorous and controllable:

```text
P0 merge blockers
  ├─ explicit Topling config failure semantics
  ├─ unsupported platform fallback
  ├─ Store multi-session open_http lifecycle
  └─ provider-aware restart / snapshot lifecycle

P1 provider contract
  ├─ provider selection and fallback ordering
  ├─ close / cleanup behavior
  └─ invalid path / invalid YAML handling

P2 packaging and script guards
  ├─ preload is opt-in or correctly gated
  └─ no accidental Topling dependency on standard RocksDB startup
```

### P0 - must cover before merge

1. **Explicit `rocksdb.option_path` failure must fail fast**
   - Missing file, unreadable file, invalid YAML, and path outside the allowed config root.
   - Expected result: if users explicitly configure `option_path`, startup fails clearly instead of silently falling back to standard RocksDB.

2. **Unsupported platform fallback**
   - At minimum, assert the platform gate for Linux arm64/aarch64 does not enter Topling preload/provider selection.
   - Expected result: non-supported OS/arch uses standard RocksDB and does not preload `librocksdbjni-linux64.so`.

3. **Store multi-session `open_http` lifecycle**
   - Create/open more than one Store `RocksDBSession` with `rocksdb.open_http=true`.
   - Expected result: Topling HTTP server starts at most once, or Store rejects the ambiguous configuration with a deterministic error.

4. **Provider-aware lifecycle paths**
   - Existing Topling DB restart: CF discovery must work through a provider-aware path.
   - Snapshot verify/load: read-only verification must initialize Topling config before opening the snapshot.
   - Expected result: normal open, restart, snapshot verify, and snapshot restore all follow the same provider contract.

### P1 - provider contract tests

5. **`RocksDBProviderLoader` selection and fallback**
   - Topling available + configured -> Topling selected.
   - Topling unavailable + not configured -> standard selected.
   - Topling unavailable + explicitly configured -> fail fast, not silent fallback.

6. **Provider cleanup**
   - Successful Topling open then close calls provider-specific cleanup (`closeAllDB()` path).
   - Failed open after partial registration cleans CF handles, RocksDB handle, and repo mapping.

### P2 - packaging / script regression tests

7. **Preload script gating**
   - Non-Linux exits cleanly.
   - Linux non-x86_64 exits cleanly.
   - Linux x86_64 only proceeds when Topling is intentionally enabled.

8. **Standard RocksDB startup remains independent**
   - With no `rocksdb.option_path`, standard RocksDB startup should not require Topling native resources, GitHub Packages availability, or runtime downloads.

Without these tests, CI only proves a narrow compile/happy-path case. It does not prove that the new storage provider is safe across fallback, restart, restore, or multi-DB Store scenarios.

VGalaxies (commented):
**Review summary**

- **Blocking:** yes
- **Summary:** The Topling package cannot start as built, while the default RocksDB path gains startup and lifecycle regressions.
- **Evidence:**
  - full `git diff origin/master...HEAD`, `git diff --check`, changed-script `bash -n` pass, and targeted static call-path inspection
  - runtime Maven verification was unavailable because offline dependency resolution failed before reactor artifacts resolved

hugegraph-server/hugegraph-dist/src/assembly/static/bin/common-topling.sh
```diff
@@ -0,0 +1,308 @@
...
+        *)
+            echo "[common-topling] Skip ToplingDB native preload on unsupported platform: $os_name/$machine_arch" >&2
+            return 0
+            ;;
+    esac
+
+    local top="$(cd "$lib_dir"/../ && pwd)"
+
+    local jar_file
+    jar_file=$(ls -1 "$lib_dir"/rocksdbjni*.jar 2>/dev/null | sort -V | tail -n1 || true)
```
VGalaxies, line 274: **High: Topling distributions cannot find their JNI JAR**

`hugegraph-server/hugegraph-dist/src/assembly/static/bin/common-topling.sh:274`

**Evidence**
- The `toplingdb` profile sets the artifact ID to `toplingdb-jni` at `pom.xml:423`, producing `toplingdb-jni-<version>.jar`, but the preloader searches only `rocksdbjni*.jar`.

**Impact**
- A Linux distribution built with `-Ptoplingdb` exits during initialization or startup before launching Java.

**Requested fix**
- Search for `toplingdb-jni*.jar` when Topling is selected, or pass the resolved engine JAR explicitly, and add a packaged startup smoke test.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -99,6 +99,8 @@ if [[ $PRELOAD == "true" ]]; then
     sed -i -e '/registerBackends/d; /serverStarted/d' "${SCRIPTS}/${EXAMPLE_SCRIPT}"
 fi
 
+source $BIN/preload-topling.sh
```
VGalaxies, line 102: **High: Standard RocksDB startup unconditionally requires Topling prerequisites**

`hugegraph-server/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh:102`

**Evidence**
- The default provider remains `standard`, but every startup sources the strict preloader. That path requires `unzip` and downloads jemalloc when it is not installed; failures terminate the sourced script. `init-store.sh:50` has the same behavior.

**Impact**
- Minimal or offline Linux installations using standard RocksDB can no longer initialize or start.

**Requested fix**
- Run the preload path only when the packaged engine/provider requires Topling native preloading; leave standard RocksDB JNI loading unchanged.

hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/OpenedRocksDB.java
```diff
@@ -90,7 +91,8 @@ public void close() {
         }
         this.cfHandles.clear();
 
-        this.rocksdb.close();
+        // Use RocksDBProviderLoader to close RocksDB
+        RocksDBProviderLoader.closeRocksDB(this.rocksdb);
```
VGalaxies, line 95: **High: Snapshot restore bypasses provider-specific cleanup**

`hugegraph-server/hugegraph-rocksdb/src/main/java/org/apache/hugegraph/backend/store/rocksdb/OpenedRocksDB.java:95`

**Evidence**
- Normal closure now passes through `RocksDBProviderLoader`, but `resumeSnapshot()` calls `forceCloseRocksDB()`, which directly closes the underlying `RocksDB` object at `RocksDBStdSessions.java:226` before reopening through the provider.

**Impact**
- Topling’s repository mapping, native resources, and HTTP server can survive snapshot replacement, causing leaks or a port-binding failure during reopen.

**Requested fix**
- Make force-close close the `OpenedRocksDB` wrapper through the provider, including column-family and repository cleanup, and test Topling snapshot restoration with HTTP enabled.

hugegraph-store/hg-store-rocksdb/src/main/java/org/apache/hugegraph/rocksdb/access/RocksDBSession.java
```diff
@@ -450,8 +454,11 @@ private void openRocksDB(String dbDataPath, long version) {
...
+                                                      columnFamilyHandleList,
+                                                      hugeConfig.get(RocksDBOptions.OPTION_PATH),
+                                                      hugeConfig.get(RocksDBOptions.OPEN_HTTP));
```
VGalaxies, line 461: **High: HTTP startup is requested for every Store database**

`hugegraph-store/hg-store-rocksdb/src/main/java/org/apache/hugegraph/rocksdb/access/RocksDBSession.java:461`

**Evidence**
- Every session receives the global `OPEN_HTTP` value. Store creates separate sessions for metadata and partition databases, while each Topling repository invokes `startHttpServer()` using the same configured port.

**Impact**
- With `rocksdb.open_http=true`, opening the second database attempts to bind the same port and prevents normal Store startup or partition access.

**Requested fix**
- Manage the HTTP server once per Store process, or pass `true` only to one designated session.

hugegraph-pd/hg-pd-dist/src/assembly/static/bin/start-hugegraph-pd.sh
```diff
@@ -63,9 +63,17 @@ PID_FILE="$BIN/pid"
...
 . "$BIN"/util.sh
 
+PARENT_DIR="$(cd "$TOP"/../ && pwd)"
+SERVER_VERSION_DIR="${SERVER_VERSION_DIR:-$(find_hugegraph_server_dir "$PARENT_DIR")}"
+
 ensure_path_writable "$LOGS"
 ensure_path_writable "$PLUGINS"
 
+# preload rocksdb/toplingdb
+if [ -n "$SERVER_VERSION_DIR" ] && [ -e "$SERVER_VERSION_DIR/bin/preload-topling.sh" ]; then
```
VGalaxies, line 73: **Medium: Standalone PD and Store packages omit required preload support**

`hugegraph-pd/hg-pd-dist/src/assembly/static/bin/start-hugegraph-pd.sh:73`

**Evidence**
- PD sources the preloader only from a sibling server distribution; Store does the same at `start-hugegraph-store.sh:70`. Their assembly descriptors package no component-local Topling preload scripts.

**Impact**
- Enabling the documented Topling configuration in a standalone PD or Store distribution leaves its native libraries and resources unprepared.

**Requested fix**
- Package component-local preload support that uses each distribution’s own engine artifact, or explicitly reject Topling configuration for standalone packages.

hugegraph-rocksdb-provider/src/test/java/org/apache/hugegraph/rocksdb/provider/ToplingRocksDBProviderTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Test: isAvailable() returns false when org.rocksdb.SidePluginRepo is not on the classpath.
+     * In a standard test environment (using vanilla rocksdbjni), ToplingDB features
+     * are not available. This is the expected state for most developers.
+     */
+    @Test
+    public void testIsNotAvailableWithoutSidePluginRepo() {
+        assertFalse(provider.isAvailable());
```
VGalaxies, line 53: **Medium: The Topling build profile contradicts its tests**

`hugegraph-rocksdb-provider/src/test/java/org/apache/hugegraph/rocksdb/provider/ToplingRocksDBProviderTest.java:53`

**Evidence**
- `-Ptoplingdb` supplies `SidePluginRepo`, but this test unconditionally asserts that it is unavailable. Loader tests likewise require Topling selection to fail.

**Impact**
- The documented `mvn clean package -Ptoplingdb` path cannot run its test suite successfully and provides no positive coverage of the supported runtime.

**Requested fix**
- Isolate unavailable-provider tests with a controlled classloader and add positive tests under the Topling profile.

hugegraph-rocksdb-provider/src/main/java/org/apache/hugegraph/rocksdb/provider/ToplingRocksDBProvider.java
```diff
@@ -0,0 +1,584 @@
...
+public class ToplingRocksDBProvider extends AbstractRocksDBProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ToplingRocksDBProvider.class);
+
+    private static final String PROVIDER_NAME = "topling";
+    private static final String SIDE_PLUGIN_REPO_CLASS = "org.rocksdb.SidePluginRepo";
+
+    // Validation constants migrated from RocksDBOptions
+    private static final Pattern SAFE_PATH_PATTERN =
+            Pattern.compile("^[a-zA-Z0-9/_.-]+\\.yaml$");
```
VGalaxies, line 58: **Medium: Valid `.yml` configuration files disable Topling silently**

`hugegraph-rocksdb-provider/src/main/java/org/apache/hugegraph/rocksdb/provider/ToplingRocksDBProvider.java:58`

**Evidence**
- The path regex accepts only `.yaml`, although the later extension check explicitly permits both `.yaml` and `.yml`. The validation exception is caught and converted into standard RocksDB fallback.

**Impact**
- Selecting Topling with a valid `conf/*.yml` file silently ignores its configuration and features.

**Requested fix**
- Accept both extensions in the initial regex, for example `\\.ya?ml$`, and add a `.yml` regression test.

WaterWhisperer:
Since Easy Migrate has replaced the old SPI design, close this PR

PR #3025, opened 2026-05-13
perf(docker): improve all images build cache efficiency

### LINKED ISSUE #2977: [Task] Improve Docker build cache efficiency for pd/store/server images
## Background

The Docker images for `hugegraph-pd`, `hugegraph-store`, and `hugegraph-server` currently build with this pattern:

1. `COPY . /pkg`
2. `RUN mvn package ...`
3. Copy build outputs into runtime image

This causes low cache efficiency:

- Any repository file change (even non-Java files) can invalidate the `COPY` layer.
- The expensive `mvn package` step is re-run too often.
- CI time stays high even when Java code is unchanged.

## Goal

Improve the Docker build cache hit rate and reduce unnecessary Maven rebuild time, **without changing image runtime behavior**.

## Scope (Starter Task / Phase A)

Please focus on the low-risk, high-impact subset:

- Improve `.dockerignore` to exclude files that should not affect the image build cache (for example: IDE files, some markdown/text files. etc).
- Refactor Dockerfile layer order so dependency-related files are copied before source files where practical.
- Use BuildKit Maven cache mount (`--mount=type=cache,target=/root/.m2`) in Maven build steps.
- Keep existing image entrypoints/ports/runtime behavior unchanged.

## Out of Scope (Do NOT include in this issue)

- Full artifact pipeline split ("build once, package many images").
- Cross-repo CI architecture redesign.
- New release logic or tag strategy changes.

## Deliverables

- A PR with Dockerfile and/or `.dockerignore` updates.
- A short before/after note from CI showing cache behavior and build time delta.

## Acceptance Criteria

- A second build with no Java source changes shows a clear improvement in build time and/or cache hits.
- Existing image smoke/integration checks still pass.
- No functional regression in generated images.

## Suggested Starting Points

- `hugegraph-pd/Dockerfile`
- `hugegraph-store/Dockerfile`
- `hugegraph-server/Dockerfile`
- `hugegraph-server/Dockerfile-hstore`

If the scope starts growing, keep this issue focused on Phase A and open follow-up issues for larger refactors.

bitflicker64:
## Purpose of the PR

close #2977

## Main Changes

**`.dockerignore` (new file)**
- Excludes Docker build-context noise explicitly because `.dockerignore` does not inherit `.gitignore`
- Filters local build outputs, extracted release directories, archives, IDE/OS files, logs, generated artifacts, env-local files, git internals, compose files, and docs

**All 4 Dockerfiles**
- Added `--mount=type=cache,target=/root/.m2` on `mvn package` — deps cached across repeated builds
- Kept `vim` and `cron` in the runtime images to preserve existing container debugging workflows and the current `start-hugegraph.sh -m true` monitor path
- Removed ineffective build-time `service cron start`, `rm /var/lib/dpkg/info/libc-bin.*`, and other dead cleanup/debug commands from runtime stages

> Note: `dependency:go-offline` not used — inter-module deps like `hg-pd-client` are not on Maven Central and would cause it to fail.

## Verifying these changes

```markdown
# Build all 4 images
docker build -f hugegraph-pd/Dockerfile        -t hugegraph/pd:test .  
docker build -f hugegraph-store/Dockerfile     -t hugegraph/store:test .  
docker build -f hugegraph-server/Dockerfile    -t hugegraph/server:test .  
docker build -f hugegraph-server/Dockerfile-hstore -t hugegraph/server-hstore:test .  
  
# Verify runtime contents and cron in each image
docker run --rm --entrypoint /bin/bash hugegraph/pd:test    -c "ls /hugegraph-pd/    && which cron"  
docker run --rm --entrypoint /bin/bash hugegraph/store:test -c "ls /hugegraph-store/ && which cron"  
docker run --rm --entrypoint /bin/bash hugegraph/server:test -c "ls /hugegraph-server/ && which cron"  
  
# Full stack health check (pull_policy: always requires --pull never for local images)
HUGEGRAPH_VERSION=test docker compose -f docker/docker-compose.yml up --pull never  
# All 3 containers should show (healthy) in docker ps 
```
```markdown
- [x] Tested and verified:
  - All 4 Dockerfiles build successfully (Maven deps cached via `--mount=type=cache`)
  - Runtime contents and `cron` confirmed in all images
  - Full stack health check passed — all 3 containers came up `(healthy)`:
    - `hugegraph/pd:test` — healthy on port 8620
    - `hugegraph/store:test` — healthy on port 8520
    - `hugegraph/server:test` — healthy on port 8080
```

- [ ] Dependencies (add/update license info & regenerate_known_dependencies.sh)
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects (typed here)
- [x] Nope

## Documentation Status

- [x] `Doc - No Need`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,34 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# NOTE: This file intentionally stays minimal.
+# Most patterns (IDE files, build artifacts, logs, OS files) are already
+# covered by .gitignore. Only Docker-specific extras are listed here.
+
+# Git internals (never tracked by git, must be explicit here)
+.git
+.gitignore
+.gitattributes
+.github
+
+# Docker compose files not needed in build context
+**/docker-compose*.yml
+**/docker-compose*.yaml
+
+# Docs not needed in build context
+**/*.md
+docs/
--- a/.dockerignore
+++ b/.dockerignore
@@ -40,7 +40,7 @@ apache-hugegraph-*/
 **/*.class
 **/gen-java/
 **/upload-files/
-**/dist/
+**/target/dist/
 **/build/
 **/node_modules/
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -23,7 +23,7 @@
 # Pre-extracted release dirs / archives
 apache-hugegraph-*/
 **/*.tar
-**/*.tar.gz*
+**/*.tar.gz
 **/*.zip
 **/*.war
 
@@ -40,7 +40,6 @@ apache-hugegraph-*/
 **/*.class
 **/gen-java/
 **/upload-files/
-**/target/dist/
 **/build/
 **/node_modules/
 
@@ -51,7 +50,7 @@ apache-hugegraph-*/
 # Git internals
 .git
 .gitignore
-.gitattribut
+.gitattributes
 .github
 
 # Compose / docs not needed in build context
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,21 +15,47 @@
 # limitations under the License.
 #
 
-# NOTE: This file intentionally stays minimal.
-**/target/
-# Most patterns (IDE files, build artifacts, logs, OS files) are already
-# covered by .gitignore. Only Docker-specific extras are listed here.
+# IMPORTANT: .dockerignore does NOT inherit .gitignore — patterns must be restated.
 
-# Git internals (never tracked by git, must be explicit here)
+# Build output
+**/target/
+
+# Pre-extracted release dirs / archives
+apache-hugegraph-*/
+**/*.tar
+**/*.tar.gz*
+**/*.zip
+**/*.war
+
+# IDE / OS
+.idea/
+.vscode/
+**/*.iml
+**/*.iws
+**/.DS_Store
+
+# Build / runtime artifacts
+**/logs/
+**/*.log
+**/*.class
+**/gen-java/
+**/upload-files/
+**/dist/
+**/build/
+**/node_modules/
+
+# Env files
+.env.local
+.env.*.local
+
+# Git internals
 .git
 .gitignore
-.gitattributes
+.gitattribut
 .github
 
-# Docker compose files not needed in build context
+# Compose / docs not needed in build context
 **/docker-compose*.yml
 **/docker-compose*.yaml
-
-# Docs not needed in build context
 **/*.md
 docs/
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,6 +16,7 @@
 #
 
 # NOTE: This file intentionally stays minimal.
+**/target/
 # Most patterns (IDE files, build artifacts, logs, OS files) are already
 # covered by .gitignore. Only Docker-specific extras are listed here.
 
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -10,8 +10,7 @@ on:
 
 jobs:
   build-server:
-    # TODO: we need test & replace it to ubuntu-24.04 or ubuntu-latest
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     env:
       USE_STAGE: 'false' # Whether to include the stage repository.
       TRAVIS_DIR: hugegraph-server/hugegraph-dist/src/assembly/travis
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build-server:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     env:
       USE_STAGE: 'false' # Whether to include the stage repository.
       TRAVIS_DIR: hugegraph-server/hugegraph-dist/src/assembly/travis
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   build-server:
+    # TODO: we need test & replace it to ubuntu-24.04 or ubuntu-latest
     runs-on: ubuntu-22.04
     env:
       USE_STAGE: 'false' # Whether to include the stage repository.
--- a/hugegraph-server/Dockerfile
+++ b/hugegraph-server/Dockerfile
@@ -16,18 +16,24 @@
 #
 
 # Dockerfile for HugeGraph Server
+# NOTE: Build with DOCKER_BUILDKIT=1 to enable cache mounts and --parents COPY.
 # 1st stage: build source code
 FROM maven:3.9.0-eclipse-temurin-11 AS build
 
-COPY . /pkg
 WORKDIR /pkg
+
+# Copy pom files first — layer is only invalidated when pom files change,
+# keeping the Maven cache mount warm across source-only changes.
+COPY --parents **/pom.xml ./
 ARG MAVEN_ARGS
 
-RUN mvn package $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true && pwd && ls -l && rm \
-    ./hugegraph-server/*.tar.gz && rm ./hugegraph-pd/*.tar.gz && rm ./hugegraph-store/*.tar.gz
+COPY . .
+RUN --mount=type=cache,target=/root/.m2 \
+    mvn package $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
+    && rm ./hugegraph-server/*.tar.gz ./hugegraph-pd/*.tar.gz ./hugegraph-store/*.tar.gz
 
 # 2nd stage: runtime env
-# Note: ZGC (The Z Garbage Collector) is only supported on ARM-Mac with java > 13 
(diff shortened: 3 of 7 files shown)


imbajin (commented):
Direction is right but a few blockers remain. See inline for net-new findings; +1'd Copilot's threads on the labs syntax / pom-only layer / root-pom glob (those are correct). Body-level: README/BUILDING should also note the new BuildKit ≥ 0.20 requirement, and removing `vim` is a behavioral change worth calling out in the commit body for ops folks who exec into containers.

.dockerignore
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# NOTE: This file intentionally stays minimal.
+**/target/
+# Most patterns (IDE files, build artifacts, logs, OS files) are already
+# covered by .gitignore. Only Docker-specific extras are listed here.
```
imbajin: ‼️ The premise here is incorrect — Docker's build context exclusion does **not** inherit `.gitignore`. As a result, locally-extracted release dirs (`apache-hugegraph-*/`, measured at ~790 MB on a normal dev checkout), `.idea/`, `node_modules/`, `logs/`, `*.iml`, `*.class`, `gen-java/`, `upload-files/`, `.vscode/`, `*.tar.gz`, `dist/`, `build/` all leak into every build's context, defeating the cache work in this PR (and risking IDE / secret leakage into image layers).

```suggestion
**/target/

# IMPORTANT: .dockerignore does NOT inherit .gitignore — patterns must be restated.
# Pre-extracted release dirs / archives
apache-hugegraph-*/
**/*.tar
**/*.tar.gz*
**/*.zip
**/*.war

# IDE / OS
.idea/
.vscode/
**/*.iml
**/*.iws
**/.DS_Store

# Build / runtime artifacts
**/logs/
**/*.log
**/*.class
**/gen-java/
**/upload-files/
**/dist/
**/build/
**/node_modules/

# Env files
.env.local
.env.*.local

# Git internals
.git
.gitignore
.gitattributes
.github

# Compose / docs not needed in build context
**/docker-compose*.yml
**/docker-compose*.yaml
**/*.md
docs/
```

.github/workflows/server-ci.yml
```diff
@@ -10,8 +10,7 @@ on:
 
 jobs:
   build-server:
-    # TODO: we need test & replace it to ubuntu-24.04 or ubuntu-latest
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
imbajin: ⚠️ Two concerns on this line:

1. **Scope creep**: bumping `ubuntu-22.04 → ubuntu-latest` is unrelated to the Docker cache work in this PR. The original `# TODO` was a deliberate guard (Ubuntu 24.04 changed default Java, removed legacy packages). Suggest reverting in this PR and splitting into a focused follow-up.
2. **`ubuntu-latest` drifts** when GitHub flips the alias — pin explicitly:

```suggestion
    runs-on: ubuntu-24.04
```

Flagging here since it's the only CI file touched: there is **no CI workflow that exercises any Dockerfile** (verified — zero `docker build` references under `.github/workflows/`). PR's verification is "run `docker build` locally". Future regressions will ship silently. Consider adding a build-only smoke job:

```yaml
docker-build:
  runs-on: ubuntu-24.04
  strategy:
    matrix:
      dockerfile:
        - hugegraph-pd/Dockerfile
        - hugegraph-store/Dockerfile
        - hugegraph-server/Dockerfile
        - hugegraph-server/Dockerfile-hstore
  steps:
    - uses: actions/checkout@v4
    - uses: docker/setup-buildx-action@v3
    - run: docker buildx build -f ${{ matrix.dockerfile }} --load .
```

hugegraph-server/Dockerfile
```diff
@@ -39,25 +46,18 @@ ENV JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseContainerSupport -XX:Max
...
-    && rm /var/lib/dpkg/info/libc-bin.* \
-    && apt-get -q clean \
-    && apt-get -q update \
+# 1. Install runtime dependencies and configure server
+RUN apt-get -q update \
     && apt-get -q install -y --no-install-recommends --no-install-suggests \
        dumb-init \
        procps \
        curl \
        lsof \
```
imbajin, line 54: 🧹 Pre-existing issue (not a regression from this PR, but worth fixing while in the area): the `cron` package stays installed in the runtime image (~3 MB), but no cron daemon is ever started — `dumb-init` only execs `docker-entrypoint.sh`. This means `start-hugegraph.sh -m true` silently fails to fire its scheduled monitor job, even though `crontab_append` returns success. Either:
- Drop `cron` entirely (saves ~3 MB and removes the false impression that monitor mode works), or
- Start cron in `docker-entrypoint.sh` (`cron &` before the server start) so the feature actually works.

Same applies to `Dockerfile-hstore` and the PD / Store Dockerfiles.

.github/workflows/server-ci.yml
bitflicker64: Reverted it back , I was mainly testing what would break with `ubuntu-latest`. Also added smoke test 🫡

hugegraph-server/Dockerfile
bitflicker64: removed cron from the runtime images for now since properly enabling it would expand the scope of this PR. happy to work on enabling this feature later in a follow up pr if u want

bitflicker64:
`**/dist/`matches `org/apache/hugegraph/dist/ ` a real Java source package directory.

The full path that gets excluded is:
```
hugegraph-server/hugegraph-dist/src/main/java/org/apache/hugegraph/dist/ 
```
The last path component is dist, so `**/dist/` matches it and wipes the entire `org.apache.hugegraph.dist` package from the Docker build context before mvn package ever runs. so used  `**/target/dist/  ` instead

imbajin:
Thanks for iterating on the Docker cache changes. The cache direction looks good, and besides the monitor/runtime compatibility point below plus the small `.dockerignore` typo, I don't see other blockers from my side.

I think we should keep the runtime behavior unchanged in this PR.

Today the image still exposes the legacy monitor path:

```text
start-hugegraph.sh -m true
        -> start-monitor.sh
        -> crontab registers monitor-hugegraph.sh every minute
        -> monitor-hugegraph.sh checks process + /versions
        -> restart HugeGraphServer if the process is gone or REST is unhealthy
```

After removing `cron`, that path is broken:

```text
start-hugegraph.sh -m true
        -> start-monitor.sh
        -> crontab: command not found
        -> monitor is not registered
```

I agree that the crontab-based monitor is not the ideal long-term model for Docker. For container users, the better direction is probably a Docker-native lifecycle model:

```text
docker-entrypoint.sh
        -> start HugeGraphServer
        -> watch server pid and optionally probe /versions
        -> exit the container when the server dies or stays unhealthy
        -> Docker restart policy restarts the container
```

That can then be paired with:

```text
docker run --restart unless-stopped ...
Dockerfile HEALTHCHECK ...
```

But this is a runtime behavior/design change, not just a build-cache optimization. It also needs documentation, because standalone Docker healthcheck only marks the container as unhealthy and does not restart it by itself; users still need a restart policy, or an entrypoint watchdog that exits the container after repeated failures.

So for this PR, I suggest the lowest-risk path:

1. Keep `cron` in the runtime image for now, so the existing `start-hugegraph.sh -m true` monitor path is not broken.
2. Remove only the build-time `service cron start` cleanup if needed, since starting a service during image build does not keep it running in containers.
3. Fix the small `.dockerignore` typo: `.gitattribut` -> `.gitattributes`.
4. Leave the Docker-native watchdog / healthcheck / restart-policy design to a follow-up PR with docs.

In short:

```text
This PR:
  improve Docker build cache
  keep runtime behavior unchanged

Follow-up PR:
  replace legacy cron monitor with Docker-native watchdog / healthcheck docs
```

bitflicker64:
> Thanks for iterating on the Docker cache changes. The cache direction looks good, and besides the monitor/runtime compatibility point below plus the small `.dockerignore` typo, I don't see other blockers from my side.
> 
Thanks for the review !!
> I think we should keep the runtime behavior unchanged in this PR.
> 
> Today the image still exposes the legacy monitor path:
> ...........
> ...........
> ...........
> But this is a runtime behavior/design change, not just a build-cache optimization. It also needs documentation, because standalone Docker healthcheck only marks the container as unhealthy and does not restart it by itself; users still need a restart policy, or an entrypoint watchdog that exits the container after repeated failures.
> 
Makess. sensee
> So for this PR, I suggest the lowest-risk path:
> 
> 1. Keep `cron` in the runtime image for now, so the existing `start-hugegraph.sh -m true` monitor path is not broken.
> 2. Remove only the build-time `service cron start` cleanup if needed, since starting a service during image build does not keep it running in containers.
> 3. Fix the small `.dockerignore` typo: `.gitattribut` -> `.gitattributes`.
> 4. Leave the Docker-native watchdog / healthcheck / restart-policy design to a follow-up PR with docs.
> 
should i also remove `**/target/dist/` since `**/target/` is alr covered also wanted to ask about `**/*.tar.gz*` should we keep it `**/*.tar.gz` to be safee ? 
> In short:
> 
> ```
> This PR:
>   improve Docker build cache
>   keep runtime behavior unchanged
> 
> Follow-up PR:
>   replace legacy cron monitor with Docker-native watchdog / healthcheck docs
> ```
ill make a follow up pr as soon as my end sems exams are overrr 🫡

imbajin:
Yes, `**/target/dist/` can be removed since `**/target/` already covers it.

For the archive patterns, I think it is fine to exclude local `*.tar.gz` / `*.tgz` / similar generated archives from the Docker build context. The Dockerfiles should rebuild the distribution packages inside the build stage, so the image build should not depend on pre-existing local release archives.

The only thing I would like to confirm is validation: please make sure all four Docker images still build successfully and the new images can start normally after these ignore rules are applied. If that is verified, excluding these archive files sounds good to me.

bitflicker64:
> The only thing I would like to confirm is validation: please make sure all four Docker images still build successfully and the new images can start normally after these ignore rules are applied. If that is verified, excluding these archive files sounds good to me.
>
validated all 4 images locally ,all build successfully and the runtime contents look correct (artifacts in place, cron installed). The .dockerignore archive patterns don't exclude anything the build actually needs since everything gets rebuilt from source inside the build stage anyway.


`edit: tagged the new images and tested with compose all three containers come up healthy `

imbajin (approved):
Thanks for the follow-up validation. The Docker build CI is green, the four images were validated locally, and the compose runtime check confirms the containers come up healthy. I also updated the PR description to match the current implementation: `vim` / `cron` are kept, while the ineffective build-time `service cron start` and dead cleanup/debug commands are removed.

Looks good to me now.

PR #3026, opened 2026-05-14
fix(server): keep schema ~create_time as Date after reload

### LINKED ISSUE #3013: [Bug] schema ~create_time userdata round-trips Date as String after cache reload
### Bug Type (问题类型)

logic (逻辑设计问题)

### Before submit

### Environment (环境信息)

- Server Version: 1.0.0 (Apache Release Version)

HugeGraph automatically writes internal schema userdata `Userdata.CREATE_TIME` (`"~create_time"`) as a `java.util.Date` when creating schema elements.                
   
  ### Expected                                                                                                                                                          
                             
  After a schema element is reloaded from backend storage, this value should still be a `java.util.Date`, consistent with existing schema tests that cast               
  `schemaElement.userdata().get(Userdata.CREATE_TIME)` to `Date`.
                                                                                                                                                                        
  ### Actual                 

  After a backend/store reload, `schemaElement.userdata().get(Userdata.CREATE_TIME)` can be returned as a `String` timestamp instead of `java.util.Date`.               
   
  This is exposed more clearly after PR #3011 because cross-JVM schema-cache-clear events cause remote server nodes to invalidate and reload schema from backend        
  storage. The publishing JVM may not observe the issue because it can keep using the in-memory schema object and, with the self-echo skip, does not reload its own
  freshly-created schema.                                                                                                                                               
                             
  ### Suspected cause

  The schema userdata serialization path writes `Date` values as strings, but the schema deserialization path does not convert the internal `~create_time` value back to
   `java.util.Date`.
                                                                                                                                                                        
  ### Scope                  

  Out of scope for PR #3011. That PR only changes runtime schema-cache invalidation propagation. It does not change persisted schema data or fix schema userdata        
  serialization round-trips.
                                                                                                                                                                        
  ### Minimal repro sketch   

  1. Start two HugeGraph server JVMs sharing the same graph / meta / backend.                                                                                           
  2. On **Server A**, create a schema element (e.g. a property key or vertex label).
  3. Ensure **Server B** reloads that schema from backend storage after a schema-cache-clear event.                                                                     
  4. On **Server B**, inspect `schemaElement.userdata().get(Userdata.CREATE_TIME)`.                                                                                     
  5. **Expected type:** `java.util.Date` — **actual type:** `String`.

```javascript

```

```javascript

```

dpol1:
- close #3013

  `Userdata.CREATE_TIME` (`"~create_time"`) is written as `java.util.Date`
  by `SchemaTransaction.setCreateTimeIfNeeded()`, then persisted through backend
  serializers as JSON. On reload, userdata is deserialized through raw
  `Map.class` paths, so Jackson cannot restore the value type and
  `schemaElement.userdata().get(Userdata.CREATE_TIME)` can come back as `String`.

  That breaks callers and existing tests that expect `~create_time` to keep the
  server-side `Date` contract after a schema is reloaded from backend storage.

  ## Main Changes

  - Added `Userdata.normalizeValue(key, value)` as a static helper on the `Userdata` class.
    - Converts `Userdata.CREATE_TIME` string values back with `DateUtil.parse(...)`.
    - Leaves existing `Date` values and all other userdata keys unchanged.
    - Throws a clearer `IllegalArgumentException` if a `~create_time` string is invalid.

  - Routed both userdata setters through `Userdata.put(key, value)`, which applies
    normalization before storing:
    - `userdata(String, Object)` covers serializer reload paths such as
      `TextSerializer`, `BinarySerializer`, `MysqlSerializer`, and
      `CassandraSerializer`.
    - `userdata(Userdata)` covers schema `fromMap()` hydration paths for
      `PropertyKey`, `VertexLabel`, `EdgeLabel`, and `IndexLabel`.

  - Added a null check to `userdata(Userdata)` for consistent argument validation.

  - Added regression coverage for schema userdata normalization and serializer
    round trips.

  This PR is intentionally limited to the server-side `SchemaElement`. The mirror
  class in `hugegraph-struct` has a similar userdata shape and can be handled in a
  separate follow-up PR.

  ## Verifying these changes

      - `SchemaElementTest` covers single-key normalization, bulk normalization,
        idempotency for `Date`, invalid `~create_time` strings, null bulk userdata,
        and non-`CREATE_TIME` keys.
      - `TextSerializerTest#testPropertyKeyUserdataCreateTimeRoundTripsAsDate`
        verifies the text serializer schema round trip keeps `CREATE_TIME` as `Date`.
      - `BinarySerializerTest#testPropertyKeyUserdataCreateTimeRoundTripsAsDate`
        verifies the binary serializer schema round trip keeps `CREATE_TIME` as `Date`.
      - The new test class is registered in `UnitTestSuite`.

  - [ ]  Dependencies (add/update license (https://hugegraph.apache.org/docs/contribution-guidelines/contribute/#321-check-licenses) info & ../install-dist/scripts/dependency/regenerate_known_dependencies.sh)
  - [ ]  Modify configurations
  - [ ]  The public API
  - [x]  Nope

  The public SchemaElement.userdata(...) signatures are unchanged. The behavior change is limited to internal userdata    normalization for Userdata.CREATE_TIME, handled via `Userdata.normalizeValue` / `Userdata.put()`.

  ## Documentation Status

  - [ ]  Doc - TODO
  - [ ]  Doc - Done
  - [x]  Doc - No Need

## Notes

- This PR fixes the server-side schema reload path for `Userdata.CREATE_TIME` (`~create_time`) only.
- `~create_time` is a server-reserved userdata key; valid string values are normalized back to `java.util.Date`, while malformed values now fail with `IllegalArgumentException`.
- Follow-up: #3028  tracks preserving typed `Userdata.DEFAULT_VALUE` (`~default_value`) reloads and checking the same raw userdata deserialization gap in `hugegraph-struct`.

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
@@ -103,7 +103,7 @@ public abstract class SchemaElement implements Nameable, Typeable,
     public void userdata(String key, Object value) {
         E.checkArgumentNotNull(key, "userdata key");
         E.checkArgumentNotNull(value, "userdata value");
-        this.userdata.put(key, Userdata.normalizeValue(key, value));
+        this.userdata.put(key, value);
     }
 
     /**
@@ -112,11 +112,7 @@ public abstract class SchemaElement implements Nameable, Typeable,
      */
     public void userdata(Userdata userdata) {
         E.checkArgumentNotNull(userdata, "userdata");
-        for (Map.Entry<String, Object> e : userdata.entrySet()) {
-            this.userdata.put(e.getKey(),
-                              Userdata.normalizeValue(e.getKey(),
-                                                      e.getValue()));
-        }
+        this.userdata.putAll(userdata);
     }
 
     public void removeUserdata(String key) {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
@@ -27,7 +27,6 @@ import org.apache.hugegraph.backend.id.IdGenerator;
 import org.apache.hugegraph.type.Nameable;
 import org.apache.hugegraph.type.Typeable;
 import org.apache.hugegraph.type.define.SchemaStatus;
-import org.apache.hugegraph.util.DateUtil;
 import org.apache.hugegraph.util.E;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 
@@ -97,40 +96,27 @@ public abstract class SchemaElement implements Nameable, Typeable,
         return Collections.unmodifiableMap(this.userdata);
     }
 
+    /**
+     * Add userdata. String values of {@link Userdata#CREATE_TIME} are
+     * normalized to {@link java.util.Date} and malformed strings are rejected.
+     */
     public void userdata(String key, Object value) {
         E.checkArgumentNotNull(key, "userdata key");
         E.checkArgumentNotNull(value, "userdata value");
-        this.userdata.put(key, normalizeUserdataValue(key, value));
+        this.userdata.put(key, Userdata.normalizeValue(key, value));
     }
 
+    /**
+     * Add userdata in bulk. String values of {@link Userdata#CREATE_TIME} are
+     * normalized to {@link java.util.Date} and malformed strings are rejected.
+     */
     public void userdata(Userdata userdata) {
         E.checkArgumentNotNull(userdata, "userdata");
         for (Map.Entry<String, Object> e : userdata.entrySet()) {
             this.userdata.put(e.getKey(),
-                              normalizeUserdataValue(e.getKey(), e.getValue()));
+                              Userdata.normalizeValue(e.getKey(),
+                                                      e.getValue()));
         }
-    }
-
-    /**
-     * Normalize internal userdata values whose runtime type can diverge from
-     * their serialized form. The only such key today is
-     * {@link Userdata#CREATE_TIME}: it is written as a {@link java.util.Date}
-     * but persisted as a formatted JSON string by the backend serializers,
-     * and Jackson cannot re-type a value to {@code Date} when the target is
-     * a raw {@code Map}. This method restores the original type after
-     * deserialization. Idempotent for values already of the expected type.
-     */
-    private static Object normalizeUserdataValue(String key, Object value) {
-        if (Userdata.CREATE_TIME.equals(key) && value instanceof String) {
-            try {
-                return DateUtil.parse((String) value);
-            } catch (RuntimeException e) {
-                throw new IllegalArgumentException(String.format(
-                          "Invalid userdata '%s' value: '%s'",
-                          Userdata.CREATE_TIME, value), e);
-            }
-        }
-        return value;
     }
 
     public void removeUserdata(String key) {
@@ -139,6 +125,7 @@ public abstract class SchemaElement implements Nameable, Typeable,
     }
 
     public void removeUserdata(Userdata userdata) {
+        E.checkArgumentNotNull(userdata, "userdata");
         for (String key : userdata.keySet()) {
             this.userdata.remove(key);
         }
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
@@ -27,6 +27,7 @@ import org.apache.hugegraph.backend.id.IdGenerator;
 import org.apache.hugegraph.type.Nameable;
 import org.apache.hugegraph.type.Typeable;
 import org.apache.hugegraph.type.define.SchemaStatus;
+import org.apache.hugegraph.util.DateUtil;
 import org.apache.hugegraph.util.E;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 
@@ -99,11 +100,37 @@ public abstract class SchemaElement implements Nameable, Typeable,
     public void userdata(String key, Object value) {
         E.checkArgumentNotNull(key, "userdata key");
         E.checkArgumentNotNull(value, "userdata value");
-        this.userdata.put(key, value);
+        this.userdata.put(key, normalizeUserdataValue(key, value));
     }
 
     public void userdata(Userdata userdata) {
-        this.userdata.putAll(userdata);
+        E.checkArgumentNotNull(userdata, "userdata");
+        for (Map.Entry<String, Object> e : userdata.entrySet()) {
+            this.userdata.put(e.getKey(),
+                              normalizeUserdataValue(e.getKey(), e.getValue()));
+        }
+    }
+
+    /**
+     * Normalize internal userdata values whose runtime type can diverge from
+     * their serialized form. The only such key today is
+     * {@link Userdata#CREATE_TIME}: it is written as a {@link java.util.Date}
+     * but persisted as a formatted JSON string by the backend serializers,
+     * and Jackson cannot re-type a value to {@code Date} when the target is
+     * a raw {@code Map}. This method restores the original type after
+     * deserialization. Idempotent for values already of the expected type.
+     */
+    private static Object normalizeUserdataValue(String key, Object value) {
+        if (Userdata.CREATE_TIME.equals(key) && value instanceof String) {
+            try {
+                return DateUtil.parse((String) value);
+            } catch (RuntimeException e) {
+                throw new IllegalArgumentException(String.format(
+                          "Invalid userdata '%s' value: '%s'",
+                          Userdata.CREATE_TIME, value), e);
+            }
+        }
+        return value;
     }
 
     public void removeUserdata(String key) {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
@@ -62,9 +62,16 @@ public class Userdata extends HashMap<String, Object> {
      * value to {@code Date} when the target is a raw {@code Map}. This method
      * restores the original type after deserialization. Idempotent for values
      * already of the expected type.
+     * <p>
+     * An empty string is passed through unchanged: it is the key-only
+     * placeholder used by the {@code eliminate()}/{@code DELETE} builder flow
+     * (e.g. {@code .userdata(CREATE_TIME, "").eliminate()}), where the value is
+     * ignored and only the key drives {@code removeUserdata}. Parsing it would
+     * fail before the eliminate path can apply its key-only semantics.
      */
     public static Object normalizeValue(String key, Object value) {
-        if (CREATE_TIME.equals(key) && value instanceof String) {
+        if (CREATE_TIME.equals(key) && value instanceof String &&
+            !((String) value).isEmpty()) {
             try {
                 return DateUtil.parse((String) value);
             } catch (RuntimeException e) {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
@@ -35,8 +35,22 @@ public class Userdata extends HashMap<String, Object> {
     }
 
     public Userdata(Map<String, Object> map) {
-        for (Map.Entry<String, Object> e : map.entrySet()) {
-            this.put(e.getKey(), normalizeValue(e.getKey(), e.getValue()));
+        this.putAll(map);
+    }
+
+    /**
+     * Normalizes the value before storing so the {@link #CREATE_TIME}-is-Date
+     * invariant holds regardless of how entries are added.
+     */
+    @Override
+    public Object put(String key, Object value) {
+        return super.put(key, normalizeValue(key, value));
+    }
+
+    @Override
+    public void putAll(Map<? extends String, ?> map) {
+        for (Map.Entry<? extends String, ?> e : map.entrySet()) {
+            this.put(e.getKey(), e.getValue());
         }
     }
 
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
@@ -22,6 +22,7 @@ import java.util.Map;
 
 import org.apache.hugegraph.exception.NotAllowException;
 import org.apache.hugegraph.type.define.Action;
+import org.apache.hugegraph.util.DateUtil;
 
 public class Userdata extends HashMap<String, Object> {
 
@@ -34,7 +35,31 @@ public class Userdata extends HashMap<String, Object> {
     }
 
     public Userdata(Map<String, Object> map) {
-        this.putAll(map);
+        for (Map.Entry<String, Object> e : map.entrySet()) {
+            this.put(e.getKey(), normalizeValue(e.getKey(), e.getValue()));
+        }
+    }
+
+    /**
+     * Normalize internal userdata values whose runtime type can diverge from
+     * their serialized form. The only such key today is {@link #CREATE_TIME}:
+     * it is written as a {@link java.util.Date} but persisted as a formatted
+     * JSON string by the backend serializers, and Jackson cannot re-type a
+     * value to {@code Date} when the target is a raw {@code Map}. This method
+     * restores the original type after deserialization. Idempotent for values
+     * already of the expected type.
+     */
+    public static Object normalizeValue(String key, Object value) {
+        if (CREATE_TIME.equals(key) && value instanceof String) {
+            try {
+                return DateUtil.parse((String) value);
+            } catch (RuntimeException e) {
+                throw new IllegalArgumentException(String.format(
+                          "Invalid userdata '%s' value: '%s'",
+                          CREATE_TIME, value), e);
+            }
+        }
+        return value;
     }
 
     public static void check(Userdata userdata, Action action) {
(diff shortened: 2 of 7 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
```diff
@@ -99,11 +100,37 @@ public Map<String, Object> userdata() {
...
+    /**
+     * Normalize internal userdata values whose runtime type can diverge from
+     * their serialized form. The only such key today is
+     * {@link Userdata#CREATE_TIME}: it is written as a {@link java.util.Date}
+     * but persisted as a formatted JSON string by the backend serializers,
+     * and Jackson cannot re-type a value to {@code Date} when the target is
+     * a raw {@code Map}. This method restores the original type after
+     * deserialization. Idempotent for values already of the expected type.
+     */
+    private static Object normalizeUserdataValue(String key, Object value) {
```
imbajin: ⚠️ Consider moving `normalizeUserdataValue` onto `Userdata` itself (next to `CREATE_TIME` / `DEFAULT_VALUE` constants and the existing `check()` validator). Three benefits:

1. **Cohesion** — the rule is *about* userdata semantics, not about `SchemaElement`.
2. **Reuse for the announced follow-up** — the PR notes that `hugegraph-struct` has a mirror `Userdata` with the same shape; if normalization lives on `Userdata`, the struct fix becomes a one-line delegation.
3. **Future userdata readers** — anything that hydrates a `Userdata` from raw JSON gets the fix for free, not just the two `SchemaElement` setters.

Not blocking — the current placement works. But shifting it would prevent re-implementing this for every userdata-bearing class.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/SchemaElement.java
```diff
@@ -99,11 +100,37 @@ public Map<String, Object> userdata() {
     public void userdata(String key, Object value) {
```
imbajin, line 103: ⚠️ The two public `userdata(...)` setters now silently coerce `String → Date` for `~create_time` and throw `IllegalArgumentException` on a malformed string. The detailed Javadoc lives on the private helper, where IDE tooltips / public-API readers won't see it.

A one-line note on each public setter (or on the class header) calling out the implicit conversion would prevent surprise for callers reading the public API.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/SchemaElementTest.java
```diff
@@ -0,0 +1,125 @@
...
+            schema.userdata(Userdata.CREATE_TIME, "not-a-date");
+        }, e -> {
+            Assert.assertContains(Userdata.CREATE_TIME, e.getMessage());
+            Assert.assertContains("not-a-date", e.getMessage());
+        });
+    }
+
+    @Test
+    public void testSingleSetterLeavesOtherStringKeysUntouched() {
+        SchemaElement schema = newSchema();
```
imbajin, line 102: ⚠️ The PR specifically wraps `DateUtil.parse(...)`'s `RuntimeException` as the cause of the thrown `IllegalArgumentException`. Worth locking that contract in:

```suggestion
        }, e -> {
            Assert.assertContains(Userdata.CREATE_TIME, e.getMessage());
            Assert.assertContains("not-a-date", e.getMessage());
            Assert.assertNotNull(e.getCause());
        });
```

Protects against a future refactor that drops the cause chain (which would silently degrade error diagnostics).

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/SchemaElementTest.java
```diff
@@ -0,0 +1,125 @@
...
+    @Test
+    public void testBulkSetterRejectsNullUserdata() {
+        SchemaElement schema = newSchema();
+
+        Assert.assertThrows(IllegalArgumentException.class, () -> {
+            schema.userdata(null);
+        }, e -> {
+            Assert.assertContains("userdata", e.getMessage());
+        });
+    }
```
imbajin, line 211: ⚠️ Consider one more test that asserts `schema.userdata(Userdata.CREATE_TIME, null)` still throws `IllegalArgumentException` from the existing `E.checkArgumentNotNull(value, ...)`.

Reason: the new helper sits between the null check and the `put`. A future refactor that reorders these (e.g., parse first, null-check second) would silently break the null contract — currently nothing pins it down for the `CREATE_TIME` key specifically.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/serializer/TextSerializerTest.java
```diff
@@ -0,0 +1,58 @@
...
+        original.userdata(Userdata.CREATE_TIME, created);
+
+        BackendEntry entry = ser.writePropertyKey(original);
+        PropertyKey reloaded = ser.readPropertyKey(objects.graph(), entry);
+
+        Object value = reloaded.userdata().get(Userdata.CREATE_TIME);
+        Assert.assertTrue("CREATE_TIME should be a Date after round-trip, " +
+                          "was " + (value == null ? "null" : value.getClass()),
+                          value instanceof Date);
+        Assert.assertEquals(created, value);
```
imbajin, line 54: ⚠️ Two coverage gaps worth one small change each:

1. **Only `PropertyKey` is round-tripped.** `VertexLabel`, `EdgeLabel`, and `IndexLabel` reach the helper through the *bulk* setter (`fromMap` → `userdata(new Userdata(...))`), which is a different control path than `readPropertyKey`'s per-entry single setter. Adding even one assertion for any of those three would harden coverage cheaply.
2. **Near-duplicate test classes.** This file and the new `BinarySerializerTest#testPropertyKeyUserdataCreateTimeRoundTripsAsDate` differ only in serializer type. A `@Parameterized` test or a shared helper in `BaseUnitTest` taking `Function<HugeConfig, AbstractSerializer>` would remove ~50 lines and stop the two from drifting.

Neither blocks merge; (1) is the more valuable of the two.

imbajin (commented):
Reviewed correctness, design, and tests. **No blocking issues** — fix is correct and narrowly scoped, all backend reload paths (Text / Binary / Mysql / Cassandra / HBase / Postgres / Palo / Hstore + the four `fromMap` paths) route through the new helper, and `DateUtil.parse` is symmetric with the `"yyyy-MM-dd HH:mm:ss.SSS"` format used by `JsonUtil`/`HugeGraphSONModule`. Inline comments cover important refinements (helper placement, public setter docs, test depth). A few additional minor notes below that don't have a clean inline target:

- 🧹 **`removeUserdata(Userdata)` (`SchemaElement.java:114-118`) lacks a null check** — adding `E.checkArgumentNotNull(userdata, ...)` keeps the bulk setters symmetric with the new check on `userdata(Userdata)`.
- 🧹 **Behavior change worth a release note**: any REST client that posts `~create_time` as a JSON string in user-data now silently receives a `Date` (or an `IllegalArgumentException` if malformed). `~create_time` is a server-reserved key (`~` prefix, written by `SchemaTransaction.setCreateTimeIfNeeded`), so risk is small, but flagging it in the PR body / changelog is helpful.
- 🧹 **Follow-up tracking**: `Userdata.DEFAULT_VALUE` has the same Jackson type-erasure problem (e.g., a property key whose data type is `Date` will reload its default value as `String`). The PR description already defers the `hugegraph-struct` mirror — please add `DEFAULT_VALUE` to the same follow-up list, since the root cause and fix shape are identical.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
```diff
@@ -34,7 +35,31 @@ public Userdata() {
...
+    /**
+     * Normalize internal userdata values whose runtime type can diverge from
+     * their serialized form. The only such key today is {@link #CREATE_TIME}:
+     * it is written as a {@link java.util.Date} but persisted as a formatted
+     * JSON string by the backend serializers, and Jackson cannot re-type a
+     * value to {@code Date} when the target is a raw {@code Map}. This method
+     * restores the original type after deserialization. Idempotent for values
+     * already of the expected type.
+     */
+    public static Object normalizeValue(String key, Object value) {
```
imbajin, line 72: ⚠️ `normalizeValue` is invoked from the `Userdata(Map)` constructor but **not** from `put()` / `putAll()`.

Callers can bypass normalization by writing directly to a `Userdata` instance:
```java
Userdata ud = new Userdata();
ud.put(Userdata.CREATE_TIME, "2026-01-01 00:00:00.000"); // still a String
```
The object then holds an inconsistent type until it is passed to `SchemaElement.userdata(Userdata)`. This pattern already appears in `SchemaElementTest#testBulkSetterNormalizesCreateTimeAndKeepsOtherEntries`, which calls `bulk.put(CREATE_TIME, formatted)` directly and relies on `SchemaElement` to fix it up.

If the invariant is that `CREATE_TIME` must *always* be `Date` inside a `Userdata`, the guarantee should live in `put()` as well:
```java
@Override
public Object put(String key, Object value) {
    return super.put(key, normalizeValue(key, value));
}
```

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
```diff
@@ -34,7 +35,31 @@ public Userdata() {
     }
 
     public Userdata(Map<String, Object> map) {
```
imbajin, line 37: ⚠️ All four `fromMap()` paths (`PropertyKey`, `VertexLabel`, `EdgeLabel`, `IndexLabel`) call:
```java
schema.userdata(new Userdata((Map<String, Object>) entry.getValue()));
```
This normalizes `CREATE_TIME` twice: once here in `Userdata(Map)`, and once again in `SchemaElement.userdata(Userdata)`. Normalization is currently idempotent (`Date → Date` is a no-op), so there is no bug today. However, the redundancy makes the invariant harder to reason about and would silently break if normalization ever becomes non-idempotent. Consider consolidating to a single layer (e.g., normalize only in `SchemaElement`, or only in `Userdata.put()` as noted above).

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/serializer/TextSerializerTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
```
imbajin, line 55: ⚠️ The PR description states that `MysqlSerializer` and `CassandraSerializer` are also affected by the same `CREATE_TIME` deserialization gap:
> *"`userdata(String, Object)` covers serializer reload paths such as TextSerializer, BinarySerializer, **MysqlSerializer**, and **CassandraSerializer**"*

Only `TextSerializer` and `BinarySerializer` have round-trip tests added. Adding equivalent `testPropertyKeyUserdataCreateTimeRoundTripsAsDate` tests for `MysqlSerializer` and `CassandraSerializer` would close this coverage gap for the stated fix surface.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/SchemaElementTest.java
```diff
@@ -0,0 +1,202 @@
...
+import org.apache.hugegraph.schema.Userdata;
+import org.apache.hugegraph.schema.VertexLabel;
+import org.apache.hugegraph.testutil.Assert;
+import org.apache.hugegraph.unit.FakeObjects;
+import org.apache.hugegraph.util.DateUtil;
+import org.junit.Test;
+
+public class SchemaElementTest {
+
+    private static final String DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
```
imbajin: 🧹 `DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS"` is hardcoded here and duplicated identically in `TextSerializerTest` line 36. `DateUtil` uses this format internally but does not expose it as a public constant. If the default format ever changes, both tests will silently diverge.

Consider exposing a `DateUtil.FORMAT_MS` constant, or use the single-arg `DateUtil.parse(str)` overload in tests to avoid coupling to a format literal at all.

imbajin (commented):
No blocking issues. The core fix (normalizing `CREATE_TIME` String→Date on deserialization) is correct and the main round-trip paths are tested. Four non-blocking observations below.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/serializer/TextSerializerTest.java
dpol1: Holding off, with reasoning:

  `MysqlSerializer.readUserdata` (L167) and `CassandraSerializer.readUserdata`
  (L222) both call the same `schema.userdata(key, value)` setter as
  `TextSerializer`/`BinarySerializer`. That setter now routes through the
  single `Userdata.put` normalization path, which is already exercised by:

  - `TextSerializerTest` / `BinarySerializerTest` round-trips
  - `SchemaElementTest` (constructor, single setter, bulk, `fromMap`)

  There's no existing unit-test harness for the SQL serializers — adding one
  means backend/mock infra for **zero new code paths**. The marginal coverage
  doesn't justify it.

  Better closure for the shared logic: a direct `UserdataTest.normalizeValue`
  unit test, serializer-agnostic. Happy to add that as a follow-up; proposing
  not to gate this PR on it. Pushback welcome if you see a path difference
  I'm missing.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java
dpol1: Done. `Userdata` now overrides `put()` (and `putAll()`, since HashMap's
  `putAll` skips the overridden `put`):

  ```java
  @Override
  public Object put(String key, Object value) {
      return super.put(key, normalizeValue(key, value));
  }
```
  So a raw new Userdata().put(CREATE_TIME, "...") is normalized too, not
  just the Map constructor.

File: hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/SchemaElementTest.java (hunk 9 above)
Comment (dpol1) [reply hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/SchemaElementTest.java]: Address by 8110bc4

File: hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java (hunk 7 above)
Comment (dpol1) [reply hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/Userdata.java:37]: Address by https://github.com/apache/hugegraph/commit/8110bc4cd3138470c432618b06bb739910e21861

File: hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/serializer/TextSerializerTest.java [hunk 10]
```diff
@@ -0,0 +1,56 @@
...
```
imbajin, line 55: ⚠️ Only `PropertyKey` is round-tripped here. `VertexLabel`, `EdgeLabel`, and `IndexLabel` each have their own `writeXxx`/`readXxx` methods and reach the normalization through the bulk setter (`fromMap` → `userdata(Userdata)`), a different path from `PropertyKey` (single-key setter). A regression in those paths would not be caught by the existing tests.

Would it be possible to add at minimum one `VertexLabel` round-trip test here as the representative bulk-setter path?

dpol1: @imbajin not quite — `readUserdata` is a single shared method (L909). 
`readVertexLabel`/`readEdgeLabel`/`readPropertyKey`/`readIndexLabel` all call it, and it uses the single-key setter for every type. So a VertexLabel round-trip here runs the exact same path as PropertyKey, not the bulk one.

The bulk path (`fromMap → userdata(Userdata)`) is the schema-builder side, already pinned by `SchemaElementTest#testVertexLabelFromMapNormalizesCreateTimeString`. And after this PR both setters go through `Userdata.put`, so it's one normalization point. IMHO a VertexLabel serializer test would just re-run the identical loop, no new path covered. I'll still add one if you want the `readVertexLabel` wiring guarded — just say so.

imbajin: Good point, I agree that the serializer round-trip path is shared here. A VertexLabel serializer round-trip would mostly re-run the same `readUserdata()` loop, so I do not think we need to insist on that test.

I found a separate edge case while re-checking the current implementation: moving normalization into `Userdata.put()` also affects eliminate inputs, where the value is only a placeholder and `removeUserdata(Userdata)` deletes by key.

```text
schema.xxx(...)
  .userdata(Userdata.CREATE_TIME, "")
  .eliminate()
      |
      v
Userdata.put("~create_time", "")
      |
      v
DateUtil.parse("") -> IllegalArgumentException
      |
      v
never reaches removeUserdata(...)
```

Existing eliminate tests use placeholder values like `.userdata("icon", "").eliminate()` / `.userdata("max", "").eliminate()`, so this seems like a real behavior change for deleting `~create_time` via the builder or REST remove action.

Could we either skip normalization for eliminate/delete inputs, or explicitly reject removal of `Userdata.CREATE_TIME` with a dedicated check? Right now it fails as a date parse error before the eliminate path can apply its key-only semantics.

dpol1: Option 2 won't work here — the parse failure is in the builder accumulating userdata via `userdata(k, v)`, before `eliminate()` / `Userdata.check(ELIMINATE)` ever runs, so there's no action context at that point.

Went with option 1: `normalizeValue` now leaves a blank ~create_time untouched, same as the existing `.userdata("icon", "").eliminate()` placeholder convention. Non-empty bad values still throw on the add path, so the contract is unchanged. Added a PropertyKeyCoreTest reproducing your scenario end-to-end (fails before, passes after).

Address by 1977a44

dpol1:
@imbajin WDYT about this approach?

imbajin (approved):
LGTM, thanks for addressing the eliminate placeholder case. The current approach looks good to me.

One small follow-up: the PR description still mentions `SchemaElement.normalizeUserdataValue`, but the final implementation is centered on `Userdata.normalizeValue` / `Userdata.put()`. It would be good to update the description so it matches the current code.

dpol1:
Changed

dpol1:
Thanks @imbajin for lead the direction of this PR with your sharp and precise review

PR #3034, opened 2026-05-21
fix(server): avoid extracting text range filters

### LINKED ISSUE #2935: [Bug] Unexpected ArrayIndexOutOfBoundsException for filter-step and repeat()-step
### Bug Type (问题类型)

gremlin (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0
- Backend: RocksDB x nodes

### Expected behavior
```
g.V().has('vp4', lt('')).repeat(__.out('el2')).emit().times(1).count()
g.V().match(__.as('start').has('vp4', lt('')).union(__.out('el2')).as('m')).select('m').count()
```
### Actual behavior
+ The first query thrown a exception: ```java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0```
### Additional Context
+ We moved the```has(...)``` step and ```repeat()```step in first query into the ```match()```step and replace the ```repeat()``` with ```union()``` step, which should not affect the result.
+ When I executed these two queries on Janusgraph and Tinkerpop, they both returned the same results and were consistent with the results of the second query.
+ I have implemented a fuzzing tool and am using it to test HugeGraph. To reduce the burden on developers, I have simplified the test cases as much as possible and isolated irrelevant information.

```javascript
none
```

```javascript
none
```

LegendPei:
## Purpose of the PR
- close #2935

This PR fixes the query-planning path exposed by #2935.

For top-level traversals like `g.V().has("vp4", P.lt("")).repeat(...).count()`,
HugeGraph may extract the text range predicate into a backend query. However,
range predicates on text properties should not be planned as backend range
index queries. This can make the direct traversal fail while an equivalent
`match()` traversal returns normally.

This change keeps such text range predicates in the traversal layer instead of
pushing them down to the backend query, making the direct traversal consistent
with the equivalent `match()` form.

## Main Changes

- Avoid extracting `Text` property predicates with `gt/gte/lt/lte` into
  `HugeGraphStep` / `HugeVertexStep` backend queries.
- Preserve normal extraction for system properties and non-text property
  predicates.
- Add a regression test for the reported traversal shape:
  `has("vp4", P.lt("")).repeat(__.out("el2")).emit().times(1).count()`.
- Verify that the direct traversal returns the same result as the equivalent
  `match()` traversal.

## Verifying these changes

      - `mvn -pl hugegraph-server/hugegraph-test -am -P core-test,memory -Dtest=CountStrategyCoreTest#testRepeatAfterTextRangeFilterWithEmptyResult -DfailIfNoTests=false "-Drat.skip=true" "-Dcheckstyle.skip=true" test`
    - `mvn -pl hugegraph-server/hugegraph-test -am -P core-test,rocksdb -Dtest=CountStrategyCoreTest#testRepeatAfterTextRangeFilterWithEmptyResult -DfailIfNoTests=false "-Drat.skip=true" "-Dcheckstyle.skip=true" test`

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
@@ -20,12 +20,10 @@ package org.apache.hugegraph.traversal.optimize;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.regex.Matcher;
@@ -44,7 +42,6 @@ import org.apache.hugegraph.backend.query.Query;
 import org.apache.hugegraph.exception.NotFoundException;
 import org.apache.hugegraph.exception.NotSupportException;
 import org.apache.hugegraph.iterator.FilterIterator;
-import org.apache.hugegraph.schema.IndexLabel;
 import org.apache.hugegraph.schema.PropertyKey;
 import org.apache.hugegraph.schema.SchemaLabel;
 import org.apache.hugegraph.structure.HugeElement;
@@ -203,48 +200,41 @@ public final class TraversalUtil {
     private static boolean extractHasContainers(HugeGraphStep<?, ?> newStep,
                                                 HasContainerHolder holder) {
         HugeGraph graph = TraversalUtil.tryGetGraph(newStep);
-        List<HasContainer> hasContainers = new ArrayList<>(holder.getHasContainers());
-        HugeType type = newStep.returnsVertex() ? HugeType.VERTEX : HugeType.EDGE;
-        Set<Id> labelIds = collectLabelIds(graph, type, newStep.getHasContainers(),
-                                           hasContainers);
-        for (HasContainer has : hasContainers) {
-            if (!canExtractHasContainer(graph, type, labelIds, has)) {
-                continue;
-            }
+        if (!canExtractHasContainers(graph, holder)) {
+            return false;
+        }
+        for (HasContainer has : holder.getHasContainers()) {
             if (!GraphStep.processHasContainerIds(newStep, has)) {
                 newStep.addHasContainer(has);
             }
-            holder.removeHasContainer(has);
         }
-        return holder.getHasContainers().isEmpty();
+        return true;
     }
 
     private static boolean extractHasContainers(HugeVertexStep<?> newStep,
                                                 HasContainerHolder holder) {
         HugeGraph graph = TraversalUtil.tryGetGraph(newStep);
-        List<HasContainer> hasContainers = new ArrayList<>(holder.getHasContainers());
-        HugeType type = newStep.returnsVertex() ? HugeType.VERTEX : HugeType.EDGE;
-        Set<Id> labelIds = collectLabelIds(graph, type, newStep.getHasContainers(),
-                                           hasContainers);
-        for (HasContainer has : hasContainers) {
-            if (!canExtractHasContainer(graph, type, labelIds, has)) {
-                continue;
-            }
-            newStep.addHasContainer(has);
-            holder.removeHasContainer(has);
+        if (!canExtractHasContainers(graph, holder)) {
+            return false;
         }
-        return holder.getHasContainers().isEmpty();
+        for (HasContainer has : holder.getHasContainers()) {
+            newStep.addHasContainer(has);
+        }
+        return true;
+    }
+
+    private static boolean canExtractHasContainers(HugeGraph graph,
+                                                   HasContainerHolder holder) {
+        for (HasContainer has : holder.getHasContainers()) {
+            if (!canExtractHasContainer(graph, has)) {
+                return false;
+            }
+        }
+        return true;
     }
 
     static boolean canExtractHasContainer(HugeGraph graph,
                                           HasContainer has) {
-        return canExtractHasContainer(graph, HugeType.UNKNOWN, null, has);
-    }
-
-    private static boolean canExtractHasContainer(HugeGraph graph,
-                                                  HugeType type,
-                                                  Set<Id> labelIds,
-                                                  HasContainer has) {
         if (isSysProp(has.getKey())) {
             return true;
         }
@@ -256,9 +246,6 @@ public final class TraversalUtil {
         try {
             pkey = graph.propertyKey(has.getKey());
         } catch (NotFoundException e) {
-            return false;
-        }
-        if (!hasIndexForProperty(graph, type, labelIds, pkey, has)) {
             return false;
         }
         if (!pkey.dataType().isText()) {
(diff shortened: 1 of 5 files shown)


imbajin (commented):
The original correctness fix is clear, but the latest version now broadens the PR into a more aggressive HasStep split-and-remove optimization. I left one scope/risk comment suggesting we keep this PR on the minimal safe path and move the broader optimization to a follow-up.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -166,38 +170,219 @@ public static void trySetGraph(Step<?, ?> step, HugeGraph graph) {
...
+        Set<Id> labelIds = collectLabelIds(graph, type, newStep.getHasContainers(),
+                                           hasContainers);
+        for (HasContainer has : hasContainers) {
+            if (!canExtractHasContainer(graph, type, labelIds, has)) {
+                continue;
+            }
+            if (!GraphStep.processHasContainerIds(newStep, has)) {
+                newStep.addHasContainer(has);
+            }
+            holder.removeHasContainer(has);
```
imbajin: ‼️ I think we should scope this PR back down before merging. The original bug is a focused correctness issue: text range predicates such as `has("vp4", P.lt(""))` should not be pushed into the backend query path. The latest implementation now goes further and splits a mixed `HasStep`, pushes selected containers down, removes them from the original `HasStep`, and adds new label/index-availability logic in `TraversalUtil`.

```text
Original bug boundary

Text range predicate
        |
        v
wrongly pushed to backend
        |
        v
exception / mismatch with match()
```

```text
Latest implementation boundary

HasStep(vp4 < "", age = 1)
        |
        +-- push age = 1 into HugeGraphStep / HugeVertexStep
        |
        +-- remove age = 1 from the original HasStep
        |
        +-- rely on new label/index checks to prove backend can enforce it
```

That is a much broader optimizer change than the original fix, and once a container is removed from the original `HasStep`, correctness depends on `canExtractHasContainer()` exactly matching what the backend planner can really execute. The new tests cover several cases, but this still expands the review surface into composite-index matching, label context, paging/sysprop interactions, and future index-planner behavior.

For this PR, I would prefer one of these smaller and safer shapes:

```text
Option 1: minimal correctness fix
  - keep unsafe Text + range predicates in the traversal layer
  - do not split mixed HasSteps
  - avoid adding backend index-availability logic here

Option 2: low-risk partial prefiltering
  - optionally push safe containers down as redundant backend pre-filters
  - keep the original HasStep unchanged unless all containers are extractable
  - correctness remains guarded by the full traversal-layer filter
```

```text
Safer partial prefiltering shape

Backend prefilter:
  age = 1

Traversal filter still intact:
  vp4 < "" AND age = 1

Result:
  backend may reduce candidates, but traversal still owns correctness
```

The current split-and-remove optimization can still be valuable, but I think it should be a separate follow-up PR with dedicated tests for composite-index prefix matching, label-less queries, `~page` / paging semantics when some user filters remain in traversal, and parity with the backend index planner. Keeping this PR minimal would make the #2935 fix easier to reason about and safer to merge.

LegendPei: Okay, sorry. I was too obsessed with fixing the review issue of copilot earlier, which led to a slight deviation in direction. I'll adjust my focus now.

imbajin: Thanks for adjusting the direction. To clarify my previous comment: I do not mean that we should choose freely between the two shapes. For this PR, I would prefer the minimal option A.

The original issue is small and focused: `Text` range predicates should not be extracted into the backend query path. So the safest fix here is to add the narrow condition that blocks only that unsafe case, while keeping the existing extraction behavior unchanged for the rest.

```text
Preferred shape for this PR

HasStep(vp4 < "", age = 1)
        |
        v
if any extracted container is Text + range:
  keep this HasStep in traversal layer
  avoid backend text-range extraction

No new index-planner logic
No mixed HasStep split/remove behavior
No broader optimizer contract change
```

This may leave a small performance loss for the uncommon case where one `has()` step mixes a text-range predicate with other safe predicates, but it keeps the patch proportional to the bug and avoids introducing new correctness risks around label context, index matching, composite index prefixes, paging, or unique/search/numeric index behavior.

PS: the more aggressive split-and-remove optimization can still be considered later

LegendPei: Thanks for the clarification. I agree with option A for this PR.

I have scoped the local changes back to the minimal correctness fix:
- no mixed `HasStep` split/remove behavior
- no new index-planner or label/index availability logic
- if a `HasStep` contains a `Text` range predicate (`gt/gte/lt/lte`), keep the whole `HasStep` in the traversal layer
- keep the existing extraction behavior for other cases as much as possible

This intentionally accepts the small potential performance loss for mixed `HasStep`s, so the fix stays proportional to #2935. I’ll push the narrowed patch after one final local check.

LegendPei: Now this fix adopts a conservative strategy. When encountering the same HasStep mixed with text range and other security conditions, the entire HasStep will be retained, possibly with less backend push-down optimization

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -166,38 +167,103 @@ public static void trySetGraph(Step<?, ?> step, HugeGraph graph) {
...
+        try {
+            pkey = graph.propertyKey(has.getKey());
+        } catch (NotFoundException e) {
+            return false;
+        }
+        if (!pkey.dataType().isText()) {
+            return true;
+        }
+
+        List<P<Object>> predicates = new ArrayList<>();
```
imbajin, line 255: ⚠️ Could we add a small regression case for the other text-range predicate shapes here?

The new guard walks the collected predicates and blocks `gt/gte/lt/lte` on `Text` properties, but the current tests mostly exercise `P.lt("")`. It would be useful to cover at least one upper-bound variant (`gt/gte/lte`) and one connective range form such as `between()` / `inside()`, so the `collectPredicates()` path is protected too.

For example, the expected behavior should stay the same: text range predicates are kept in the traversal layer instead of being extracted into the backend query.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -166,38 +167,103 @@ public static void trySetGraph(Step<?, ?> step, HugeGraph graph) {
...
-                traversal.removeStep(step);
             }
-            step = step.getNextStep();
+            step = nextStep;
         } while (step instanceof HasStep || step instanceof NoOpBarrierStep);
     }
 
+    private static boolean extractHasContainers(HugeGraphStep<?, ?> newStep,
+                                                HasContainerHolder holder) {
+        HugeGraph graph = TraversalUtil.tryGetGraph(newStep);
```
imbajin, line 202: ⚠️ Could we add one traversal-level test for the `tryGetGraph(newStep) == null` path?

The helper tests cover the direct `canExtractHasContainer(null, ...)` behavior, but they do not exercise `extractHasContainer()` through a real child traversal shape where TinkerPop may leave the child traversal without an attached graph. A small case around a `__` traversal, for example a `union()` / child traversal containing `has("vp4", P.lt(...))`, would better guard the null-graph fallback that this patch relies on.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -166,38 +167,103 @@ public static void trySetGraph(Step<?, ?> step, HugeGraph graph) {
...
                 }
-                TraversalHelper.copyLabels(step, step.getPreviousStep(), false);
-                traversal.removeStep(step);
             }
-            step = step.getNextStep();
+            step = nextStep;
         } while (step instanceof HasStep || step instanceof NoOpBarrierStep);
     }
 
+    private static boolean extractHasContainers(HugeGraphStep<?, ?> newStep,
```
imbajin, line 200: 🧹 Could we add one edge-start regression case as well?

This patch changes the `HugeGraphStep` extraction path for both vertex and edge returns, but the new traversal tests are centered on `V()` / `out()` vertex filtering. A small `g.E().has(textProp, P.lt(...))` case would confirm that text range predicates are also kept out of backend extraction when the `HugeGraphStep` is returning edges.

imbajin (commented):
The current patch looks scoped back to the safer minimal fix. I left three non-blocking test-coverage suggestions to better guard the text-range extraction behavior and related fallback paths.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
LegendPei: Thanks, added the suggested regression coverage without changing the production logic.
﻿
The new tests cover:
- `P.gte(...)` for another single text-range predicate shape
- `P.between(...)` for the connective/`collectPredicates()` path
- an `extractHasContainer()` traversal-level case with no attached graph, covering the `tryGetGraph(newStep) == null` fallback
- an edge-start `g.E().has("ep4", P.lt(""))` case
﻿
So the implementation remains scoped to the minimal option A: keep the whole `HasStep` in the traversal layer when it contains a `Text` range predicate, and avoid backend text-range extraction.

imbajin (approved):
+1 from my side. The current version is scoped back to the safer minimal fix: keep `Text` range predicates out of backend extraction and leave the affected `HasStep` in the traversal layer. The previous review concerns around the broader split/remove optimization and missing regression coverage have been addressed, and the current CI is green.

There are still a few possible follow-up optimizations, but I do not think they should block this PR:

- Mixed `HasStep` optimization: when a `HasStep` contains both a `Text` range predicate and other safe predicates, the whole step is kept in the traversal layer. This is conservative and correct for this bug fix, but a later PR could explore redundant backend pre-filtering while preserving the traversal-layer filter for correctness.
- Additional edge cases: the current tests cover the important paths (`V()`, `out()`, `E()`, null graph fallback, `lt/gte/between`, and suite registration). Future coverage could add more predicate combinations such as `gt/lte`, `inside/outside`, `not(...)`, and `NoOpBarrierStep` boundaries.
- Performance follow-up: if mixed text-range filters become common in real workloads, we could evaluate whether a narrower pushdown strategy is worth adding with dedicated planner/index tests.

These are follow-up improvements rather than merge blockers. The PR now looks proportional to the original correctness issue.

PR #3035, opened 2026-05-21
fix(server): normalize typed DEFAULT_VALUE after JSON reload

### LINKED ISSUE #3028: [Bug] Preserve typed DEFAULT_VALUE and struct userdata after JSON reload
### Bug Type (问题类型)

logic (逻辑设计问题)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0
- Backend: schema metadata reload paths; not backend-specific
- OS: N/A
- Data Size: N/A

### Expected

`Userdata.DEFAULT_VALUE` (`~default_value`) should preserve the expected runtime type after schema metadata is serialized and reloaded.

For example, a `PropertyKey` with `DataType.DATE` should reload its default value as `java.util.Date`, not as a generic JSON `String`.

### Actual

Schema userdata is deserialized through raw `Map` / `Object` paths, so Jackson cannot infer the original runtime type.

For `~default_value`, a `Date` default value can be serialized as a formatted string and later reloaded as `String`.

This may affect multiple schema reload paths, including:

- `SchemaMetaManager -> PropertyKey.fromMap(...)`
- JSON-based backend serializers such as Text/Binary/Table/Cassandra-style userdata reload paths
- the mirror userdata model in `hugegraph-struct`

### Possible fix direction

`PropertyKey` already knows its `dataType`, so `~default_value` could be normalized after the `PropertyKey` is reconstructed, using the same conversion logic already used for property values.

### Scope

- Investigate `Userdata.DEFAULT_VALUE` reload behavior, especially for `DataType.DATE`.
- Normalize confirmed typed default values after schema reload.
- Check whether the same gap exists in `hugegraph-struct`.
- Add regression tests for the confirmed reload path(s).

```javascript

```

```javascript

```

dpol1:
## Purpose of the PR

- fix #3028

Normalizes `PropertyKey` default values to their declared data type upon retrieval. Previously, values stored in `userdata` could lose their original type during serialization and deserialization (for example, a `DATE` default reloaded as a `String`), leading to type mismatches when applying defaults.

The `defaultValue()` method in both `hugegraph-core` and `hugegraph-struct` now converts deserialized representations back to their expected runtime types using the existing type-aware converters based on `PropertyKey.dataType()`. This ensures that typed defaults (such as `DATE`) remain correctly typed after a JSON round-trip and when applied to vertices and edges.

This change is verified with tests covering schema parsing, vertex property assignment, and both binary and text serializers.

## Main Changes

- **Core `PropertyKey.defaultValue()` normalization**
  - In `hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/PropertyKey.java`, `defaultValue()` now:
    - reads `Userdata.DEFAULT_VALUE` from `userdata`,
    - if non-null, normalizes it via `validValue(value)`, which routes through the existing type-aware converters (`convValue` / `convSingleValue` → `dataType().valueToDate/Number/UUID/Blob`).
  - This ensures that defaults reloaded from JSON as `String` (for typed keys such as `DATE`) are converted back to their runtime type before being used by consumers like `HugeElement.updateToDefaultValueIfNone()`.
  - SET Cardinality fix: Updated convValue() in both core and struct to map Cardinality.SET to a LinkedHashSet instead of passively preserving the JSON parser's ArrayList.

- **Struct `PropertyKey.defaultValue()` mirror**
  - Applies the same normalization logic to `hugegraph-struct/src/main/java/org/apache/hugegraph/struct/schema/PropertyKey.java` so that struct `PropertyKey` behaves consistently with core:
    - `defaultValue()` also normalizes the stored `~default_value` according to the `DataType`.
  - Adds a focused unit test `PropertyKeyTest.testDefaultValueNormalizedToDate()` in `hugegraph-struct` that constructs a `DATE` property key with a `String` default and asserts that `defaultValue()` returns a `Date`.
  - SET Cardinality fix: Updated convValue() in both core and struct to map Cardinality.SET to a LinkedHashSet instead of passively preserving the JSON parser's ArrayList.

- **End-to-end vertex default application**
  - Adds `testAddVertexWithDateDefaultValue()` to `VertexCoreTest`:
    - defines a `DATE` property key with `Userdata.DEFAULT_VALUE` set to a `Date`,
    - creates a vertex without explicitly setting that property,
    - commits, reloads the vertex from the backend (forcing schema reload),
    - asserts that the applied property is a `Date` with the expected value, not a `String`.
  - This reproduces and locks in the user-facing symptom from #3028 through `HugeElement.updateToDefaultValueIfNone()`.

- **Serializer round-trip for `DEFAULT_VALUE`**
  - Extends `TextSerializerTest` and `BinarySerializerTest` with:
    - `testPropertyKeyDefaultValueRoundTripsAsDate()`:
      - serializes a `DATE` `PropertyKey` whose `DEFAULT_VALUE` is a `Date`,
      - deserializes it back,
      - asserts that `defaultValue()` on the reloaded key returns a `Date` equal to the original.
  - Ensures that both text and binary serializers preserve the typed default after a full round-trip.

- **Schema parsing normalization (core)**
  - Extends `SchemaElementTest` with:
    - `testPropertyKeyFromMapNormalizesDateDefaultValue()`:
      - builds a `PropertyKey` from a `Map` with `DATA_TYPE=DATE` and `~default_value` as a formatted date `String`,
      - asserts that `defaultValue()` returns a `Date`.
    - `testPropertyKeyFromMapNormalizesDateSetDefaultValue()`:
      - same as above but for `Cardinality.SET`, with `~default_value` as a collection of date strings,
      - asserts that `defaultValue()` returns a `Collection` whose elements are all `Date` instances.
  - This validates both the scalar and collection branches of `validValue()` for `DATE` defaults.

## Verifying these changes

  - `hugegraph-core`:
    - `VertexCoreTest.testAddVertexWithDateDefaultValue()`
    - `SchemaElementTest.testPropertyKeyFromMapNormalizesDateDefaultValue()`
    - `SchemaElementTest.testPropertyKeyFromMapNormalizesDateSetDefaultValue()`
  - Serializers:
    - `TextSerializerTest.testPropertyKeyDefaultValueRoundTripsAsDate()`
    - `BinarySerializerTest.testPropertyKeyDefaultValueRoundTripsAsDate()`
  - `hugegraph-struct`:
    - `PropertyKeyTest.testDefaultValueNormalizedToDate()`
    - `PropertyKeyTest.testSetDefaultValueCollapsesDuplicatesAndReturnsSet()`

All existing tests remain green.

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [x] Other affects (typed here)
  - Tightens the behavior of `PropertyKey.defaultValue()` for malformed or type-incompatible defaults: such values are now validated and may trigger conversion errors instead of leaking unexpected raw `String` values at read time.
- [ ] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/PropertyKey.java
```diff
@@ -121,7 +121,12 @@ public void defineDefaultValue(Object value) {
 
     public Object defaultValue() {
         // TODO add a field default_value
-        return this.userdata().get(Userdata.DEFAULT_VALUE);
+        Object value = this.userdata().get(Userdata.DEFAULT_VALUE);
+        // Userdata is reloaded from JSON as a raw Map, so a typed default
+        // value (e.g. Date) comes back as a String. Normalize it to the
+        // runtime type expected by this property key's data type. Idempotent
+        // for values already of the expected type.
+        return value == null ? null : this.validValue(value);
```
imbajin: ⚠️ **Preserve SET cardinality when normalizing defaults**

`defaultValue()` now delegates `~default_value` normalization to `validValue()`, but JSON-reloaded SET defaults arrive as an `ArrayList`. The existing `convValue()` then chooses the output collection from the input type, so a `Cardinality.SET` default can still come back as an `ArrayList`. The new SET test only asserts `Collection`, so it misses this.

This can install list-valued defaults for SET properties after schema reload, preserving duplicates and breaking cardinality semantics. Please make both core and struct `PropertyKey.convValue()` choose the target container from `this.cardinality` (`LinkedHashSet` for SET, `ArrayList` for LIST), and add a regression assertion that a SET default returns a `Set` and collapses duplicates.

imbajin (commented):
Found one SET-cardinality regression in the default-value normalization path.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/PropertyKey.java
dpol1: Good catch, thanks! You were right about the container semantics. I reverted the broader change in `convValue()` to avoid breaking other paths, and instead only normalize the `defaultValue()` when the cardinality is `SET`, wrapping it in a `LinkedHashSet` so duplicates are collapsed but order is preserved. I also updated the tests to inject duplicates and verify that `SET` defaults behave as expected while `SINGLE/LIST` stay unchanged. Latest fix is in a0fe682.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/PropertyKey.java
```diff
@@ -121,7 +121,23 @@ public void defineDefaultValue(Object value) {
...
+            return null;
+        }
+
+        // Userdata is reloaded from JSON as a raw Map, so a typed default
+        // value (e.g. Date) comes back as a String. Normalize it to the
+        // runtime type expected by this property key's data type. Idempotent
+        // for values already of the expected type.
+        Object normalized = this.validValue(value);
+
+        // For SET cardinality, ensure we return a Set container and collapse duplicates
```
imbajin: ⚠️ **Cover SET defaults through server round trips**

This branch now special-cases `Cardinality.SET` after `validValue()` so JSON-reloaded defaults return a `LinkedHashSet`, but the server-side serializer and graph tests added in this PR still cover only SINGLE date defaults. Please add a Binary/Text serializer round-trip, and preferably one graph-level default-injection regression, for a `valueSet().asDate()` default so the new SET container/duplicate-collapse behavior is locked at the server boundary.

imbajin (commented):
Blocking: no. Summary: The SET default normalization path still needs server-side regression coverage. Evidence: local schema/serializer tests passed (23 tests) and vertex default test passed (1 test), but the added server tests exercise only SINGLE date defaults.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/PropertyKey.java
dpol1: Hey @imbajin, pushed the missing server tests (`VertexCoreTest` + serializers). I used duplicate `String` arrays to properly mimic the JSON reload and verify the `Set` collapse.

About Copilot's comments: I switched to `validValueOrThrow()` (it was reasonable), but ignored the tip to change `convValue()`. Forcing a `Set` there breaks the `<V> V` contract and throws `ClassCastException`s in existing tests. Keeping the fix inside `defaultValue()` is much safer.

Good to merge?

imbajin (approved):
+1 from my side. The current head fixes the typed DEFAULT_VALUE reload path for both core and struct, including DATE defaults and SET duplicate-collapse behavior, and the CI checks are green.

Non-blocking follow-up: it may still be worth canonicalizing the normalized DEFAULT_VALUE back into userdata, or adding hasSameContent/TableSerializer regression coverage, so raw schema-content comparisons are also type-stable after serializer reloads. I do not think this should block this PR because the user-facing defaultValue/default-application path is covered.

PR #3039, opened 2026-05-26
fix(server): handle match() in no index case

### LINKED ISSUE #2936: [Bug] NoIndexException triggered by match()-step
### Bug Type (问题类型)

gremlin (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0
- Backend: RocksDB x nodes
- Data Size:  10 vertices, 20 edges

### Expected behavior
```gremlin
g.V().has('vp4', neq('J2O')).has('vl1', 'vp2',gte(false)).has('vp2').has('vl0', 'vp3',gt(4592737712018141718)).out().count()

g.V().has('vp4', neq('J2O')).has('vl1', 'vp2',gte(false)).match(__.as('start0').has('vp2').has('vl0', 'vp3',gt(4592737712018141718)).repeat(__.out()).times(1).as('m0')).select('m0').count()
```
### Actual behavior
+ The first query returned the results normally.
+ The second query thrown a exception: ```org.apache.hugegraph.exception.NoIndexException: Don't accept query based on properties [vp4, vp2] that are not indexed in label 'vl1', may not match secondary/range/not-equal condition```
### 
+ We moved the```has(...)``` step and ```out()```step in first query into the ```match()```step and replace the ```out()``` with ```repeat()``` step, which should not affect the result.
+ When I executed these two queries on Janusgraph and Tinkerpop, they both returned the same results and were consistent with the results of the first query.
+ I have implemented a fuzzing tool and am using it to test HugeGraph. To reduce the burden on developers, I have simplified the test cases as much as possible and isolated irrelevant information. I hope this work can be helpful for further improving the stability of HugeGraph.

```javascript
none
```

```javascript
none
```

LegendPei:
## Purpose of the PR

- close #2936

This PR fixes a `NoIndexException` triggered by `match()` traversals when preceding `has()` filters contain index-required predicates such as `neq()` or range predicates but the related properties are not indexed.

In this case, HugeGraph extracted these filters into `HugeGraphStep` and pushed them into backend `ConditionQuery`, causing backend index validation to reject the traversal before TinkerPop could evaluate the filters normally.

## Main Changes

- Avoid extracting `HasStep` filters into `HugeGraphStep` when:
  - the following traversal step is `match()`, and
  - the `has()` predicate contains `neq`, `gt`, `gte`, `lt`, or `lte`.

- Keep those filters in the traversal so they are evaluated by the normal TinkerPop filter path instead of being forced into backend index query planning.

- Add a regression test covering the `match()` + no-index condition scenario from #2936.

## Verifying these changes

    - `mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,rocksdb -Dtest=CountStrategyCoreTest#testMatchWithNoIndexConditionMatchesDirectTraversal -DfailIfNoTests=false`
    - `mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,rocksdb -Dtest=CountStrategyCoreTest -DfailIfNoTests=false`

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -179,7 +193,103 @@ public static void extractHasContainer(HugeGraphStep<?, ?> newStep,
...
+
+    private static boolean hasUsableMatchIndex(HugeGraph graph,
+                                               HugeGraphStep<?, ?> step,
+                                               HasContainer has) {
+        if (isSysProp(has.getKey())) {
+            return true;
+        }
+
+        PropertyKey pkey = graph.propertyKey(has.getKey());
+        if (pkey.dataType() != DataType.BOOLEAN) {
```
VGalaxies: Non-blocking: `hasUsableMatchIndex()` currently treats only boolean properties with secondary/unique indexes as safe to push down before `match()`. That seems narrower than the predicate set being handled (`gt/gte/lt/lte/neq`) and may stop range-indexed numeric/text predicates from using the backend index. Please add a regression test for an indexed non-boolean range predicate before `match()`, or generalize the index-type check.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/CountStrategyCoreTest.java
```diff
@@ -125,4 +183,164 @@ public void testWhereCountGteNegativeDoesNotBuildInvalidRange() {
...
+                                                        .hasLabel(P.neq("vl0"))
+                                                        .match(__.<Vertex>as("s")
+                                                                 .has("vp2")
+                                                                 .as("m"))
+                                                        .<Vertex>select("m")
+                                                        .count();
+
+        HugeGraphStep<?, ?> graphStep = applyAndGetGraphStep(traversal);
+        Assert.assertEquals(1, graphStep.getHasContainers().size());
+        Assert.assertFalse(hasRemainingHasStep(traversal, T.label.getAccessor()));
```
imbajin: ⚠️ **Execute the system-property traversal in this regression test**

This test only calls `applyStrategies()` and checks the optimized step shape, so the new `isSysProp()` allow-path can still break at runtime without failing here. Please execute the traversal, or compare it with an equivalent direct traversal, so `hasLabel(P.neq(...))` before `match()` is covered end to end.

imbajin (commented):
Blocking: no. Summary: One test coverage gap remains for the new system-property match path. Evidence: local `mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,memory -Dtest=CountStrategyCoreTest -DfailIfNoTests=false -ntp` passed 12 tests, but this added test only applies strategies and never executes the traversal.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -179,7 +193,142 @@ public static void extractHasContainer(HugeGraphStep<?, ?> newStep,
...
                 traversal.removeStep(step);
             }
-        } while (step instanceof HasStep || step instanceof NoOpBarrierStep);
+            step = nextStep;
+        }
+    }
+
+    private static boolean followedByMatchStep(Step<?, ?> step) {
+        Step<?, ?> next = step.getNextStep();
+        while (next instanceof HasStep || next instanceof NoOpBarrierStep) {
```
imbajin: ‼️ **Treat transparent steps before `match()` consistently**

`followedByMatchStep()` only skips `HasStep` and `NoOpBarrierStep`, so a traversal like `g.V().has("vp4", P.neq("J2O")).identity().match(...)` no longer looks match-adjacent here and the unsafe range/neq predicate can still be extracted into `HugeGraphStep`. Other extraction paths in this file already treat `IdentityStep` as transparent, so please skip it here too and add a regression with an intervening `identity()` before `match()`.

imbajin (commented):
Blocking: yes. Summary: A transparent step between has() and match() can still bypass the new no-index guard. Evidence: static review of TraversalUtil.followedByMatchStep() on current head c6f649 and local CountStrategyCoreTest passed 13 tests without covering this shape.

imbajin:
image

address the conflicts, thx~

imbajin (commented):
Blocking: yes. Summary: The new match-specific extraction path can throw when a traversal has no graph context. Evidence: local regression test `TraversalUtilOptimizeTest#testExtractHasContainerBeforeMatchWithoutGraphKeepsHasStep` failed with a NullPointerException at `TraversalUtil.java:277`; existing targeted tests passed 30 tests.

---
🔗 Latest-head CI also has visible failures; please check the failing server/hstore jobs: https://github.com/apache/hugegraph/actions/runs/26890724385/job/79315653294

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/TraversalUtil.java
```diff
@@ -181,6 +204,203 @@ public static void extractHasContainer(HugeGraphStep<?, ?> newStep,
...
+        return false;
+    }
+
+    private static List<HasContainer> extractUsableHasContainers(
+            HugeGraphStep<?, ?> step, HasContainerHolder holder) {
+        List<HasContainer> extracted = new ArrayList<>();
+        HugeGraph graph = tryGetGraph(step);
+        for (HasContainer has : holder.getHasContainers()) {
+            if (hasMatchIndexSensitivePredicate(has) &&
+                !hasUsableMatchIndex(graph, step, has)) {
```
imbajin: ‼️ **Handle missing graph before checking match indexes**

When `extractUsableHasContainers()` evaluates a match-sensitive predicate, this branch calls `hasUsableMatchIndex(graph, step, has)` even when `tryGetGraph(step)` returned `null`. The existing non-match path keeps such has steps local, but a graph-less traversal shaped like `__.V().has("age", P.gt(18)).match(...)` now throws `NullPointerException` at `graph.propertyKey(...)` instead. Please treat `graph == null` as not usable here, matching `hasUnusableMatchPredicate()` above, and keep the `HasStep` in the traversal.

Evidence: a local regression test adding that traversal to `TraversalUtilOptimizeTest` fails with `Cannot invoke "org.apache.hugegraph.HugeGraph.propertyKey(String)" because "graph" is null`; the existing targeted suite still passes 30 tests, so this path is currently unprotected.

imbajin (approved):
LGTM. Current head addresses the previous match/no-index blockers, all latest GitHub checks pass, and the focused local suite passed 31 tests for CountStrategyCoreTest and TraversalUtilOptimizeTest.

PR #3010, opened 2026-04-26
chore(ci): optimize rerun workflow and add macOS RocksDB coverage

contrueCT:
## Purpose of the PR

Optimize the Rerun CI workflow and add macOS coverage for the single-node RocksDB server path.

The rerun workflow now only starts the rerun decision job when an upstream CI workflow actually fails, avoiding unnecessary skipped rerun workflow records after successful or non-failure CI completions.

This PR also adds macOS RocksDB server checks in parallel with the existing Linux Server CI, covering both Intel and Apple Silicon runners for compile, core test, server startup, and API test coverage.

## Main Changes

- Optimize `.github/workflows/rerun-ci.yml`
  - Add a job-level condition to `decide-rerun-action` so it only runs when `github.event.workflow_run.conclusion == 'failure'`
  - Remove redundant non-failure shell handling from the decision script
  - Keep the existing supported events, retry limit, delay, permissions, and `gh run rerun --failed` behavior unchanged
- Add macOS RocksDB coverage to `.github/workflows/server-ci.yml`
  - Add `build-server-macos-rocksdb` for `macos-15-intel` and `macos-15`
  - Run compile, RocksDB core tests, and RocksDB API tests on macOS
- Improve macOS compatibility in CI helper scripts
  - Make `start-server.sh` avoid Bash 4 associative arrays and use portable backend serializer selection
  - Add macOS-compatible `sed -i` handling
  - Make `run-api-test.sh` and `build-report.sh` resolve paths from the script/repo root and use `curl`/`wget` fallback downloads
- Stabilize Apple Silicon RocksDB API tests
  - Pass scoped `SERVER_JAVA_OPTIONS` for `macos-15` (`-Xms512m -Xmx2g`) through the existing server startup `-j` option path
  - Keep the JaCoCo javaagent behavior intact

## Verifying these changes

- Ran `git diff --check`
- Ran `bash -n` for the changed shell scripts
- Verified GitHub Actions on this PR:
  - `build-server-macos-rocksdb (macos-15-intel)` passed
  - `build-server-macos-rocksdb (macos-15, -Xms512m -Xmx2g)` passed
  - Existing Linux Server CI jobs for `memory`, `rocksdb`, and `hbase` passed
  - Commons, PD/Store/HStore, License Checker, 3rd-party, and CodeQL checks passed

## Does this PR potentially affect the following parts?

[...]

Issue #3006, opened 2026-04-23
[Bug] README “Build from Source” instructions outdated. "gremlin-console.sh" fails on Mac M4 (Apple Silicon)


vaijosh:
### Bug Type (问题类型)

other exception / error (其他异常报错)

### Before submit

- [x] 我已经确认现有的 [Issues](https://github.com/apache/hugegraph/issues) 与 [FAQ](https://hugegraph.apache.org/docs/guides/faq/) 中没有相同 / 重复问题 (I have confirmed and searched that there are no similar problems in the historical issue and documents)

### Environment (环境信息)

- Server Version: 1.0.0 (Apache Release Version)
- Backend: RocksDB x nodes, HDD or SSD 
- OS: xx CPUs, xx G RAM, Ubuntu 2x.x / CentOS 7.x 
- Data Size:  xx vertices, xx edges 

### Expected & Actual behavior (期望与实际表现)

Description: 
- The README’s “Build from Source” instructions are outdated and do not match the current directory structure and startup instructions after building from source. As a result, users cannot start the server as described.
- Additionally, the bundled Gremlin Console (bin/gremlin-console.sh) fails to start on Mac M4 (Apple Silicon) due to native library (Jansi) incompatibility and Java version issues.

Details:
- After building from source, the server scripts are not in the location described in the README, causing startup failures.
- On Mac M4, running Gremlin Console results in errors like:   java.lang.UnsatisfiedLinkError: Could not load library. Reasons: [no jansi in java.library.path ... (fat file, but missing compatible architecture (have 'x86_64,i386', need 'arm64'))] and the gremlin console works only with Java 11

Proposed Fix:
- Update the README to reflect the correct post-build directory structure and server startup steps.
- Add troubleshooting notes for Apple Silicon and Java 17+ users regarding Gremlin Console

### Vertex/Edge example (问题点 / 边数据举例)

```javascript

```

### Schema [VertexLabel, EdgeLabel, IndexLabel] (元数据结构)

```javascript

```

vaijosh:
Many thanks @imbajin for reviewing and merging the changes.

PR #3040, opened 2026-05-26
chore(ci): add gremlin-console smoke test

### LINKED ISSUE #3015: [Improve] add macOS Apple Silicon smoke test for Gremlin Console
## Background

This follow-up task is related to issue #3006.

Issue #3006 reported that `bin/gremlin-console.sh` can fail on Mac M4 / Apple Silicon. However, the exact cause has not been fully isolated yet and should not be assumed in advance.

PR #3010 is intentionally focused on rerun CI behavior and macOS RocksDB server/API checks, so the Gremlin Console Apple Silicon validation is tracked separately in this follow-up task.

Related context:
- issue #3006 tracks the user-facing problem around outdated build/startup docs and `bin/gremlin-console.sh` failures on Mac M4 / Apple Silicon
- PR #3010 adds macOS RocksDB CI coverage, but does not cover the `bin/gremlin-console.sh` entry point

## Problem

`bin/gremlin-console.sh` is a different entry point from the HugeGraph server startup path and has its own runtime behavior on macOS Apple Silicon.

At the moment, this path is not covered by a dedicated macOS Apple Silicon smoke test in CI, which means regressions or platform-specific startup failures may not be detected early.

## Proposal

Add a focused follow-up for Gremlin Console on macOS Apple Silicon, for example:

- add a lightweight macOS smoke test for `bin/gremlin-console.sh`
- validate a tiny non-interactive invocation such as:
  - `gremlin-console.sh -- -e <script>`
- investigate the actual startup/runtime failure mode on Apple Silicon

## Expected Outcome

- CI can catch obvious regressions in the console startup path

contrueCT:
## Purpose of the PR

- close #3015 

Add a focused Gremlin Console smoke test for the server distribution entry point.

This follows up on #3015 and the review note from #3010. The existing macOS RocksDB CI covers server compile, core tests, startup, and API tests, but it does not directly exercise `bin/gremlin-console.sh`. This PR adds a lightweight non-interactive smoke test for that path on macOS and Linux.

Related context: #3006 reported user-facing build/startup issues and mentioned Gremlin Console behavior on Apple Silicon, but this PR only adds CI validation and does not assume a specific Jansi or Java compatibility root cause.

## Main Changes

- Add `run-gremlin-console-smoke-test.sh`
  - Generates a tiny Groovy script
  - Runs `bin/gremlin-console.sh -- -e <script>`
  - Verifies the expected `gremlin-console-smoke-ok` marker
  - Cleans temporary script/log files with `trap`
- Run the smoke test in Linux Server CI only for the RocksDB matrix entry
- Run the same smoke test in macOS RocksDB CI on both Intel and Apple Silicon runners

## Verifying these changes

    - Ran `bash -n hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh`
    - Ran `git diff --check`

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-dist/src/assembly/travis/run-api-test.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/travis/run-api-test.sh
@@ -20,6 +20,7 @@ set -ev
 BACKEND=$1
 REPORT_DIR=$2
 REPORT_FILE=$REPORT_DIR/jacoco-api-test-for-raft.xml
+RUN_GREMLIN_CONSOLE_SMOKE_TEST=${3:-false}
 
 TRAVIS_DIR=$(cd "$(dirname "$0")" && pwd)
 REPO_ROOT=$(cd "$TRAVIS_DIR/../../../../.." && pwd)
@@ -97,6 +98,11 @@ $TRAVIS_DIR/start-server.sh $SERVER_DIR $BACKEND $JACOCO_PORT || (cat $SERVER_DI
 
 # run api-test
 mvn test -pl hugegraph-server/hugegraph-test -am -P api-test,$BACKEND || (cat $SERVER_DIR/logs/hugegraph-server.log && exit 1)
+
+if [ "$RUN_GREMLIN_CONSOLE_SMOKE_TEST" == "true" ]; then
+    $TRAVIS_DIR/run-gremlin-console-smoke-test.sh "$SERVER_DIR" || \
+        (cat "$SERVER_DIR/logs/hugegraph-server.log" && exit 1)
+fi
 
 $TRAVIS_DIR/build-report.sh $BACKEND $JACOCO_PORT $REPORT_FILE
 
--- a/hugegraph-server/hugegraph-dist/src/assembly/travis/run-api-test.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/travis/run-api-test.sh
@@ -100,7 +100,7 @@ $TRAVIS_DIR/start-server.sh $SERVER_DIR $BACKEND $JACOCO_PORT || (cat $SERVER_DI
 mvn test -pl hugegraph-server/hugegraph-test -am -P api-test,$BACKEND || (cat $SERVER_DIR/logs/hugegraph-server.log && exit 1)
 
 if [ "$RUN_GREMLIN_CONSOLE_SMOKE_TEST" == "true" ]; then
-    $TRAVIS_DIR/run-gremlin-console-smoke-test.sh "$SERVER_DIR" || \
+    bash "$TRAVIS_DIR/run-gremlin-console-smoke-test.sh" "$SERVER_DIR" || \
         (cat "$SERVER_DIR/logs/hugegraph-server.log" && exit 1)
 fi
 
--- a/hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
@@ -38,23 +38,27 @@ fi
 
 TMP_DIR=${TMPDIR:-/tmp}
 TMP_WORK_DIR=$(mktemp -d "${TMP_DIR}/hugegraph-gremlin-console-smoke.XXXXXX")
-SMOKE_REMOTE_CONFIG="${TMP_WORK_DIR}/remote.yaml"
+REMOTE_CONFIG_BACKUP="${TMP_WORK_DIR}/remote.yaml.orig"
 SMOKE_SCRIPT="${TMP_WORK_DIR}/smoke.groovy"
 SMOKE_LOG="${TMP_WORK_DIR}/smoke.log"
 
 cleanup() {
+    if [ -f "$REMOTE_CONFIG_BACKUP" ]; then
+        cp "$REMOTE_CONFIG_BACKUP" "$REMOTE_CONFIG"
+    fi
     rm -rf "$TMP_WORK_DIR"
 }
 trap cleanup EXIT
 
-cp "$REMOTE_CONFIG" "$SMOKE_REMOTE_CONFIG"
-cat >> "$SMOKE_REMOTE_CONFIG" <<EOF
+cp "$REMOTE_CONFIG" "$REMOTE_CONFIG_BACKUP"
+cat >> "$REMOTE_CONFIG" <<EOF
+
 username: admin
 password: pa
 EOF
 
 cat > "$SMOKE_SCRIPT" <<EOF
-:remote connect tinkerpop.server ${SMOKE_REMOTE_CONFIG}
+:remote connect tinkerpop.server conf/remote.yaml
 :> def count = g.V().count().next(); if (count < 0L) throw new IllegalStateException("Unexpected vertex count: " + count); "${SMOKE_MARKER}-" + count
 EOF
 
--- a/hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
@@ -62,6 +62,7 @@ EOF
 cat > "$SMOKE_SCRIPT" <<EOF
 :remote connect tinkerpop.server conf/remote.yaml
 :> def count = g.V().count().next(); if (count < 0L) throw new IllegalStateException("Unexpected vertex count: " + count); "${SMOKE_MARKER}-" + count
+println(result[0].object)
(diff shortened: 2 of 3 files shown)


hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
```diff
@@ -0,0 +1,55 @@
...
+    exit 1
+fi
+
+if [ ! -x "$SERVER_DIR/bin/gremlin-console.sh" ]; then
+    echo "Gremlin Console script is not executable: $SERVER_DIR/bin/gremlin-console.sh"
+    exit 1
+fi
+
+TMP_DIR=${TMPDIR:-/tmp}
+SMOKE_SCRIPT=$(mktemp "${TMP_DIR}/hugegraph-gremlin-console-smoke.XXXXXX.groovy")
```
VGalaxies: **Major: Non-portable `mktemp` template breaks macOS smoke test**

`hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh:34`

**Evidence**
- `.github/workflows/server-ci.yml:166` runs this script on `macos-15` and `macos-15-intel`, but lines 34-35 use templates ending in `.groovy`/`.log` after `XXXXXX`. macOS `mktemp(1)` documents replacing only appended/trailing `X`s: https://keith.github.io/xcode-man-pages/mktemp.1.html

**Impact**
- The new macOS CI smoke-test step can fail before Gremlin Console is launched, making the PR introduce unreliable CI coverage.

**Requested fix**
- Use a portable trailing-`X` template, for example create a temp directory with `mktemp -d "${TMPDIR:-/tmp}/hugegraph-gremlin-console-smoke.XXXXXX"` and place `smoke.groovy` / `smoke.log` inside it, or omit the file extensions from the `mktemp` templates.

VGalaxies (commented):
**Review summary**

- **Blocking:** yes
- **Summary:** The new smoke-test script is not portable to the macOS CI jobs added by this PR.
- **Evidence:**
  - `git diff --check origin/master...HEAD` clean
  - `bash -n .../run-gremlin-console-smoke-test.sh` clean
  - static review of changed workflow/script

hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
contrueCT: Thanks for catching this. You're right — the original mktemp template is not portable to macOS because the Xs are not trailing. I updated the helper to create a temporary directory with a portable mktemp -d template and place smoke.groovy / smoke.log inside it. The fix has been pushed to this PR.

hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
```diff
@@ -0,0 +1,56 @@
...
+TMP_WORK_DIR=$(mktemp -d "${TMP_DIR}/hugegraph-gremlin-console-smoke.XXXXXX")
+SMOKE_SCRIPT="${TMP_WORK_DIR}/smoke.groovy"
+SMOKE_LOG="${TMP_WORK_DIR}/smoke.log"
+
+cleanup() {
+    rm -rf "$TMP_WORK_DIR"
+}
+trap cleanup EXIT
+
+cat > "$SMOKE_SCRIPT" <<EOF
```
VGalaxies, line 60: **Medium: Smoke test does not exercise HugeGraph integration**

`hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh:43`

**Evidence**
- The generated script only evaluates `1 + 1`, prints `gremlin-console-smoke-ok`, and runs it with `bin/gremlin-console.sh -- -e "$SMOKE_SCRIPT"`. It never connects through `conf/remote.yaml` or executes a remote traversal; in the Linux job it is also invoked after `run-api-test.sh`, which stops the server.

**Impact**
- CI can pass even if the packaged Gremlin Console cannot connect to HugeGraph, the remote config is broken, or Gremlin Server auth/traversal compatibility regresses.

**Requested fix**
- Run the smoke while HugeGraph is running and execute a harmless remote operation through the packaged config, for example `:remote connect tinkerpop.server conf/remote.yaml` with the CI auth settings and an assertion on `g.V().count()` or an equivalent remote expression.

VGalaxies (commented):
**Review summary**

- **Blocking:** no
- **Summary:** The new Gremlin Console smoke test only validates local console startup, not connectivity to HugeGraph.
- **Evidence:**
  - Reviewed `git diff origin/master...HEAD`
  - `git diff --check origin/master...HEAD` passed
  - `rg` found no `remote.yaml`, `:remote`, or remote traversal usage in the new script

hugegraph-server/hugegraph-dist/src/assembly/travis/run-gremlin-console-smoke-test.sh
contrueCT: Thanks, updated the smoke test to run while HugeGraph is still running inside run-api-test.sh. It now copies the packaged conf/remote.yaml, appends the CI admin credentials, connects with :remote connect tinkerpop.server, and submits a harmless remote g.V().count().next() check before validating the smoke marker. The separate post-API-test smoke step has been removed.

hugegraph-server/hugegraph-dist/src/assembly/travis/run-api-test.sh
```diff
@@ -98,6 +99,11 @@ $TRAVIS_DIR/start-server.sh $SERVER_DIR $BACKEND $JACOCO_PORT || (cat $SERVER_DI
 # run api-test
 mvn test -pl hugegraph-server/hugegraph-test -am -P api-test,$BACKEND || (cat $SERVER_DIR/logs/hugegraph-server.log && exit 1)
 
+if [ "$RUN_GREMLIN_CONSOLE_SMOKE_TEST" == "true" ]; then
+    $TRAVIS_DIR/run-gremlin-console-smoke-test.sh "$SERVER_DIR" || \
```
imbajin: ‼️ **Make the smoke helper executable or invoke it through bash**

Evidence: this line executes `$TRAVIS_DIR/run-gremlin-console-smoke-test.sh` directly, but the new file is checked in as mode `100644` (`git ls-tree HEAD .../run-gremlin-console-smoke-test.sh`) and local `test -x` returns false. The latest `build-server (rocksdb, 11)` job reaches this line after the API tests pass and then fails with `Permission denied`.

Impact: every CI path that enables the smoke test fails before validating Gremlin Console. Please either commit the helper with executable mode (`chmod +x`) or call it explicitly with `bash` here and in the macOS path.

imbajin (commented):
Blocking: yes. Summary: The new Gremlin Console smoke helper is invoked directly but is checked in without the executable bit. Evidence: latest-head CI fails with Permission denied at run-api-test.sh:103, and git ls-tree shows mode 100644.
---
🔗 Please check the failed latest-head jobs, for example https://github.com/apache/hugegraph/actions/runs/27470916921/job/81205575322.

hugegraph-server/hugegraph-dist/src/assembly/travis/run-api-test.sh
contrueCT: Thanks, you were right about the invocation issue. The helper was added as a non-executable script, and the original workflow invoked it through bash. When moving the smoke test into run-api-test.sh to keep it inside the live server window, the call accidentally became direct execution.

I updated run-api-test.sh to invoke the helper through bash again, kept the smoke test inside the API-test server lifecycle, and adjusted the console script so it works under Gremlin Console's -e evaluation model. The latest Server CI is green on Linux rocksdb and both macOS runners, with the smoke marker emitted after connecting to the live Gremlin Server.

imbajin (approved):
LGTM, +1. The previous CI smoke-test issues have been addressed on the current head, and the relevant Linux/macOS RocksDB jobs are green with the smoke marker emitted after connecting to the live Gremlin Server.

Non-blocking follow-up: the new smoke test now validates the packaged `bin/gremlin-console.sh` startup path and a live remote traversal, which covers the main CI gap. One possible future improvement is to align it even more closely with the user-facing console flow documented in `README.md`, i.e. `:remote connect tinkerpop.server conf/remote.yaml` followed by `:> ...`, or to reuse `scripts/remote-connect.groovy` from the Docker healthcheck path. The current driver-based script is enough for this PR's smoke coverage, so I do not think this should block merge.

PR #3044, opened 2026-05-31
fix(server): fix foreground mode exit code propagation in startup scripts

### LINKED ISSUE #3043: [Bug] Docker containers never restart when Java process crashes — broken cron monitor + tail -f /dev/null supervision
### Before submit

---

### Environment

- HugeGraph version: 1.7.0+
- Related PR: #3025

---

**Expected:**

**Actual:**

```
# Java crashes at T=0:30
$ docker ps

```

---

### Root Cause Analysis

```
  HugeGraph gets restarted
```

```bash
# ... post-startup checks ...
```

---

### Impact

```
          Users get: Connection refused

  T=0:30  Java crashes
  T=0:31  New container starts
```

---

```
restserver.url=127.0.0.1:8080
```

---

### Related

- Follow-up to #3025

### Related Issues  
  
| Issue | Description |  
|---|---|  

bitflicker64:
### Problem  
  
In foreground mode (`-d false`), `start-hugegraph.sh` had a structural  
bug: all post-branch logic ran **unconditionally after** the daemon/foreground  
`if/else` block.  
  
In foreground mode the script blocks at `hugegraph-server.sh` until Java  
exits. After Java exits:  
- `PID="$!"` captures an empty string (no background job)  
- `bin/pid` is written with an empty string  
- `wait_for_startup` fails immediately (empty PID, Java already dead)  
- `disown` fails (no background jobs)  
- `OPEN_MONITOR` fires unconditionally — even in foreground mode  
- The script exits 0, losing Java's exit code entirely  
  
This means a Java crash in foreground mode is invisible to the process  
supervisor (Docker restart policy, systemd, etc.).  
  
### Fix  
  
- Move all post-branch logic (`PID="$!"`, pid file write, `trap`,  
  `wait_for_startup`, `disown`, `OPEN_MONITOR`) inside the  
  `DAEMON == "true"` branch where it belongs.  
- In the foreground branch: background Java with `&`, capture `$!`,  
  write the pid file, then `wait $PID` (so the script blocks while Java  
  runs) and `exit $?` (so Java's exit code propagates out).  
- Gate `OPEN_MONITOR` inside the daemon branch — monitor setup after a  
  blocking foreground call makes no sense.  
  
### Also included  
  
- Fix `restserver.url` in `rest-server.properties` to include the  
  `http://` scheme prefix (was `127.0.0.1:8080`, now  
  `http://127.0.0.1:8080`). The `check_port` and URL parsing in  
  `util.sh` expect a full URL.  
  
### Tests  
  
New script: `hugegraph-server/hugegraph-dist/src/assembly/travis/test-start-hugegraph.sh`  
  
| Test | Unmodified code | After fix |  
|---|---|---|  
| Test 1 — daemon mode regression | PASS | PASS |  
| Test 2 — foreground mode blocks | PASS | PASS |  
| Test 3 — exit code propagates (137 on SIGKILL) | **FAIL** | PASS |  
| Test 4 — `-m true` registers cron in daemon mode | PASS | PASS |  
  
Wired into `server-ci.yml` after the Compile step, rocksdb backend only  
(foreground mode behavior is backend-independent).  
  
### What this does NOT include  
  
Docker entrypoint changes and runtime restart validation (kill Java  
inside container → container exits → Docker restarts it) are in the  
follow-up PR for chunks 4–8.  
  
Relates to #3043 (chunk 1 of 3 startup script fixes)  
close #3043

### DIFF
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
@@ -103,33 +103,36 @@ if [[ $DAEMON == "true" ]]; then
     echo "Starting HugeGraphServer in daemon mode..."
     "${BIN}"/hugegraph-server.sh "${CONF}/${GREMLIN_SERVER_CONF}" "${CONF}"/rest-server.properties \
     "${OPEN_SECURITY_CHECK}" "${USER_OPTION}" "${GC_OPTION}" "${OPEN_TELEMETRY}" &
+
+    PID="$!"
+    # Write pid to file
+    echo "$PID" > "$PID_FILE"
+
+    trap 'kill $PID; exit' SIGHUP SIGINT SIGQUIT SIGTERM
+
+    wait_for_startup ${PID} 'HugeGraphServer' "$REST_SERVER_URL/graphs" "${SERVER_STARTUP_TIMEOUT_S}" || {
+        if [[ "${STDOUT_MODE:-false}" == "true" ]]; then
+            echo "See 'docker logs' for HugeGraphServer log output." >&2
+        else
+            echo "See $LOGS/hugegraph-server.log for HugeGraphServer log output." >&2
+        fi
+        exit 1
+    }
+    disown
+
+    if [ "$OPEN_MONITOR" == "true" ]; then
+        if ! "$BIN"/start-monitor.sh; then
+            echo "Failed to open monitor, please start it manually"
+        fi
+        echo "An HugeGraphServer monitor task has been append to crontab"
+    fi
 else
     echo "Starting HugeGraphServer in foreground mode..."
     "${BIN}"/hugegraph-server.sh "${CONF}/${GREMLIN_SERVER_CONF}" "${CONF}"/rest-server.properties \
-    "${OPEN_SECURITY_CHECK}" "${USER_OPTION}" "${GC_OPTION}" "${OPEN_TELEMETRY}"
-fi
-
-PID="$!"
-# Write pid to file
-echo "$PID" > "$PID_FILE"
-
-trap 'kill $PID; exit' SIGHUP SIGINT SIGQUIT SIGTERM
-
-wait_for_startup ${PID} 'HugeGraphServer' "$REST_SERVER_URL/graphs" "${SERVER_STARTUP_TIMEOUT_S}" || {
-    if [[ "${STDOUT_MODE:-false}" == "true" ]]; then
-        echo "See 'docker logs' for HugeGraphServer log output." >&2
-    else
-        echo "See $LOGS/hugegraph-server.log for HugeGraphServer log output." >&2
-    fi
-    if [[ $DAEMON == "true" ]]; then
-        exit 1
-    fi
-}
-disown
-
-if [ "$OPEN_MONITOR" == "true" ]; then
-    if ! "$BIN"/start-monitor.sh; then
-        echo "Failed to open monitor, please start it manually"
-    fi
-    echo "An HugeGraphServer monitor task has been append to crontab"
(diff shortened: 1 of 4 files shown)


hugegraph-server/hugegraph-dist/src/assembly/static/bin/start-hugegraph.sh
```diff
@@ -103,33 +103,36 @@ if [[ $DAEMON == "true" ]]; then
...
-    fi
-    echo "An HugeGraphServer monitor task has been append to crontab"
+else
+    echo "Starting HugeGraphServer in foreground mode..."
+    "${BIN}"/hugegraph-server.sh "${CONF}/${GREMLIN_SERVER_CONF}" "${CONF}"/rest-server.properties \
+    "${OPEN_SECURITY_CHECK}" "${USER_OPTION}" "${GC_OPTION}" "${OPEN_TELEMETRY}" &
+    PID="$!"
+    # Write pid to file
+    echo "$PID" > "$PID_FILE"
+    wait $PID
```
imbajin, line 137: ‼️ **Forward termination signals in foreground mode**

Evidence: the daemon branch installs a trap, but the new foreground branch backgrounds Java and only waits on `$PID` here. If Docker or systemd sends TERM to the wrapper, the signal is not forwarded to Java, so shutdown can skip HugeGraph's graceful path and wait for the supervisor's hard kill. Please install the same TERM/INT/QUIT/HUP forwarding before `wait $PID`, and cover that path in `test-start-hugegraph.sh`.

imbajin (commented):
Blocking: yes. Summary: Foreground mode still lacks graceful shutdown signal forwarding. Evidence: static review of start-hugegraph.sh lines 130-137; bash -n passed.
---
🔗 Please also check the latest `codecov/project` failure: https://app.codecov.io/gh/apache/hugegraph/pull/3044

bitflicker64:
@imbajin Added trap 'kill $PID; wait $PID; exit $?' SIGHUP SIGINT SIGQUIT SIGTERM before wait $PID in the foreground branch, and added Test 5 to cover the SIGTERM forwarding path. Also fixed the crontab -r issue in Test 4.

hugegraph-server/hugegraph-dist/src/assembly/travis/test-start-hugegraph.sh
```diff
@@ -0,0 +1,430 @@
...
+# Kill server, clear pid file, kill ports, clear crontab monitor entry
+cleanup() {
+    info "Cleaning up..."
+    "$STOP_SCRIPT" >/dev/null 2>&1 || true
+    if [[ -s "$PID_FILE" ]]; then
+        kill "$(cat "$PID_FILE")" 2>/dev/null || true
+    fi
+    rm -f "$PID_FILE"
+    # kill anything still holding server ports so check_port doesn't fail next test
+    lsof -ti :8080 | xargs kill -9 2>/dev/null || true
```
VGalaxies, line 80: Non-blocking: this CI test now depends on `lsof` and `crontab` being present and usable on the runner. Please add a small preflight/skip or install the tools explicitly; otherwise startup-script regressions and CI environment failures will look the same.

bitflicker64: Added preflight check for all required external tools (lsof, crontab, curl, java) .suite exits with a clear SKIP message if any are missing rather than failing mid test.

bitflicker64:
ill start working on chunk 2 now ?

imbajin (approved):
+1 from my side. The current head addresses the foreground signal-forwarding and exit-code propagation issue, and CI is green.

Non-blocking follow-up: the new startup regression test currently skips with exit 0 when required tools are missing, so future CI hardening could distinguish "skipped due to environment" from "tested and passed". macOS coverage for the startup script can also be considered separately.

PR #3047, opened 2026-06-04
fix(pd,store): fg mode exit code propagation in startup scripts

### LINKED ISSUE #3043: [Bug] Docker containers never restart when Java process crashes — broken cron monitor + tail -f /dev/null supervision

bitflicker64:
Fix foreground mode in start-hugegraph-pd.sh and start-hugegraph-store.sh (chunks 2–3 of #3043)  
  
## Purpose of the PR  
  
- relates to #3043  
- close #3043
  
## Main Changes  
  
**Problem**  
  
Both `start-hugegraph-pd.sh` and `start-hugegraph-store.sh` had no foreground  
branch — the scripts always backgrounded Java with `exec ... &`, wrote `$!` to  
the pid file, and exited 0, losing Java's exit code entirely. Docker/systemd  
supervisors never saw a non-zero exit, so containers were never restarted on  
Java crash.  
  
**Fix**  
  
Add `DAEMON="true"` default and `-d true|false` flag to getopts in both scripts.  
Daemon branch keeps `exec ... &` with `$!` as before. Foreground branch writes  
`$$` to the pid file **before** `exec` (exec replaces the shell with Java, so  
`$$` == Java's PID after exec), then `exec java` without `&` so the process  
blocks and Java's exit code propagates out directly.  
  
No trap needed in the foreground branch — `exec` replaces the shell process with  
Java, so signals from Docker/systemd go directly to Java without a wrapper to  
forward through (unlike chunk 1 where `& + wait` required an explicit trap).  
  
**Tests**  
  
| Script | Test file | Tests | Baseline | After fix |  
|---|---|---|---|---|  
| `start-hugegraph-pd.sh` | `test-start-hugegraph-pd.sh` | 4 (daemon regression, foreground blocks, exit code 137, SIGTERM 143) | 3 passed, 9 failed | 12 passed, 0 failed |  
| `start-hugegraph-store.sh` | `test-start-hugegraph-store.sh` | 4 (same) | 3 passed, 5 failed | 11 passed, 0 failed |  
  
Both test scripts wired into `pd-store-ci.yml` (no backend guard needed — pd/store jobs always run).  
  
Note: Store health check is skipped when PD is not running — handled gracefully with a warning, not a failure.  
  
**What's NOT in this PR**  
  
Docker entrypoints, HEALTHCHECK, and cron monitor removal are in chunks 4–8 (separate PR).  
  
## Verifying these changes  
  
    - `test-start-hugegraph-pd.sh $PD_DIR` — 12 assertions, all pass after fix  
    - `test-start-hugegraph-store.sh $STORE_DIR` — 11 assertions, all pass after fix  
  
- [ ] Dependencies  
- [ ] Modify configurations  
- [ ] The public API  
- [ ] Other affects  
- [x] Nope  
  
## Documentation Status  
  
- [x] `Doc - No Need`

### DIFF
--- a/.github/workflows/pd-store-ci.yml
+++ b/.github/workflows/pd-store-ci.yml
@@ -107,6 +107,12 @@ jobs:
         run: |
           mvn clean package -U -Dmaven.javadoc.skip=true -Dmaven.test.skip=true -ntp --fail-at-end
 
+      - name: Run start-hugegraph-pd.sh foreground mode tests
+        run: |
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          PD_DIR=hugegraph-pd/apache-hugegraph-pd-$VERSION/
+          $TRAVIS_DIR/test-start-hugegraph-pd.sh $PD_DIR
+
       - name: Prepare env and service
         run: |
           $TRAVIS_DIR/start-pd.sh
@@ -162,6 +168,12 @@ jobs:
         # todo remove --fail-at-end after test
         run: |
           mvn clean package -U -Dmaven.javadoc.skip=true -Dmaven.test.skip=true -ntp --fail-at-end
+
+      - name: Run start-hugegraph-store.sh foreground mode tests
+        run: |
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          STORE_DIR=hugegraph-store/apache-hugegraph-store-$VERSION/
+          $TRAVIS_DIR/test-start-hugegraph-store.sh $STORE_DIR
 
       - name: Prepare env and service
         run: |
--- a/hugegraph-pd/hg-pd-dist/src/assembly/static/bin/start-hugegraph-pd.sh
+++ b/hugegraph-pd/hg-pd-dist/src/assembly/static/bin/start-hugegraph-pd.sh
@@ -26,14 +26,18 @@ fi
 if [ -z "$OPEN_TELEMETRY" ];then
   OPEN_TELEMETRY="false"
 fi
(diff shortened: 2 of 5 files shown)


imbajin (approved):
I re-reviewed the current head a6c38e8 and did not find any blocking issue in the PD/Store foreground-mode changes.

The implementation preserves the existing daemon behavior by default, while `-d false` writes the current shell PID before `exec`, so the Java process owns the same PID and exit/signal behavior propagates to the caller. The new CI foreground suites also exercise the important process semantics: blocking behavior, pid-file behavior, SIGKILL exit propagation, and SIGTERM termination for both PD and Store.

A couple of non-blocking follow-ups may still be worth considering:
1. The PD/Store README files do not document the new `-d true|false` option yet.
2. The Store foreground test mainly validates the process model; it intentionally tolerates health-check timeout when PD is not running, while the later workflow steps still cover the normal PD+Store path.

Overall this looks merge-safe to me as the foreground-mode chunk for #3043. The Docker entrypoints still need to be wired to this foreground mode in a follow-up change before the original Docker restart/lifecycle issue is fully closed.

PR #3048, opened 2026-06-04
ci: add codecov.yml with 0.5% threshold to avoid false failures on shell-only PRs

bitflicker64:
## Purpose of the PR

- relates to #3043

Shell scripts and test scripts added in recent PRs (`test-start-hugegraph.sh`, `test-start-hugegraph-pd.sh`, `test-start-hugegraph-store.sh`) have no Java coverage. Codecov reports a large coverage drop and fails the PR check even though actual Java coverage is unchanged.

## Main Changes

- Add `codecov.yml` at repo root:
  - `ignore` block excludes all `*.sh` files and the `travis/` scripts directory from coverage calculation
  - `threshold: 0.5%` for both project and patch so small rounding noise doesn't fail the check

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects
- [x] Nope

## Documentation Status

- [x] `Doc - No Need`

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,12 @@
+ignore:  
+  - "**/*.sh"  
+  - "hugegraph-server/hugegraph-dist/src/assembly/travis/**"  
+  
+coverage:  
+  status:  
+    project:  
+      default:  
+        threshold: 0.5%  
+    patch:  
+      default:  
+        threshold: 0.5%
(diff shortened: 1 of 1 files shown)


bitflicker64:
Hi @imbajin, is this PR relevant/needed for the project?

Recent shell-only PRs have been failing Codecov because non-Java files skew the stats. This adds a codecov.yml to ignore .sh files so they stop blocking builds.

PR #3049, opened 2026-06-04
feat(server) : make buffer max capacity configurable

### LINKED ISSUE #3046: Maybe the max buffer capacity should be customized
### Feature Description (功能描述)

In some extreme big data query tasks, if the amount of data I need to export is very large, MAX_BUFFER_CAPACITY will limit the export. This parameter is located in BytesBuffer under org.apache.hugegraph.backend.serializer in hugegraph-core. It defaults to 128MB and does not support customization. It is desirable to enable customization through configuration files.

LegendPei:
## Purpose of the PR

- close #3046

This PR makes the max capacity of one serializer buffer configurable instead of always using the hardcoded default 128MB limit.

The default behavior remains unchanged, while users who need to serialize larger objects can tune the buffer limit through configuration. The new option is still bounded to avoid unbounded memory allocation.

## Main Changes

    - xxx

- [x]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [x]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
@@ -234,7 +234,7 @@ public class StandardHugeGraph implements HugeGraph {
 
         LockUtil.init(this.spaceGraphName());
 
-        BytesBuffer.setMaxBufferCapacity(
+        BytesBuffer.initMaxBufferCapacity(
                 config.get(CoreOptions.SERIALIZER_BUFFER_MAX_CAPACITY));
         MemoryManager.setMemoryMode(
                 MemoryManager.MemoryMode.fromValue(config.get(CoreOptions.MEMORY_MODE)));
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
@@ -234,8 +234,11 @@ public class StandardHugeGraph implements HugeGraph {
 
         // Init process-wide static configs before lock, so that validation
         // failures won't leave stale lock groups in LockManager.
+        boolean explicitBufferCapacity = config.containsKey(
+                CoreOptions.SERIALIZER_BUFFER_MAX_CAPACITY.name());
         BytesBuffer.initMaxBufferCapacity(
-                config.get(CoreOptions.SERIALIZER_BUFFER_MAX_CAPACITY));
+                config.get(CoreOptions.SERIALIZER_BUFFER_MAX_CAPACITY),
+                explicitBufferCapacity);
         MemoryManager.setMemoryMode(
                 MemoryManager.MemoryMode.fromValue(config.get(CoreOptions.MEMORY_MODE)));
         MemoryManager.setMaxMemoryCapacityInBytes(config.get(CoreOptions.MAX_MEMORY_CAPACITY));
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
@@ -232,8 +232,8 @@ public class StandardHugeGraph implements HugeGraph {
         this.readMode = GraphReadMode.OLTP_ONLY;
         this.schedulerType = config.get(CoreOptions.SCHEDULER_TYPE);
 
-        LockUtil.init(this.spaceGraphName());
-
+        // Init process-wide static configs before lock, so that validation
+        // failures won't leave stale lock groups in LockManager.
         BytesBuffer.initMaxBufferCapacity(
                 config.get(CoreOptions.SERIALIZER_BUFFER_MAX_CAPACITY));
         MemoryManager.setMemoryMode(
@@ -242,6 +242,8 @@ public class StandardHugeGraph implements HugeGraph {
         MemoryManager.setMaxMemoryCapacityForOneQuery(
                 config.get(CoreOptions.ONE_QUERY_MAX_MEMORY_CAPACITY));
         RoundUtil.setAlignment(config.get(CoreOptions.MEMORY_ALIGNMENT));
+
+        LockUtil.init(this.spaceGraphName());
 
         try {
             this.storeProvider = this.loadStoreProvider();
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
@@ -48,6 +48,7 @@ import org.apache.hugegraph.backend.id.IdGenerator;
 import org.apache.hugegraph.backend.id.SnowflakeIdGenerator;
 import org.apache.hugegraph.backend.query.Query;
 import org.apache.hugegraph.backend.serializer.AbstractSerializer;
+import org.apache.hugegraph.backend.serializer.BytesBuffer;
 import org.apache.hugegraph.backend.serializer.SerializerFactory;
 import org.apache.hugegraph.backend.store.BackendFeatures;
 import org.apache.hugegraph.backend.store.BackendProviderFactory;
@@ -233,6 +234,8 @@ public class StandardHugeGraph implements HugeGraph {
 
         LockUtil.init(this.spaceGraphName());
 
+        BytesBuffer.setMaxBufferCapacity(
+                config.get(CoreOptions.SERIALIZER_BUFFER_MAX_CAPACITY));
         MemoryManager.setMemoryMode(
                 MemoryManager.MemoryMode.fromValue(config.get(CoreOptions.MEMORY_MODE)));
         MemoryManager.setMaxMemoryCapacityInBytes(config.get(CoreOptions.MAX_MEMORY_CAPACITY));
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
@@ -145,6 +145,9 @@ public final class BytesBuffer extends OutputStream {
                         capacity);
 
         if (!explicit) {
+            if (maxBufferCapacity == null) {
+                maxBufferCapacity = MAX_BUFFER_CAPACITY;
+            }
             return;
         }
 
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
@@ -338,7 +338,6 @@ public final class BytesBuffer extends OutputStream {
 
     public BytesBuffer writeBigBytes(byte[] bytes) {
         if (bytes.length > BLOB_LEN_MAX) {
-            // TODO: note the max blob size depends on max buffer capacity
             E.checkArgument(false,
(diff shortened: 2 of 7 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
```diff
@@ -233,6 +234,8 @@ public StandardHugeGraph(HugeConfig config) {
 
         LockUtil.init(this.spaceGraphName());
 
+        BytesBuffer.setMaxBufferCapacity(
```
imbajin: ‼️ I think a process-wide buffer limit is the right scope here, because `BytesBuffer` is a static low-level utility and many call sites do not carry graph context. The issue is that this currently behaves like a graph-level config that each `StandardHugeGraph` constructor can overwrite, so the last opened graph silently changes the limit for all graphs in the same JVM. Please make the process-wide semantic explicit: initialize the value once, document `serializer.buffer_max_capacity` as JVM/process-wide, and reject conflicting values when another graph is opened with a different setting. A focused multi-graph test for this conflict/last-opened behavior would be enough; no need to refactor the whole serializer path into per-graph state for this PR.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
```diff
@@ -174,13 +192,15 @@ private void require(int size) {
         }
 
         // Extra capacity as buffer
-        int newCapacity = size + this.buffer.limit() + DEFAULT_CAPACITY;
-        if (newCapacity > MAX_BUFFER_CAPACITY) {
+        long newCapacity = (long) size + this.buffer.limit() +
```
imbajin: ⚠️ Non-blocking, but this checks the padded growth target against the configured max, not the actual required size. For example, with `max=128`, a full 64-byte buffer rejects a 1-byte write because `64 + 1 + DEFAULT_CAPACITY == 129`, even though the final content size would only be 65. Consider checking `position + size > maxCapacity` for rejection, then allocating `min(position + size + DEFAULT_CAPACITY, maxCapacity)` when the write itself still fits.

imbajin (commented):
The implementation is close and the process-wide scope is the right fit for BytesBuffer. Please avoid last-opened-graph-wins behavior by making that process-wide semantic explicit and rejecting conflicting graph configs.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
```diff
@@ -233,6 +234,8 @@ public StandardHugeGraph(HugeConfig config) {
 
         LockUtil.init(this.spaceGraphName());
 
+        BytesBuffer.initMaxBufferCapacity(
```
VGalaxies, line 239: **Medium: Clean up lock groups when buffer capacity init fails**

`hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java:237`

**Evidence**
- `LockUtil.init(this.spaceGraphName())` runs at line 235, then the new `BytesBuffer.initMaxBufferCapacity(...)` can throw at line 237 on invalid or conflicting `serializer.buffer_max_capacity`. The constructor cleanup begins only inside later `try` blocks, so this exception bypasses `LockUtil.destroy(...)`. `LockManager.create()` throws when a lock group already exists, so a later retry for the same graph can fail before reaching the corrected config.

**Impact**
- A single failed graph open caused by this new option can leave process-wide lock state behind and prevent loading that graph again without restarting the process.

**Requested fix**
- Move the buffer-capacity initialization before `LockUtil.init(...)`, or wrap this initialization path so `LockUtil.destroy(this.spaceGraphName())` runs for any exception after lock initialization.

VGalaxies (commented):
**Review summary**

- **Blocking:** yes
- **Summary:** The PR leaves graph lock groups registered when the new serializer buffer capacity initialization fails, which can block a corrected retry in the same process.
- **Evidence:**
  - Static review
  - `git diff --check origin/master...HEAD` passed. `mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test -Dtest=BytesBufferTest` failed before reaching the target tests because `hugegraph-pd` ran surefire with no tests

imbajin (commented):
Blocking: yes. Summary: The process-wide serializer buffer cap is still sourced from each graph config, so mixed custom/default graph configs can fail multi-graph startup. Evidence: static review of GraphManager local graph loading plus BytesBufferTest passed locally.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
```diff
@@ -231,15 +232,19 @@ public StandardHugeGraph(HugeConfig config) {
         this.readMode = GraphReadMode.OLTP_ONLY;
         this.schedulerType = config.get(CoreOptions.SCHEDULER_TYPE);
 
-        LockUtil.init(this.spaceGraphName());
-
+        // Init process-wide static configs before lock, so that validation
+        // failures won't leave stale lock groups in LockManager.
+        BytesBuffer.initMaxBufferCapacity(
```
imbajin, line 239: ‼️ **Keep the process-wide cap from conflicting with default graph configs**

Evidence: `GraphManager` loads every file under `conf/graphs` in one JVM, and each `StandardHugeGraph` calls `BytesBuffer.initMaxBufferCapacity(config.get(...))`. With this option stored as a graph option, the first graph that sets `serializer.buffer_max_capacity=268435456` freezes the static cap, then any later graph that omits the option still passes the default `134217728` and fails with the new conflicting-value check.

Impact: enabling this option for one graph can prevent a normal multi-graph server from starting unrelated graphs, even though a process-wide setting should be shared by the process.

Requested fix: move the source to server-wide config, or treat an unset/default graph value as inheriting the already-initialized process value, and add a `GraphManager`/`MultiGraphsTest` case covering one custom graph plus one default graph.

VGalaxies (commented):
**Review summary**

- **Blocking:** yes
- **Summary:** The process-wide serializer buffer cap can be frozen by a graph that only uses the default, making later explicit graph configs fail depending on load order.
- **Evidence:**
  - Static review of `git diff origin/master...HEAD`
  - verified changed-side line numbers with `nl -ba`

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
```diff
@@ -122,6 +126,40 @@ public static BytesBuffer wrap(byte[] array, int offset, int length) {
...
+
+    public static synchronized void initMaxBufferCapacity(int capacity) {
+        E.checkArgument(capacity >= DEFAULT_CAPACITY &&
+                        capacity <= MAX_BUFFER_CAPACITY_UPPER_BOUND,
+                        "Max buffer capacity must be in range [%s, %s], " +
+                        "but got %s",
+                        DEFAULT_CAPACITY, MAX_BUFFER_CAPACITY_UPPER_BOUND,
+                        capacity);
+
+        if (maxBufferCapacity == null) {
```
VGalaxies, line 151: **High: Default config can block later explicit buffer cap**

`hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java:142`

**Evidence**
- `StandardHugeGraph` initializes `BytesBuffer` for every graph using `config.get(CoreOptions.SERIALIZER_BUFFER_MAX_CAPACITY)`, whose default is `BytesBuffer.MAX_BUFFER_CAPACITY`. `initMaxBufferCapacity()` stores the first value unconditionally when `maxBufferCapacity == null`, and only treats `MAX_BUFFER_CAPACITY` as inherit on later calls. Therefore a default graph opened first fixes the process-wide value at 128 MiB, and a later graph explicitly setting 256 MiB hits the conflict path.

**Impact**
- Multi-graph deployments can succeed or fail based on graph load order; a graph with a valid explicit `serializer.buffer_max_capacity` may fail to load after any graph that omitted the option.

**Requested fix**
- Distinguish omitted/default values from explicit configuration before initializing the process-wide cap, or validate/select the single process-wide value before opening graphs. Add coverage for default-first then custom-second graph initialization.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
```diff
@@ -122,6 +126,43 @@ public static BytesBuffer wrap(byte[] array, int offset, int length) {
...
+    public static synchronized void initMaxBufferCapacity(int capacity,
+                                                          boolean explicit) {
+        E.checkArgument(capacity >= DEFAULT_CAPACITY &&
+                        capacity <= MAX_BUFFER_CAPACITY_UPPER_BOUND,
+                        "Max buffer capacity must be in range [%s, %s], " +
+                        "but got %s",
+                        DEFAULT_CAPACITY, MAX_BUFFER_CAPACITY_UPPER_BOUND,
+                        capacity);
+
+        if (!explicit) {
```
imbajin, line 147: ‼️ **Freeze the process-wide cap before any graph can use it**

Evidence: `StandardHugeGraph` now passes `explicit=false` when a graph omits `serializer.buffer_max_capacity`, and `BytesBuffer.initMaxBufferCapacity()` returns without setting `maxBufferCapacity`. If that default-config graph is already open, a later graph with an explicit value can still set the static JVM-wide cap, so buffers created after that point for the first graph use the later graph's limit instead of the default it started with.

Impact: the serializer limit for a running graph can change solely because another graph is created later, leaving multi-graph behavior dependent on load order.

Requested fix: select or freeze the single process-wide value before any graph is opened, or reject late explicit initialization once any graph has started using the default process value. Please add a regression test covering a default graph that remains open while a later explicit graph is opened.

imbajin (commented):
Blocking: yes. Summary: A default-config graph can already be running when a later explicit graph mutates the JVM-wide serializer buffer cap. Evidence: static review of `StandardHugeGraph`/`BytesBuffer`; `mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test -Dtest=BytesBufferTest -DfailIfNoTests=false` passed.
---
🔗 Please also check the failing codecov/project status: https://app.codecov.io/gh/apache/hugegraph/pull/3049

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/serializer/BytesBuffer.java
```diff
@@ -122,6 +126,46 @@ public static BytesBuffer wrap(byte[] array, int offset, int length) {
...
+            if (maxBufferCapacity == null) {
+                maxBufferCapacity = MAX_BUFFER_CAPACITY;
```
imbajin: ‼️ **Make the process-wide cap independent of graph load order**

Evidence: the default-path branch now stores `MAX_BUFFER_CAPACITY` when the first graph omits `serializer.buffer_max_capacity`, while an explicit graph opened first still lets later default graphs inherit the custom cap. That means the same set of graph configs can either start or fail depending only on which graph file is opened first.

Impact: a valid explicit process-wide serializer cap can be rejected in multi-graph deployments after any default-config graph initializes first.

Requested fix: select the single process-wide value before opening individual graphs, or let omitted/default graph configs defer initialization until an explicit value is seen; add coverage for both explicit-first/default-second and default-first/explicit-second ordering.

imbajin (commented):
Blocking: yes. Summary: Found one load-order regression in the current head. Evidence: static review of BytesBuffer init semantics plus git diff --check origin/master...HEAD.

imbajin (approved):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test -Dtest=BytesBufferTest -DfailIfNoTests=false; git diff --check origin/master...HEAD; latest-head checks successful.

PR #3050, opened 2026-06-04
fix: gremlin-console.sh fails on Mac M/ARM CPU #3050

vaijosh:
### What I changed
- Added runtime dependency `org.fusesource.jansi:jansi:2.4.0` in `hugegraph-server/hugegraph-dist/pom.xml`.
- Updated release docs:
  - `install-dist/release-docs/LICENSE`
  - `install-dist/release-docs/NOTICE`
  - `install-dist/release-docs/licenses/LICENSE-jansi-2.4.0.txt`

### Validation performed
```bash
 ✘ vjoshi@DNWW7WJ4LJ  ~/SourceCode/HUGEGRAPH/hugegraph   3031-AppleSiliconFix  export RUNNER_TEMP="$(mktemp -d)"     
SERVER_DIR="hugegraph-server/apache-hugegraph-server-${VERSION}"
printf "println \"gremlin-console-smoke-ok\"\n" > "$RUNNER_TEMP/gremlin-console-smoke.groovy"
"${SERVER_DIR}/bin/gremlin-console.sh" -- -e "$RUNNER_TEMP/gremlin-console-smoke.groovy"

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/Users/vjoshi/SourceCode/HUGEGRAPH/hugegraph/hugegraph-server/apache-hugegraph-server-1.7.0/lib/groovy-2.5.14-indy.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
gremlin-console-smoke-ok

Manually checked if we can launch gremlin console.

### DIFF
--- a/hugegraph-server/hugegraph-dist/pom.xml
+++ b/hugegraph-server/hugegraph-dist/pom.xml
@@ -33,6 +33,7 @@
         <assembly.dir>${project.basedir}/src/assembly</assembly.dir>
         <assembly.descriptor.dir>${assembly.dir}/descriptor</assembly.descriptor.dir>
         <assembly.static.dir>${assembly.dir}/static</assembly.static.dir>
+        <jansi.version>2.4.0</jansi.version>
     </properties>
 
     <dependencies>
@@ -122,6 +123,15 @@
             <groupId>io.grpc</groupId>
             <artifactId>grpc-stub</artifactId>
             <version>${grpc.version}</version>
+        </dependency>
+
+        <!-- Added updated jar to avoid java.lang.NoClassDefFoundError: org/fusesource/jansi/AnsiConsole when launching
+             gremlin console on arm64 -->
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>${jansi.version}</version>
+            <scope>runtime</scope>
         </dependency>
     </dependencies>
 
--- a/install-dist/release-docs/licenses/LICENSE-jansi-2.4.0.txt
+++ b/install-dist/release-docs/licenses/LICENSE-jansi-2.4.0.txt
@@ -1,0 +1,16 @@
+Copyright (c) 2007-2021, the original author(s)
+All rights reserved.
+
+This software is dual-licensed under the Apache License, Version 2.0 and the GNU Lesser General Public License, version 3 or later (LGPL-3.0-or-later).
+
+You may obtain a copy of the Apache License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+You may obtain a copy of the GNU Lesser General Public License at
+
+    http://www.gnu.org/licenses/lgpl-3.0.html
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+
+This product includes software developed by the Jansi project (http://fusesource.github.io/jansi/).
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: hugegraph-server/hugegraph-dist/pom.xml [hunk 1]
```diff
@@ -123,6 +124,15 @@
             <artifactId>grpc-stub</artifactId>
             <version>${grpc.version}</version>
         </dependency>
+
+        <!-- Added updated jar to avoid java.lang.NoClassDefFoundError: org/fusesource/jansi/AnsiConsole when launching
+             gremlin console on arm64 -->
+        <dependency>
```
imbajin, line 130: ‼️ **Update the known dependency inventory**

Evidence: this PR adds `org.fusesource.jansi:jansi:2.4.0`, and the dependency-check job diffs the generated inventory with `install-dist/scripts/dependency/known-dependencies.txt`; the failed job shows `+jansi-2.4.0.jar`, while the tracked file has no such entry.

Impact: the release dependency gate will keep failing for this PR, and the newly bundled jar is not recorded in the deterministic dependency inventory.

Requested fix: regenerate and commit `install-dist/scripts/dependency/known-dependencies.txt`, or add the missing `jansi-2.4.0.jar` entry in the correct sorted position.

install-dist/release-docs/licenses/LICENSE-jansi-2.4.0.txt
```diff
@@ -0,0 +1,16 @@
+Copyright (c) 2007-2021, the original author(s)
```
imbajin, line 1: ‼️ **Vendor the full license text**

Evidence: this new file is a 16-line summary with URLs, but the surrounding `install-dist/release-docs/licenses/LICENSE-*.txt` files vendor complete upstream license texts. The new `LICENSE` entry also points readers here for license details, which this summary does not provide.

Impact: the release docs for the newly bundled jar are incomplete, which is an ASF release/legal risk even if the runtime dependency is correct.

Requested fix: replace this file with the full upstream license text for the license option HugeGraph is relying on for distribution, and align the `LICENSE`/`NOTICE` wording with that choice.

hugegraph-server/hugegraph-dist/pom.xml
imbajin, line 130: ⚠️ **Add a distro regression check**

Evidence: the fix is purely in distro packaging, so the normal unit/core/api suites do not exercise the `gremlin-console.sh` classpath. I could not find an automated check that launches the packaged console or at least asserts the tarball contains `lib/jansi-*.jar`.

Impact: this exact classpath regression can silently return after a dependency or assembly change and only be caught by users starting the packaged console.

Requested fix: add a distro-level smoke check, for example launching `bin/gremlin-console.sh -- -e <script>` from the assembled server, or at minimum asserting the packaged archive contains the Jansi jar.

imbajin (commented):
Blocking: yes. Summary: the new packaged dependency is missing required release/dependency inventory updates and a distro regression guard. Evidence: dependency-check reports `+jansi-2.4.0.jar`, dependency:tree shows the new runtime jar, and the added license file is only a short summary.

---
🔗 Please also check the latest failed dependency-check job: https://github.com/apache/hugegraph/actions/runs/26973199772/job/79593480638

hugegraph-server/hugegraph-dist/pom.xml
vaijosh: Done. Thanks @imbajin.

imbajin (approved):
LGTM. The explicit Jansi runtime dependency is justified here because the bundled Gremlin Console path still relies on an older JLine/Jansi stack that lacks macOS arm64 native support. The PR keeps the fix scoped to the distribution package, updates the required release/dependency metadata, and the relevant CI checks are green. No separate user-facing documentation is needed since the behavior remains transparent to users: bin/gremlin-console.sh should simply work on Apple Silicon.

vaijosh:
Many thanks @imbajin for reviewing the PR.

PR #3051, opened 2026-06-05
fix(docker): supervise Java process in entrypoints instead of tail -f /dev/null

### LINKED ISSUE #3043: [Bug] Docker containers never restart when Java process crashes — broken cron monitor + tail -f /dev/null supervision

bitflicker64:
## Purpose of the PR

- relates to #3043
- close #3043

All three `docker-entrypoint.sh` files used `tail -f /dev/null` to keep the container alive. When Java crashed, `tail` kept running and the container stayed up with no Java inside — Docker restart policy never fired.

## Main Changes

**PD and Store:** replace daemon start + `tail -f /dev/null` with `-d false`, which triggers the foreground branch added in #3047. `exec java` replaces the shell with Java directly, so the entrypoint blocks and exits with Java's real exit code when Java exits.

**Server:** keep daemon start (`wait-partition.sh` must run after Java is already up in the background). Replace `tail -f /dev/null` with `tail --pid=$(cat ./bin/pid) -f /dev/null`. When Java exits, `tail --pid` returns, entrypoint exits 1, restart policy fires.

Note: server entrypoint always exits 1 — `start-hugegraph.sh` calls `disown` before returning so Java is not a child of the entrypoint shell and its real exit code is not available. This is sufficient for `--restart=on-failure`.

| Component | Before | After |
|---|---|---|
| PD | daemon start + `tail -f /dev/null` | `-d false` → `exec java` blocks |
| Store | daemon start + `tail -f /dev/null` | `-d false` → `exec java` blocks |
| Server | daemon start + `tail -f /dev/null` | daemon start + `tail --pid` supervision |

## Verifying these changes

  - `docker run --restart=on-failure` → `kill -9` Java inside container → container exits and Docker restarts it automatically
  - Verified locally: 3 kill-9 tests on server container, PD restart loop confirmed

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects
- [x] Nope

## Documentation Status

- [x] `Doc - TODO` — README/docs update for Docker deployment lifecycle will follow in a separate PR

### DIFF
--- a/hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
+++ b/hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
@@ -91,4 +91,8 @@ if [[ "${ACTUAL_BACKEND}" == "hstore" ]]; then
     ./bin/wait-partition.sh || log "WARN: partitions not assigned yet"
 fi
 
-tail -f /dev/null
+PID=$(cat ./bin/pid 2>/dev/null || true)
+if [[ -n "$PID" ]]; then
+    tail --pid="$PID" -f /dev/null
+    exit 1
+fi
--- a/hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
+++ b/hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
@@ -93,6 +93,7 @@ fi
 
 PID=$(cat ./bin/pid 2>/dev/null || true)
 if [[ -n "$PID" ]]; then
+    trap 'kill -TERM "$PID" 2>/dev/null; while kill -0 "$PID" 2>/dev/null; do sleep 1; done; exit 0' TERM INT
     tail --pid="$PID" -f /dev/null
     exit 1
 fi
--- a/hugegraph-pd/hg-pd-dist/docker/docker-entrypoint.sh
+++ b/hugegraph-pd/hg-pd-dist/docker/docker-entrypoint.sh
@@ -82,5 +82,4 @@ log "  pd.initial-store-list=${HG_PD_INITIAL_STORE_LIST}"
 log "  pd.initial-store-count=${HG_PD_INITIAL_STORE_COUNT}"
 log "  pd.data-path=${HG_PD_DATA_PATH}"
 
-./bin/start-hugegraph-pd.sh -j "${JAVA_OPTS:-}"
-tail -f /dev/null
+./bin/start-hugegraph-pd.sh -d false -j "${JAVA_OPTS:-}"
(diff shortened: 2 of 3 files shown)


hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
```diff
@@ -91,4 +91,8 @@ if [[ "${ACTUAL_BACKEND}" == "hstore" ]]; then
     ./bin/wait-partition.sh || log "WARN: partitions not assigned yet"
 fi
 
-tail -f /dev/null
+PID=$(cat ./bin/pid 2>/dev/null || true)
+if [[ -n "$PID" ]]; then
+    tail --pid="$PID" -f /dev/null
```
imbajin, line 97: ⚠️ **Forward shutdown to HugeGraphServer**

Evidence: PD and Store now execute their foreground scripts with `-d false`, but the server entrypoint still starts `start-hugegraph.sh` in daemon mode and then blocks on `tail --pid="$PID" -f /dev/null` at this line. Docker sends SIGTERM to PID 1, so this shell/tail path can exit without signaling the Java PID; the server then waits for Docker's forced kill instead of a graceful shutdown.

Please add a TERM/INT trap that kills and waits for `$PID`, or otherwise run the server foreground path after the post-start checks, so `docker stop` shuts HugeGraphServer down cleanly.

imbajin (commented):
Blocking: no. Summary: server Docker entrypoint still needs graceful shutdown forwarding. Evidence: static review of hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh; bash -n passed for all three changed entrypoints.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
bitflicker64: Good catch added a `TERM/INT` trap before the `tail --pid` block in f20b8a4:

```bash
PID=$(cat ./bin/pid 2>/dev/null || true)
if [[ -n "$PID" ]]; then
    trap 'kill -TERM "$PID" 2>/dev/null; while kill -0 "$PID" 2>/dev/null; do sleep 1; done; exit 0' TERM INT
    tail --pid="$PID" -f /dev/null
    exit 1
fi
 ```
When `dumb-init` forwards SIGTERM from `docker stop`, the trap fires: sends SIGTERM to Java, polls with `kill -0` until Java exits cleanly, then exits 0. The `tail --pid` crash-detection path still exits 1 so restart policy fires on a crash.

imbajin (approved):
+1 from my side. The previous server SIGTERM concern is addressed at the current head: PD/Store now use the foreground exec path, and Server keeps the hstore post-start check while supervising the Java PID with crash detection plus TERM/INT forwarding. CI is green and I don’t see remaining blockers.

PR #3052, opened 2026-06-05
chore(docker): add HEALTHCHECK & clean Dockerfiles

### LINKED ISSUE #3043: [Bug] Docker containers never restart when Java process crashes — broken cron monitor + tail -f /dev/null supervision

bitflicker64:
## Purpose of the PR

- relates to #3043
- close #3043

`docker ps` always shows `Up` even when Java has crashed inside the container because all three Dockerfiles had no `HEALTHCHECK`. Operators cannot distinguish a healthy container from a zombie. Also, `cron` was installed in all three images but is never used in Docker deployments — the cron-based monitor (`-m true`) is for VM/bare-metal only.

## Main Changes

- Add `HEALTHCHECK` to `hugegraph-server/Dockerfile`, `hugegraph-pd/Dockerfile`, `hugegraph-store/Dockerfile`:
  - Server: `curl http://localhost:8080/versions`
  - PD: `curl http://localhost:8620/v1/health`
  - Store: `curl http://localhost:8520/v1/health`
  - Fallback: if HTTP is not yet up but Java is alive (`kill -0` on pid file), report healthy — avoids false unhealthy during startup
  - Endpoints match what is already used in `docker/docker-compose.yml`
- Remove `cron` from `apt-get install` in all three Dockerfiles — shrinks image size and reduces attack surface

## Verifying these changes

  - `docker run` container → `docker inspect --format='{{.State.Health.Status}}'` → reports `healthy` once Java is up
  - `kill -9` Java inside container → status transitions to `unhealthy` within `--interval * --retries` seconds

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects
- [x] Nope

## Documentation Status

- [x] `Doc - No Need`

### DIFF
--- a/hugegraph-server/Dockerfile
+++ b/hugegraph-server/Dockerfile
@@ -67,8 +67,7 @@ EXPOSE 8080
 VOLUME /hugegraph-server
 
 HEALTHCHECK --interval=15s --timeout=10s --start-period=90s --retries=3 \
-    CMD curl -fsS http://localhost:8080/versions >/dev/null \
-        || kill -0 "$(cat ./bin/pid 2>/dev/null)" 2>/dev/null
+    CMD curl -fsS http://localhost:8080/versions >/dev/null
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["./docker-entrypoint.sh"]
--- a/hugegraph-server/Dockerfile
+++ b/hugegraph-server/Dockerfile
@@ -53,7 +53,6 @@ RUN apt-get -q update \
        curl \
        lsof \
        vim \
-       cron \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* \
     && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
@@ -66,6 +65,10 @@ RUN chmod 755 ./docker-entrypoint.sh
 
 EXPOSE 8080
 VOLUME /hugegraph-server
+
+HEALTHCHECK --interval=15s --timeout=10s --start-period=90s --retries=3 \
+    CMD curl -fsS http://localhost:8080/versions >/dev/null \
+        || kill -0 "$(cat ./bin/pid 2>/dev/null)" 2>/dev/null
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["./docker-entrypoint.sh"]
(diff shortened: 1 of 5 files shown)


hugegraph-server/Dockerfile
```diff
@@ -67,5 +66,9 @@ RUN chmod 755 ./docker-entrypoint.sh
 EXPOSE 8080
 VOLUME /hugegraph-server
 
+HEALTHCHECK --interval=15s --timeout=10s --start-period=90s --retries=3 \
+    CMD curl -fsS http://localhost:8080/versions >/dev/null \
+        || kill -0 "$(cat ./bin/pid 2>/dev/null)" 2>/dev/null
```
imbajin: ⚠️ **Keep the healthcheck tied to the HTTP endpoint**

Evidence: the existing compose healthchecks use `curl ... || exit 1` for `/versions` and `/v1/health`, but the new Dockerfile healthchecks return success whenever `./bin/pid` still points to a live JVM. This same fallback is added to the server, hstore, PD, and store images.

Impact: after the start period, a container can remain `healthy` even when the REST endpoint is down or wedged, which weakens the health signal this PR is adding. Please make the steady-state Dockerfile healthcheck fail on HTTP failure, or keep any process-only fallback outside the steady-state HEALTHCHECK path.

imbajin (commented):
Blocking: yes. Summary: The image healthchecks currently mark a live JVM as healthy even when the configured HTTP health endpoint is failing. Evidence: static review of hugegraph-server/Dockerfile:69-71 and existing compose healthchecks in docker/docker-compose.yml:49/77/99.

---
CI/status: codecov/project is failing on the latest head; please check 🔗 https://github.com/apache/hugegraph/runs/79772635318.

hugegraph-server/Dockerfile
bitflicker64: Good point --start-period=90s already covers the startup window, so the || kill -0 fallback is redundant and weakens the signal. Removing it from all four Dockerfiles so the HEALTHCHECK is purely HTTP-based. Pushing now.

imbajin (approved):
+1 from my side. The previous HTTP healthcheck concern is addressed at the current head: all four Dockerfiles now use HTTP-only HEALTHCHECK commands, and docker-build CI verifies that built images keep .Config.Healthcheck configured. CI is green and I don’t see remaining blockers.

PR #3055, opened 2026-06-08
chore(ci): exit 77 when tools missing to distinguish skip from pass

### LINKED ISSUE #3043: [Bug] Docker containers never restart when Java process crashes — broken cron monitor + tail -f /dev/null supervision

### LINKED ISSUE #3054: ci(test): distinguish "skipped due to missing tools" from "tested and passed" in startup script tests
## Problem

The three startup regression test scripts exit `0` when required tools are missing from the environment:

```bash
# test-start-hugegraph.sh, test-start-hugegraph-pd.sh, test-start-hugegraph-store.sh
if ! command -v "$tool" >/dev/null 2>&1; then
    echo "SKIP: required tool '$tool' not found — skipping test suite"
    exit 0
fi
```

Exit code `0` is indistinguishable from "all tests ran and passed". If a CI runner is missing `lsof`, `curl`, `java`, or `crontab`, the step silently reports success and no one notices the tests never ran.

This was flagged as a non-blocking follow-up in the review of #3044 and #3047.

## Expected behavior

When required tools are missing, the test scripts should exit with a distinct code (e.g. `77`, the conventional POSIX skip code used by Automake and TAP) so CI can tell the difference between:

- `exit 0` — tested and passed
- `exit 77` — skipped due to missing environment
- `exit 1` — tested and failed

The CI workflow steps should handle `77` explicitly and emit a `::notice::` annotation instead of failing.

## Files to change

**Test scripts** — change `exit 0` to `exit 77` in the tool-check block:

- `hugegraph-server/hugegraph-dist/src/assembly/travis/test-start-hugegraph.sh` (lines 148–153)
- `hugegraph-server/hugegraph-dist/src/assembly/travis/test-start-hugegraph-pd.sh` (lines 132–137)
- `hugegraph-server/hugegraph-dist/src/assembly/travis/test-start-hugegraph-store.sh` (lines 113–127)

**CI workflow steps** — wrap the test call to handle exit 77:

- `.github/workflows/server-ci.yml` — "Run start-hugegraph.sh foreground mode tests" step (lines 73–79)
- `.github/workflows/pd-store-ci.yml` — "Run start-hugegraph-pd.sh foreground mode tests" step (lines 110–114) and the equivalent Store step

**CI wrapper pattern:**
```yaml
run: |
  set +e
  $TRAVIS_DIR/test-start-hugegraph.sh $SERVER_DIR
  EXIT=$?
  set -e
  if [ $EXIT -eq 77 ]; then
    echo "::notice::Startup tests skipped — required tools not available in this environment"
    exit 0
  fi
  exit $EXIT
```

## Related

- Parent issue: #3043 
- Flagged in review of #3044 (imbajin comment: "the new startup regression test currently skips with exit 0 when required tools are missing")
- Flagged in review of #3047 (same pattern in PD/Store test scripts)

bitflicker64:
## Purpose of the PR

- relates to #3043
- follow-up to #3044, #3047
- close #3054 
- close #3043
The three startup test scripts (`test-start-hugegraph.sh`, `test-start-hugegraph-pd.sh`, `test-start-hugegraph-store.sh`) exit `0` when required tools (`lsof`, `curl`, `java`) are not found. Exit code `0` is indistinguishable from a passing test run — CI shows green even though no tests ran.

## Main Changes

**Test scripts:**
- Change `exit 0` → `exit 77` in all tool-missing skip paths
- Change `exit 0` → `exit 77` in the Store `ulimit -n` skip path

**CI workflows (`server-ci.yml`, `pd-store-ci.yml`):**
- Wrap each test step with `set +e` / exit code capture
- Treat exit `77` as a visible `::notice::` annotation (not a failure)
- All other non-zero exits still fail the step

## Verifying these changes

  - Remove `lsof` from the runner path → CI step shows `::notice::... skipped` instead of silently passing
  - Normal run (all tools present) → exit `0`, step passes as before
  - Test failure → exit `1`, step fails as before

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects
- [x] Nope

## Documentation Status

- [x] `Doc - No Need`

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -70,21 +70,31 @@ jobs:
         run: |
           mvn clean compile -U -Dmaven.javadoc.skip=true -ntp
 
-      - name: Run start-hugegraph.sh foreground mode tests
+      - name: Check startup test prerequisites
+        id: server-preflight
         if: ${{ env.BACKEND == 'rocksdb' }}
+        run: |
+          for tool in lsof crontab curl java; do
+            if ! command -v "$tool" >/dev/null 2>&1; then
+              echo "can_run=false" >> "$GITHUB_OUTPUT"
+              echo "skip_reason=missing tool: $tool" >> "$GITHUB_OUTPUT"
+              exit 0
+            fi
+          done
+          echo "can_run=true" >> "$GITHUB_OUTPUT"
+
+      - name: Run start-hugegraph.sh foreground mode tests
+        if: ${{ env.BACKEND == 'rocksdb' && steps.server-preflight.outputs.can_run == 'true' }}
         run: |
           mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
           SERVER_DIR=hugegraph-server/apache-hugegraph-server-$VERSION/
-          set +e
           $TRAVIS_DIR/test-start-hugegraph.sh $SERVER_DIR
-          EXIT=$?
-          set -e
-          if [ $EXIT -eq 77 ]; then
-            echo "::notice::Startup tests skipped — required tools not available"
-            exit 0
-          fi
-          exit $EXIT
+
+      - name: Startup tests skipped (missing prerequisites)
+        if: ${{ env.BACKEND == 'rocksdb' && steps.server-preflight.outputs.can_run == 'false' }}
+        run: |
+          echo "::notice::Server startup tests skipped — ${{ steps.server-preflight.outputs.skip_reason }}"
 
       - name: Run unit test
         run: |
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -76,7 +76,15 @@ jobs:
           mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
           SERVER_DIR=hugegraph-server/apache-hugegraph-server-$VERSION/
+          set +e
           $TRAVIS_DIR/test-start-hugegraph.sh $SERVER_DIR
+          EXIT=$?
+          set -e
+          if [ $EXIT -eq 77 ]; then
+            echo "::notice::Startup tests skipped — required tools not available"
+            exit 0
+          fi
+          exit $EXIT
 
       - name: Run unit test
         run: |
--- a/.github/workflows/pd-store-ci.yml
+++ b/.github/workflows/pd-store-ci.yml
@@ -111,7 +111,15 @@ jobs:
         run: |
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
           PD_DIR=hugegraph-pd/apache-hugegraph-pd-$VERSION/
+          set +e
           $TRAVIS_DIR/test-start-hugegraph-pd.sh $PD_DIR
+          EXIT=$?
+          set -e
+          if [ $EXIT -eq 77 ]; then
+            echo "::notice::PD startup tests skipped — required tools not available"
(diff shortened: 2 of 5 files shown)


bitflicker64:
@dosu review?

imbajin (commented):
Blocking: yes. Summary: The startup test wrapper still reports skipped suites as successful GitHub Actions steps. Evidence: static review of the workflow wrapper and local dummy-dist runs where the scripts returned 77.
---
Latest-head CI has visible failures; please also check dependency-check / build-commons (11): 🔗 https://github.com/apache/hugegraph/actions/runs/27120670959/job/80036825383

.github/workflows/server-ci.yml
```diff
@@ -76,7 +76,15 @@ jobs:
           mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
           SERVER_DIR=hugegraph-server/apache-hugegraph-server-$VERSION/
+          set +e
           $TRAVIS_DIR/test-start-hugegraph.sh $SERVER_DIR
+          EXIT=$?
+          set -e
+          if [ $EXIT -eq 77 ]; then
+            echo "::notice::Startup tests skipped — required tools not available"
+            exit 0
```
imbajin: ‼️ **Report missing-prerequisite startup tests as skipped, not successful**

Evidence: this wrapper turns the test script `77` into `exit 0` after emitting a notice. I reproduced the script side with a dummy dist and restricted `PATH`: all three startup scripts return `77` when `lsof` is unavailable, but the workflow wrapper converts that case to a successful step. In GitHub Actions, `::notice::` only creates an annotation; the final `exit 0` still marks the startup-test step as success.

Impact: the PR makes the logs clearer, but the check remains green exactly like a real passing startup suite. That keeps the original ambiguity for maintainers who inspect only the check result.

Requested fix: split this into a preflight step that writes `can_run=false` and a specific `skip_reason` to `$GITHUB_OUTPUT`, then guard the real startup-test step with `if: steps.<preflight>.outputs.can_run == 'true'`. Add a separate notice/summary step for the skip path. While doing that, keep the reason specific so the store `ulimit -n < 1024` case is not reported as missing tools.

bitflicker64: Switched to the preflight/guard pattern as requested. The preflight step writes can_run and skip_reason to $GITHUB_OUTPUT; the test step is guarded by if: steps.<id>.outputs.can_run == 'true' so it shows as grey Skipped (not green Success) when prerequisites are missing. The Store preflight distinguishes missing tool: <name> from ulimit -n is <N> (store requires >= 1024) with separate skip_reason values. Pushing now.

imbajin (approved):
+1 from my side. The previous concern has been addressed.

One non-blocking improvement for a future follow-up: the preflight checks are now duplicated between the workflow YAML and the startup scripts, so it may be worth consolidating them into a shared helper later to avoid drift when prerequisites change.

bitflicker64:
> +1 from my side. The previous concern has been addressed.
> 
> One non-blocking improvement for a future follow-up: the preflight checks are now duplicated between the workflow YAML and the startup scripts, so it may be worth consolidating them into a shared helper later to avoid drift when prerequisites change.

suree that makes sense should i add a TODO ?

imbajin:
> suree that makes sense should i add a TODO ?

I’m fine either way~

PR #3056, opened 2026-06-09
docs: document -d flag and Docker process supervision model 

### LINKED ISSUE #3043: [Bug] Docker containers never restart when Java process crashes — broken cron monitor + tail -f /dev/null supervision

bitflicker64:
## Purpose of the PR

- Part of #3043 (Chunk 10) 
- close #3043
- Follow-up to #3047 (added `-d` flag to PD/Store startup scripts) and #3051/#3052 (Docker entrypoint supervision + HEALTHCHECK)

Documents the new `-d true|false` daemon flag and the Docker-native process supervision model introduced in the preceding PRs.

## Main Changes

**`hugegraph-pd/README.md`:**
- Add `-d DAEMON` to the startup options signature
- Document: daemon mode (`true` = daemon, `false` = foreground; default: `true`); set to `false` under Docker or a process supervisor

**`hugegraph-store/README.md`:**
- Same additions as PD

**`hugegraph-server/hugegraph-dist/docker/README.md`:**
- Add section explaining HEALTHCHECK endpoints (`:8080/versions`, `:8620/v1/health`, `:8520/v1/health`) and the Java process supervision model
- Note that the cron-based monitor (`-m true`) is for VM/bare-metal only, not Docker

## Verifying these changes

  - `docker inspect <image> | grep -A5 Healthcheck` — confirms HEALTHCHECK is present
  - `docker stop <container>` — container exits cleanly (SIGTERM forwarded to Java)
  - `kill -9 <java-pid>` inside container — container exits, restart policy fires

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects
- [x] Nope

## Documentation Status

- [x] `Doc - Updated` (in-repo README files only; `hugegraph-doc` follow-up is tracked separately)

### DIFF
--- a/hugegraph-server/hugegraph-dist/docker/README.md
+++ b/hugegraph-server/hugegraph-dist/docker/README.md
@@ -143,7 +143,4 @@ native `HEALTHCHECK` instructions. `docker ps` shows real health status:
 | `hugegraph/hugegraph-pd` | `GET /v1/health` on port 8620 |
 | `hugegraph/hugegraph-store` | `GET /v1/health` on port 8520 |
 
-The entrypoints supervise the Java process directly — when Java exits, the
-container exits and Docker's `restart: unless-stopped` policy brings it back
-automatically. The old cron-based monitor (`-m true`) is for VM/bare-metal
-deployments only and is not used in Docker images.
+The entrypoints supervise the Java process directly — when Java exits, the container exits. If started with a restart policy (the provided compose files use `restart: unless-stopped`), Docker will bring it back automatically. The old cron-based monitor (`-m true`) is for VM/bare-metal deployments only and is not used in Docker images.
--- a/hugegraph-server/hugegraph-dist/docker/README.md
+++ b/hugegraph-server/hugegraph-dist/docker/README.md
@@ -130,3 +130,20 @@ HUGEGRAPH_VERSION=1.7.0 docker compose -f docker-compose-3pd-3store-3server.yml 
 
 See [docker/README.md](../../../docker/README.md) for the full setup guide,
 environment variable reference, and troubleshooting.
+
+## 6. Process Supervision & Health Checks
+
+All four HugeGraph Docker images (PD, Store, Server, Server-hstore) include
+native `HEALTHCHECK` instructions. `docker ps` shows real health status:
+
+| Image | Health endpoint |
+|---|---|
+| `hugegraph/hugegraph` | `GET /versions` on port 8080 |
+| `hugegraph/hugegraph-hstore` | `GET /versions` on port 8080 |
+| `hugegraph/hugegraph-pd` | `GET /v1/health` on port 8620 |
+| `hugegraph/hugegraph-store` | `GET /v1/health` on port 8520 |
+
+The entrypoints supervise the Java process directly — when Java exits, the
+container exits and Docker's `restart: unless-stopped` policy brings it back
+automatically. The old cron-based monitor (`-m true`) is for VM/bare-metal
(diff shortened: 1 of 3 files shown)


hugegraph-server/hugegraph-dist/docker/README.md
```diff
@@ -130,3 +130,20 @@ HUGEGRAPH_VERSION=1.7.0 docker compose -f docker-compose-3pd-3store-3server.yml
...
+| Image | Health endpoint |
+|---|---|
+| `hugegraph/hugegraph` | `GET /versions` on port 8080 |
+| `hugegraph/hugegraph-hstore` | `GET /versions` on port 8080 |
+| `hugegraph/hugegraph-pd` | `GET /v1/health` on port 8620 |
+| `hugegraph/hugegraph-store` | `GET /v1/health` on port 8520 |
+
+The entrypoints supervise the Java process directly — when Java exits, the
+container exits and Docker's `restart: unless-stopped` policy brings it back
+automatically. The old cron-based monitor (`-m true`) is for VM/bare-metal
```
imbajin: ⚠️ **Scope the restart-policy behavior**

Evidence: this says Docker will bring the container back automatically, but `restart: unless-stopped` is supplied by the compose files, not by the image or entrypoint. A user following the direct `docker run` examples will not get automatic restarts unless they also pass a Docker restart policy.

Impact: operators can expect self-recovery that is not actually configured for non-compose deployments.

Requested fix: qualify this as applying when the container is started with a restart policy, for example the provided compose files use `restart: unless-stopped`.

imbajin (commented):
Blocking: no. Summary: The Docker supervision docs overstate restart behavior for direct docker run users. Evidence: static review of hugegraph-server/hugegraph-dist/docker/README.md and the compose files.
---
Latest-head CI has a visible failure; please also check codecov/project: 🔗 https://app.codecov.io/gh/apache/hugegraph/pull/3056

hugegraph-server/hugegraph-dist/docker/README.md
bitflicker64: Fixed qualified the restart-policy statement: "If started with a restart policy (the provided compose files use restart: unless-stopped), Docker will bring it back automatically." Direct docker run users need to pass --restart unless-stopped explicitly. Pushed.

imbajin (approved):
+1, the doc update matches the current Docker supervision behavior.

One non-blocking follow-up: since the README also has direct `docker run` examples, it may be helpful to add a small `--restart unless-stopped` example or note there as well. The new section already scopes restart behavior correctly, but an inline example would make it harder for users to miss.

bitflicker64:
> +1, the doc update matches the current Docker supervision behavior.
> 
> One non-blocking follow-up: since the README also has direct `docker run` examples, it may be helpful to add a small `--restart unless-stopped` example or note there as well. The new section already scopes restart behavior correctly, but an inline example would make it harder for users to miss.

ill make a follow up  pr later with the chore fixes 🫡

PR #3058, opened 2026-06-10
fix(core): ref-count store event listener to fix owner-first close leak

### LINKED ISSUE #3033: [Bug] storeEventListenStatus leaks shared store listener when a non-owner CachedGraphTransaction closes first
### Bug Type (问题类型)

logic (逻辑设计问题)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0 (Apache Release Version)

## TL;DR

**Component:** `CachedGraphTransaction` and `CachedSchemaTransaction` (cache layer, store event listener lifecycle)

**Bug:** When multiple `CachedGraphTransaction` instances share a graph and a non-owner transaction closes first, the shared store event listener is orphaned — it remains registered with the backend provider but becomes untracked, so it is never cleaned up.

**Impact:** On long-running HugeGraph servers, each leaked listener holds references to cache structures and continues receiving store events indefinitely, resulting in a slow memory leak and unnecessary event processing that accumulates over the JVM lifetime.

### Expected

When several `CachedGraphTransaction` instances share the same graph, closing a non‑owner transaction must not touch the shared store listener registered by the first (owner) transaction. The store listener should only be unregistered when the owner closes.

### Actual

`unlistenChanges()` calls `storeEventListenStatus.remove(graphName)` unconditionally. The first transaction to close — owner or not — pops the tracking entry. If it is a non‑owner, it then calls `provider().unlisten(this.storeEventListener)` on its own per‑instance listener, which was never registered with the provider, so the unlisten is a no‑op. The owner's real store listener stays registered but is now untracked. When the owner closes later, `remove(graphName)` returns `null`, the real `unlisten(...)` call is skipped, and the listener leaks for the rest of the JVM lifetime.

The same pattern is present in `CachedSchemaTransaction.java`.

### Reproduction

1. Open two `CachedGraphTransaction` instances `T1` (owner) and `T2` (non‑owner) for the same `spaceGraphName`.
2. `T1.listenChanges()` registers `T1.storeEventListener` via `store().provider().listen(...)` and inserts the tracking entry into `storeEventListenStatus`.
3. `T2.listenChanges()` sees the entry already there and skips registration.
4. Close `T2` first. `storeEventListenStatus.remove(graphName)` returns `true`. `provider().unlisten(T2.storeEventListener)` runs as a no‑op because that instance was never registered.
5. Close `T1`. `storeEventListenStatus.remove(graphName)` returns `null`. The real `provider().unlisten(T1.storeEventListener)` is skipped.
6. `T1.storeEventListener` stays registered with the provider indefinitely.

### Root cause

Same owner‑first‑close shape that PR #3017 fixed for `graphCacheEventListeners` via the `CacheListenerHolder` ref‑count. `storeEventListenStatus` was left as‑is, with the gap flagged in the in‑code TODO:

- [`CachedGraphTransaction`](https://github.com/apache/hugegraph/blob/master/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java#L238-L246)
- mirror in [`CachedSchemaTransaction`](https://github.com/apache/hugegraph/blob/master/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransaction.java)

The unit test [`CachedGraphTransactionTest`](https://github.com/apache/hugegraph/blob/master/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java#L106-L109) papers over it with `storeListeners.putIfAbsent(graphName, true)` after closing the secondary transaction, so the test teardown can still unlisten the primary listener.

### Suggested fix

Two options surfaced in the PR #3017 review:

1. Apply the same `CacheListenerHolder` ref‑counted holder pattern used for the cache listener to the store listener.
2. Stop registering and unregistering on transaction lifecycle. The cache is already a `CacheManager` singleton per `spaceGraphName`, so the store listener lifetime is the graph, not the transaction. Move register/unregister to graph open/dispose.

Option 2 is preferred: it removes the owner concept and the whole race class. Option 1 is the smaller mechanical change.

### Scope

- Apply the chosen fix to `CachedGraphTransaction` and `CachedSchemaTransaction`.
- Remove the `putIfAbsent` workaround from `CachedGraphTransactionTest`.
- Add a regression test that exercises non‑owner‑first‑close ordering and asserts the owner's listener is still unregistered when the owner closes.

### References

- PR #3017 review thread on `CachedGraphTransactionTest.java:236` (TODO follow‑up note).
- PR #3017 review thread on `CachedGraphTransaction.java:210` (owner‑first‑close discussion).
- Introduced as a non‑blocking TODO in PR #3017 (cache event alignment).

```javascript

```

```javascript

```

dpol1:
## Purpose of the PR

close apache/hugegraph#3033

`storeEventListenStatus` had the same owner-first-close bug #3017 fixed
for graph cache listeners: non-owner `close()` dropped the tracking entry
and skipped `unlisten()` as a no-op, leaving the owner's listener
registered but untracked — leak + no store-event cache invalidation.

## Main Changes

- **`CachedGraphTransaction`**: add `StoreListenerHolder` inner class
  (listener, provider, refCount) + `STORE_EVENT_LISTENERS` static
  registry; acquire/release via `ConcurrentMap.compute()` with
  provider-identity guard for graph close/reopen. TODO block removed.
  Drop write-only `storeEventListener` instance field — listener
  ownership moved to `StoreListenerHolder`, field was assigned but
  never read.
- **`GraphTransaction`**: remove `storeEventListenStatus` declaration
  and orphaned `ConcurrentHashMap` import.
- **`CachedGraphTransactionTest`**: delete
  `restoreStoreListenerStatusForKnownTeardownBug` workaround; repoint
  reflection helper to `STORE_EVENT_LISTENERS`; add 5 regression tests:
  non-owner close keeps listener registered, last close removes listener,
  owner-first close with real `STORE_CLEAR` fires through provider and
  surviving tx clears cache, reopen re-registers a fresh holder. Provider
  is pooled across reopen — provider-replacement branch stays defensive,
  not exercised.

`CachedSchemaTransaction*` unaffected — per-instance, balanced 1:1.

### Breaking change (internal API)

This removes the `protected static final ConcurrentHashMap<String, Boolean>
storeEventListenStatus` field from `GraphTransaction`. The removal is
intentional and mirrors #3017, which already removed the sibling
`graphCacheListenStatus` (same `protected static`, same 1.7.0 vintage) with no
compatibility shim when graph-cache listening moved to the ref-counted holder.

- **Scope**: internal listen-tracking state consumed only by the in-tree
  `CachedGraphTransaction`; no external extension point uses it.
- **No deprecated shim**: a `@Deprecated` no-op field cannot preserve the old
  per-graph boolean semantics (now replaced by ref-counting), so it would be
  dead, misleading static state.
- **Migration**: none required unless external code subclasses
  `GraphTransaction` and reads this internal field directly. Store-event
  listening is now handled transparently by `STORE_EVENT_LISTENERS`.

## Verifying these changes

  - RED/GREEN TDD: new tests failed on master (`NoSuchFieldException` on
    `STORE_EVENT_LISTENERS`), green after fix.
  - 13/13 pass: `CachedGraphTransactionTest`
  - 36/36 pass: `CachedGraphTransactionTest`, `CachedSchemaTransactionTest`,
    `CacheManagerTest`, `CacheTest`

- [x] Other affects (typed here): removes the `protected static`
  `storeEventListenStatus` field from `GraphTransaction` — an intentional,
  internal-only breaking API change (see "Breaking change (internal API)"
  above), mirroring #3017. No known external use.

## Documentation Status

- [x] `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
@@ -28,7 +28,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -140,8 +139,6 @@ public class GraphTransaction extends IndexableTransaction {
 
     private final int verticesCapacity;
     private final int edgesCapacity;
-    protected static final ConcurrentHashMap<String, Boolean> storeEventListenStatus =
-            new ConcurrentHashMap<>();
 
     public GraphTransaction(HugeGraphParams graph, BackendStore store) {
         super(graph, store);
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java
@@ -30,6 +30,8 @@ import org.apache.hugegraph.backend.cache.Cache;
 import org.apache.hugegraph.backend.cache.CachedGraphTransaction;
 import org.apache.hugegraph.backend.id.Id;
 import org.apache.hugegraph.backend.id.IdGenerator;
+import org.apache.hugegraph.backend.store.BackendStoreProvider;
+import org.apache.hugegraph.event.EventHub;
 import org.apache.hugegraph.event.EventListener;
 import org.apache.hugegraph.schema.VertexLabel;
 import org.apache.hugegraph.structure.HugeEdge;
@@ -108,6 +110,10 @@ public class CachedGraphTransactionTest extends BaseUnitTest {
     private static int holderRefCount(Object holder) {
         Integer refCount = Whitebox.getInternalState(holder, "refCount");
         return refCount;
+    }
+
+    private static BackendStoreProvider holderProvider(Object holder) {
+        return Whitebox.getInternalState(holder, "provider");
     }
 
     private HugeVertex newVertex(Id id) {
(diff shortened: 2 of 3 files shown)


hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java
```diff
@@ -265,18 +255,58 @@ public void testClosingNonOwnerKeepsGraphCacheListenerRegistered()
...
+        second.close();
+
+        // Non-owner close must decrement the refcount, not drop the entry
+        Assert.assertSame(holder, storeListeners.get(graphName));
+        Assert.assertEquals(refCount, holderRefCount(holder));
+        Assert.assertSame(registered, holderListener(holder));
+    }
+
+    @Test
+    public void testLastCloseRemovesStoreListener() throws Exception {
```
imbajin, line 298: ⚠️ **Cover the owner-first store listener path**

The new store-listener tests cover the non-owner close and final-release cases, but they do not exercise the owner-first path that this ref-counted lifecycle is meant to protect. Please add a store-side equivalent of `testCacheListenerSurvivesOwnerClose()`: close the owner transaction first, trigger a store event such as `STORE_CLEAR` / `STORE_TRUNCATE`, and verify the surviving transaction still observes it through the provider listener. That would also give the provider-replacement branch in `CachedGraphTransaction` a useful regression guard if the test can reopen the same graph name with a new provider.

imbajin (commented):
Blocking: no. Summary: Please add one store-listener lifecycle test for the owner-first close path. Evidence: static review of CachedGraphTransactionTest; local Maven test was blocked by unresolved ${revision} dependency while current CI checks pass.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/cache/CachedGraphTransactionTest.java
dpol1: Address by b3cfd87

imbajin (dismissed):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: CachedGraphTransactionTest passed locally and latest-head CI checks are green.

dpol1:
@imbajin This PR covers the CGT fix. A follow-up improvement for CachedSchemaTransaction will be tracked separately. Not a bug or a hurry issue, but an improvement that will be calmly implemented.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -140,8 +139,6 @@ public class GraphTransaction extends IndexableTransaction {
 
     private final int verticesCapacity;
     private final int edgesCapacity;
```
VGalaxies, line 141: **Low: Protected listener state field was removed**

`hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java:141`

**Evidence**
- The diff removes `protected static final ConcurrentHashMap<String, Boolean> storeEventListenStatus` from `GraphTransaction`; `origin/master` exposed it at old lines 143-144.

**Impact**
- External subclasses or same-package extensions compiled against this protected member can fail to compile, and already-compiled code that references the field can fail linkage after upgrade.

**Requested fix**
- Keep a deprecated compatibility field, or explicitly document this as a breaking internal API change in the release notes.

VGalaxies (commented):
**Review summary**

- **Blocking:** no
- **Summary:** One low-severity compatibility risk remains from removing a protected field.
- **Evidence:**
  - `git diff --unified=0 origin/master...HEAD -- hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java`

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
dpol1: @VGalaxies The removal is intentional and scoped. This PR only drops `storeEventListenStatus`, mirroring #3017, which already removed the sibling `graphCacheListenStatus` from this same class (also `protected static final`, same 1.7.0 vintage) when it moved graph-cache listening to the ref-counted `GRAPH_CACHE_EVENT_LISTENERS` holder. The `// TODO (follow-up): storeEventListenStatus has the same owner-first close bug…` comment #3017 left in `CachedGraphTransaction` scoped exactly this change.

I'd prefer not to keep a deprecated compatibility field:
  - It's internal listen-tracking state consumed only by the in-tree `CachedGraphTransaction`; no external extension point uses it, and #3017 already set the precedent of removing the equivalent field outright without a shim.
  - A `@Deprecated` no-op field can't preserve the old per-graph boolean semantics (now replaced by ref-counting), so it would be dead, misleading static state.

The PR description already notes this as an intentional internal-API removal; I can expand that into an explicit breaking-change/release-note line if you'd prefer. Does that work?

dpol1:
Addressed - Also updated PR description

imbajin (approved):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: git diff --check; CachedGraphTransactionTest passed locally (13 tests); latest-head checks pass.

Issue #3059, opened 2026-06-11
[Question] How is memory reclamation handled in RocksDB storage mode?


cichengzi:
### Problem Type (问题类型)

None

### Before submit

- [x] 我已经确认现有的 [Issues](https://github.com/apache/hugegraph/issues) 与 [FAQ](https://hugegraph.apache.org/docs/guides/faq/) 中没有相同 / 重复问题 (I have confirmed and searched that there are no similar problems in the historical issue and documents)

### Environment (环境信息)

- Server Version: 1.5.0 (Apache Release Version)
- Backend: RocksDB
- OS: 4 CPUs, 32 G RAM, Ubuntu 22.0
- Data Size: 450w+ vertices, 1000w+ edges

### Your Question (问题描述)

I'd like to know how HugeGraph 1.5.0 handles memory reclamation and control in RocksDB storage mode. After repeatedly creating graphs, adding nodes and edges, and performing Gremlin queries, memory usage remains consistently high. What can I do to get that memory reclaimed and released? Even after I executed a Gremlin query task with a large amount of data, all the HugeGraph backend interfaces became unusable. No matter which interface I called, it would eventually result in an InterruptedException or Timeout.

### Vertex/Edge example (问题点 / 边数据举例)

```javascript

```

### Schema [VertexLabel, EdgeLabel, IndexLabel] (元数据结构)

```javascript

```

cichengzi:
I currently have multiple machines, two with 32GB RAM and one with 64GB RAM. However, in actual use, as graphs are continuously created and tasks are executed, memory usage increases rapidly until it almost completely consumes all memory, and memory is not reclaimed for a long time. I've actually made some JVM memory settings in `hugegraph-server.sh` as follows:

JAVA_OPTIONS="-Xms16g -Xmx16g -XX:MetaspaceSize=1024m -XX:MaxMetaspaceSize=2047m -XX:MaxDirectMemorySize=2g -XX:+UseG1GC -XX:MaxGCPauseMillis=200 -XX:InitiatingHeapOccupancyPercent=40"

# Turn on security check
exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \

org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF} \
    >> ${OUTPUT} 2>&1

cichengzi:
[...]

Issue #3057, opened 2026-06-09
[Question] After I completed a query that retrieved a particularly large amount of data, my HugeGraph front end could no longer display the task list, showing "Failed to update/query TaskStore: java.lang.InterruptedException".


cichengzi:
### Problem Type (问题类型)

rest-api (结果不合预期)

### Before submit

- [x] 我已经确认现有的 [Issues](https://github.com/apache/hugegraph/issues) 与 [FAQ](https://hugegraph.apache.org/docs/guides/faq/) 中没有相同 / 重复问题 (I have confirmed and searched that there are no similar problems in the historical issue and documents)

### Environment (环境信息)

- Server Version: 1.5.0 (Apache Release Version)
- Backend: RocksDB
- OS: 4 CPUs, 32 G RAM, Ubuntu 22.0
- Data Size:  450w+ vertices, 1000w+ edges

### Your Question (问题描述)

My Gremlin query retrieved a large amount of data; the original query results were stored as JSON, totaling about 1GB. Now I can no longer query the task through the task list. Furthermore, I can no longer retrieve results individually via an API like http://${IP}:${PORT}/graphs/${GRAPH_NAME}/tasks/${TASK_ID}. What should I do?

{
    "exception": "class org.apache.hugegraph.HugeException",
    "message": "Failed to update/query TaskStore: java.lang.InterruptedException",
    "cause": "java.lang.InterruptedException"
}

### Vertex/Edge example (问题点 / 边数据举例)

```javascript

```

### Schema [VertexLabel, EdgeLabel, IndexLabel] (元数据结构)

```javascript

```

cichengzi:
However, in my extreme scenario, I need to support the export of data exceeding 2GB. Your configuration item here is of type int, right? Also, after adjusting it, if I view the task list in the Hugegraph front end, the total recalled data of the tasks exceeds 2GB, and I still cannot view the tasks. Furthermore, I need to query the results of the task multiple times normally, and this method is not feasible.

cichengzi:
None of the methods you mentioned are very suitable for me.

imbajin:
Thanks for the clarification. I think this case should not be solved by increasing `task.result_size_limit`.

HugeGraph already has pagination/cursor-like APIs for some data access paths, for example:

- [Vertex API pagination](https://hugegraph.apache.org/docs/clients/restful-api/vertex/#method--url-6)
- Traverser shard scan APIs:
[...]

PR #3060, opened 2026-06-17
fix(server): avoid loading huge task results for metadata queries

contrueCT:
﻿﻿﻿

## Purpose of the PR

Fix task metadata/recovery paths that can still load and decompress large task results, making `GET /tasks`, task restore, and `DELETE /tasks/{id}?force=true` unusable when historical tasks contain huge result payloads.

This is related to #3057 and #3059. The PR focuses on metadata-only access and cleanup paths, while large Gremlin result export/chunking can be handled separately.
 

## Main Changes

- Add scheduler APIs that allow task reads with or without result payloads.
- Add metadata-only task deserialization that avoids reading `~task_result`.
- Keep `GET /tasks/{id}` default behavior compatible, and add `with_result=false`.
- Make `GET /tasks`, task restore, and delete paths use metadata-only reads.
- Add lightweight task vertex deletion to avoid `GraphTransaction.removeVertex()` force-loading indexed task vertices.
- Avoid loading separated `~taskresult` vertices when cleaning distributed task results.
- Return a result-stripped copy for in-memory tasks when `withResult=false`.

## Verifying these changes

    - `mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,rocksdb -Dtest=TaskAndResultSchedulerTest -DfailIfNoTests=false -Dcheckstyle.skip=true -Drat.skip=true`
    - `mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test -Dtest=TaskCoreTest#testTaskWithoutResult -DfailIfNoTests=false -Dcheckstyle.skip=true -Drat.skip=true`
    - `mvn test-compile -pl hugegraph-server/hugegraph-test -am -P api-test,rocksdb -Dtest=TaskApiTest -DskipTests=true -Dcheckstyle.skip=true -Drat.skip=true`

- [ ]  Modify configurations
- [x]  The public API (adds optional `with_result` query parameter for `GET /tasks/{id}`; default behavior remains compatible)
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need` (the optional `with_result` query parameter is exposed by OpenAPI annotations, and the default behavior remains compatible)

### DIFF
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/job/TaskAPI.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/job/TaskAPI.java
@@ -102,12 +102,13 @@ public class TaskAPI extends API {
             limit = NO_LIMIT;
             List<Id> idList = ids.stream().map(IdGenerator::of)
                                  .collect(Collectors.toList());
-            iter = scheduler.tasks(idList);
+            iter = scheduler.tasks(idList, false);
         } else {
             if (status == null) {
-                iter = scheduler.tasks(null, limit, page);
+                iter = scheduler.tasks(null, limit, page, false);
             } else {
-                iter = scheduler.tasks(parseStatus(status), limit, page);
+                iter = scheduler.tasks(parseStatus(status), limit, page,
+                                       false);
             }
         }
 
@@ -136,12 +137,17 @@ public class TaskAPI extends API {
                                    @Parameter(description = "The graph name")
                                    @PathParam("graph") String graph,
                                    @Parameter(description = "The task id")
-                                   @PathParam("id") long id) {
+                                   @PathParam("id") long id,
+                                   @Parameter(description = "Whether to load task result")
+                                   @DefaultValue("true")
+                                   @QueryParam("with_result")
+                                   boolean withResult) {
         LOG.debug("Graph [{}] get task: {}", graph, id);
 
         TaskScheduler scheduler = graph(manager, graphSpace, graph)
                 .taskScheduler();
-        return scheduler.task(IdGenerator.of(id)).asMap();
+        return scheduler.task(IdGenerator.of(id), withResult)
+                        .asMap(true, withResult);
     }
 
     @DELETE
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
@@ -308,8 +308,9 @@ public class DistributedTaskScheduler extends TaskAndResultScheduler {
             if (vertex == null) {
                 return null;
             }
-            HugeTask<V> result = HugeTask.fromVertex(vertex);
-            this.tx().removeVertex(vertex);
+            HugeTask<V> result = HugeTask.fromVertex(vertex, false);
+            this.deleteTaskResultFromTx(id);
+            this.tx().removeTaskVertex(vertex);
             return result;
         });
     }
@@ -629,7 +630,7 @@ public class DistributedTaskScheduler extends TaskAndResultScheduler {
                     // 1. start task can be from schedule() & cronSchedule()
                     // 2. recheck the status of task, in case one same task
                     // called by both methods at same time;
-                    HugeTask<Object> queryTask = task(this.task.id());
+                    HugeTask<Object> queryTask = task(this.task.id(), false);
                     if (queryTask != null &&
                         !TaskStatus.NEW.equals(queryTask.status())) {
                         return;
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
@@ -306,11 +306,13 @@ public class DistributedTaskScheduler extends TaskAndResultScheduler {
             Iterator<Vertex> vertices = this.tx().queryTaskInfos(id);
             HugeVertex vertex = (HugeVertex) QueryResults.one(vertices);
             if (vertex == null) {
+                this.deleteTaskResultFromTx(id);
                 return null;
             }
             HugeTask<V> result = HugeTask.fromVertex(vertex, false);
-            this.deleteTaskResultFromTx(id);
+            // Prefer a retryable orphan result over a live task missing result.
             this.tx().removeTaskVertex(vertex);
+            this.deleteTaskResultFromTx(id);
             return result;
         });
     }
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
@@ -310,9 +310,10 @@ public class DistributedTaskScheduler extends TaskAndResultScheduler {
                 return null;
             }
             HugeTask<V> result = HugeTask.fromVertex(vertex, false);
-            // Prefer a retryable orphan result over a live task missing result.
-            this.tx().removeTaskVertex(vertex);
+            // Keep the task vertex as a retryable tombstone until its result
+            // vertex is removed; cronSchedule() can rediscover DELETING tasks.
             this.deleteTaskResultFromTx(id);
+            this.tx().removeTaskVertex(vertex);
             return result;
         });
     }
@@ -325,6 +326,12 @@ public class DistributedTaskScheduler extends TaskAndResultScheduler {
             this.updateStatus(id, null, TaskStatus.DELETING);
             return null;
         } else {
+            HugeTask<?> task = this.taskWithoutResult(id);
+            if (task != null && task.status() != TaskStatus.DELETING) {
+                initTaskParams(task);
+                task.overwriteStatus(TaskStatus.DELETING);
+                this.save(task);
+            }
             return this.deleteFromDB(id);
         }
     }
@@ -590,7 +597,7 @@ public class DistributedTaskScheduler extends TaskAndResultScheduler {
         }
     }
 
-    private boolean isLockedTask(String taskId) {
+    protected boolean isLockedTask(String taskId) {
         return MetaManager.instance().isLockedTask(graphSpace,
                                                    graphName, taskId);
     }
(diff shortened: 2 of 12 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/TaskAndResultScheduler.java
```diff
@@ -103,41 +103,65 @@ public <V> void save(HugeTask<V> task) {
...
+            }
         }
 
         return task;
     }
 
     @Override
     public <V> Iterator<HugeTask<V>> tasks(List<Id> ids) {
-        return this.tasksWithoutResult(ids);
+        return this.tasks(ids, false);
```
imbajin, line 137: ‼️ **Preserve the existing `tasks(...)` result contract**

The new `TaskScheduler` default methods keep the old behavior as `tasks(ids) -> tasks(ids, true)`, but this override changes the task/result-backed scheduler to return metadata-only tasks by default. Existing Java callers of `scheduler.tasks(...).next().result()` can now silently get `null` while `scheduler.task(id)` still returns the result. Please keep the no-result path opt-in by removing these overrides or delegating them to `withResult=true`, and pass `false` only from the metadata-only call sites such as REST list, restore, delete, and scheduler scans.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/TaskCoreTest.java
```diff
@@ -115,6 +116,79 @@ public void testTask() throws TimeoutException {
         });
     }
 
+    @Test
+    public void testTaskWithoutResult() throws TimeoutException {
+        HugeGraph graph = graph();
+        TaskScheduler scheduler = graph.taskScheduler();
+        CountDownLatch latch = new CountDownLatch(1);
+
+        TaskCallable<String> callable = new TaskCallable<String>() {
```
imbajin: ⚠️ **Use a reloadable task callable in this regression test**

This anonymous `TaskCallable` is persisted by class name, then the worker reloads it through `TaskCallable.fromClass()`. It has no public no-arg constructor, so the task fails to load as `TaskCallable$1`, and `scheduler.waitUntilTaskCompleted(id, 10)` times out. I reproduced it with `mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test -Dtest=TaskCoreTest#testTaskWithoutResult -DfailIfNoTests=false -ntp`. Please use an existing reloadable callable or add a small static callable class for this test.

imbajin (commented):
Blocking: yes. Summary: Found a scheduler API compatibility regression and a failing targeted test. Evidence: `TaskCoreTest#testTaskWithoutResult` failed locally; `codecov/project` is also failing.

---
🔗 Please also check the failed `codecov/project` status: https://app.codecov.io/gh/apache/hugegraph/pull/3060

imbajin (dismissed):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: git diff --check; TaskCoreTest#testTaskWithoutResult passed; latest GitHub checks passed.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
```diff
@@ -308,8 +308,9 @@ protected <V> HugeTask<V> deleteFromDB(Id id) {
             if (vertex == null) {
                 return null;
             }
-            HugeTask<V> result = HugeTask.fromVertex(vertex);
-            this.tx().removeVertex(vertex);
+            HugeTask<V> result = HugeTask.fromVertex(vertex, false);
+            this.deleteTaskResultFromTx(id);
```
imbajin: Please make the cleanup expectation explicit here. This path now deletes `~taskresult` before deleting `~task`, but `TaskTransaction`/`TaskAndResultTransaction` run with auto-commit and `removeTaskVertex()` commits in `afterWrite()`. If the second delete fails after this call succeeds, we can leave a live task whose result has already been removed. Could you either make the two removals failure-safe/atomic, or add a targeted test/explanation showing why this intermediate state is acceptable/recoverable?

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/TaskCoreTest.java
```diff
@@ -115,6 +115,88 @@ public void testTask() throws TimeoutException {
...
+        task = new HugeTask<>(id, null, new MetadataResultCallable());
+        task.type("test");
+        task.name("metadata-task");
+        scheduler.schedule(task);
+
+        scheduler.waitUntilTaskCompleted(id, 10);
+
+        HugeTask<?> taskWithResult = scheduler.task(id, true);
+        Assert.assertEquals("\"metadata-result\"", taskWithResult.result());
+
```
imbajin, line 162: This currently proves the `withResult` switch for a small result, but it does not reproduce the original failure mode: loading/decompressing a large historical task result. Please add a regression case with a genuinely large/compressed result and assert that metadata-only reads (`task(id, false)`, `tasks(..., false)`, and ideally restore/scheduler metadata paths) do not touch `task_result`.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/job/TaskAPI.java
```diff
@@ -136,12 +137,17 @@ public Map<String, Object> get(@Context GraphManager manager,
                                    @Parameter(description = "The graph name")
                                    @PathParam("graph") String graph,
                                    @Parameter(description = "The task id")
-                                   @PathParam("id") long id) {
+                                   @PathParam("id") long id,
+                                   @Parameter(description = "Whether to load task result")
+                                   @DefaultValue("true")
```
imbajin, line 142: Since this introduces a new public query parameter, please update the PR checklist/docs status accordingly, or briefly explain why the generated OpenAPI annotation is sufficient and no user-facing docs are needed.

imbajin (commented):
Thanks, the current head looks directionally good and preserves the default `GET /tasks/{id}` behavior. Could you please add or clarify a bit more regression coverage before merge?

Two areas seem worth covering:

1. A regression case with a genuinely large/compressed task result, verifying metadata-only paths such as `GET /tasks`, task restore/scheduler loops, and `with_result=false` do not load/decompress `task_result`.
2. The distributed task/result delete path: please either make the `~task` + `~taskresult` cleanup failure-safe/atomic, or add a targeted test/explanation for the current two-step delete path. Since `TaskTransaction` auto-commits after `removeTaskVertex()`, a failure between deleting `~taskresult` and deleting `~task` could otherwise leave a live task whose result has already been removed.

Minor checklist note: `with_result` is a new query parameter, so the PR checklist/docs status may need to reflect that public API surface, even if the runtime OpenAPI annotation already exposes it.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/DistributedTaskScheduler.java
```diff
@@ -306,10 +306,13 @@ protected <V> HugeTask<V> deleteFromDB(Id id) {
...
             HugeVertex vertex = (HugeVertex) QueryResults.one(vertices);
             if (vertex == null) {
+                this.deleteTaskResultFromTx(id);
                 return null;
             }
-            HugeTask<V> result = HugeTask.fromVertex(vertex);
-            this.tx().removeVertex(vertex);
+            HugeTask<V> result = HugeTask.fromVertex(vertex, false);
+            // Prefer a retryable orphan result over a live task missing result.
+            this.tx().removeTaskVertex(vertex);
```
imbajin: ‼️ **Make orphan result cleanup reachable from production retries**

This failure mode is now explicitly tested, but the recovery path in the test calls `deleteFromDBForTest(id)` directly after the task vertex has already been removed. In production, `cronSchedule()` only discovers delete work by scanning `DELETING` task vertices, while `TaskAndResultTransaction` is auto-commit and `removeTaskVertex()` commits through `afterWrite()`. If `deleteTaskResultFromTx()` fails after this line, the remaining `~taskresult` vertex has no task vertex/status index entry left for the scheduler to rediscover, so the large result can leak permanently. Please either keep a retryable tombstone until result cleanup succeeds, add a production orphan-result cleanup path, or keep the task vertex until the result delete has succeeded; the regression test should exercise that production recovery path rather than only a direct helper call.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/TaskAndResultScheduler.java
```diff
@@ -103,41 +103,65 @@ public <V> void save(HugeTask<V> task) {
...
+            }
         }
 
         return task;
     }
 
     @Override
     public <V> Iterator<HugeTask<V>> tasks(List<Id> ids) {
-        return this.tasksWithoutResult(ids);
+        return this.tasks(ids, true);
```
imbajin: ⚠️ **Preserve metadata-only defaults for bulk task scans**

Before this PR, the task/result-backed scheduler's no-flag `tasks(ids)` and `tasks(status, limit, page)` overloads returned metadata-only tasks; the result-loading path was only in the protected `queryTask(...)` helpers. This now routes legacy bulk callers through `withResult=true`, which means any caller not audited in this PR can again load/decompress historical huge task results. Please keep the no-flag bulk overloads metadata-only for this scheduler and require callers that need payloads to opt in through the new `withResult=true` overloads; add a regression covering the no-flag bulk behavior so this boundary stays explicit.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/TaskScheduler.java
```diff
@@ -47,12 +47,25 @@ public interface TaskScheduler {
 
     <V> HugeTask<V> delete(Id id, boolean force);
 
-    <V> HugeTask<V> task(Id id);
+    default <V> HugeTask<V> task(Id id) {
+        return this.task(id, true);
+    }
 
-    <V> Iterator<HugeTask<V>> tasks(List<Id> ids);
+    <V> HugeTask<V> task(Id id, boolean withResult);
```
imbajin: ⚠️ **Keep existing TaskScheduler implementers source-compatible**

The old no-argument methods were changed to defaults, but the new `withResult` overloads are abstract. Any existing downstream `TaskScheduler` implementation that compiled against the old interface now has to add three new methods before it can recompile, even if it does not support separated task results. Please make the new overloads default methods that delegate to the old methods, and let only the built-in schedulers override them where `withResult=false` is meaningful.

imbajin (commented):
Blocking: yes. Summary: The current head still has task-result cleanup and scheduler compatibility regressions to resolve. Evidence: git diff --check passed; TaskCoreTest#testTaskWithoutResult and TaskAndResultSchedulerTest passed locally; latest GitHub checks have no visible failures.

imbajin (approved):

PR #3062, opened 2026-06-22
  fix(core): self-heal etcd meta watch on transport reconnect

### LINKED ISSUE #3036: [Improvement] Re-register HStore schema cache clear listener after a Meta reconnect
- Follow-up to #3011.

That PR made `CachedSchemaTransactionV2` listen for `schema-cache-clear` meta
events and drop the matching local V2 caches, so a schema change on one server
no longer leaves other nodes serving stale schema in HStore / multi-server mode.

The catch is the listener registers exactly once per JVM. The gRPC watch behind
it is process-wide and there is no `unlisten`. If the Meta transport reconnects
and the old watch gets dropped, the listener goes deaf and nothing brings it
back. There is already a manual recovery hook,
`resetMetaListenerForReconnect()`, but nothing calls it, because Meta exposes no
reconnect signal to hang it on.

The two halves of the gap, both in `CachedSchemaTransactionV2`:

- The listener-lifetime comment explaining why recovery is not automatic today:
  https://github.com/apache/hugegraph/blob/master/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java#L52-L58
- The `resetMetaListenerForReconnect()` hook that has no caller and the javadoc
  saying it must be wired to a Meta reconnect callback:
  https://github.com/apache/hugegraph/blob/master/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java#L254-L270

### What should happen

After the Meta transport reconnects, the JVM-wide `schema-cache-clear` listener
comes back on its own and keeps delivering events for every graph. No operator
action.

### What happens now

`MetaManager` / `MetaDriver` expose `listen` and `keepAlive` but no reconnect
callback. A dropped watch goes unnoticed, events stop arriving, and the node can
keep serving stale schema until someone calls `resetMetaListenerForReconnect()`
by hand. Nothing does.

### Proposed fix

Two parts, same theme, easiest to do in one go:

1. Give `MetaManager` / `MetaDriver` a reconnect callback (something like
   `listenReconnect` / `onTransportReconnect`) that fires when the transport
   reconnects and the previous watch is gone.
2. Point `CachedSchemaTransactionV2` at that callback so it re-registers the
   listener. `resetMetaListenerForReconnect()` stops being a manual entry point
   and becomes the callback target.

### How to verify

Two servers against HStore/Meta. Force a Meta reconnect (restart Meta or kill
the connection), then change a schema on server A and assert server B clears its
V2 caches and stops returning stale schema.

dpol1:
## Purpose of the PR

- close #3036

`CachedSchemaTransactionV2` registers a JVM-global watch on the meta store so that a schema change on one node clears the schema cache on the other nodes. That watch goes through `EtcdMetaDriver.listen` / `listenPrefix`, which handed jetcd the bare `Consumer<WatchResponse>` overload. That overload discards `onError` and `onCompleted`, so when jetcd ends a watch (e.g. after a transport reconnect tears down the gRPC stream) the listener stopped receiving events with no log line and no exception. The node kept serving stale schema and nothing reported it. `PdMetaDriver` does not have this problem because its `KvClient` already re-subscribes on error.

## Main Changes

- **`EtcdMetaDriver.listen` / `listenPrefix`** now register a `Watch.Listener` instead of the bare `Consumer` overload, surfacing both `onError` and `onCompleted`.
  - `onError` **only logs** at WARN level. jetcd 0.5.9 (`WatchImpl.WatcherImpl.handleError`) already retries recoverable errors internally by notifying `onError` and rescheduling `resume()` on the same watcher; re-subscribing from `onError` would therefore open a duplicate watch on every transient reconnect.
  - `onCompleted` **schedules a re-subscribe** after a 1 s backoff on a daemon thread. `onCompleted` is jetcd's terminal-close signal: the old watcher is already removed and will not recover, so a replacement watch is safe and necessary. This mirrors the self-heal `PdMetaDriver` already gets from `KvClient`.
  - If the re-subscribe itself throws (e.g. the endpoint is still unreachable), it is retried with the same backoff instead of abandoning recovery after a single failure.
- **Removed `CachedSchemaTransactionV2.resetMetaListenerForReconnect()`**. It was a manual stopgap added in #3011 with no callers, and it never shipped in a tagged release. Now that the driver self-heals, the JVM-global register-once flag is correct to stay set; the manual reset has nothing left to do. The lifecycle comment on `metaEventListenerRegistered` is updated to describe the self-heal behaviour.
- The `MetaDriver` interface is unchanged, so neither implementor needs edits beyond `EtcdMetaDriver`.

Known limitation: the re-subscribe opens a fresh watch without a stored revision, so any cache-clear events emitted during the short reconnect window are not replayed. This is the same behaviour as `PdMetaDriver` / `KvClient`. The change removes the permanent silent failure; it does not add gap replay.

## Verifying these changes

  - New `EtcdMetaDriverTest` (JUnit + Mockito) mocks the jetcd `Client`/`Watch` via the package-private test constructor (no live etcd needed) and asserts:
    1. `onCompleted` triggers a re-subscribe (a second `watch(...)` call).
    2. `onError` does **not** trigger a re-subscribe (only logs), so no duplicate watch is opened.
    3. The re-created prefix watch preserves the original key and `isPrefix` `WatchOption`.
    4. `onNext` still delivers events to the consumer.
    5. A re-watch that throws once is retried and eventually succeeds (no permanent loss).
  - Local run: `EtcdMetaDriverTest` 5/5, `CachedSchemaTransactionTest` 18/18, `MetaManagerSchemaCacheClearEventTest` 6/6 after the stopgap removal.
  - Red-green: commenting out the re-subscribe in `onCompleted` makes the recovery tests fail; restoring it makes them pass.

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [ ] Other affects (typed here)
- [x] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriver.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriver.java
@@ -25,6 +25,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
 import org.apache.commons.io.FileUtils;
@@ -33,7 +36,9 @@ import org.apache.hugegraph.meta.lock.EtcdDistributedLock;
 import org.apache.hugegraph.meta.lock.LockResult;
 import org.apache.hugegraph.type.define.CollectionType;
 import org.apache.hugegraph.util.E;
+import org.apache.hugegraph.util.Log;
 import org.apache.hugegraph.util.collection.CollectionFactory;
+import org.slf4j.Logger;
 
 import com.google.common.base.Strings;
 
@@ -42,6 +47,7 @@ import io.etcd.jetcd.Client;
 import io.etcd.jetcd.ClientBuilder;
 import io.etcd.jetcd.KV;
 import io.etcd.jetcd.KeyValue;
+import io.etcd.jetcd.Watch;
 import io.etcd.jetcd.kv.GetResponse;
 import io.etcd.jetcd.lease.LeaseKeepAliveResponse;
 import io.etcd.jetcd.options.DeleteOption;
@@ -57,8 +63,23 @@ import io.netty.handler.ssl.SslProvider;
 
 public class EtcdMetaDriver implements MetaDriver {
 
+    private static final Logger LOG = Log.logger(EtcdMetaDriver.class);
+
     private final Client client;
     private final EtcdDistributedLock lock;
+
+    // Re-subscribes a dropped watch off the jetcd callback thread; single
+    // daemon thread, process-lifetime (no close()), so JVM shutdown reclaims it.
+    private final ScheduledExecutorService reWatchExecutor =
+            Executors.newSingleThreadScheduledExecutor(r -> {
+                Thread thread = new Thread(r, "etcd-meta-rewatch");
+                thread.setDaemon(true);
+                return thread;
+            });
+
+    // Backoff before re-subscribing a dropped watch. Package-private and
+    // mutable only so tests can set it to 0; never reassigned in production.
+    long reWatchDelayMs = 1000L;
 
     public EtcdMetaDriver(String trustFile, String clientCertFile,
                           String clientKeyFile, Object... endpoints) {
@@ -74,6 +95,13 @@ public class EtcdMetaDriver implements MetaDriver {
         ClientBuilder builder = this.etcdMetaDriverBuilder(endpoints);
         this.client = builder.build();
         this.lock = EtcdDistributedLock.getInstance(this.client);
+    }
+
+    // Package-private constructor for tests: inject a mock Client and skip lock
+    // setup (watch tests never touch the distributed lock).
+    EtcdMetaDriver(Client client) {
+        this.client = client;
+        this.lock = null;
     }
 
     private static ByteSequence toByteSequence(String content) {
@@ -303,9 +331,8 @@ public class EtcdMetaDriver implements MetaDriver {
     @SuppressWarnings("unchecked")
     @Override
     public <T> void listen(String key, Consumer<T> consumer) {
-
-        this.client.getWatchClient().watch(toByteSequence(key),
-                                           (Consumer<WatchResponse>) consumer);
+        this.watchKey(toByteSequence(key), WatchOption.DEFAULT,
+                      (Consumer<WatchResponse>) consumer);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


imbajin (approved):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: git diff --check passed; EtcdMetaDriverTest passed (4 tests); latest GitHub checks passed.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriver.java
```diff
@@ -314,9 +341,35 @@ public <T> void listen(String key, Consumer<T> consumer) {
...
+     * this surfaces {@code onError}/{@code onCompleted}: when the underlying
+     * watch terminates (e.g. a transport reconnect drops the gRPC stream) it
+     * re-subscribes after a short backoff, so the listener is not silently lost.
+     * Mirrors the re-subscribe behaviour PdMetaDriver already gets from KvClient.
+     */
+    private void watchKey(ByteSequence key, WatchOption option,
+                          Consumer<WatchResponse> consumer) {
+        Watch.Listener listener = Watch.listener(
+                consumer,
+                throwable -> this.scheduleReWatch(key, option, consumer, throwable),
```
imbajin: ‼️ Critical: this creates duplicate active watches on retryable jetcd reconnects.

HugeGraph uses `jetcd-core:0.5.9`, and `WatchImpl.handleError()` already calls the application listener's `onError(...)` and then internally `reschedule()`s the same `WatcherImpl` for retryable gRPC errors. With this PR, `onError` also schedules a fresh `watchKey(...)`, while the original watcher is still resumable and not closed/tracked here.

That means one transient transport reconnect can leave two active watchers for the same key/prefix, and repeated reconnects can multiply meta-event delivery and retained watch resources. This affects all `EtcdMetaDriver.listen/listenPrefix` consumers, not only schema cache clear.

Please avoid opening a fresh watch from retryable `onError`. Either rely on jetcd's built-in reschedule path, or track the returned `Watch.Watcher` and replace it only after a confirmed terminal close/cancel path where jetcd will not resume the original watcher. The test should model jetcd's real `onError -> reschedule same watcher` behavior, not just a mocked listener callback.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriver.java
```diff
@@ -314,9 +341,35 @@ public <T> void listen(String key, Consumer<T> consumer) {
...
+        // Watcher intentionally not closed: process-lifetime watch, recreated
+        // on self-heal; the prior watcher's stream has already terminated.
+        this.client.getWatchClient().watch(key, option, listener);
+    }
 
+    private void scheduleReWatch(ByteSequence key, WatchOption option,
+                                 Consumer<WatchResponse> consumer,
+                                 Throwable cause) {
+        LOG.warn("etcd meta watch dropped, re-subscribing", cause);
+        this.reWatchExecutor.schedule(
```
imbajin: ‼️ Critical: a failed delayed re-watch attempt stops recovery permanently.

The scheduled task calls `watchKey(...)` directly. If `watch(...)` throws once while the watch client is still closed/unavailable, the task fails and no further retry is scheduled. Since this PR also removes the upper-layer reset hook and leaves `metaEventListenerRegistered` permanently true, the schema-cache-clear listener can again be lost forever after a reconnect failure.

Please wrap the scheduled body, log failed retry attempts, and keep retrying with bounded/backoff behavior while the driver is active. Add a regression test where the first replacement `watch(...)` call throws and a later attempt succeeds.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriverTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        Mockito.verify(watch, Mockito.timeout(2000).times(2))
+               .watch(Mockito.any(ByteSequence.class),
+                      Mockito.any(WatchOption.class),
+                      Mockito.any(Watch.Listener.class));
+    }
+
+    @Test
+    public void testListenPrefixReSubscribesOnError() {
+        Watch watch = Mockito.mock(Watch.class);
```
imbajin, line 77: ⚠️ Important: this test only proves a second `watch(...)` call happens; it does not prove the re-created prefix watch preserves the original key and prefix `WatchOption`.

Please capture the second invocation's `ByteSequence` and `WatchOption` and assert it still watches `"prefix"` with prefix semantics enabled. Otherwise a future regression could downgrade this to an exact-key watch while the test still passes.

imbajin (changes requested):
Blocking issues found in the etcd watch recovery path: retryable jetcd watch errors can now create duplicate active watchers, and a failed delayed re-watch attempt can stop recovery permanently. Please fix these before merging; I also left one non-blocking test-hardening note.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriver.java
dpol1: Address by d2bd2a2

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriver.java
dpol1: Address by d2bd2a2

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/meta/EtcdMetaDriverTest.java
dpol1: Address by d2bd2a2

imbajin (approved):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: git diff --check origin/master...HEAD; EtcdMetaDriverTest passed 5 tests.

Non-blocking follow-ups:
- Please update the PR description to match the latest implementation: `onError` now only logs because jetcd retries recoverable errors itself, while re-watch is scheduled from terminal `onCompleted`; the test count is now 5, not 4.
- Consider reducing recoverable watch `onError` logging from WARN with stack trace to INFO/DEBUG or a rate-limited log, since jetcd is expected to self-retry those transient failures.
- If HugeGraph treats public core methods as a compatibility surface, consider keeping `CachedSchemaTransactionV2.resetMetaListenerForReconnect()` as a deprecated no-op or documented fallback. Given it was an uncalled stopgap that did not ship in a tagged release, I do not consider this blocking.

PR #3068, opened 2026-06-25
fix(core): keep unsafe range filters local

### LINKED ISSUE #2929: [Bug] Two query behaviors that produce the same result are inconsistent.
### Bug Type (问题类型)

gremlin (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0
- Backend: RocksDB x nodes

### Expected behavior
For the two query below, the same result should be returned：
```gremlin
g.E().has('ep4', lt(0.32696354)).and(__.hasLabel('el2')).count()

g.E().has('ep4', lt(0.32696354)).match(__.as('start1').and(__.hasLabel('el2')).as('m1')).select('m1').count()
```
### Actual behavior
+ The first query thrown a exception: ```org.apache.hugegraph.exception.NoIndexException: Don't accept query based on properties [ep4] that are not indexed in label 'el2', may not match secondary/range condition```

+ When I executed these two queries on Janusgraph and Tinkerpop, they both returned the same results.

```javascript

```

```javascript

```

LegendPei:
## Purpose of the PR

- close #2929 

## Main Changes

- Avoid pushing unsafe non-indexed range/neq predicates into `HugeGraphStep` when followed by match/connective label filters.
- Still push down usable predicates, such as label filters and indexed conditions, to keep backend filtering where it is safe.
- Keep remaining unsafe predicates as local `HasStep` filters.
- Prevent count extraction from crossing a remaining local `HasStep`, so local filters are evaluated before `count()`.
- Add a regression test for connective label filtering with a non-indexed range predicate.

## Verifying these changes

    - xxx

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugeConnectiveLabelStepStrategy.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugeConnectiveLabelStepStrategy.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hugegraph.traversal.optimize;
+
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Set;
+
+import org.apache.tinkerpop.gremlin.process.traversal.Step;
+import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.TraversalStrategy;
+import org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.AndStep;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.HasStep;
+import org.apache.tinkerpop.gremlin.process.traversal.step.util.HasContainer;
+import org.apache.tinkerpop.gremlin.process.traversal.strategy.AbstractTraversalStrategy;
+import org.apache.tinkerpop.gremlin.process.traversal.strategy.optimization.InlineFilterStrategy;
+import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper;
+import org.apache.tinkerpop.gremlin.structure.T;
+
+public final class HugeConnectiveLabelStepStrategy
+        extends AbstractTraversalStrategy<TraversalStrategy.OptimizationStrategy>
+        implements TraversalStrategy.OptimizationStrategy {
+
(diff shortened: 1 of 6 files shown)


imbajin (approved):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: `git diff --check origin/master...refs/remotes/pr/3068`; JDK 11 `CountStrategyCoreTest` passed 26 tests; latest-head checks passed.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugeConnectiveLabelStepStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+        return INSTANCE;
+    }
+
+    private static void markConnectiveLabelSteps(Traversal.Admin<?, ?> traversal,
+                                                Set<Traversal.Admin<?, ?>> visited) {
+        if (!visited.add(traversal)) {
+            return;
+        }
+
+        for (AndStep<?> step : TraversalHelper.getStepsOfClass(AndStep.class,
```
imbajin, line 71: ‼️ `OrStep` connective label filters still miss this optimization path.

This strategy only scans `AndStep`, but label-only `or()` children can hit the same unsafe range + connective-label shape as the fixed `and(__.hasLabel(...))` case. I verified the PR head still fails for a traversal like:

```java
g.E().has("ep4", P.lt(0.32696354F))
     .or(__.hasLabel("el2"), __.hasLabel("el3"))
     .count()
```

It still throws `NoIndexException` for the non-indexed `ep4` range under one of the labels. Please generalize the marker pass to cover label-only `OrStep` as well, or operate on the relevant connective step abstraction while keeping the same safety checks, and add a regression test for `or(__.hasLabel("el2"), __.hasLabel("el3"))`.

imbajin (changes requested):
Blocking: yes. The current head still misses the same unsafe range + connective label class for label-only `or()` filters. I verified this on a temporary detached worktree by adding a minimal regression traversal using `or(__.hasLabel("el2"), __.hasLabel("el3"))`; the PR head still throws `NoIndexException`, while the existing `and(__.hasLabel("el2"))` regression passes.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugeConnectiveLabelStepStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+        for (Step<?, ?> step : traversal.getSteps()) {
+            TraversalUtil.markConnectiveLabelStep(step);
+        }
+    }
+
+    private static boolean hasOnlyLabelContainers(HasStep<?> step) {
+        if (step.getHasContainers().isEmpty()) {
+            return false;
+        }
+        for (HasContainer has : step.getHasContainers()) {
```
imbajin, line 125: ⚠️ Related edge case to guard while fixing the marker logic: this check only verifies that each container key is `T.label`, but it does not constrain the label predicate itself.

That means label predicates such as `hasLabel(P.neq("el2"))`, `hasLabel(P.without("el2"))`, or other broad/negative forms can also be marked as connective label steps and then passed to `TraversalUtil.extractLabelHasContainers()`. That is a wider class than the concrete positive label filters covered by the new regression test (`hasLabel("el2")`).

Please consider only marking/extracting bounded positive label predicates, for example label equality or a finite positive label set, and leave negative/unsupported label predicates local. A small regression around `and(__.hasLabel(P.neq("el2")))` or `P.without(...)` would make this boundary explicit.

imbajin (approved):
Blocking: no. Summary: No obvious issues found in the current head. Evidence: git diff --check origin/master...HEAD; JDK 11 `mvn -pl hugegraph-server/hugegraph-test -am -P core-test,rocksdb -Dtest=CountStrategyCoreTest,TraversalUtilOptimizeTest -DfailIfNoTests=false test` passed 51 tests; latest-head checks passed.

Issue #3079, opened 2026-07-06
[Feature] Implement Cloud Storage Support for HugeGraph


vaijosh:
### Feature Description (功能描述)

**Motivation**
in modern cloud-native environments, compute instances are ephemeral—frequently created and destroyed based on autoscaling policies. HugeGraph currently relies on local disk storage (RocksDB), which creates significant risks for data durability and complicates operational lifecycle management in these dynamic environments. We propose integrating a pluggable cloud-native storage tier to offload data to object storage, effectively decoupling compute from storage to ensure data persistence, reliability, and operational flexibility.

**High-Level Requirements:**

- Decoupled Storage: Enable offloading and retrieving HugeGraph data from cloud object storage (e.g., S3, GCS, Azure Blob).
- Extensible Architecture: Implement a provider-agnostic framework using an SPI (Service Provider Interface) mechanism. This design ensures the core storage engine remains decoupled, allowing the community to add new storage backends without requiring modifications to the core codebase.
- Initial Cloud Support: Provide out-of-the-box support for few popular cloud object storage services e.g. AWS S3, Azure Data Lake Storage Gen2. Google Cloud Storage (GCS) etc.

**Functional Requirements:**

- Automated Data Offloading: Mechanism to persist local SST data to configured cloud object storage.
- Intelligent Hydration: Ability to recover/hydrate data from cloud storage to local compute nodes upon startup or during read-miss scenarios.
- Configuration: Standardized property-based configuration for buckets, endpoints, and caching strategies.

**Non-Functional Requirements:**
 
- Performance: Minimize read/write latency overhead via local caching layers.
- Scalability: Support elastic scaling of compute nodes without requiring complex manual data migration.
- Maintainability: Enable the addition of new storage backends via an SPI-based plugin model to minimize technical debt.

Discussion thread  : https://github.com/apache/hugegraph/discussions/3080
RFC Document :[ RFC: Cloud Storage Support for HugeGraph](https://docs.google.com/document/d/1yzxKX9cpNKqTgYifWc0-zR2otAUeKGba5DSUUzAGSd8/edit?usp=sharing)
PoC Code: In Branch [HStore+CloudStorage](https://github.com/vaijosh/hugegraph/tree/Hstore%2BCloudStorage) in my fork https://github.com/vaijosh/hugegraph

PR #3081, opened 2026-07-06
feat(store): implement pluggable cloud storage for HStore

vaijosh:
# Purpose

* **Closes:** #3079

This PR introduces a cloud storage architecture for **HugeGraph Store (HStore)**. In cloud-native environments, storage nodes are ephemeral, making reliance on local disk storage a single point of failure. This implementation decouples the storage layer from local disk dependencies, allowing SST files to be offloaded to durable, scalable cloud storage providers.

### Pluggable Architecture

* Created the `CloudStorageProvider` SPI interface to enable extensible storage backends.
* Added `CloudStorageProviderFactory` for seamless provider discovery and lifecycle management.

### S3 Provider Implementation

* Introduced the `hg-store-cloud-s3` module, leveraging AWS SDK v2 for production-ready S3/S3 compatible storage interaction.

### Lifecycle & Event Integration

* Integrated `CloudStorageEventListener` with `RocksDBFactory`. This hooks into critical SST lifecycle events (`onTableFileCreated`, `onTableFileDeleted`) to ensure synchronized state between local RocksDB and cloud storage.
* Implemented startup hydration (`onDBCreated`) to backfill pre-existing files and read-miss on-demand hydration to ensure data availability.

### Infrastructure & Testing

* Added a comprehensive local development environment via `docker/cloud-storage/` using MinIO.
* Included an integration test script (`test-graph-queries-and-sst.sh`) to verify end-to-end data durability and query consistency.

## Implementation Highlights

* **SPI Integration**: Standardized service discovery via `META-INF/services`.
* **RocksDB Hooking**: Registered a singleton `RocksdbEventListener` within `RocksDBFactory` to intercept file operations without modifying core RocksDB logic.
* **Read-Miss Protection**: Added a guard window mechanism in `onReadMiss` to prevent redundant hydration requests during high-concurrency read scenarios.

## Verifying These Changes

* [x] **Unit Tests**: Coverage for config parsing (`CloudStorageConfigTest`), factory registration, and event listener logic.
* [x] **Integration Tests**: Verified using the new `docker-compose` setup with MinIO.
* [x] **Configuration**: Verified `application.yml` bindings for credentials, bucket management, and sync intervals.

## Impact

* **Dependencies**: Added AWS SDK v2; updated `LICENSE` and `NOTICE` accordingly.
* **Configurations**: Added `cloud.storage` namespace to `application.yml` (disabled by default).
* **Public API**: No breaking changes to existing public-facing client APIs.

## Documentation

* Architecture and configuration details are documented in `hugegraph-store/docs/pluggable-cloud-storage-architecture.md`.


VGalaxies (commented):
**Review summary**

- **Blocking:** yes
- **Summary:** The change cannot currently guarantee recoverable RocksDB state, may silently disable cloud storage, and contains S3 retry and deployment defects.
- **Evidence:**
  - Static analysis of `git diff origin/master...HEAD`
  - shell syntax and POM XML checks passed
  - `git diff --check` reported whitespace-only errors

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,445 @@
...
+     * run were never uploaded (e.g. cloud storage was enabled after the last shutdown).
+     */
+    private void uploadExistingSstFiles(CloudStorageProvider provider, String dbName,
+                                        String dbPath) {
+        Path root = Paths.get(dbPath);
+        if (!root.toFile().isDirectory()) {
+            return;
+        }
+        try (Stream<Path> paths = Files.walk(root)) {
+            paths.filter(p -> p.toString().endsWith(".sst"))
```
VGalaxies, line 2667: **High: SST-only uploads cannot restore a RocksDB database**

`hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java:298`

**Evidence**
- Production upload paths publish only `*.sst`, while the startup-hydration test manually seeds `CURRENT` and `MANIFEST`. RocksDB requires these metadata files to identify its consistent live file set.

**Impact**
- A node starting with an empty data directory cannot reconstruct the advertised cloud backup and may open an empty database or fail.

**Requested fix**
- Upload and restore an atomic RocksDB checkpoint containing `CURRENT`, `MANIFEST`, options, and referenced SST files.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,445 @@
...
+        }
+        List<String> downloaded = downloadMissingSstFiles(provider,
+                                                          session.getGraphName(),
+                                                          session.getDbPath());
+        if (downloaded.isEmpty()) {
+            return false;
+        }
+        try {
+            Map<byte[], List<String>> sstByCf = new HashMap<>();
+            sstByCf.put(table.getBytes(StandardCharsets.UTF_8), downloaded);
```
VGalaxies: **High: Read-miss hydration ingests every SST into one column family**

`hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java:168`

**Evidence**
- `downloadMissingSstFiles()` collects every missing SST under the database prefix, then the complete list is assigned to the single table whose read missed. The upload key retains no column-family identity, although RocksDB column families maintain separate LSM trees.

**Impact**
- A multi-column-family store can ingest another table’s data into the wrong table or fail hydration.

**Requested fix**
- Persist the originating column family for each SST and group/filter files by column family before ingestion.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,445 @@
...
+     */
+    @Override
+    public void onTableFileDeleted(String dbName, String cfName, String filePath) {
+        CloudStorageProvider provider = CloudStorageProviderFactory.getActiveProvider();
+        if (provider == null) {
+            return;
+        }
+        String remoteKey = toRelativeKey(filePath);
+        try {
+            provider.deleteFile(remoteKey);
```
VGalaxies, line 1984: **High: Compaction can delete the last recoverable cloud generation**

`hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java:250`

**Evidence**
- A replacement-SST upload failure is only queued or moved to the DLQ, but subsequent deletion callbacks unconditionally delete the old input SST objects.

**Impact**
- Disk loss between the failed replacement upload and manual recovery leaves neither the old nor new SST generation in cloud storage.

**Requested fix**
- Retain old remote objects until a complete replacement generation/checkpoint has been durably committed.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,445 @@
...
+    @Override
+    public void onDBOpening(String dbName, String dbPath) {
+        if (!startupHydrationEnabled) {
+            return;
+        }
+        CloudStorageProvider provider = CloudStorageProviderFactory.getActiveProvider();
+        if (provider == null) {
+            return;
+        }
+        preHydrateDbFiles(provider, dbName, dbPath);
```
VGalaxies, line 675: **High: Recreated databases can hydrate deleted data**

`hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java:144`

**Evidence**
- The listener hydrates the existing remote prefix during every open but implements neither `onDBDeleteBegin` nor `onDBDeleted`; directory deletion outside RocksDB generates no table-file deletion callbacks.

**Impact**
- Recreating a graph or partition at the same path can download and ingest stale data from the deleted database.

**Requested fix**
- Delete or tombstone the remote database generation during database destruction and make hydration honor that generation marker.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,445 @@
...
+    @Override
+    public void onTableFileCreated(String dbName, String cfName,
+                                   String filePath, long fileSize) {
+        CloudStorageProvider provider = CloudStorageProviderFactory.getActiveProvider();
+        if (provider == null) {
+            return;
+        }
+        String remoteKey = toRelativeKey(filePath);
+        try {
+            provider.uploadFile(filePath, remoteKey);
```
VGalaxies: **High: Cloud upload blocks the RocksDB event thread**

`hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java:217`

**Evidence**
- `RocksDBFactory` invokes listeners inline, and this callback synchronously performs the entire S3 upload, including multipart retries and sleeps. RocksDB documents that these callbacks run on the triggering thread and should not perform extended work.

**Impact**
- Slow or unavailable object storage can block flush/compaction threads and cause write stalls.

**Requested fix**
- Hand uploads to a bounded executor and pin each SST using a checkpoint, hard link, or spool copy before returning from the callback.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java
```diff
@@ -116,6 +138,69 @@ public void init() {
...
+                     + "(dataRoot='{}', startupHydration={}, readMissHydration=true, "
+                     + "readMissGuardWindowMs={}, uploadRetryMaxAttempts={}, "
+                     + "uploadRetryInitialDelayMs={}, uploadRetryMaxDelayMs={})",
+                     cfg.getProvider(), resolvedDataRoot,
+                     cfg.isStartupHydrationEnabled(),
+                     cfg.getReadMissGuardWindowMs(),
+                     cfg.getUploadRetryMaxAttempts(),
+                     cfg.getUploadRetryInitialDelayMs(),
+                     cfg.getUploadRetryMaxDelayMs());
+        } catch (Exception e) {
```
VGalaxies, line 265: **High: Enabled cloud storage fails open**

`hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java:188`

**Evidence**
- Every provider initialization or configuration exception is logged and swallowed, so startup continues without registering the cloud listener.

**Impact**
- A store configured with `cloud.storage.enabled=true` can remain healthy while uploading no data.

**Requested fix**
- Fail startup when an explicitly enabled provider cannot initialize, or require an explicit fail-open option and expose the disabled state through readiness.

docker/cloud-storage/entrypoints/store-entrypoint.sh
```diff
@@ -0,0 +1,93 @@
...
+    "port": "$(json_escape "${HG_STORE_GRPC_PORT}")"
+  },
+  "raft": { "address": "$(json_escape "${HG_STORE_RAFT_ADDRESS}")" },
+  "server": { "port": "$(json_escape "${HG_STORE_REST_PORT}")" },
+  "app": { "data-path": "$(json_escape "${HG_STORE_DATA_PATH}")" },
+  "cloud": {
+    "storage": {
+      "enabled": "$(json_escape "${HG_CLOUD_STORAGE_ENABLED}")",
+      "provider": "$(json_escape "${HG_CLOUD_STORAGE_PROVIDER}")",
+      "bucket": "$(json_escape "${HG_CLOUD_STORAGE_BUCKET}")",
```
VGalaxies: **High: S3 properties are emitted under the wrong namespace**

`docker/cloud-storage/entrypoints/store-entrypoint.sh:67`

**Evidence**
- The entrypoint writes `cloud.storage.bucket`, `region`, and credentials, while `AppConfig` reads provider properties only from `cloud.storage.s3.*`; the resulting property map is empty.

**Impact**
- The supplied Docker stack starts without an active S3 provider and uploads no SST files.

**Requested fix**
- Nest these properties inside `"s3": { ... }` and assert successful provider initialization during startup.

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
```diff
@@ -0,0 +1,687 @@
...
+            }
+
+            InputStream partStream = new LimitedInputStream(fis, partLen);
+            UploadPartResponse resp = s3Client.uploadPart(
+                    UploadPartRequest.builder()
+                                    .bucket(bucket).key(fullKey)
+                                    .uploadId(uploadId).partNumber(partNumber)
+                                    .contentLength(partLen)
+                                    .build(),
+                    RequestBody.fromInputStream(partStream, partLen));
```
VGalaxies: **High: Multipart request bodies cannot be replayed for SDK retries**

`hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java:538`

**Evidence**
- `RequestBody.fromInputStream()` receives `LimitedInputStream`, which has no mark/reset support. AWS SDK retries require a replayable stream and can fail with `IllegalStateException`, which the surrounding retry loop does not catch.

**Impact**
- A transient failure during a large upload can abort the multipart operation instead of retrying the affected part.

**Requested fix**
- Supply a `ContentStreamProvider` that opens, seeks, and bounds a fresh file stream for each SDK attempt.

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
```diff
@@ -0,0 +1,687 @@
...
+            return resp.eTag();
+        } catch (SdkClientException e) {
```
VGalaxies: **High: S3 service failures bypass the provider retry contract**

`hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java:540`

**Evidence**
- Provider operations catch only `SdkClientException`; S3 service responses use `S3Exception`/`AwsServiceException`, which are separate exception types. The outer part retry catches only `IOException`.

**Impact**
- 429/5xx responses bypass part-level retries, while callers can receive unchecked exceptions despite the provider’s `IOException` contract.

**Requested fix**
- Catch service exceptions, classify retryable versus permanent status codes, and translate them consistently into retryable or non-retryable cloud exceptions.

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
```diff
@@ -0,0 +1,687 @@
...
+                s3Client.abortMultipartUpload(
+                        AbortMultipartUploadRequest.builder()
+                                                  .bucket(bucket).key(fullKey)
+                                                  .uploadId(uploadId).build());
+                log.warn("S3 multipart upload aborted: key={} uploadId={}", fullKey, uploadId);
+            } catch (Exception abortEx) {
+                log.warn("S3 multipart abort failed: key={} uploadId={} reason={}",
+                         fullKey, uploadId, abortEx.getMessage());
+            }
+            throw new IOException(
```
VGalaxies, line 807: **Medium: Multipart wrapping erases the direct-DLQ marker**

`hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java:506`

**Evidence**
- `uploadOnePartWithRetry()` throws `CloudStorageNonRetryableException`, but the enclosing catch wraps it as a plain `IOException`; the queue recognizes non-retryable failures only by type.

**Impact**
- `multipart-exhausted-direct-dlq=true` still permits expensive whole-file retries.

**Requested fix**
- Abort the multipart upload, then rethrow `CloudStorageNonRetryableException` unchanged.

install-dist/release-docs/LICENSE
```diff
@@ -740,6 +758,33 @@ non-Apache-2.0 license text is required for a bundled component.
     https://central.sonatype.com/artifact/org.yaml/snakeyaml/1.28 -> Apache 2.0
     https://central.sonatype.com/artifact/org.yaml/snakeyaml/2.2 -> Apache 2.0
     https://central.sonatype.com/artifact/org.zeroturnaround/zt-zip/1.14 -> Apache 2.0
+    https://central.sonatype.com/artifact/software.amazon.awssdk/annotations/2.25.60 -> Apache 2.0
```
VGalaxies: **Medium: Release metadata lists different dependency versions**

`install-dist/release-docs/LICENSE:761`

**Evidence**
- The provider pins AWS SDK `2.33.8`, and `known-dependencies.txt` records `2.33.8` with Netty `4.1.126.Final`; LICENSE adds AWS `2.25.60` and Netty `4.1.108.Final`.

**Impact**
- Binary release licensing metadata does not describe the actual bundled artifacts.

**Requested fix**
- Regenerate the inventory from the packaged distribution and update LICENSE/NOTICE to the exact resolved versions and components.

docker/cloud-storage/scripts/test-graph-queries-and-sst.sh
```diff
@@ -0,0 +1,378 @@
...
+log "starting stores and server..."
+docker compose -f "$COMPOSE_FILE" up -d store0 store1 store2 server
+log "waiting for stores..."
+wait_svc "store0" 180
+wait_svc "store1" 180
+wait_svc "store2" 180
+wait_svc "server" 180
+log "waiting for graph backend..."
+wait_http "$GRAPH_API_BASE/graph/vertices" 60
+log "✓ SUCCESS: Cloud storage infrastructure ready"
```
VGalaxies, line 1406: **Medium: The advertised SST test never verifies cloud storage**

`docker/cloud-storage/scripts/test-graph-queries-and-sst.sh:378`

**Evidence**
- The script reports success after service and HTTP health checks; it never loads graph data, triggers or counts SST uploads, checks MinIO, tests recovery, or produces the reports promised by the README.

**Impact**
- The validation passes even when the provider is disabled by the entrypoint namespace defect.

**Requested fix**
- Fail unless graph writes succeed, SST objects appear in every expected bucket, and a recovery read succeeds from cloud-backed state.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
vaijosh: Thanks @VGalaxies for sharing the feedback.In recent commit I have implemented this.
To be frank earlier it was planned but I was working on some additional resiliancy mechanism to make sure that DB is consistent even during Sync failures and cluster crash scenarios.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
vaijosh: Thanks @VGalaxies . Yes I have addressed it in recent commit.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
vaijosh: Thanks @VGalaxies.
I have addressed this review comment in recent commit #7564613

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
vaijosh: Thanks @VGalaxies.
I have addressed it in commit #32e407e

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
vaijosh: Thanks @VGalaxies.
I have addressed it in commit #32e407e

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
vaijosh: Thanks @VGalaxies.
I have addressed it in commit #32e407e

install-dist/release-docs/LICENSE
vaijosh: Thanks @VGalaxies.
I have addressed it in commit #32e407e

docker/cloud-storage/scripts/test-graph-queries-and-sst.sh
vaijosh: Thanks @VGalaxies.
I have addressed it in commit #32e407e

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
vaijosh: Thanks @VGalaxies.
I have addressed it in commit #32e407e

VGalaxies (commented):
### 🚨 Review summary

> [!IMPORTANT]
> The cloud recovery path has blocking consistency, configuration, deletion-safety, release-metadata, and verification gaps.

#### 📊 Risk dashboard

| Signal | Result |
| --- | --- |
| 🚦 Review gate | **Blocked** |
| 🔎 Actionable findings | **9** (8 High, 1 Medium) |
| 🧪 Verification coverage | **5** checks |
| ⚠️ Confirmed / limited checks | **2 / 1** |

#### 🔬 Coverage details

<details>
<summary>⚠️ <strong>S3 failure semantics</strong> — Confirmed</summary>

**Trace:** S3CloudStorageProvider upload, download, delete, and exception-classification paths, CloudStorageEventListener purge and hydration flows, AWS SDK 2.33.8 retry-code and file-transformer behavior

**Conclusion:** Confirmed ignored per-object deletion errors, retryable error-code misclassification, and non-atomic download recovery risks; lower-impact lifecycle leads were dropped.

</details>

<details>
<summary>⚠️ <strong>Release and E2E integrity</strong> — Confirmed</summary>

**Trace:** install-dist release LICENSE and known-dependencies inventory, cloud-storage Docker artifact build and recovery workflow, Maven provider packaging and test activation

**Conclusion:** Confirmed the release dependency inventory mismatch and E2E false-green paths caused by suppressed build and wipe failures.

</details>

<details>
<summary>🟡 <strong>Focused unit suites</strong> — Limited</summary>

**Trace:** hg-store-common cloud configuration, provider, and factory tests, hg-store-cloud-s3 exception-classification tests, hg-store-node listener, retry, metrics, tracker, configuration, and callback tests, hg-store-core BusinessHandlerImplTest

**Conclusion:** Surefire reports recorded 184 tests with 0 failures, 0 errors, and 0 skipped. The suites do not cover concurrent metadata publication, comma-separated data roots, or partial S3 batch deletion.

</details>

<details>
<summary>✅ <strong>Maven reactor</strong> — Clear</summary>

**Trace:** Full 44-project reactor with the cloud-s3 profile, hugegraph-store module ordering and dependency graph

**Conclusion:** Maven validate completed successfully for all 44 reactor projects; the suspected duplicate cloud-s3 module activation did not reproduce.

</details>

<details>
<summary>✅ <strong>Shell syntax</strong> — Clear</summary>

**Trace:** pd-entrypoint.sh, store-entrypoint.sh, test-graph-queries-and-sst.sh

**Conclusion:** bash -n returned successfully for all three changed shell entrypoints and workflows.

</details>

> [!WARNING]
> **Verification limits**
> - No live Docker and MinIO recovery drill was run, so cluster recovery and purge behavior were not exercised end to end.
> - The test environment used Java 17; the legacy JaCoCo agent emitted unsupported class-version instrumentation errors, so coverage instrumentation was not verified even though test assertions passed.
> - S3SingleLargeFileE2ETest was intentionally not run because its default path generates a 20 GiB file.
> - The independent recovery-consistency run timed out after 900 seconds and produced no usable result.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,1318 @@
...
+     * <p>By the time a compaction event fires, the MANIFEST already reflects the post-compaction
+     * live SST set — a checkpoint captures that consistent state without needing a flush. In
+     * {@code wal} mode the WAL tail is included in the checkpoint, so un-flushed data is still
+     * recoverable. In {@code flush} mode un-flushed MemTable data written since the last sync is
+     * not captured, but that is acceptable here because the goal is to publish a MANIFEST that
+     * does not reference any cloud object that is about to be deleted.
+     *
+     * <p>Package-private for testability.
+     */
+    boolean syncMetadataSnapshotInline(CloudStorageProvider provider, String dbName) {
```
VGalaxies, line 2418: ### 🚨 High · Serialize metadata publication per database

#### 🧪 Reproduction path

1. Prepare two snapshots for one database whose manifests represent successive generations.
2. Block the older publication after capture.
3. Allow the newer snapshot to publish CURRENT and prune metadata.
4. Release the older publication and observe it overwrite CURRENT and prune the newer manifest.

Two upload-dispatch threads and the RocksDB compaction callback can concurrently enter syncMetadataSnapshotInline() for the same database. There is no per-database lock or generation check, while every invocation overwrites the same CURRENT key and then prunes metadata not belonging to its own snapshot.

> [!IMPORTANT]
> **Impact:** A slower, older snapshot can publish after a newer snapshot and delete the newer manifest. Disaster recovery can therefore roll back to stale data or find CURRENT pointing at metadata that another publication removed.

#### 🛠️ Suggested change

Serialize capture, publication, and pruning per database, and reject snapshots older than the last published RocksDB generation. Add a deterministic concurrency test with deliberately reordered uploads.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java
```diff
@@ -116,6 +145,119 @@ public void init() {
...
+    private void initCloudStorage() {
+        CloudStorageConfig cfg = cloudStorageSpringConfig.toCloudStorageConfig();
+        if (!cfg.isEnabled()) {
+            log.info("Cloud storage disabled (cloud.storage.enabled=false)");
+            return;
+        }
+        try {
+            CloudStorageProviderFactory.initialize(cfg);
+            String resolvedDataRoot =
+                    Paths.get(dataPath).toAbsolutePath().normalize().toString();
```
VGalaxies: ### 🚨 High · Support every configured HStore data root

#### 🔄 Contract mismatch

| Expected behavior | Actual behavior |
| --- | --- |
| Each configured data directory maps to stable relative cloud keys and restores files to that same directory. | All configured paths are interpreted as one literal filesystem path containing commas. |

HStore supports comma-separated app.data-path values and PartitionManager explicitly splits them. Cloud initialization instead passes the entire comma-separated string to Paths.get() and constructs one CloudStorageEventListener data root.

> [!IMPORTANT]
> **Impact:** For multi-directory stores, SST paths do not match the synthetic combined root. Upload keys fall back to container-specific absolute paths, hydration resolves them beneath a nonexistent comma-named directory, and the DLQ is persisted in the wrong location.

#### 🛠️ Suggested change

Parse every configured data root and resolve each database or SST against its actual matching root. Preserve that root association when constructing keys and hydration destinations.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java
```diff
@@ -116,6 +145,119 @@ public void init() {
...
+                     + "readMissHydration=true, "
+                     + "readMissGuardWindowMs={}, uploadRetryMaxAttempts={}, "
+                     + "uploadRetryInitialDelayMs={}, uploadRetryMaxDelayMs={})",
+                     cfg.getProvider(), resolvedDataRoot, storeScopePrefix,
+                     cfg.isStartupHydrationEnabled(),
+                     cfg.getReadMissGuardWindowMs(),
+                     cfg.getUploadRetryMaxAttempts(),
+                     cfg.getUploadRetryInitialDelayMs(),
+                     cfg.getUploadRetryMaxDelayMs());
+        } catch (Exception e) {
```
VGalaxies, line 265: ### 🚨 High · Do not silently disable explicitly enabled cloud storage

#### 🔄 Contract mismatch

| Expected behavior | Actual behavior |
| --- | --- |
| An invalid enabled cloud-storage configuration prevents the node from advertising healthy service. | Initialization errors are logged and startup continues with cloud callbacks absent. |

The catch-all around provider initialization, retry-queue creation, and listener registration logs every exception and lets PostConstruct complete. An unknown provider, missing S3 bucket, or later setup failure therefore leaves the node running without an active listener.

> [!IMPORTANT]
> **Impact:** Operators can enable cloud recovery and receive normal service while no SST or metadata is mirrored. The durability failure remains invisible until local storage is lost.

#### 🛠️ Suggested change

Fail startup or mark the node unhealthy when enabled cloud storage cannot initialize. If degraded local-only operation is required, make it an explicit configuration policy and expose its state through health checks and metrics.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
```diff
@@ -0,0 +1,864 @@
...
+                if (!toDelete.isEmpty()) {
+                    try {
+                        DeleteObjectsResponse deleteResp = s3Client.deleteObjects(
+                                DeleteObjectsRequest.builder()
+                                                   .bucket(bucket)
+                                                   .delete(software.amazon.awssdk.services.s3.model.Delete.builder()
+                                                                                                          .objects(toDelete)
+                                                                                                          .build())
+                                                   .build());
+                        totalDeleted += deleteResp.deleted().size();
```
VGalaxies, line 484: ### 🚨 High · Treat partial DeleteObjects responses as purge failures

#### 🔄 Contract mismatch

| Expected behavior | Actual behavior |
| --- | --- |
| A database purge succeeds only after all stale objects are deleted while the tombstone remains protective. | Partial deletion is reported as success and may remove the tombstone before stale objects. |

deletePrefix() counts deleteResp.deleted() but never inspects deleteResp.errors(). S3 can return HTTP success containing per-key failures, so the fallback runs only when the entire request throws. Database deletion also places the _DELETED tombstone inside the same prefix being purged.

> [!IMPORTANT]
> **Impact:** The tombstone can be deleted successfully while stale SST or metadata keys remain. A later database open then sees no deletion marker and can hydrate data from the deleted generation.

#### 🛠️ Suggested change

Inspect and retry every per-object error, report any keys that remain, and verify the prefix is empty. Preserve the tombstone until successful verification, preferably in a generation namespace outside the data being purged.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
```diff
@@ -0,0 +1,864 @@
...
+                return new IOException(message, ase);
+            }
+            return new CloudStorageNonRetryableException(message, ase);
+        }
+
+        // Client-side network/IO/timeout failures are retryable by default.
+        return new IOException("S3 " + operation + " failed for key='" + key + "'", e);
+    }
+
+    private static boolean isRetryableServiceFailure(AwsServiceException e) {
```
VGalaxies, line 993: ### ⚠️ Medium · Honor AWS retryable error codes

#### 🔄 Contract mismatch

| Expected behavior | Actual behavior |
| --- | --- |
| Errors classified as retryable by the AWS SDK remain eligible for provider and whole-file retries. | Retryable error codes with non-listed HTTP statuses are treated as permanent. |

isRetryableServiceFailure() checks throttling and selected HTTP statuses only. AWS SDK 2.33.8 explicitly marks PriorRequestNotComplete, RequestTimeout, RequestTimeoutException, and InternalError as retryable, but an HTTP 400 RequestTimeout is classified here as CloudStorageNonRetryableException.

> [!WARNING]
> **Impact:** Transient S3 failures bypass the configured whole-file retry policy and move an SST directly toward the DLQ, leaving it local-only until another compaction happens to retry it.

#### 🛠️ Suggested change

Use the AWS SDK retry-condition logic or include its retryable error-code set, with focused HTTP 400 RequestTimeout and PriorRequestNotComplete tests.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
```diff
@@ -0,0 +1,864 @@
...
+    }
+
+    @Override
+    public void downloadFile(String remoteKey, String localPath) throws IOException {
+        String fullKey = buildKey(remoteKey);
+        long startNs = System.nanoTime();
+        Path destinationPath = Paths.get(localPath);
+        try {
+
+            s3Client.getObject(
```
VGalaxies, line 649: ### 🚨 High · Publish downloaded files atomically

#### 🔄 Contract mismatch

| Expected behavior | Actual behavior |
| --- | --- |
| Only complete downloads become visible at RocksDB paths, and retries replace stale temporary files safely. | Downloads expose destination files before completion and existing partial artifacts block or bypass recovery. |

The S3 client writes directly to the supplied destination. Startup hydration skips any destination that already exists, while read-miss hydration reuses a fixed .hydrate path. AWS SDK ResponseTransformer.toFile does not replace an existing destination.

> [!IMPORTANT]
> **Impact:** A process termination or cleanup failure during download can leave a partial final SST that startup treats as complete, or a stale .hydrate file that permanently blocks retries. RocksDB recovery then fails or opens against corrupted data.

#### 🛠️ Suggested change

Always download to a unique sibling temporary file, validate successful completion, and atomically replace the destination. Remove stale temporary artifacts on failure and startup.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

install-dist/release-docs/LICENSE
```diff
@@ -444,28 +452,38 @@ non-Apache-2.0 license text is required for a bundled component.
     https://central.sonatype.com/artifact/io.netty/netty-all/4.1.42.Final -> Apache 2.0
     https://central.sonatype.com/artifact/io.netty/netty-all/4.1.44.Final -> Apache 2.0
     https://central.sonatype.com/artifact/io.netty/netty-all/4.1.61.Final -> Apache 2.0
+    https://central.sonatype.com/artifact/io.netty/netty-buffer/4.1.108.Final -> Apache 2.0
```
VGalaxies: ### 🚨 High · Regenerate the release dependency inventory

#### 📐 Measured result

| Metric | Observed | Required / target |
| --- | --- | --- |
| bundled artifacts absent at their resolved version | 17 | 0 |
| incorrect Netty 4.1.108 entries | 10 | 0 |

known-dependencies.txt adds ten Netty 4.1.126.Final artifacts, while LICENSE adds their 4.1.108.Final versions. LICENSE also omits commons-codec-1.17.1, http-auth-aws-eventstream-2.33.8, httpcore-4.4.16, retries-2.33.8, retries-spi-2.33.8, slf4j-api-1.7.36, and url-connection-client-2.33.8.

> [!IMPORTANT]
> **Impact:** The Apache binary-release LICENSE describes dependencies that are not bundled and omits dependencies that are bundled, blocking an accurate release.

#### 🛠️ Suggested change

Regenerate LICENSE from the resolved distribution, replace the ten incorrect Netty versions, add every omitted artifact, and compare the packaged JAR inventory against the release documents.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

docker/cloud-storage/scripts/test-graph-queries-and-sst.sh
```diff
@@ -0,0 +1,917 @@
...
+volumes:
+  hg-minio-data:
+  hg-pd-data:
+  hg-store0-data:
+  hg-store1-data:
+  hg-store2-data:
+YAML
+prepare_artifacts
+log "building local cloud-storage images..."
+docker compose -f "$COMPOSE_FILE" build --no-cache 2>&1 | grep -E "^(Building|FINISHED|\[|Successfully|Error)" || true
```
VGalaxies: ### 🚨 High · Abort the E2E workflow when image builds fail

#### 🔄 Contract mismatch

| Expected behavior | Actual behavior |
| --- | --- |
| Any PD or Store image build failure stops the workflow before containers start. | Every build failure is converted to success and cached images may be reused. |

The script enables pipefail but appends || true to docker compose build. It then starts services without --build, so cached cloud-storage-local images remain eligible for use.

> [!IMPORTANT]
> **Impact:** The workflow can pass against stale images after the pull request code failed to compile or package, producing a false recovery result.

#### 🛠️ Suggested change

Preserve and check the compose build exit status while filtering output, abort before compose up on failure, and record the image IDs used by the test.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

docker/cloud-storage/scripts/test-graph-queries-and-sst.sh
```diff
@@ -0,0 +1,917 @@
...
+
+# Stop a store, wipe its RocksDB state-machine data (db/ + metadata graph) while
+# preserving raft/ and snapshot/, then start it again. This models local
+# state-machine loss where the node still rejoins its raft group but must
+# re-hydrate its RocksDB data from cloud on open (pre-hydration path).
+wipe_store_rocksdb_state() {
+    local idx="$1"
+    local vol="${COMPOSE_PROJECT_NAME}_hg-store${idx}-data"
+    docker compose -f "$COMPOSE_FILE" stop "store${idx}" >/dev/null 2>&1 || true
+    docker run --rm --entrypoint /bin/sh -v "${vol}:/s" "$MINIO_MC_IMAGE" -c '
```
VGalaxies, line 481: ### 🚨 High · Verify that the recovery test actually wipes local state

#### 🔄 Contract mismatch

| Expected behavior | Actual behavior |
| --- | --- |
| Recovery verification proceeds only after the active Store volume's RocksDB state was demonstrably removed. | Destructive setup failures are ignored, so unchanged local data can satisfy the recovery assertion. |

Stopping the store, mounting and deleting the volume contents, and restarting the store all ignore failures. A wrong volume name can create and wipe a new empty volume while the real Store volume and its data remain untouched.

> [!IMPORTANT]
> **Impact:** The before-and-after vertex counts can match because the original local RocksDB data was never removed, allowing the recovery test to pass without reading anything from cloud storage.

#### 🛠️ Suggested change

Fail closed on stop, volume mount, deletion, and restart errors. Assert that expected non-Raft directories existed before deletion and are absent before restarting the Store.

---
<sub>🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**</sub>

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java
vaijosh: Addressed

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
vaijosh: Addressed

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java
vaijosh: Addressed

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
vaijosh: Addressed

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
vaijosh: Addressed

install-dist/release-docs/LICENSE
vaijosh: Addressed

docker/cloud-storage/scripts/test-graph-queries-and-sst.sh
vaijosh: Addressed

docker/cloud-storage/scripts/test-graph-queries-and-sst.sh
vaijosh: Addressed

vaijosh:
> ### 🚨 Review summary
> Important
> 
> The cloud recovery path has blocking consistency, configuration, deletion-safety, release-metadata, and verification gaps.
> 
> #### 📊 Risk dashboard
> Signal	Result
> 🚦 Review gate	**Blocked**
> 🔎 Actionable findings	**9** (8 High, 1 Medium)
> 🧪 Verification coverage	**5** checks
> ⚠️ Confirmed / limited checks	**2 / 1**
> #### 🔬 Coverage details
> ⚠️ **S3 failure semantics** — Confirmed
> **Trace:** S3CloudStorageProvider upload, download, delete, and exception-classification paths, CloudStorageEventListener purge and hydration flows, AWS SDK 2.33.8 retry-code and file-transformer behavior
> 
> **Conclusion:** Confirmed ignored per-object deletion errors, retryable error-code misclassification, and non-atomic download recovery risks; lower-impact lifecycle leads were dropped.
> 
> ⚠️ **Release and E2E integrity** — Confirmed
> **Trace:** install-dist release LICENSE and known-dependencies inventory, cloud-storage Docker artifact build and recovery workflow, Maven provider packaging and test activation
> 
> **Conclusion:** Confirmed the release dependency inventory mismatch and E2E false-green paths caused by suppressed build and wipe failures.
> 
> 🟡 **Focused unit suites** — Limited
> **Trace:** hg-store-common cloud configuration, provider, and factory tests, hg-store-cloud-s3 exception-classification tests, hg-store-node listener, retry, metrics, tracker, configuration, and callback tests, hg-store-core BusinessHandlerImplTest
> 
> **Conclusion:** Surefire reports recorded 184 tests with 0 failures, 0 errors, and 0 skipped. The suites do not cover concurrent metadata publication, comma-separated data roots, or partial S3 batch deletion.
> 
> ✅ **Maven reactor** — Clear
> **Trace:** Full 44-project reactor with the cloud-s3 profile, hugegraph-store module ordering and dependency graph
> 
> **Conclusion:** Maven validate completed successfully for all 44 reactor projects; the suspected duplicate cloud-s3 module activation did not reproduce.
> 
> ✅ **Shell syntax** — Clear
> **Trace:** pd-entrypoint.sh, store-entrypoint.sh, test-graph-queries-and-sst.sh
> 
> **Conclusion:** bash -n returned successfully for all three changed shell entrypoints and workflows.
> 
> Warning
> 
> **Verification limits**
> 
> * No live Docker and MinIO recovery drill was run, so cluster recovery and purge behavior were not exercised end to end.
> * The test environment used Java 17; the legacy JaCoCo agent emitted unsupported class-version instrumentation errors, so coverage instrumentation was not verified even though test assertions passed.
> * S3SingleLargeFileE2ETest was intentionally not run because its default path generates a 20 GiB file.
> * The independent recovery-consistency run timed out after 900 seconds and produced no usable result.
> 
> 🤖 Codex review · **GPT-5.6 Sol** · effort: **xhigh**

Hi @VGalaxies.
Thanks for reviewing the changes. I have addressed all the review comments. Can you please review it again?

vaijosh:
@VGalaxies 
I am not sure why code coverage check is failing. 
We have
The Diff has  94.79% coverage  (agains target 42.32%)  [codecov/patch]**(https://github.com/apache/hugegraph/pull/3081/checks?check_run_id=89150435397)Successful in 1s — 94.79% of diff hit (target 42.32%)**

Not sure why the coverage is suddenly dropped at project level  to 30.60% (-11.72%) compared to e960cc5 [codecov/project](https://github.com/apache/hugegraph/pull/3081/checks?check_run_id=89150433373)Failing after 1s — 30.60% (-11.72%) compared to e960cc5

hugegraph-store/pom.xml
```diff
@@ -302,5 +307,13 @@
...
             </build>
         </profile>
+
+        <!-- Build the S3 cloud-storage plugin:  mvn install -P cloud-s3 -DskipTests -->
+        <profile>
+            <id>cloud-s3</id>
+            <modules>
+                <module>hg-store-cloud-s3</module>
+            </modules>
+        </profile>
```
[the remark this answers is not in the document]
vaijosh: Addressed in latest commit

hugegraph-store/hg-store-common/src/main/java/org/apache/hugegraph/store/cloud/CloudStorageProviderFactory.java
```diff
@@ -0,0 +1,212 @@
...
+     *
+     * @param config cloud storage configuration
+     * @return the initialized provider, or {@code null} when
+     *         {@link CloudStorageConfig#isEnabled()} is {@code false}
+     * @throws IllegalArgumentException if no provider matching {@code config.getProvider()}
+     *                                  was found on the classpath
+     */
+    public static synchronized CloudStorageProvider initialize(CloudStorageConfig config) {
+        if (!config.isEnabled()) {
+            // Disabling cloud storage must deactivate any currently active provider: otherwise a
```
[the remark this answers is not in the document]
vaijosh: Addressed in latest commit

hugegraph-store/hg-store-rocksdb/src/main/java/org/apache/hugegraph/rocksdb/access/RocksDBSession.java
```diff
@@ -693,6 +764,70 @@ public void saveSnapshot(String snapshotPath) throws DBStoreException {
...
+            checkpoint.createCheckpoint(tempDir);
+        } catch (final Exception e) {
+            try {
+                FileUtils.deleteDirectory(new File(tempDir));
+            } catch (IOException ignore) {
+                // best-effort cleanup of a partial checkpoint
+            }
+            log.error("Fail to create metadata checkpoint at {}", tempDir, e);
+            throw new DBStoreException(
+                    String.format("Fail to create metadata checkpoint at %s", tempDir));
```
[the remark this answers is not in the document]
vaijosh: Addressed in latest commit

hugegraph-store/hg-store-rocksdb/src/main/java/org/apache/hugegraph/rocksdb/access/SessionOperatorImpl.java
```diff
@@ -208,7 +208,14 @@ public void deleteRange(byte[] keyFrom, byte[] keyTo) throws DBStoreException {
     @Override
     public byte[] get(String table, byte[] key) throws DBStoreException {
         try (CFHandleLock cf = this.getLock(table)) {
-            return rocksdb().get(cf.get(), key);
+            byte[] value = rocksdb().get(cf.get(), key);
+            if (value != null) {
+                return value;
+            }
+            if (RocksDBFactory.getInstance().onReadMiss(this.session, table, key)) {
```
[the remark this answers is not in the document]
vaijosh: Addressed in latest commit

hugegraph-store/hg-store-cloud-s3/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+          Required in docker/cloud-storage runtime: CloudStorageEventListener hooks into
+          RocksDB event callback types (FlushJobInfo/TableFileCreationInfo).
+          Keep this aligned with hg-store-rocksdb to avoid NoClassDefFoundError.
+        -->
+        <dependency>
+            <groupId>org.rocksdb</groupId>
+            <artifactId>rocksdbjni</artifactId>
+            <version>7.7.3</version>
+            <scope>runtime</scope>
+        </dependency>
```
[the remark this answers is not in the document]
vaijosh: Addressed

hugegraph-store/hg-store-common/src/main/java/org/apache/hugegraph/store/cloud/CloudStorageProviderFactory.java
```diff
@@ -0,0 +1,217 @@
...
+            } catch (IOException e) {
+                log.warn("Error closing previous cloud storage provider", e);
+            }
+        }
+
+        provider.init(config);
+        activeProvider = provider;
+        log.info("Cloud storage provider '{}' initialized. bucket={}",
+                 name, config.getProviderProperties().getOrDefault("bucket", "N/A"));
+        return provider;
```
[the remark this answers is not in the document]
vaijosh: Addressed

vaijosh:
> ## Pull request overview
> Copilot reviewed 58 out of 60 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (1)
> **hugegraph-store/hg-store-rocksdb/src/test/java/org/apache/hugegraph/rocksdb/access/RocksDBSessionTest.java:24**
> 
> * This test file now has unused imports (e.g., static import `fail`, `java.nio.file.Files/Path`, and `org.junit.Test`). In Java, unused imports fail compilation, and this class currently has no active `@Test` methods (the annotations are commented out), so these imports should be removed (or the tests re-enabled).

Implemented this review comment. Removed unused imports.

vaijosh:
Hi @VGalaxies @imbajin 
Please don't merge the PR. 
I can see that JDK17 changes are in master branch. So I will need to test my changes with JRE17. I will update here once my testing is done.

vaijosh:
> Hi @VGalaxies @imbajin Please don't merge the PR. I can see that JDK17 changes are in master branch. So I will need to test my changes with JRE17. I will update here once my testing is done.

Please ignore. I was bit confused due to merge conflict I got in docker/README.md due to changes in 9aaad115. It seems  only reference to JDK17 was removed from README.md. But we are not yet moved to JDK17.

vaijosh:
@imbajin @VGalaxies 
I have committed few more changes, fixed multiple review comments by Claude Code and Github Copilot.

Tested following scenarios 
1. SST and metadata synching is happening properly when Flush happens - End to End test and Manual steps in README.md
2. During Compaction the new files are synced and few files are deleted.  - End to End tests.
3. During disk failures, the rehydration happens via Cloud storage. - End to End test and Manual steps in README.md
4. During delete scenarios the TOMBSTONE is written and when delete is complete the data is deleted. - End to End test
5. Load IBM AML large data  ( Around 677609 Vertices and 17843793 Edgest) - aml_sql_showcase_hugegraph.ipynb

vaijosh:
@VGalaxies @imbajin 
Is anyone currently reviewing this PR? I understand it's a substantial change and will take some time to review, which is completely fine. I just wanted to check in to ensure it hasn't stalled and is still on the radar.

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
```diff
@@ -0,0 +1,1100 @@
...
+     *   S3 multipart part 1/41 uploaded: size=512.0 MB | elapsed=6 230 ms | throughput=82.18 MB/s
+     *   S3 multipart part 2/41 uploaded: size=512.0 MB | elapsed=6 050 ms | throughput=84.63 MB/s
+     *   ...
+     *   S3 multipart upload completed: key=hugegraph/hgstore-data/000099.sst | parts=41
+     * </pre>
+     *
+     * <p>If any part fails the multipart upload is aborted (to avoid incomplete-upload storage
+     * charges) and an {@link IOException} is thrown.
+     */
+    private void uploadMultipart(java.nio.file.Path path, long fileSize,
```
bitflicker64, line 736: ⚠️ This multipart driver and its retry layer are re-implementing the SDK you already depend on.

`init()` (L176-248) builds the `S3Client` with no `overrideConfiguration(...)`, so AWS SDK v2 defaults are live: every `uploadPart` is already retried 3× with full-jitter exponential backoff on 5xx, throttling and IO errors. `uploadOnePartWithRetry` (L880) then wraps that in a second loop, so a part can be attempted up to 12 times before anyone hears about it.

Smallest fix, no new dependency: delete `uploadOnePartWithRetry`, `retryBackoffMs`, `sleepQuietly`, `MAX_PART_UPLOAD_RETRIES` and `MAX_RETRY_BACKOFF_MS`, and set the count you want on the client instead — `ClientOverrideConfiguration.builder().retryPolicy(RetryPolicy.builder().numRetries(n).build())`.

Lazier still: `software.amazon.awssdk:s3-transfer-manager` from the same 2.33.8 BOM, over an `S3AsyncClient` built with `.multipartEnabled(true)`, collapses `uploadFile` to

```java
tm.uploadFile(UploadFileRequest.builder()
        .source(path)
        .putObjectRequest(b -> b.bucket(bucket).key(fullKey))
        .build())
  .completionFuture().join();
```

That deletes `uploadMultipart`, `uploadOnePart`, `openBoundedPartStream`, `LimitedInputStream` and the three `cloud.storage.s3.multipart-part-retry-*` knobs in both `application.yml` files — roughly 400 of this file's 1100 lines.

Separately, `abortStaleMultipartUploads` (L259) can go too: its own javadoc already tells operators to configure the S3 `AbortIncompleteMultipartUpload` lifecycle rule, which is the platform doing this for free and doesn't need an init-time `ListMultipartUploads` sweep.

.licenserc.yaml
```diff
@@ -66,6 +66,7 @@ header: # `header` section is configurations for source codes license header.
     - '**/*.log'
     - '**/*.txt'
     - '**/*.json'
+    - '**/*.ipynb'
```
bitflicker64, line 69: ⚠️ This repo-wide header exemption exists to land one file that isn't part of the feature.

The only `.ipynb` in the PR is `docker/cloud-storage/notebooks/aml_sql_showcase_hugegraph.ipynb` — 1720 lines, 43 cells of anti-money-laundering Gremlin traversals that download `LI-Large_Trans.csv` / `HI-Large_accounts.csv` from Kaggle. It drives the Gremlin/REST API; it never touches SST offload, hydration or the cloud provider. It's also absent from the PR description: "Infrastructure & Testing" commits to the MinIO compose stack and `test-graph-queries-and-sst.sh`, both of which are here and both of which do cover the end-to-end durability check.

Drop the notebook and revert this line. As written it permanently exempts every future notebook in the tree from the ASF header check, which is a real cost paid for a demo that belongs in hugegraph-doc or a gist.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,3150 @@
...
+                    + linkEx.getMessage(), linkEx);
+        }
+    }
+
+    /**
+     * Blocks the calling (RocksDB flush/compaction) thread while the pending-upload backlog exceeds
+     * {@link #backpressureHighWatermark}, up to {@link #BACKPRESSURE_MAX_WAIT_MS}. This is the
+     * durability-tier backpressure: it keeps at-risk local-only data bounded.
+     */
+    private void applyBackpressure(String dbName) {
```
bitflicker64, line 1858: ⚠️ Backpressure is inert as shipped, and nobody asked for it.

`upload-backpressure-high-watermark` is `0` in both `application.yml` files, and this method and `dlqEnqueueRateBacklog` (L1910) both return immediately when it's `<= 0`. Nothing in #3079 or the PR description asks for ingestion throttling, and the yaml comment this PR adds argues against switching it on: "the throttle parks RocksDB's flush/compaction thread (up to 30s per SST event), which under a sustained cloud outage can stall memtable flushes / stop writes for the partition."

Deleting the path removes `applyBackpressure`, `pendingUploadBacklog`, `dlqEnqueueRateBacklog`, the `dlqRateLock` monitor plus the four `lastDlqRateSample*` / `dlqRateBacklogContribution` fields, `BACKPRESSURE_MAX_WAIT_MS`, `BACKPRESSURE_POLL_MS`, `DLQ_ENQUEUE_RATE_WINDOW_MS`, the `backpressureHighWatermark` constructor parameter threaded through two constructors, `CloudUploadRetryQueue.getDlqEnqueuedTotal()`, and a config key in two files.

The `cloud_storage_retry_queue_size` and `cloud_storage_dlq_persistence_healthy` gauges you already register surface the same "durability is degrading" signal without parking a RocksDB flush thread. Add the throttle back when an operator asks for one and can say what it should do.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java
```diff
@@ -312,6 +636,113 @@ public class RocksdbConfig {
...
+        }
+
+        /**
+         * Reads all {@code cloud.storage.<provider>.*} keys from the Spring Environment
+         * and returns them as a flat map with the provider sub-prefix stripped.
+         *
+         * <p>For example, with {@code provider=s3}, the YAML key
+         * {@code cloud.storage.s3.bucket} becomes {@code bucket} in the returned map.
+         */
+        private Map<String, String> readProviderProperties() {
```
bitflicker64, line 723: ⚠️ Spring Boot already does this in one call.

This method casts to `AbstractEnvironment`, streams `getPropertySources()`, filters to `EnumerablePropertySource`, flat-maps `getPropertyNames()`, string-prefix-filters, de-dupes and re-reads each key. `hg-store-node` is on spring-boot 2.5.14, so:

```java
Map<String, String> props = Binder.get(environment)
        .bind("cloud.storage." + provider, Bindable.mapOf(String.class, String.class))
        .orElseGet(Collections::emptyMap);
```

Same kebab-case keys the `S3CloudStorageConfig.KEY_*` constants expect (`bucket`, `access-key`, ...), about 25 lines down to 3, and the `AbstractEnvironment` / `EnumerablePropertySource` imports go with it.

Worth noting it isn't only shorter: `Binder` applies relaxed binding, so `CLOUD_STORAGE_S3_SECRET_KEY` and friends resolve. A raw `key.startsWith("cloud.storage.s3.")` scan over property *names* won't match environment-variable-style names, which is exactly how credentials tend to arrive in the container deployment this PR ships.

bitflicker64 (commented):
Blocking: no. Summary: The SPI/factory shape is out of scope here because issue #3079 explicitly asks for it, but the biggest simplification left is inside the S3 provider — `init()` sets no `overrideConfiguration`, so the AWS SDK is already retrying every `uploadPart` with jittered backoff, and the hand-rolled multipart driver plus its second retry layer (~400 of that file's 1100 lines) can go. Evidence: local diff of the exact head against the PR base (`git diff 401e627c30a7a0e0db5548a9431b6acfe82e43a1 eddcd868cc97d574e75f0a5709899b0a4600b222`, 62 files / +24239 / -84); read `S3CloudStorageProvider.init()` L176-248 and the multipart path L691-956, `CloudStorageEventListener.applyBackpressure`/`dlqEnqueueRateBacklog` L1858-1942 against `upload-backpressure-high-watermark: 0` in both `application.yml` files, `AppConfig.readProviderProperties()` L723-737 against spring-boot 2.5.14 in `hg-store-node/pom.xml`, and the `**/*.ipynb` exemption added at `.licenserc.yaml:69`. Note: the GitHub Files API reports `additions=0` for `hg-store-rocksdb/**` and `pom.xml` at this diff size — those files do carry changes; the local diff is authoritative. All four anchors verified against right-side added lines in the exact-head diff.

hugegraph-store/hg-store-cloud-s3/src/main/java/org/apache/hugegraph/store/cloud/s3/S3CloudStorageProvider.java
vaijosh: Addressed

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
vaijosh: Addressed

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/AppConfig.java
vaijosh: Addressed.

.licenserc.yaml
vaijosh: Addressed

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
```diff
@@ -0,0 +1,3026 @@
...
+     */
+    static final String DB_TOMBSTONE_SUFFIX = "_DELETED";
+
+    /**
+     * Convenience constructor with startup hydration enabled and default read-miss guard window.
+     *
+     * @param dataRoots configured store data roots (typically parsed from comma-separated
+     *                  {@code app.data-path})
+     */
+    public CloudStorageEventListener(List<String> dataRoots) {
```
bitflicker64: ⚠️ Seven public constructors, and production uses exactly one of them: `AppConfig.initCloudStorage()` calls the full 7-arg form with a `Tuning` built from both builder methods. The other six overloads have only test callers (47 in `CloudStorageEventListenerTest`, 12 in `CloudStorageIntegrationTest`, 2 in `CloudRecoveryIntegrationTest`). Meanwhile the two tuning knobs are settable three ways: `Tuning.builder()`, the constructor path, and the `set*` methods the javadoc keeps "for tests and runtime overrides".

Requested change: one constructor that takes `metadataSyncDebounceMs` and `metadataSyncMaxUnpublished` as plain parameters. That keeps the "fully configured the moment it is constructed" guarantee this javadoc argues for, and deletes the six convenience overloads plus the `Tuning`/`Builder` pair, roughly 150 lines. Tests spell out the defaults or share one helper in the test tree. `CloudUploadRetryQueue` has the same pattern at smaller scale (four constructors, one production caller) and the same fix applies.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudSyncTracker.java
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * <p>{@link Roaring64NavigableMap} is not thread-safe, so all access to a per-DB bitmap is
+ * synchronized on the bitmap instance.
+ */
+public final class CloudSyncTracker {
+
+    /** Holds the bitmap and the epoch it was created under. */
+    private static final class DbState {
+        final long epoch;
+        final Roaring64NavigableMap bitmap;
```
bitflicker64: ⚠️ This bitmap is used as a plain set: the only operations on it anywhere in the class are `addLong`, `removeLong`, `contains` and `getLongCardinality`. Because `Roaring64NavigableMap` is not thread-safe, every access needs its `synchronized (state.bitmap)` block (five of them), and the PR adds `org.roaringbitmap:RoaringBitmap:0.9.38` as a new direct dependency to `hg-store-node/pom.xml` for it.

Requested change: `final Set<Long> confirmed = ConcurrentHashMap.newKeySet();` in `DbState`. Same four operations (`add`/`remove`/`contains`/`size`), all five synchronized blocks go away (the `compute`-based epoch check in `markBit` is what actually guards the clearDb race, and it stays), and the pom entry is deleted. Roaring compactness earns its keep at millions of entries; live SST files per DB number in the thousands, so a boxed set costs well under a megabyte.

bitflicker64 (commented):
Blocking: no. Summary: the rework since the last round is real (the multipart driver, the inert backpressure path, the notebook and the property-source scan are all gone), and the biggest simplification left is the construction surface of `CloudStorageEventListener`: seven public constructors plus a `Tuning`/`Builder` pair plus equivalent setters, when production has exactly one caller using the full path, about 150 lines that one constructor with two extra parameters replaces. Second: `CloudSyncTracker` uses a `Roaring64NavigableMap` as a plain set, which costs a new `RoaringBitmap` dependency in hg-store-node and five `synchronized` blocks that `ConcurrentHashMap.newKeySet()` makes unnecessary. Evidence: reviewed the full diff at head f5b285d (60 files, +21633/-83) with the head fetched locally; `git grep 'new CloudStorageEventListener('` shows one production call site (AppConfig) against 61 test call sites; CloudSyncTracker touches the bitmap only via addLong/removeLong/contains/getLongCardinality. The SPI/factory shape is not raised: issue #3079 explicitly asks for an SPI-based extensible architecture, and the retry queue's DLQ persistence is data-loss protection, not scaffolding.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudStorageEventListener.java
vaijosh: Thanks @bitflicker64. I have addressed this.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/cloud/CloudSyncTracker.java
vaijosh: Thanks @bitflicker64. I have addressed this.

PR #3082, opened 2026-07-06
refactor(server): disable legacy master-worker scheduler logic

Tsukilc:
## Summary

This PR soft-disables the pre-PD master-worker task scheduling path while keeping old configs and task data upgrade-safe. Scheduler selection is now backend-driven: `hstore` uses `DistributedTaskScheduler`, and other backends use the local `StandardTaskScheduler`.

## Before -> After

Before:
- `task.scheduler_type` selected the scheduler mode.
- The local scheduler used `server_info`, heartbeat, server load, and master/worker role election.
- Normal tasks moved through `SCHEDULING -> SCHEDULED -> QUEUED -> RUNNING`, with the master assigning tasks to workers.

After:
- `task.scheduler_type` is deprecated and ignored; old config files only emit a warning.
- Non-HStore graphs use a direct local path: `QUEUED -> local executor`, with no master-worker assignment.
- HStore graphs use the distributed scheduler, coordinated by task status and task locks instead of legacy `server_info`.
- `ServerInfoManager` and role-election startup are kept only as soft-disabled compatibility shims.

## Compatibility and Impact

- Existing completed task records and task results remain readable and deletable.
- Legacy `SCHEDULING` / `SCHEDULED` tasks from the old master-worker path are preserved in storage but are not restored or executed automatically. Operators should drain or cancel old in-flight master-worker tasks before upgrading if those tasks still matter.
- `server.role_election`, `server.id`, and `task.scheduler_type` no longer control scheduling behavior.
- Distributed delete/cancel handling is minimally hardened: `DELETING` is used as a retryable tombstone, running task deletion waits for the runner lifecycle, and save-after-delete resurrection is guarded.

## Follow-up Cleanup

This PR is the first soft-disable step, not the final physical deletion. Some deprecated config, role-election, and `server_info` code is intentionally left as compatibility scaffolding for this transition. After the soft-disable behavior is verified as stable, follow-up PRs should mark and remove the remaining dead paths, especially:

- unused `server_info` persistence, heartbeat, load, and worker-picking code paths;
- role-election listener/state-machine wiring that no longer participates in scheduling;
- deprecated scheduler config transfer/removal leftovers;
- transition-only compatibility tests once the legacy path is fully removed.

The intended end state is to fully remove the old master-worker scheduling system after this compatibility window, while keeping the PD/HStore distributed scheduler as the only distributed scheduling path.

### DIFF
--- a/hugegraph-cluster-test/hugegraph-clustertest-dist/src/assembly/static/conf/hugegraph.properties.template
+++ b/hugegraph-cluster-test/hugegraph-clustertest-dist/src/assembly/static/conf/hugegraph.properties.template
@@ -45,7 +45,6 @@ store=hugegraph
 pd.peers=$PD_PEERS_LIST$
 
 # task config
-task.scheduler_type=local
 task.schedule_period=10
 task.retry=0
 task.wait_timeout=10
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/config/ServerOptions.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/config/ServerOptions.java
@@ -556,9 +556,9 @@ public class ServerOptions extends OptionHolder {
     public static final ConfigOption<String> SERVER_ID =
             new ConfigOption<>(
                     "server.id",
-                    "The id of hugegraph-server.",
-                    disallowEmpty(),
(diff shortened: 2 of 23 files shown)


imbajin (approved):
TODO: cleanup legacy code/configs in phrase2

PR #3096, opened 2026-07-19
feat: support secure Hubble monitoring targets

imbajin:
## Purpose

Provide the minimum Server/PD compatibility and security boundary needed by Hubble native
monitoring and GraphSpace administration. Matching Hubble work:
https://github.com/apache/hugegraph-toolchain/pull/743

## Core before → after

| Area | Before | After |
|---|---|---|
| Store monitoring | A Store registration address could become an HTTP monitoring target even if it was a gRPC address or malformed. | Publish only a validated `host + rest.port` REST target (IPv4, hostname, or bracketed IPv6). |
| Hubble Gremlin query | The two different `/gremlin` endpoints were easy to confuse. | Hubble logs in once and calls HugeGraph **REST** `/gremlin` with its server-side token. |
| Direct Gremlin Server access | Its token boundary was unclear. | The separate TinkerPop Gremlin Server HTTP endpoint does **not** accept HugeGraph tokens in this PR; existing account-password authentication remains. |
| GraphSpace authorization | Group, target, membership, and grant operations lacked a complete GraphSpace boundary. | Scoped operations validate GraphSpace, role, and referenced-resource ownership; cross-space references are rejected. |
| Secrets and auth metadata | PD config text could expose `secretKey`; generated role grants were editable; token cache/log handling was broad. | Redact `secretKey` and tokens, hide generated grants, and evict only the logged-out token's role-cache entry. |
| Graph configuration | Local graph limits could be copied too broadly into PD configs. | Propagate only explicit safe cache/buffer limits, without overwriting supplied values. |

```text
Hubble browser
  │  log in once
  ▼
HugeGraph REST API `/gremlin` ── token ──> execute query                 ✅

Direct client ──────────────────> TinkerPop Gremlin Server HTTP `/gremlin`
                                    └─ HugeGraph token unsupported here  ✅
```

## Scope

This PR does not add PD/Store credential verification, a token issuer, token support for the
separate TinkerPop Gremlin Server endpoint, alerting, notifications, or rolling-upgrade support.

## Verification — head `f3e0ceff7`

- Server auth/scoping: **35 passed**; PD Store discovery: **1 passed**.
- Affected root reactors passed Checkstyle and RAT; `git diff --check` passed.
- Independent read-only review found no blocking correctness or security issue.


imbajin (commented):
Blocking: yes. Summary: The scoped auth API has proxy-boundary, standalone-mode, and graph-space isolation regressions, and monitoring discovery can emit invalid HTTP targets. Evidence: six independent static review lanes plus latest-head GitHub checks.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/AuthManager.java
```diff
@@ -53,6 +53,11 @@ public interface AuthManager {
 
     HugeGroup deleteGroup(Id id);
 
+    default HugeGroup deleteGroup(String graphSpace, Id id) {
```
imbajin, line 60: ‼️ These scoped overloads are not overridden by `HugeGraphAuthProxy.AuthManagerProxy`, which is the manager returned to REST calls when authentication is enabled. Scoped group deletion therefore always reaches this default exception, while target/belong/access calls fall back to unscoped proxy methods and lose the requested graph space. Please override and delegate every scoped overload in the proxy while preserving its permission, creator, and cache behavior, and cover non-default graph-space CRUD through the authenticated proxy.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/GraphSpaceGroupAPI.java
```diff
@@ -0,0 +1,299 @@
...
+
+    static void ensureManager(GraphManager manager, String graphSpace) {
+        ensurePdModeEnabled(manager);
+        ensureAuthManager(manager, graphSpace);
+    }
+
+    static void ensureAuthManager(GraphManager manager, String graphSpace) {
+        E.checkArgument(manager.graphSpace(graphSpace) != null,
+                        "The graph space '%s' does not exist", graphSpace);
+        checkManagerPermission(manager.authManager(), graphSpace,
```
imbajin: ‼️ This gate is also used by the existing target, belong, and access routes, but standalone `StandardAuthManager` returns `false` for both `isAdminManager()` and `isSpaceManager()`. As a result, even the built-in admin is denied for `/graphspaces/DEFAULT/auth/{targets,belongs,accesses}` after upgrade. Please apply this manager-role gate only to the V2/PD mode or provide equivalent standalone admin semantics, with REST regressions for standalone DEFAULT auth enabled and disabled.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/GraphSpaceGroupAPI.java
```diff
@@ -0,0 +1,299 @@
...
+                               HugeGraphAuthProxy.username());
+    }
+
+    static void checkBelongReferences(AuthManager authManager,
+                                      String graphSpace,
+                                      HugeBelong belong) {
+        if (!(authManager instanceof StandardAuthManagerV2)) {
```
imbajin: ‼️ In authenticated requests the supplied manager is `AuthManagerProxy`, so this concrete-type check returns before validating membership or group scope. A space manager can consequently reference a user who is not a member of this space, or a group from another space, when creating belongs/accesses; lower-level metadata checks only verify global existence. Please expose scope-aware validation through `AuthManager` and delegate it through the proxy instead of gating on the concrete class, then add cross-space REST tests.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/AccessAPI.java
```diff
@@ -114,20 +141,35 @@ public String list(@Context GraphManager manager,
...
-            Id id = UserAPI.parseId(target);
-            belongs = manager.authManager().listAccessByTarget(id, limit);
+            accesses = authManager.listAccessByTarget(graphSpace, target,
+                                                       -1L);
         } else {
-            belongs = manager.authManager().listAllAccess(limit);
+            accesses = authManager.listAllAccess(graphSpace, -1L);
         }
-        return manager.serializer().writeAuthElements("accesses", belongs);
+        accesses = accesses.stream()
```
imbajin, line 168: ‼️ Filtering only on `access.graphSpace()` includes the built-in role accesses stored in the same space. Those records do not resolve to a scoped business group, yet `checkScopedGroupReference()` treats a missing group as valid and GET/DELETE do not call it, so a space manager can enumerate or delete default SPACE/SPACE_MEMBER/analyst/observer authorization links. Please require a verified scoped-group source for this business API and keep built-in role metadata hidden and immutable; cover list/get/update/delete with a built-in-role access.

hugegraph-pd/hg-pd-service/src/main/java/org/apache/hugegraph/pd/service/SDConfigService.java
```diff
@@ -215,34 +215,14 @@ private Set<String> getStoreAddresses() {
...
-                String hostPart =
-                        host.contains(":") && !host.startsWith("[") ? "[" + host + "]" : host;
-                address = hostPart + ":" + port.get().trim();
-            }
-        } catch (Throwable t) {
-            log.error("Failed to extract the REST address of store, cause by:", t);
-        }
-        return address;
+        String restAddress = StoreRestAddressUtil.getRestAddress(store);
+        return restAddress != null ? restAddress : address;
```
imbajin: ⚠️ `StoreRestAddressUtil` returns null when `rest.port` is absent or invalid, but this fallback publishes the Store gRPC address as a monitoring target while the discovery scheme remains HTTP. Prometheus will then request `/actuator/prometheus` from a gRPC port, and this also disagrees with `StoreAPI`, which exposes a null REST address. Please omit stores without a validated REST endpoint (or only fall back when the address is known to be HTTP) and add missing/invalid-port service-discovery tests.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/auth/HugeTarget.java
```diff
@@ -226,10 +246,15 @@ public Map<String, Object> asMap() {
 
         Map<String, Object> map = new HashMap<>();
 
+        map.put(Hidden.unHide(P.GRAPHSPACE), this.graphSpace);
```
imbajin, line 249: ⚠️ New nodes now persist `graphspace` (and optionally `target_description`) in shared target metadata, but the previous `HugeTarget.property()` throws `AssertionError` for both unknown keys. During a rolling upgrade, an older server reading a target written by a new server can therefore fail target loading and dependent authorization. Please version or feature-gate the new persisted format, avoid writing fields that old readers cannot accept where the scope can be derived from the metadata namespace, or explicitly block mixed-version operation; add an old-reader/new-writer compatibility test.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/WsAndHttpBasicAuthHandler.java
```diff
@@ -103,28 +107,21 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
-                } catch (IllegalArgumentException iae) {
-                    sendError(ctx, msg);
-                    return;
-                }
-                String authorization = new String(userPass,
-                                                  StandardCharsets.UTF_8);
-                String[] split = authorization.split(":");
-                if (split.length != 2) {
+                final Map<String, String> credentials = new HashMap<>();
+                if (header.startsWith(BASIC_AUTH_PREFIX)) {
```
imbajin: ⚠️ HTTP authentication scheme names are case-insensitive, but these `startsWith("Basic ")` / `startsWith("Bearer ")` checks reject valid `basic` or `bearer` variants with 401. Please split the scheme from credentials, compare the scheme with `equalsIgnoreCase`, preserve the credential payload unchanged, and add case-variant tests.

imbajin: Fixed the standards issue for Basic by parsing the scheme separately and comparing it case-insensitively. Bearer handling on the internal Gremlin HTTP endpoint has been removed from this PR; Bearer remains on the public REST API to avoid duplicate authentication.

zyxxoo (approved):

PR #3100, opened 2026-07-20
fix(server): skip unsafe count optm for nested predicates

### LINKED ISSUE #3085: [Bug] Wrong result for where(count().is()
### Bug Type (问题类型)

gremlin (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version:
  - core: 1.7.0
  - gremlin: 3.5.1
  - api: 0.71.0.0
- Docker Image: `hugegraph/hugegraph:latest`
- Backend: default Docker standalone backend
- Data Size: 2 vertices, 1 edge

HugeGraph returns wrong results when a `where()` child traversal uses `count().is(...)`.

Minimal graph:
(a:A)-[:E]->(b:A)

Bug-triggering query:
g.V('a')
 .where(__.both('E').count().is(not(outside(1,18).and(gte(0)))))
 .count()

Expected:
[1]

Actual:
[0]

The dual predicate is also wrong:
g.V('a')
 .where(__.both('E').count().is(outside(1,18).and(gte(0))))
 .count()

Expected:
[0]

Actual:
[1]

For count = 1, outside(1,18) is false, so outside(1,18).and(gte(0)) should be false, and not(...) should be true.

```javascript
**Vertex/Edge example**



curl -fsS --compressed -X DELETE \
  "http://127.0.0.1:8080/graphs/hugegraph/clear?confirm_message=I'm%20sure%20to%20delete%20all%20data"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"id":"a","label":"A","properties":{}}' \
  "http://127.0.0.1:8080/graphs/hugegraph/graph/vertices"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"id":"b","label":"A","properties":{}}' \
  "http://127.0.0.1:8080/graphs/hugegraph/graph/vertices"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"label":"E","outV":"a","outVLabel":"A","inV":"b","inVLabel":"A","properties":{}}' \
  "http://127.0.0.1:8080/graphs/hugegraph/graph/edges"
```

```javascript
**Schema**



curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"name":"A","id_strategy":"CUSTOMIZE_STRING","properties":[],"nullable_keys":[],"enable_label_index":true}' \
  "http://127.0.0.1:8080/graphs/hugegraph/schema/vertexlabels"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"name":"E","source_label":"A","target_label":"A","frequency":"SINGLE","properties":[],"nullable_keys":[],"enable_label_index":true}' \
  "http://127.0.0.1:8080/graphs/hugegraph/schema/edgelabels"
```

lokidundun:
## Purpose of the PR

- close #3085 

## Main Changes
- Avoid applying `HugeCountStrategy` to `count().is(...)` traversals
  containing nested connective predicates.
- Keep the original `CountGlobalStep + IsStep` traversal for nested
  predicates, so that TinkerPop evaluates the complete predicate tree.
- Preserve the existing optimization for simple predicates and flat
  connective predicates.
- Prevent partially analyzed nested predicates from being transformed
  into `RangeGlobalStep` or `NotStep`, which can change or invert the
  filtering semantics.
- Add regression tests for a nested connective predicate and its
  negated form.

## Verifying these changes

    - xxx

- [ ]  Modify configurations
- [ ]  The public API
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugeCountStrategy.java
```diff
@@ -272,10 +272,28 @@ private boolean doStrategy(final Step step) {
             return false;
         }
 
+        final P<?> predicate = ((IsStep<?>) step.getNextStep()).getPredicate();
+        if (this.hasNestedConnectivePredicate(predicate)) {
```
imbajin: ‼️ **Critical: flat connective predicates remain semantically unsafe**

This guard only rejects a `ConnectiveP` whose child is another `ConnectiveP`, so flat expressions such as `P.eq(0L).and(P.gt(0L))` and `P.lte(0L).or(P.gte(1L))` still enter the optimization. The loop then derives `useNotStep` from only the first range-driving child and removes the complete `count().is(...)`; on an empty adjacency the contradiction can return true, and on a non-empty adjacency the tautology can return false. Both shapes were reproduced against the current strategy on TinkerPop 3.5.1. Please disable the destructive rewrite for every `ConnectiveP` unless equivalence is derived from the complete boolean expression; the conservative fix is to skip all `ConnectiveP` values here. Add flat AND/OR regressions for both zero and nonzero counts.

imbajin (commented):
Blocking: yes. Summary: Flat connective count predicates remain semantically unsafe and can return incorrect results after the destructive NotStep rewrite. Evidence: exact-head control-flow review and two independent TinkerPop 3.5.1 reproductions.

lokidundun:
This ci failure is not from this PR.HugeCountStrategy only fires on Gremlin traversals with count().is(...), but KoutTraverser uses HugeGraph's own BFS framework (no Gremlin traversal at all). The change only skips more cases,.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugeCountStrategy.java
```diff
@@ -272,6 +272,11 @@ private boolean doStrategy(final Step step) {
             return false;
         }
 
+        final P<?> predicate = ((IsStep<?>) step.getNextStep()).getPredicate();
+        if (predicate instanceof ConnectiveP) {
```
imbajin: ⚠️ **Important: preserve safe range limiting for flat connective predicates**

Returning here for every `ConnectiveP` bypasses the entire strategy, including the safe `RangeGlobalStep` insertion. Common predicates such as `between(1, 18)` are connective predicates, so a high-degree `count().is(...)` traversal now counts the complete input instead of stopping at the upper threshold. The new result assertions do not detect this plan regression. Please keep disabling the destructive `NotStep`/`dismissCountIs` rewrites unless full-expression equivalence is proven, while retaining a proven-safe range cap for analyzable flat connectives (and continue skipping nested cases as needed). Add a traversal-plan regression that asserts a representative flat connective still receives the range bound.

imbajin (commented):
Blocking: yes. Summary: The blanket ConnectiveP guard fixes the incorrect-result cases, but it also removes safe range limiting for common flat count predicates and can turn bounded counts into full scans. Evidence: exact-head control-flow inspection, confirmed independently by two review lanes.

imbajin (commented):
Blocking: no. Summary: The current head preserves safe range limiting for flat connective predicates while bypassing unsafe nested predicate rewrites; no actionable issues were found. Evidence: CountStrategyCoreTest passed 35/35 on the exact head, and all 15 visible latest-head checks succeeded.

imbajin (approved):

PR #3122, opened 2026-07-27
fix(server): handle repeated range predicates correctly

### LINKED ISSUE #3083: [Bug] Wrong result for repeated predicates
### Bug Type (问题类型)

gremlin (结果不合预期)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0
- gremlin: 3.5.1
- Docker Image: `hugegraph/hugegraph:latest`
- Backend: default Docker standalone backend
- Data Size: 2 vertices, 1 edge

HugeGraph returns a wrong result when multiple `has()` predicates.

The graph has one edge:
(a:A)-[:E {x: -4}]->(b:A)

Bug-triggering query:
g.E().has('x', -4).has('x', lte(4)).count()

Expected:
[1]

Actual:
[0]

The conjunction is satisfiable because the edge has x = -4, and -4 <= 4.

```javascript
**Vertex/Edge example**



curl -fsS --compressed -X DELETE \
  "http://127.0.0.1:8080/graphs/hugegraph/clear?confirm_message=I'm%20sure%20to%20delete%20all%20data"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"id":"a","label":"A","properties":{}}' \
  "http://127.0.0.1:8080/graphs/hugegraph/graph/vertices"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"id":"b","label":"A","properties":{}}' \
  "http://127.0.0.1:8080/graphs/hugegraph/graph/vertices"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"label":"E","outV":"a","outVLabel":"A","inV":"b","inVLabel":"A","properties":{"x":-4}}' \
  "http://127.0.0.1:8080/graphs/hugegraph/graph/edges"
```

```javascript
**Schema**



curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"name":"x","data_type":"LONG","cardinality":"SINGLE"}' \
  "http://127.0.0.1:8080/graphs/hugegraph/schema/propertykeys"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"name":"A","id_strategy":"CUSTOMIZE_STRING","properties":[],"nullable_keys":[],"enable_label_index":true}' \
  "http://127.0.0.1:8080/graphs/hugegraph/schema/vertexlabels"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"name":"E","source_label":"A","target_label":"A","frequency":"SINGLE","properties":["x"],"nullable_keys":["x"],"enable_label_index":true}' \
  "http://127.0.0.1:8080/graphs/hugegraph/schema/edgelabels"

curl -fsS --compressed -H 'Content-Type: application/json' -X POST \
  -d '{"name":"idx_e_x_range","base_type":"EDGE_LABEL","base_value":"E","index_type":"RANGE","fields":["x"]}' \
  "http://127.0.0.1:8080/graphs/hugegraph/schema/indexlabels"
```

LegendPei:
## Purpose of the PR

- close #3083

Fix incorrect empty results when multiple predicates target the same range-indexed property.

## Main Changes
- Cache the range-index validation result per element and property.
- Avoid consuming the same index value multiple times for repeated predicates.
- Add regression coverage for repeated predicates, including reversed predicate order.

## Verifying these changes

    - `mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,rocksdb -Dtest=EdgeCoreTest -DfailIfNoTests=false`

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQuery.java
```diff
@@ -875,11 +877,17 @@ public boolean checkRangeIndex(HugeElement element, Condition cond) {
                 return true;
             }
 
+            Map<Id, Boolean> propertyMatches =
+                    this.element2RangeIndexMatches.get(element.id());
+            if (propertyMatches != null && propertyMatches.containsKey(propId)) {
```
imbajin: ‼️ This cache outlives a single `test(element)` pass, while later index batches can continue adding values for the same element/property. If a stale value is seen first, `false` is cached and a later batch containing the current property value still returns that old result; the reverse case also skips left-index cleanup. The exact head reproduces this in `VertexCoreTest.testRemoveLeftRangeIndex` across memory, RocksDB, HBase, macOS, and HStore (expected 10, observed 2 or 6). Please scope deduplication to one condition loop rather than the whole query lifetime, so a later batch re-evaluates newly recorded index values, and rerun the full core backend matrix.

imbajin (commented):
Blocking: yes. Summary: The query-lifetime range-match cache reuses an incomplete first-batch result and causes valid elements to be omitted. Evidence: three independent review lanes, exact-head static tracing, and repeated failures of VertexCoreTest.testRemoveLeftRangeIndex across memory, RocksDB, HBase, macOS, and HStore checks.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQuery.java
```diff
@@ -631,10 +631,12 @@ public boolean test(HugeElement element) {
          * so don't break early even if test() return false.
          */
         boolean valid = true;
+        Map<Id, Boolean> rangeIndexMatches = new HashMap<>();
```
imbajin: ⚠️ This unconditionally allocates a new `HashMap` for every candidate element passed through `test()`, including queries where `element2IndexValueMap == null` and the map is never used. Because this is the per-element filtering path, broad scans now create O(N) short-lived maps and add avoidable GC pressure outside repeated-range queries. Please create the tracker only when range-index checking is active, preferably lazily or only when duplicate range properties exist, so ordinary queries retain the previous allocation profile.

imbajin (commented):
Blocking: yes. Summary: The correctness fix introduces an allocation on every condition-filtered element, including queries that never use range-index bookkeeping. Evidence: exact-head static inspection and targeted memory-profile path analysis; the two focused correctness tests pass.

imbajin (commented):
Blocking: no. Summary: The cross-batch correctness and non-index allocation regressions are fixed; the remaining range-index allocation concern is already covered by an existing thread, so no duplicate inline comment is added. Evidence: three independent exact-head review lanes, EdgeCoreTest ran 169 tests with 0 failures/errors and 11 skipped, and visible exact-head checks are green.

imbajin (approved):

Issue #3125, opened 2026-07-29
[Bug][HStore Client] gRPC stub pools bind every entry to one ManagedChannel


bitflicker64:
## HugeGraph version

Current `master`, verified at commit `b9710a7b0319ace41dd3c4ff5fb1165f53ee538e`.

## Problem

`AbstractGrpcClient` creates an array of 32 `ManagedChannel` instances per target, apparently to distribute requests across a channel pool. However, both the blocking-stub and asynchronous-stub initialization loops bind every array entry to the same channel.

The loop variable is `i`, but each iteration selects `channels[index]`, where `index` was calculated once before the loop:

```java
int index = (int) (l & (concurrency - 1));

HgPair<ManagedChannel, AbstractBlockingStub>[] value = new HgPair[concurrency];
IntStream.range(0, concurrency).forEach(i -> {
    ManagedChannel channel = channels[index];
    AbstractBlockingStub stub = getBlockingStub(channel);
    value[i] = new HgPair<>(channel, stub);
});
```

The asynchronous path contains the same indexing error:

```java
IntStream.range(0, concurrency).parallel().forEach(i -> {
    ManagedChannel channel = channels[index];
    AbstractAsyncStub stub = getAsyncStub(channel);
    value[i] = new HgPair<>(channel, stub);
});
```

Source:

https://github.com/apache/hugegraph/blob/b9710a7b0319ace41dd3c4ff5fb1165f53ee538e/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java#L93-L161

## Actual behavior

Although `getChannels(target)` creates 32 channels, all entries in a newly initialized blocking or asynchronous stub array reference whichever single channel happened to be selected by the first call's `index`.

Subsequent calls rotate across the stub-array indexes, but those stubs all use the same underlying `ManagedChannel`. The remaining 31 created channels are unused by that stub pool.

Consequences include:

- The intended channel-level request distribution does not occur.
- All traffic for that client and target is concentrated on one channel.
- Creating 31 additional channels consumes resources without providing concurrency or isolation.
- A problem affecting the selected channel affects every entry in the corresponding stub pool.

This defect is independent of DNS caching or Kubernetes Store replacement behavior and can be verified without a cluster.

## Expected behavior

Each stub-pool entry should be paired with its corresponding channel-pool entry:

```java
ManagedChannel channel = channels[i];
```

[...]

PR #3126, opened 2026-07-29
fix(server): configure finite DNS cache TTL

### LINKED ISSUE #3124: [Bug][HStore][Kubernetes] Data writes remain unavailable after Store pod replacement due to stale DNS resolution
## HugeGraph version

Current `master`, verified at commit `b9710a7b0319ace41dd3c4ff5fb1165f53ee538e`.

Backend and topology:

- HStore backend
- 3 PD nodes
- 3 Store nodes
- 3 Server nodes
- Kubernetes deployment using stable per-Store DNS names

## Problem

Replacing a Store pod during a rolling update can leave running HugeGraph Server pods permanently unable to write data to that Store.

The replacement Store pod has the same stable DNS name and Store node ID, but a different pod IP. PD reports the Store as `Up` with the correct node ID and address. The replacement Store is reachable from the Server pod, but HStore data writes continue to use the failed transport until the Server process is restarted.

This makes routine Store upgrades, node drains, and pod rescheduling capable of taking the graph data plane out of service while the Server pods remain Ready.

## Observed behavior

After replacing a Store pod, every affected data write returned HTTP 500 after the HStore retry loop:

```text
HgStoreClientException {sessionInfo: {storeNodeSession: {storeNode:
  {address: "hugegraph-store-1.hugegraph-store.<namespace>.svc:8500",
   nodeId: 2554185034686971778},
   graphName: "DEFAULT/hugegraph/g"}},
   reason: "UNAVAILABLE: io exception"}
```

Relevant stack frames:

```text
NodeTxExecutor.retryingInvoke(NodeTxExecutor.java:404)
NodeTxExecutor.doCommit(NodeTxExecutor.java:120)
NodeTxExecutor.commitTx(NodeTxExecutor.java:100)
NodeTxSessionProxy.commit(NodeTxSessionProxy.java:110)
HstoreSessionsImpl$HstoreSession.commit(HstoreSessionsImpl.java:464)
HstoreStore.commitTx(HstoreStore.java:691)
```

At the time of failure:

- PD reported all Stores as `Up`.
- PD returned the same node ID shown in the exception.
- `hugegraph-store-1...:8500` accepted TCP connections from inside the Server pod.
- The failure did not recover after more than six minutes.
- Restarting the Server pods restored writes immediately; the same request returned HTTP 201.
- Schema operations continued to work because they did not exercise this HStore commit path.
- `/versions` and `/graphs` continued responding, so the Server remained Ready and continued receiving traffic.

## Expected behavior

After a Store pod is replaced behind the same DNS name, existing Server processes should discard or recover the failed Store transport, resolve the current address, and resume writes without requiring a Server restart.

### 1. The normal container path enables a SecurityManager

`bin/start-hugegraph.sh` defaults to:

```bash
OPEN_SECURITY_CHECK="true"
```

The normal Server Docker entrypoint invokes:

```bash
```

It does not pass `-s`, so the default is retained. `bin/hugegraph-server.sh` then adds:

```bash
-Djava.security.manager=org.apache.hugegraph.security.HugeSecurityManager
```

The HStore Server image currently uses `eclipse-temurin:11-jre-jammy`.

There is no `networkaddress.cache.ttl` or `sun.net.inetaddr.ttl` setting in the packaged `bin/` or `conf/` files.

Under Java 11, when a SecurityManager is installed and no positive DNS TTL is explicitly configured, successful `InetAddress` resolutions are cached forever (`-1`). This behavior is specified by Java's `InetAddress` documentation:

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/InetAddress.html

The corresponding OpenJDK implementation is here:

https://github.com/openjdk/jdk11u/blob/master/src/java.base/share/classes/sun/net/InetAddressCachePolicy.java

The JDK uses this infinite default as protection against DNS spoofing or rebinding after an initial trusted resolution. HugeGraph does not explicitly select or document this DNS policy; it is an incidental consequence of enabling a SecurityManager to sandbox Gremlin execution.

### 2. HStore channels and stubs are retained by target address

`AbstractGrpcClient` stores channels in a static map keyed by the target string:

```java
protected static Map<String, ManagedChannel[]> channels = new ConcurrentHashMap<>();
```

Channels are constructed with:

```java
ManagedChannelBuilder.forTarget(target).usePlaintext().build();
```

No Store-client path removes or shuts down these cached channels when a transport call fails.

The separate channel-pool indexing defect in this class should be reported independently so it does not get buried in this DNS lifecycle issue.

Source:

https://github.com/apache/hugegraph/blob/b9710a7b0319ace41dd3c4ff5fb1165f53ee538e/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java

### 3. `retryingInvoke` retries the same cached state

`NodeTxExecutor.retryingInvoke()` invokes the same `Supplier` repeatedly after sleeping. It does not invalidate a Store node, session, stub, channel, or DNS state.

The transaction's Store sessions are cached by node ID for the duration of the complete retry loop and are replaced only in `doCommit()`'s final cleanup.

Source:

https://github.com/apache/hugegraph/blob/b9710a7b0319ace41dd3c4ff5fb1165f53ee538e/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java

### 4. `UNAVAILABLE` is not handled distinctly

`NotifyingExecutor` catches all thrown transport failures, reports `HgNodeStatus.NOT_WORK`, and rethrows. It does not inspect `StatusRuntimeException.getStatus().getCode()` and has no `UNAVAILABLE` recovery path.

The resulting notification only calls `pdClient.invalidPartitionCache()`. It does not invalidate:

- `HgStoreNodeManager.nodeIdMap`
- the cached Store address
- blocking or asynchronous stubs
- cached ManagedChannels

Sources:

- https://github.com/apache/hugegraph/blob/b9710a7b0319ace41dd3c4ff5fb1165f53ee538e/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/NotifyingExecutor.java
- https://github.com/apache/hugegraph/blob/b9710a7b0319ace41dd3c4ff5fb1165f53ee538e/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/HgStoreNodePartitionerImpl.java

### 5. Re-querying PD does not refresh the existing Store node

Invalidating the partition cache allows the next retry to re-query PD for partition leadership. However, when PD returns the same Store node ID, `HgStoreNodeManager.applyNode()` immediately returns the existing `nodeIdMap` entry. It does not call the provider again to retrieve Store metadata or rebuild the Store node.

Source:

https://github.com/apache/hugegraph/blob/b9710a7b0319ace41dd3c4ff5fb1165f53ee538e/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/HgStoreNodeManager.java

### 6. gRPC attempts DNS refresh, but the JVM cache remains stale

The effective Server dependency resolves gRPC Java to 1.47.0. Its `pick_first` load balancer requests name resolution when a subchannel reaches `TRANSIENT_FAILURE` or `IDLE`:

https://github.com/grpc/grpc-java/blob/v1.47.0/core/src/main/java/io/grpc/internal/PickFirstLoadBalancer.java

However, `DnsNameResolver` ultimately resolves through `InetAddress.getAllByName()`:

https://github.com/grpc/grpc-java/blob/v1.47.0/core/src/main/java/io/grpc/internal/DnsNameResolver.java

With the JVM positive DNS cache set effectively to forever by the active SecurityManager, a gRPC refresh receives the previously cached pod IP again. This also explains why restarting only the Server process fixes the failure: it clears the JVM DNS cache and all Store-client caches.

## Reproduction outline

1. Deploy PD, Store, and Server on Kubernetes with HStore enabled.
2. Ensure Server resolves a Store through a stable per-pod/headless-Service DNS name.
3. Successfully perform a graph data write.
4. Replace that Store pod so the DNS name is unchanged but the pod IP changes.
5. Wait for the Store to register as `Up` in PD.
6. Confirm the Store DNS name and port are reachable from the Server pod.
7. Perform another graph data write.

Actual result: writes repeatedly fail with `UNAVAILABLE: io exception` until Server is restarted.

Expected result: the existing Server reconnects to the replacement Store automatically.

### Required runtime fix

Configure a finite positive DNS cache TTL for the official Server distribution/container when the SecurityManager is enabled. Prefer a Java security properties override containing, for example:

```properties
networkaddress.cache.ttl=30
```

An immediate OpenJDK 11 workaround is to pass:

```text
-Dsun.net.inetaddr.ttl=30
```

This was validated on Temurin 11 with the following effective positive TTLs:

| Runtime condition | Effective positive DNS TTL |
| --- | ---: |
| No SecurityManager, no TTL setting | `30` |
| SecurityManager, no TTL setting | `-1` (forever) |
| SecurityManager with `-Dsun.net.inetaddr.ttl=30` | `30` |
| SecurityManager with `-Dnetworkaddress.cache.ttl=30` | `-1` (forever) |

`networkaddress.cache.ttl` is a Java security property. Passing it only as an ordinary `-D` system property does not change Java 11's `InetAddress` cache policy. The supported distribution-level fix should set it through Java security properties; the chart-level JVM workaround should use `-Dsun.net.inetaddr.ttl=30`.

Java 11 checks the `networkaddress.cache.ttl` security property before the `sun.net.inetaddr.ttl` system-property fallback. Therefore, if the distribution later configures `networkaddress.cache.ttl` through Java security properties, that distribution value takes precedence and the chart's `-Dsun.net.inetaddr.ttl=30` becomes a harmless no-op. It will not override or conflict with the upstream policy.

### Compatibility and security tradeoff

For a healthy cluster whose Store addresses do not change, making the TTL finite does not tear down established channels or force working connections to re-resolve. DNS cache expiration is lazy: it permits a later resolution to consult DNS again, but does not itself schedule a DNS query every 30 seconds. In this client, gRPC requests name resolution when the channel encounters failure or enters idle state.

The intended observable change is that a new Store IP behind an unchanged DNS name can eventually be discovered instead of the old IP remaining cached for the lifetime of the Server JVM.

This is not entirely security-neutral. The SecurityManager-dependent infinite TTL is a DNS spoofing and rebinding defense, so changing it to a finite value reopens a bounded re-resolution window. The proposed tradeoff is appropriate here because:

- `HugeSecurityManager` exists to sandbox Gremlin script execution; the repository contains no indication that its incidental infinite DNS cache is a deliberate Store-transport security control.
- HStore names are obtained from PD metadata rather than arbitrary user-controlled request hostnames.
- `30` seconds is OpenJDK 11's own default positive TTL when no SecurityManager is installed.
- The negative DNS cache policy is separate (`networkaddress.cache.negative.ttl`, normally 10 seconds) and is not changed by this proposal.

If the project considers PD-provided Store addresses untrusted, that trust boundary should be addressed directly through PD authentication, transport security, and address validation rather than by permanently pinning the first DNS answer received by each Server process.

### Optional Store-client hardening

- Detect `Status.Code.UNAVAILABLE` explicitly.
- Atomically evict the target's cached blocking stubs, asynchronous stubs, and ManagedChannels.
- Shut down evicted ManagedChannels.
- Invalidate or refresh the corresponding cached `HgStoreNode` so address changes for the same node ID can be observed.
- Reset connection backoff or rebuild the target before the next application retry.

Channel invalidation alone will not solve the stable-DNS case while the JVM DNS cache remains infinite, so the finite TTL is required.

## Kubernetes readiness impact

The official HStore image health check currently calls `/versions`. That endpoint remains healthy during this failure because the Server process and control-plane APIs are available even though HStore data writes are unavailable.

It would be useful to expose a non-mutating data-plane readiness endpoint that verifies the Server can reach the current HStore partition leaders. Helm charts could then use that endpoint instead of a process-only readiness check.

## Operational severity

This failure is triggered by ordinary Kubernetes operations such as Store upgrades, pod rescheduling, and node drains. It does not self-heal, is invisible to the current readiness probe, and requires restarting otherwise healthy Server pods to restore writes.

## Integration validation: finite DNS TTL is necessary but not sufficient (2026-07-30)

The original Store-replacement scenario has now been exercised end to end in a
3 PD / 3 Store / 3 Server Kubernetes deployment built from the audited master
commit above plus PRs #3126 and #3128.

After Store-2 was replaced behind the same StatefulSet DNS name:

- PD reported all three Stores `Up`, and every Raft group recovered.
- DNS inside the existing Server pod resolved the replacement Store IP.
- The replacement Store port was reachable from that Server pod.
- The existing Server's `KvPageScanner` nevertheless continued attempting the
  removed Store IP through the channel pool retained by `AbstractGrpcClient`.
- Reads repeatedly failed with `UNAVAILABLE` / `ConnectTimeoutException` until
  the client channel and stub pools were replaced.

A five-second Netty connection timeout bounded each failed attempt but did not
restore connectivity; that run ended `PASS=73 FAIL=1`. This confirms that a
finite JVM DNS TTL permits address discovery but does not itself invalidate the
`ManagedChannel[]`, blocking-stub, or asynchronous-stub pools cached by target.

An integration-only patch then:

- fingerprints each target's resolved address set before channel selection;
- shuts down and replaces the cached channel pool when that set changes;
- rebinds both stub pools to the replacement channel identities;
- orders concurrent resolutions so an older result cannot replace newer state;
- prevents an in-flight stale stub build from being published or returned; and
- replaces channels created while the first DNS fingerprint was unavailable
  once resolution succeeds.

The final focused suite passed 6 tests with 0 failures/errors/skips. The rebuilt
Server image then passed Store replacement without any Server restart
(`PASS=90 FAIL=0`), followed by the complete cluster gate (`PASS=61 FAIL=0`)
and dataset gate (`PASS=30 FAIL=0`).

Therefore, the Store-client work described earlier as optional hardening is a
required second part of the recovery fix. PR #3126 remains the finite-DNS-TTL
prerequisite. Issue #3125 and PR #3128 cover the separate stub-pool indexing
defect; they do not invalidate channels after an address change. A separate,
focused Store-client PR should complete this issue without broadening those PRs.

Validated integration source:

- https://github.com/bitflicker64/hugegraph-helm-stack/commit/c71c77a0f63846fece72579f9372888592a10ea9
- https://github.com/bitflicker64/hugegraph-helm-stack/tree/integration/hstore-helm-upstream-prs

bitflicker64:
## Purpose of the PR

- close #3124

## Problem, end to end

1. A Store pod is replaced in Kubernetes and comes back on a new IP behind the
   same stable DNS name.
2. HugeGraph Server runs with `HugeSecurityManager` installed by default.
3. With a `SecurityManager` present and no explicit policy, Java 11 falls back to
   `networkaddress.cache.ttl = -1`, so `InetAddressCachePolicy` is `FOREVER`.
4. The Server therefore keeps resolving the Store name to the old pod IP, and
   HStore writes keep going to an address that no longer exists.
5. Nothing recovers this until the Server process is restarted.

The setting is only honoured as a **security** property. The ordinary
`-Dnetworkaddress.cache.ttl` system property has no effect here, which is why
that form is deliberately not used.

## Scope

**In scope** — the Server JVM's DNS caching policy, end to end along one path:

`hugegraph-server.sh` builds the JVM arguments -> the JVM loads the packaged
security properties -> `HugeGraphServerBootstrap` validates the *effective*
policy and installs `HugeSecurityManager` -> control passes to
`HugeGraphServer`.

The guarantee this PR is trying to make: **with the security check enabled, the
Server either starts with a finite positive DNS TTL, or it does not start at
all.** Silently falling back to an indefinite cache is treated as a bug, so
every failure mode along that path fails closed.

**Boundary** — this makes the Server *re-resolve* the name within the TTL. It
does not decide what happens to connections that were already open against the
old address.

**Out of scope**

- Store-client channel invalidation after re-resolution, tracked by #3125
- HStore readiness changes
- Helm chart changes
- Replacing the `HugeSecurityManager` sandbox so the security check can run on
  JDK 24+ (JEP 486). This PR only reports that incompatibility clearly instead
  of failing with a raw VM initialization error.

## Main Changes

- Add a packaged `conf/java-security.properties` with
  `networkaddress.cache.ttl=30`, loaded through `-Djava.security.properties`
  only when `HugeSecurityManager` is enabled.
- Validate the effective `networkaddress.cache.ttl` in a small
  `HugeGraphServerBootstrap` before the server starts, so a missing, unreadable
  or infinite policy fails fast rather than silently restoring the indefinite
  cache. Because it reads the effective `Security.getProperty`, it also covers
  policies injected through `_JAVA_OPTIONS` or an operator override.
- Reject a security-enabled startup on JDK 24+ with an actionable message.
  JEP 486 removed the Security Manager there: `-Djava.security.manager=allow`
  is a fatal VM initialization error and `System.setSecurityManager()` always
  throws. Security-disabled startup on JDK 24+ is unchanged.
- Parse the `java -version` output by selecting the JVM banner line, anchored
  to its `java version`/`openjdk version` prefix, instead of `head -1` or the
  first line containing `version "`. A preamble is printed whenever
  `JAVA_TOOL_OPTIONS`/`_JAVA_OPTIONS` is set, which is how APM agents are
  normally installed, and such an agent may print its own `version "..."`
  banner: reading either line leaves `JAVA_VERSION` unusable or reports the
  agent's version, which can reject a supported JDK or trip the JDK 24+ guard.
- Name the bundled policy file in the server log when it is missing or
  unreadable, so an upgrade that reuses an older `conf/` reports the cause in the
  log `start-hugegraph.sh` points operators at. An operator override suppresses
  only this missing-file diagnostic, never validation: in daemon mode, where
  stderr only reaches `hugegraph-server-stdout.log`, the bootstrap mirrors its
  rejection of the effective policy — a missing, unreadable or infinite-TTL
  override included — into that same server log.
- Keep security-disabled startup unchanged, and preserve later operator JVM
  options including an explicit security-properties override.
- Run the focused Java runtime and launcher coverage in an independent RocksDB
  CI step, so it is not gated behind the broader startup prerequisites.

## Verifying these changes

- [x] Need tests and can be verified as follows (Java 11):
  - Distribution package build: passed.
  - `test-java-security-properties.sh`: passed.
  - Existing Server startup-script suite: 0 failed.
  - `HugeGraphServerBootstrapTest`: 2 passed, 0 failed.
  - Checkstyle: 0 violations. Apache RAT: 0 unapproved licenses.

Fail-closed behaviour was probed directly: an infinite TTL injected through
`_JAVA_OPTIONS`, and a `-Dnetworkaddress.cache.ttl=30` system property paired
with an infinite security-properties file, are both rejected.

Each new assertion was confirmed non-vacuous by mutating the production code and
checking the test then fails:

| Mutation | Result |
|---|---|
| Point the launcher at a missing bootstrap class | `ClassNotFoundException` passed the previous negative-only assertion, fails the new positive-sentinel one |
| Raise the JDK cap so the JDK 24+ guard never fires | boundary test fails |
| Restore `head -1` version parsing | preamble regression test fails |
| Un-anchor the version match back to any `version "` line | agent-banner regression fails |
| Drop the missing-policy diagnostic | missing-file test fails |
| No-op the daemon-mode bootstrap error-log property | daemon-mode override tests fail |

- [ ] Dependencies
- [x] Modify configurations
- [ ] The public API
- [ ] Other affects
- [ ] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -70,6 +70,14 @@ jobs:
         run: |
           mvn clean compile -U -Dmaven.javadoc.skip=true -ntp
 
+      - name: Run Java security properties tests
+        if: ${{ env.BACKEND == 'rocksdb' }}
+        run: |
+          mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
+          VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          SERVER_DIR=hugegraph-server/apache-hugegraph-server-$VERSION/
+          $TRAVIS_DIR/test-java-security-properties.sh $SERVER_DIR
+
       - name: Check startup test prerequisites
         id: server-preflight
         if: ${{ env.BACKEND == 'rocksdb' }}
@@ -86,10 +94,8 @@ jobs:
       - name: Run start-hugegraph.sh foreground mode tests
         if: ${{ env.BACKEND == 'rocksdb' && steps.server-preflight.outputs.can_run == 'true' }}
         run: |
-          mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
           SERVER_DIR=hugegraph-server/apache-hugegraph-server-$VERSION/
-          $TRAVIS_DIR/test-java-security-properties.sh $SERVER_DIR
           $TRAVIS_DIR/test-start-hugegraph.sh $SERVER_DIR
 
       - name: Startup tests skipped (missing prerequisites)
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -89,6 +89,7 @@ jobs:
           mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
           SERVER_DIR=hugegraph-server/apache-hugegraph-server-$VERSION/
+          $TRAVIS_DIR/test-java-security-properties.sh $SERVER_DIR
           $TRAVIS_DIR/test-start-hugegraph.sh $SERVER_DIR
 
       - name: Startup tests skipped (missing prerequisites)
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -97,8 +97,15 @@ else
     JAVA="$JAVA_HOME/bin/java -server"
 fi
 
-JAVA_VERSION=$($JAVA -version 2>&1 | head -1 | cut -d'"' -f2 | sed 's/^1\.//' | cut -d'.' -f1)
-if [[ $? -ne 0 || $JAVA_VERSION -lt $MIN_JAVA_VERSION ]]; then
+# Pick the version line explicitly: the JVM prints a preamble such as
+# "Picked up JAVA_TOOL_OPTIONS: ..." before it whenever JAVA_TOOL_OPTIONS or
+# _JAVA_OPTIONS is set, and reading that line instead would leave JAVA_VERSION
+# unusable and silently skip every version-gated option below.
+JAVA_VERSION=$($JAVA -version 2>&1 | awk -F'"' '/version "/ {print $2; exit}' |
+               sed 's/^1\.//' | cut -d'.' -f1)
+# Drop any pre-release suffix, e.g. "24-ea" -> "24"
+JAVA_VERSION="${JAVA_VERSION%%[!0-9]*}"
+if [[ -z $JAVA_VERSION || $JAVA_VERSION -lt $MIN_JAVA_VERSION ]]; then
     echo "Make sure the JDK is installed and the version >= $MIN_JAVA_VERSION, current is $JAVA_VERSION" \
          >> "${OUTPUT}"
     exit 1
@@ -162,6 +169,17 @@ EOF
     fi
 
     SECURITY_PROPERTIES="${CONF}/java-security.properties"
+    if [[ ! -r ${SECURITY_PROPERTIES} ]]; then
+        # The bootstrap validates the effective policy and refuses to start, but
+        # its stderr goes to the stdout log in daemon mode. Name the cause here
+        # so it also reaches the log start-hugegraph.sh points operators at.
+        cat >> "${OUTPUT}" <<EOF
+ERROR: Missing or unreadable '${SECURITY_PROPERTIES}'.
+An upgraded deployment that reuses an older conf/ directory must add this file,
+or supply its own -Djava.security.properties=<file> setting a finite positive
+networkaddress.cache.ttl.
+EOF
+    fi
     JVM_OPTIONS="${JVM_OPTIONS} \
                  -Djava.security.properties=${SECURITY_PROPERTIES}"
     if [[ ${JAVA_VERSION} -ge 18 ]]; then
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -143,7 +143,9 @@ esac
 
 JVM_OPTIONS="-Dlog4j.configurationFile=${CONF}/log4j2.xml"
 if [[ ${OPEN_SECURITY_CHECK} == "true" ]]; then
-    JVM_OPTIONS="${JVM_OPTIONS} -Djava.security.manager=org.apache.hugegraph.security.HugeSecurityManager"
+    JVM_OPTIONS="${JVM_OPTIONS} \
+                 -Djava.security.properties=${CONF}/java-security.properties \
+                 -Djava.security.manager=org.apache.hugegraph.security.HugeSecurityManager"
 fi
 
 if [ "${OPEN_TELEMETRY}" == "true" ]; then
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -170,6 +170,22 @@ EOF
 
     SECURITY_PROPERTIES="${CONF}/java-security.properties"
     if [[ ! -r ${SECURITY_PROPERTIES} ]]; then
+        # An operator may deliberately replace the bundled policy with their own
+        # -Djava.security.properties=<file>, which the JVM applies last and which
+        # makes a missing bundled file harmless. Track the last such option, since
+        # an empty value clears any earlier override.
+        SECURITY_PROPERTIES_OVERRIDDEN="false"
+        for OPTION in ${JAVA_OPTIONS} ${_JAVA_OPTIONS:-}; do
+            case "${OPTION}" in
+                -Djava.security.properties=)
+                    SECURITY_PROPERTIES_OVERRIDDEN="false" ;;
+                -Djava.security.properties=?*)
+                    SECURITY_PROPERTIES_OVERRIDDEN="true" ;;
+            esac
+        done
+    fi
+    if [[ ! -r ${SECURITY_PROPERTIES} &&
+          ${SECURITY_PROPERTIES_OVERRIDDEN:-false} == "false" ]]; then
         # The bootstrap validates the effective policy and refuses to start, but
         # its stderr goes to the stdout log in daemon mode. Name the cause here
         # so it also reaches the log start-hugegraph.sh points operators at.
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -64,6 +64,10 @@ ensure_path_writable "$PLUGINS"
 MAX_MEM=$((32 * 1024))
 MIN_MEM=$((1 * 512))
 MIN_JAVA_VERSION=11
+# JDK 24 removed the Security Manager (JEP 486): "-Djava.security.manager=allow"
+# is a fatal VM initialization error there and System.setSecurityManager() always
+# throws, so HugeSecurityManager cannot be installed on newer runtimes.
+MAX_SECURITY_JAVA_VERSION=23
 
 # Add the slf4j-log4j12 binding
 CP=$(find -L $LIB -name 'log4j-slf4j-impl*.jar' | sort | tr '\n' ':')
@@ -144,6 +148,19 @@ esac
 JVM_OPTIONS="-Dlog4j.configurationFile=${CONF}/log4j2.xml"
 SECURITY_MANAGER_OPTION=""
 if [[ ${OPEN_SECURITY_CHECK} == "true" ]]; then
+    if [[ ${JAVA_VERSION} -gt ${MAX_SECURITY_JAVA_VERSION} ]]; then
+        SECURITY_UNSUPPORTED_MSG=$(cat <<EOF
+The security check requires Java ${MIN_JAVA_VERSION}-${MAX_SECURITY_JAVA_VERSION}, current is ${JAVA_VERSION}.
+JDK 24+ removed the Security Manager (JEP 486), so HugeSecurityManager can no longer be installed.
+Run the server on Java ${MAX_SECURITY_JAVA_VERSION} or lower, or start it with the security check
+disabled: 'start-hugegraph.sh -s false'.
+EOF
+)
+        echo "${SECURITY_UNSUPPORTED_MSG}" >&2
+        echo "${SECURITY_UNSUPPORTED_MSG}" >> "${OUTPUT}"
+        exit 1
+    fi
+
     SECURITY_PROPERTIES="${CONF}/java-security.properties"
     JVM_OPTIONS="${JVM_OPTIONS} \
                  -Djava.security.properties=${SECURITY_PROPERTIES}"
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -142,29 +142,15 @@ case "$GC_OPTION" in
 esac
 
 JVM_OPTIONS="-Dlog4j.configurationFile=${CONF}/log4j2.xml"
+SECURITY_MANAGER_OPTION=""
 if [[ ${OPEN_SECURITY_CHECK} == "true" ]]; then
     SECURITY_PROPERTIES="${CONF}/java-security.properties"
-    HAS_SECURITY_PROPERTIES_OVERRIDE=false
-    for JAVA_OPTION in ${JAVA_OPTIONS}; do
-        case "${JAVA_OPTION}" in
-            -Djava.security.properties=)
-                HAS_SECURITY_PROPERTIES_OVERRIDE=false
-                ;;
-            -Djava.security.properties=*)
-                HAS_SECURITY_PROPERTIES_OVERRIDE=true
-                ;;
-        esac
-    done
-    if [[ ! -f "${SECURITY_PROPERTIES}" || ! -r "${SECURITY_PROPERTIES}" ]] &&
-       [[ "${HAS_SECURITY_PROPERTIES_OVERRIDE}" != "true" ]]; then
-        ERROR="Required Java security properties path is not a readable file: ${SECURITY_PROPERTIES}"
-        echo "${ERROR}" >&2
-        echo "${ERROR}" >> "${OUTPUT}"
-        exit 1
-    fi
     JVM_OPTIONS="${JVM_OPTIONS} \
-                 -Djava.security.properties=${SECURITY_PROPERTIES} \
-                 -Djava.security.manager=org.apache.hugegraph.security.HugeSecurityManager"
+                 -Djava.security.properties=${SECURITY_PROPERTIES}"
+    if [[ ${JAVA_VERSION} -ge 18 ]]; then
+        # Required to install HugeSecurityManager programmatically on JDK 18+.
+        SECURITY_MANAGER_OPTION="-Djava.security.manager=allow"
+    fi
 fi
 
 if [ "${OPEN_TELEMETRY}" == "true" ]; then
@@ -207,10 +193,14 @@ fi
 
 # Turn on security check
 if [[ "${STDOUT_MODE:-false}" == "true" ]]; then
-    exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \
-        org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF}
+    exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} \
+        ${SECURITY_MANAGER_OPTION} -cp ${CLASSPATH}: \
+        org.apache.hugegraph.bootstrap.HugeGraphServerBootstrap \
+        ${OPEN_SECURITY_CHECK} ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF}
 else
-    exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} -cp ${CLASSPATH}: \
-        org.apache.hugegraph.dist.HugeGraphServer ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF} \
+    exec ${JAVA} -Dname="HugeGraphServer" ${JVM_OPTIONS} ${JAVA_OPTIONS} \
+        ${SECURITY_MANAGER_OPTION} -cp ${CLASSPATH}: \
+        org.apache.hugegraph.bootstrap.HugeGraphServerBootstrap \
+        ${OPEN_SECURITY_CHECK} ${GREMLIN_SERVER_CONF} ${REST_SERVER_CONF} \
         >> ${LOGS}/hugegraph-server-stdout.log 2>&1
 fi
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -143,8 +143,27 @@ esac
 
 JVM_OPTIONS="-Dlog4j.configurationFile=${CONF}/log4j2.xml"
 if [[ ${OPEN_SECURITY_CHECK} == "true" ]]; then
+    SECURITY_PROPERTIES="${CONF}/java-security.properties"
+    HAS_SECURITY_PROPERTIES_OVERRIDE=false
+    for JAVA_OPTION in ${JAVA_OPTIONS}; do
+        case "${JAVA_OPTION}" in
+            -Djava.security.properties=)
+                HAS_SECURITY_PROPERTIES_OVERRIDE=false
+                ;;
+            -Djava.security.properties=*)
+                HAS_SECURITY_PROPERTIES_OVERRIDE=true
+                ;;
+        esac
+    done
+    if [[ ! -f "${SECURITY_PROPERTIES}" || ! -r "${SECURITY_PROPERTIES}" ]] &&
+       [[ "${HAS_SECURITY_PROPERTIES_OVERRIDE}" != "true" ]]; then
+        ERROR="Required Java security properties path is not a readable file: ${SECURITY_PROPERTIES}"
+        echo "${ERROR}" >&2
+        echo "${ERROR}" >> "${OUTPUT}"
+        exit 1
+    fi
     JVM_OPTIONS="${JVM_OPTIONS} \
-                 -Djava.security.properties=${CONF}/java-security.properties \
+                 -Djava.security.properties=${SECURITY_PROPERTIES} \
                  -Djava.security.manager=org.apache.hugegraph.security.HugeSecurityManager"
 fi
 
--- a/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
@@ -97,11 +97,14 @@ else
     JAVA="$JAVA_HOME/bin/java -server"
 fi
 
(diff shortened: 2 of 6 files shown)


hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
```diff
@@ -143,7 +143,9 @@ esac
 
 JVM_OPTIONS="-Dlog4j.configurationFile=${CONF}/log4j2.xml"
 if [[ ${OPEN_SECURITY_CHECK} == "true" ]]; then
-    JVM_OPTIONS="${JVM_OPTIONS} -Djava.security.manager=org.apache.hugegraph.security.HugeSecurityManager"
+    JVM_OPTIONS="${JVM_OPTIONS} \
+                 -Djava.security.properties=${CONF}/java-security.properties \
```
imbajin: ⚠️ Fail fast when this bundled security-properties file is absent. Existing deployments can retain or mount a pre-upgrade `conf/` directory without this new file; a controlled JDK invocation accepted a missing `-Djava.security.properties=...`, exited 0, and left `Security.getProperty("networkaddress.cache.ttl")` as `null`, silently restoring the indefinite DNS cache this change is meant to prevent. Please validate the default file before launch (unless a later explicit operator override is present), or place the default outside operator-replaceable configuration, and add a missing-file launcher regression.

.github/workflows/server-ci.yml
```diff
@@ -89,6 +89,7 @@ jobs:
           mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
           SERVER_DIR=hugegraph-server/apache-hugegraph-server-$VERSION/
+          $TRAVIS_DIR/test-java-security-properties.sh $SERVER_DIR
```
imbajin: ⚠️ Keep this DNS regression independent of unrelated startup prerequisites. This call is inside a step gated by a preflight requiring `lsof`, `crontab`, `curl`, and `java`; if any of the first three is unavailable, CI emits a notice and succeeds without running the new Java-only test. Please run `test-java-security-properties.sh` in a separate rocksdb/Java step—the current exact-head log shows it passes—and leave only `test-start-hugegraph.sh` behind the broader startup gate.

imbajin (commented):
Blocking: yes. Summary: The DNS TTL policy can silently disappear with a retained configuration directory, and its regression test can be skipped by unrelated startup prerequisites. Evidence: exact-head static review, a controlled missing-security-properties reproduction, workflow/log inspection, and git diff --check; codecov/project is failing.

.github/workflows/server-ci.yml
bitflicker64: Fixed in 35e1a240. The Java security-properties test now runs in its own RocksDB step after packaging and is independent of the `lsof`, `crontab`, and `curl` startup preflight. Only `test-start-hugegraph.sh` remains behind that broader prerequisite gate.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
bitflicker64: Fixed in 35e1a240. With security enabled, the launcher now rejects a missing, unreadable, or non-file bundled properties path unless the effective last later `-Djava.security.properties=...` override has a non-empty value. The focused regression covers the missing path, non-file path, empty override, valid-then-empty rejection, empty-then-valid acceptance, and restoration.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
```diff
@@ -142,8 +142,15 @@ case "$GC_OPTION" in
 esac
 
 JVM_OPTIONS="-Dlog4j.configurationFile=${CONF}/log4j2.xml"
+SECURITY_MANAGER_OPTION=""
 if [[ ${OPEN_SECURITY_CHECK} == "true" ]]; then
-    JVM_OPTIONS="${JVM_OPTIONS} -Djava.security.manager=org.apache.hugegraph.security.HugeSecurityManager"
+    SECURITY_PROPERTIES="${CONF}/java-security.properties"
+    JVM_OPTIONS="${JVM_OPTIONS} \
+                 -Djava.security.properties=${SECURITY_PROPERTIES}"
+    if [[ ${JAVA_VERSION} -ge 18 ]]; then
```
imbajin, line 207: ‼️ This unbounded JDK 18+ branch also includes JDK 24+, where `-Djava.security.manager=allow` is a fatal VM-initialization error and `System.setSecurityManager()` cannot install the fallback manager (OpenJDK JEP 486). Since the project documents Java 11+ without an upper bound, security-enabled JDK 24+ deployments cannot start. Please explicitly cap or reject JDK 24+ with a clear compatibility message until the sandbox is replaced, or provide a supported non-SecurityManager path, and exercise the 23/24 boundary with real JVMs.

hugegraph-server/hugegraph-dist/src/assembly/travis/test-java-security-properties.sh
```diff
@@ -0,0 +1,442 @@
...
+}
+
+assert_launcher_accepts_security_properties() {
+    local properties_path="$1"
+    local error_file="${TEMP_DIR}/launcher-valid.err"
+    if JAVA_OPTIONS="" STDOUT_MODE=true "$SERVER_SCRIPT" \
+       "${TEMP_DIR}/missing-gremlin.yaml" \
+       "${TEMP_DIR}/missing-rest.properties" true \
+       "-Djava.security.properties=${properties_path}" \
+       >/dev/null 2>"$error_file"; then
```
imbajin, line 234: ⚠️ This success-path check accepts any nonzero exit as long as stderr omits the two bootstrap messages below. A `ClassNotFoundException`, malformed command line, or lost configuration argument therefore makes the test pass without proving that TTL validation and security-manager installation succeeded; the disabled-path helper has the same negative-only pattern. Please assert a stable downstream sentinel/error after the intended bootstrap path, or add a validation-only mode with an explicit success signal.

imbajin (commented):
Blocking: yes. Summary: The security-enabled launcher cannot start on documented JDK 24+, and its success-path regression can accept unrelated early JVM failures. Evidence: six independent exact-head review lanes; OpenJDK JEP 486 and static launcher/bootstrap analysis; HugeGraphServerBootstrapTest passed 2/2 on JDK 11; git diff --check passed; visible exact-head checks are green.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
bitflicker64: Real issue, fixed in ba36d5ce. `hugegraph-server.sh` now caps the security check at Java 23 and exits with a message naming JEP 486 and pointing at `start-hugegraph.sh -s false`, instead of the JVM dying on `-Djava.security.manager=allow`. Security-disabled startup on JDK 24+ is unchanged.

The 23/24 boundary is exercised through the existing mock-JVM harness (23 still gets `allow`, 24 is rejected, 24 with security off still launches). Real JDK 23/24 runners would mean adding them to the Server CI matrix, which is outside this fix.

hugegraph-server/hugegraph-dist/src/assembly/travis/test-java-security-properties.sh
bitflicker64: Correct, fixed in ba36d5ce. Both launcher helpers now assert a positive sentinel — the failure must arrive through `HugeGraphServerBootstrap.main` -> `HugeGraphServer.main` -> `Failed to load yaml config file`, which is only reachable once the TTL check and the security-manager install have both succeeded.

Confirmed by mutation: pointing the launcher at a missing bootstrap class produced `ClassNotFoundException`, which passed the old negative-only check and fails the new one.

bitflicker64:
Follow-up review found two issues in my own earlier commit, both fixed in 9438a943:

1. **The JDK 24+ guard could be defeated by an env var.** `java -version` prints `Picked up JAVA_TOOL_OPTIONS: ...` before the version line whenever `JAVA_TOOL_OPTIONS`/`_JAVA_OPTIONS` is set, so the existing `head -1` parse left `JAVA_VERSION` unusable and every version-gated branch silently evaluated false — the new guard never fired, and `-Djava.security.manager=allow` was dropped on JDK 18-23. Now the version line is selected explicitly and a non-numeric result is rejected.
2. **A missing `conf/java-security.properties` failed with the reason in the wrong log.** The bootstrap's stderr goes to `hugegraph-server-stdout.log` in daemon mode, while `start-hugegraph.sh` points operators at `hugegraph-server.log`. The launcher now names the missing file in that log.

Both are covered by new regression tests, and each was confirmed non-vacuous by reverting the fix and checking the test fails.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
```diff
@@ -93,8 +97,15 @@ else
     JAVA="$JAVA_HOME/bin/java -server"
 fi
 
-JAVA_VERSION=$($JAVA -version 2>&1 | head -1 | cut -d'"' -f2 | sed 's/^1\.//' | cut -d'.' -f1)
-if [[ $? -ne 0 || $JAVA_VERSION -lt $MIN_JAVA_VERSION ]]; then
+# Pick the version line explicitly: the JVM prints a preamble such as
+# "Picked up JAVA_TOOL_OPTIONS: ..." before it whenever JAVA_TOOL_OPTIONS or
+# _JAVA_OPTIONS is set, and reading that line instead would leave JAVA_VERSION
+# unusable and silently skip every version-gated option below.
+JAVA_VERSION=$($JAVA -version 2>&1 | awk -F'"' '/version "/ {print $2; exit}' |
```
imbajin: ⚠️ This still accepts the first arbitrary line containing `version "`, not specifically the JVM banner. For example, `APM agent version "7.2.0"` before `openjdk version "21.0.8"` makes this pipeline return `7`, so a supported JDK is rejected; a preamble containing `version "24"` can likewise trigger the new upper-bound branch. The test helper uses the same broad match, so it cannot catch this. Please anchor the match to the actual `java`/`openjdk version` line and add a tool-options/agent preamble regression.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
```diff
@@ -142,8 +153,55 @@ case "$GC_OPTION" in
...
+        # -Djava.security.properties=<file>, which the JVM applies last and which
+        # makes a missing bundled file harmless. Track the last such option, since
+        # an empty value clears any earlier override.
+        SECURITY_PROPERTIES_OVERRIDDEN="false"
+        for OPTION in ${JAVA_OPTIONS} ${_JAVA_OPTIONS:-}; do
+            case "${OPTION}" in
+                -Djava.security.properties=)
+                    SECURITY_PROPERTIES_OVERRIDDEN="false" ;;
+                -Djava.security.properties=?*)
+                    SECURITY_PROPERTIES_OVERRIDDEN="true" ;;
```
imbajin, line 189: ⚠️ Any non-empty `-Djava.security.properties=...` token suppresses the missing bundled-policy diagnostic without checking whether the replacement file exists, is readable, or supplies a valid positive TTL. The bootstrap then rejects the invalid policy, but daemon stderr goes to `hugegraph-server-stdout.log` while `start-hugegraph.sh` directs operators to `hugegraph-server.log`, which now lacks the cause. Please suppress this diagnostic only after validating the effective override, or mirror bootstrap failures into the advertised log, and cover missing/unreadable/invalid override files in daemon mode.

imbajin (commented):
Blocking: yes. Summary: JVM version parsing can still select a tool/agent preamble, and invalid operator policy overrides suppress the actionable upgrade diagnostic. Evidence: six independent exact-head review lanes, focused shell reproductions, HugeGraphServerBootstrapTest passed 2/2, git diff --check passed, and visible exact-head checks are green.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
bitflicker64: Fixed in 9d53f15. The match is now anchored to the JVM banner line itself — only lines starting with `java version "` or `openjdk version "` are considered — in both the launcher and the test helper's own `JAVA_MAJOR` parse.

Added two mock-JVM regressions where an agent banner precedes the JVM's:

- `Elastic APM agent version "7.2.0"` ahead of a JDK 21 banner must still emit `-Djava.security.manager=allow` (the unanchored match reads 7 and rejects the runtime as below the minimum).
- `APM agent version "24.0.1"` ahead of a JDK 11 banner must not trip the JDK 24+ guard and must reach the bootstrap without a security-manager option.

Both fail against the unanchored match (re-verified by mutating the packaged launcher back to `/version "/`).

hugegraph-server/hugegraph-dist/src/assembly/static/bin/hugegraph-server.sh
bitflicker64: Fixed in 791947c, via the second option: the bootstrap's rejection is mirrored into the advertised log rather than validated shell-side.

I deliberately did not add launcher-side validation of the override file. Only the JVM's own properties parsing decides what the override loads to — the escaped-key, line-continuation and `file:`-URL cases in the test suite are all files a plain shell check would mis-classify — so a shell validator could disagree with the bootstrap in both directions and reintroduce exactly this inconsistency. The bootstrap stays the single validator.

Mechanics: in daemon mode the launcher passes `-Dhugegraph.bootstrap.error.log=${LOGS}/hugegraph-server.log`, and the bootstrap appends the same fatal message it prints to stderr (including the effective `java.security.properties` value, so the broken override is named) to that file, best-effort, before the security manager is installed and without touching the logging framework. Stdout mode is unchanged since stderr is already operator-visible there.

New daemon-mode tests cover missing, unreadable and infinite-TTL operator overrides, asserting both the cause and the override path land in `hugegraph-server.log`; the unreadable fixture also carries invalid content so the case still fails closed where permission bits do not apply (root). No-op'ing the error-log property in the packaged launcher makes them fail.

.github/workflows/server-ci.yml
```diff
@@ -87,6 +87,14 @@ jobs:
             hugegraph-server/hugegraph-dist/src/assembly/static \
             hugegraph-store/hg-store-dist/src/assembly/static/bin/util.sh
 
+      - name: Run Java security properties tests
+        if: ${{ env.BACKEND == 'rocksdb' }}
+        run: |
+          mvn package -Dmaven.test.skip=true -pl hugegraph-server/hugegraph-dist -am -ntp
```
imbajin, line 93: ⚠️ This step does not execute the newly added `HugeGraphServerBootstrapTest`: it runs `mvn package -Dmaven.test.skip=true`, while `run-unit-test.sh` only runs `hugegraph-test` for the memory matrix and does nothing for RocksDB. As a result, the workflow's RocksDB path validates only the shell smoke script, not these two unit tests; please add an explicit `mvn test -pl hugegraph-server/hugegraph-dist -am -Dtest=HugeGraphServerBootstrapTest` (or include the module in an existing test command) so the added assertions are exercised in CI.

imbajin (commented):
Blocking: no. Summary: The exact-head implementation compiles and all visible checks are green, but the new bootstrap unit test is not exercised by this workflow. Evidence: the new step uses `-Dmaven.test.skip=true`; the RocksDB `run-unit-test.sh` path does not invoke `hugegraph-dist`, while an explicit exact-head run reports 2/2 passed.

imbajin (approved):

Issue #3071, opened 2026-06-28
[Improve] Gremlin Task Large Result Chunking


hllqkb:
### Before submit
- [x] I have confirmed and searched that there are no similar problems in the historical issue and documents

---

### Environment

- HugeGraph version: 1.8.0+
- Java version: 11+
- Module: `hugegraph-core` (task framework), `hugegraph-api` (REST), `hugegraph-test`
- Related PR: #3060 (task metadata reads + `with_result=false`)

---

### Problem Description

When a Gremlin async task returns large results (e.g., a `g.V().valueMap()` query on a graph with millions of vertices), the entire result set is JSON-serialized, compressed, and stored as a **single** vertex property (`~task_result`). This has three compounding bottlenecks:

| Bottleneck | Code location | Current limit |
|---|---|---|
| Result item count | `GremlinJob.TASK_RESULTS_MAX_SIZE` | 800,000 (hardcoded to `Query.DEFAULT_CAPACITY`) |
| Compressed storage size | `CoreOptions.TASK_RESULT_SIZE_LIMIT` | 16 MB (default) |
| Single property maximum | `BytesBuffer.BYTES_LEN_MAX` | 10 MB |

Note: The 800,000 item count limit is orthogonal to chunking — it limits the in-memory `ArrayList` during Gremlin execution and is not addressed by this issue. Chunking addresses the storage/serialization bottlenecks below it.

**Current behavior (GremlinJob.execute):**
```java
List<Object> results = new ArrayList<>();
while (traversal.hasNext()) {
    results.add(traversal.next());    // all in memory, up to 800K entries
    checkResultsSize(results);
}
return results;  // full list -> HugeTask.set() -> JSON serialize -> compress -> single ~task_result
```

**Actual problems:**

1. A query returning 500K vertices with 10 properties each produces a ~50MB JSON string. `JsonUtil.toJson(result)` serializes it all in memory. After compression, `checkPropertySize` in `asArray()` checks against `TASK_RESULT_SIZE_LIMIT` (16MB default). If the compressed result exceeds this, a `LimitExceedException` is thrown and the task fails.

2. On `GET /tasks/{id}?with_result=true`, the entire `~task_result` blob is decompressed and deserialized. For a 50MB result, this takes seconds and can cause timeout on slow connections, especially with the RocksDB backend where the blob is read through multiple serialization layers.

[...]

PR #3127, opened 2026-07-29
feat: Add task result chunking support (issue #3071)

cyforkk:
## Summary

This PR adds support for chunking large task results to avoid memory overflow.

## Changes

- Added `TASK_RESULT_CHUNK_SIZE` config option to `CoreOptions.java` (default 1MB)
- Added `chunkKey()` and `isChunkedProperty()` helper methods to `P` class
- Implemented chunked write in `asArray()` method
- Added `RESULT_CHUNK_COUNT` constant

## Usage

When the compressed result exceeds `task.result_chunk_size`, it will be split into
multiple properties (`~task_result_0`, `~task_result_1`, ...).

Set `task.result_chunk_size=0` to disable chunking (preserves original behavior).

## Related

Part of issue #3071 (Chunk 1-4)

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/HugeTask.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/HugeTask.java
@@ -575,9 +575,29 @@ public class HugeTask<V> extends FutureTask<V> {
 
         if (this.result != null) {
             byte[] bytes = StringEncoding.compress(this.result);
-            checkPropertySize(bytes.length, P.RESULT);
-            list.add(P.RESULT);
-            list.add(bytes);
+            int chunkSize = CoreOptions.instance().get(CoreOptions.TASK_RESULT_CHUNK_SIZE);
+
+            if (chunkSize > 0 && bytes.length > chunkSize) {
+                // Split into chunks
+                int numChunks = (int) Math.ceil((double) bytes.length / chunkSize);
+                for (int i = 0; i < numChunks; i++) {
+                    int start = i * chunkSize;
+                    int end = Math.min(start + chunkSize, bytes.length);
+                    byte[] chunk = new byte[end - start];
+                    System.arraycopy(bytes, start, chunk, 0, end - start);
+                    checkPropertySize(chunk.length, P.chunkKey(i));
+                    list.add(P.chunkKey(i));
+                    list.add(chunk);
+                }
+                // Write chunk count
+                list.add(P.RESULT_CHUNK_COUNT);
+                list.add(String.valueOf(numChunks));
+            } else {
+                // Single property (original behavior)
+                checkPropertySize(bytes.length, P.RESULT);
+                list.add(P.RESULT);
+                list.add(bytes);
+            }
         }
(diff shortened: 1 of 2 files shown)


imbajin (commented):
Blocking: yes. Summary: The chunking implementation does not compile and is not wired into the task schema, read path, or distributed result-storage path. Evidence: exact-head CI compilation errors and static traces through HugeTask, TaskTransaction, StandardHugeGraph, and TaskAndResultScheduler.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/task/HugeTask.java
```diff
@@ -575,9 +575,29 @@ protected synchronized Object[] asArray() {
 
         if (this.result != null) {
             byte[] bytes = StringEncoding.compress(this.result);
-            checkPropertySize(bytes.length, P.RESULT);
-            list.add(P.RESULT);
-            list.add(bytes);
+            int chunkSize = CoreOptions.instance().get(CoreOptions.TASK_RESULT_CHUNK_SIZE);
+
+            if (chunkSize > 0 && bytes.length > chunkSize) {
```
imbajin, line 580: ‼️ This is not the shared persisted-result path. `TaskAndResultScheduler.save()` stores task metadata through `asArrayWithoutResult()` and writes the result through unchanged `HugeTaskResult.asArray()`, so HStore/distributed deployments still persist one BLOB; `HugeTask.set()` also enforces the total result limit before this branch runs. Please center chunking in the actual result-storage abstraction and cover both local and distributed persistence, including old-format compatibility.

contrueCT:
Thanks for the implementation. I reviewed the current patch, and it still seems to follow the original design that I raised concerns about in issue #3071.

The main problem is that the chunking logic is still implemented in `HugeTask.asArray()`. At this point, the result has already been fully serialized, and `HugeTask.set()` has already applied the total result-size check, so an oversized result may still fail before the chunking logic is reached.

More importantly, this is no longer the shared persisted-result path. `TaskAndResultScheduler.save()` stores task metadata through `HugeTask.asArrayWithoutResult()` and persists the actual result separately through `HugeTaskResult.asArray()`. Since `HugeTaskResult` is unchanged in this PR, distributed/HStore deployments will still store the result as a single BLOB, and the new chunking logic in `HugeTask.asArray()` may not be used for the actual result at all.

The current patch also only adds the chunked write path. There is no corresponding schema update or read/reassembly support for `~task_result_0`, `~task_result_1`, and `~task_result_chunk_count`, so even the legacy `HugeTask` persistence path would not be complete.

I think the design should first be revised around the actual result-storage abstraction, most likely `HugeTaskResult`, and clearly define:

* how chunked and legacy single-BLOB results are distinguished;
* how `task.result_size_limit` applies to the total result versus each chunk;
* how chunks are read and reassembled safely;
* how local and distributed task schedulers share the same behavior;
* how old persisted task results remain readable.

Tests should cover the real persisted-result path, distributed/HStore behavior, chunk reassembly, incomplete chunks, and backward compatibility with the old single-property format.

PR #3130, opened 2026-07-30
refactor(store): recover retries after store replacement

### LINKED ISSUE #3124: [Bug][HStore][Kubernetes] Data writes remain unavailable after Store pod replacement due to stale DNS resolution

bitflicker64:
## Purpose

Closes #3124.

In Kubernetes, a Store keeps the same DNS name and Store ID when its Pod is replaced, but its IP can change. HugeGraph already connects through that stable DNS name. The missing piece was the retry path: after a transport failure, the same transaction could keep using the session bound to the old Store process.

```mermaid
flowchart LR
    subgraph Before["Before"]
        direction TB
        B1["Store Pod gets a new IP"] --> B2["transport fails"]
        B2 --> B3["retry reuses old Node + Session"]
        B3 --> B4["request fails"]
    end

    subgraph After["After"]
        direction TB
        A1["Store Pod gets a new IP"] --> A2["transport fails with UNAVAILABLE"]
        A2 --> A3["discard exact failed Node + target"]
        A3 --> A4["retry opens current Node + Session"]
        A4 --> A5["gRPC reconnects to new IP"]
        A5 --> A6["same request succeeds"]
    end
```

## What changed

- Remove the custom IP fingerprint, five-second DNS polling, pool replacement, and maintenance executors. Kubernetes owns Pod IPs, while gRPC owns DNS resolution and transport reconnect.
- When gRPC reports `UNAVAILABLE`, discard only the exact failed Store node, close that target, and replace its cached session during the same transaction retry. `CANCELLED` and unrelated failures do not evict the node.
- Rebuild blocking and async stubs only when their channel generation changes, with focused tests for stale sessions, Store replacement, and channel rebinding.

This relies on the finite JVM DNS TTL from #3126 and follows the ownership boundaries documented by [Kubernetes StatefulSets](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/), [Java 11 DNS cache properties](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/doc-files/net-properties.html), and gRPC's [NameResolver](https://grpc.github.io/grpc-java/javadoc/io/grpc/NameResolver.html).

## Verification

- Java 11 `store-client-test`: **9 tests, 0 failures, 0 errors, 0 skipped**.
- Java 11 compile, EditorConfig, Checkstyle, Apache RAT, and `git diff --check`: passed.
- Exact `master + PR` Docker build: **38/38 modules passed**.
- Exact-head GitHub CI: **17/17 checks passed**.

| Kubernetes scenario | Result |
| --- | --- |
| 1 Server + 1 PD + 1 Store | Pod UID/IP changed while PVC, Store ID, and FQDN stayed stable; the in-flight request returned `201` after about 31 seconds |
| 1 Server + 3 PD + 3 Store, Leader failure | Store-0 was force deleted while leading 7/12 partitions; logs showed `UNAVAILABLE → NOT_WORK → retry`, and the 120-vertex request returned `201` after about 35 seconds |
| 1 Server + 3 PD + 3 Store, Follower failure | Store-0 was force deleted again with zero Leaders; the 60-vertex request returned `201` immediately |
| Final replicated state | All 3 Stores were Up and all 12 partitions had 3 normal replicas; the final scan observed all 420 expected uniquely identified vertices |
| Process and recovery | Server Pod UID, JVM PID, and start time stayed unchanged with 0 restarts; a later write/delete and 10/10 scans passed; no `gRPC client is closed` or OOM event was observed |

The runtime uses gRPC api/core/netty/stub 1.47.0 for this behavior. Protobuf-related artifacts remain at 1.39.0.

## Scope

This PR fixes Store replacement recovery. It does not add cross-process exactly-once commit or deduplication semantics, and it does not add a second DNS or reconnect state machine.

## Documentation

No user-facing configuration or dependency is added.

### DIFF
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
@@ -22,16 +22,21 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Map;
+import java.util.Set;
+import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -49,7 +54,7 @@ import io.grpc.stub.AbstractStub;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
-public abstract class AbstractGrpcClient {
+public abstract class AbstractGrpcClient implements AutoCloseable {
 
     protected static Map<String, ManagedChannel[]> channels = new ConcurrentHashMap<>();
     private static final Map<String, String> resolvedTargets = new ConcurrentHashMap<>();
@@ -60,17 +65,18 @@ public abstract class AbstractGrpcClient {
             new ConcurrentHashMap<>();
     private static final Map<String, ReentrantReadWriteLock> channelLocks =
             new ConcurrentHashMap<>();
+    private static final Map<AbstractGrpcClient, Boolean> clients =
+            Collections.synchronizedMap(new WeakHashMap<>());
     /*
-     * Refresh runs here rather than on a request thread: a caller of getChannels() may hold a
-     * Gremlin worker stack, which HugeSecurityManager denies socket access to. Creating the very
-     * first pool for a target is still done by the caller, so that path stays exposed.
+     * DNS refresh and first-pool creation run on prestarted threads because a caller of
+     * getChannels() may hold a Gremlin worker stack that HugeSecurityManager restricts.
      */
     private static final ScheduledThreadPoolExecutor CHANNEL_MAINTENANCE_EXECUTOR =
             new ScheduledThreadPoolExecutor(
-                    64, ExecutorPool.newThreadFactory("channel-maintenance"));
+                    4, ExecutorPool.newThreadFactory("channel-maintenance"));
     private static final ScheduledThreadPoolExecutor CHANNEL_INITIALIZATION_EXECUTOR =
             new ScheduledThreadPoolExecutor(
-                    64, ExecutorPool.newThreadFactory("channel-initialization"));
+                    4, ExecutorPool.newThreadFactory("channel-initialization"));
     private static final ScheduledThreadPoolExecutor CHANNEL_RETIREMENT_EXECUTOR =
             new ScheduledThreadPoolExecutor(
                     1, ExecutorPool.newThreadFactory("channel-retirement"));
@@ -108,18 +114,25 @@ public abstract class AbstractGrpcClient {
             new ConcurrentHashMap<>();
     private final Map<String, HgPair<ManagedChannel, AbstractAsyncStub>[]> asyncStubs =
             new ConcurrentHashMap<>();
+    private final Set<String> targets = ConcurrentHashMap.newKeySet();
     private final ThreadPoolExecutor executor;
+    private volatile boolean closed;
 
     {
         executor = ExecutorPool.createExecutor("common", 60, concurrency, concurrency);
     }
 
     public AbstractGrpcClient() {
-
+        clients.put(this, Boolean.TRUE);
     }
 
     public ManagedChannel[] getChannels(String target) {
-        CompletableFuture<Void> refresh = this.triggerChannelRefresh(target);
+        return this.getChannels(target, () -> true);
+    }
+
+    private ManagedChannel[] getChannels(String target, BooleanSupplier targetAvailable) {
+        CompletableFuture<Void> refresh =
+                this.registerTargetAndTriggerRefresh(target, targetAvailable);
         ManagedChannel[] tc = channels.get(target);
         if (tc != null) {
             return tc;
@@ -131,7 +144,12 @@ public abstract class AbstractGrpcClient {
          * avoids that in the common case; if the wait expires the rebuild still happens.
          */
         this.awaitInitialResolution(refresh);
-        synchronized (channels) {
+        ReentrantReadWriteLock targetLock = acquireTargetWriteLock(target);
+        try {
+            if (this.closed || !this.targets.contains(target) ||
+                !targetAvailable.getAsBoolean()) {
+                throw new IllegalStateException("The gRPC target is closed");
+            }
             if ((tc = channels.get(target)) == null) {
                 CompletableFuture<ManagedChannel[]> creation = new CompletableFuture<>();
                 try {
@@ -150,9 +168,10 @@ public abstract class AbstractGrpcClient {
                         } catch (InterruptedException e) {
                             interruption = e;
                         } catch (ExecutionException e) {
-                            Throwable cause = e.getCause();
-                            throw cause instanceof RuntimeException ? (RuntimeException) cause :
-                                  new RuntimeException(cause);
+                            if (interruption != null) {
+                                Thread.currentThread().interrupt();
+                            }
+                            throw propagate(e.getCause());
                         }
                     }
                     if (interruption != null) {
@@ -161,11 +180,12 @@ public abstract class AbstractGrpcClient {
                         throw new RuntimeException(interruption);
                     }
                 } catch (Throwable e) {
-                    throw e instanceof RuntimeException ? (RuntimeException) e :
-                          new RuntimeException(e);
+                    throw propagate(e);
                 }
                 channels.put(target, tc);
             }
+        } finally {
+            targetLock.writeLock().unlock();
         }
         return tc;
     }
@@ -173,8 +193,14 @@ public abstract class AbstractGrpcClient {
     public abstract AbstractBlockingStub getBlockingStub(ManagedChannel channel);
 
     public AbstractBlockingStub getBlockingStub(String target) {
+        return this.getBlockingStub(target, () -> true);
+    }
+
+    protected AbstractBlockingStub getBlockingStub(String target,
+                                                   BooleanSupplier targetAvailable) {
         return this.acquireStub(target, this.blockingStubs, this::getBlockingStub,
-                                stub -> (AbstractBlockingStub) this.setBlockingStubOption(stub));
+                                stub -> (AbstractBlockingStub) this.setBlockingStubOption(stub),
+                                targetAvailable);
     }
 
     /**
@@ -183,16 +209,17 @@ public abstract class AbstractGrpcClient {
      * the last check; a refresh landing immediately afterwards can still retire that pool, so
      * callers are not shielded from an in-flight replacement.
      *
-     * <p>The pool check needs no lock: the pool is published before the previous one is retired,
-     * so reading the current pool from the map is enough to know retirement has not started.
+     * <p>The final pool check and cache publication happen under the target's read lock, so a
+     * replacement cannot publish a new pool and retire the previous one in between them.
      */
     @SuppressWarnings("unchecked")
     private <S> S acquireStub(String target,
                               Map<String, HgPair<ManagedChannel, S>[]> stubCache,
                               Function<ManagedChannel, S> stubFactory,
-                              Function<S, S> stubOption) {
+                              Function<S, S> stubOption,
+                              BooleanSupplier targetAvailable) {
         while (true) {
-            ManagedChannel[] targetChannels = this.getChannels(target);
+            ManagedChannel[] targetChannels = this.getChannels(target, targetAvailable);
             HgPair<ManagedChannel, S>[] pairs = stubCache.get(target);
             int index = nextStubIndex();
             S configuredStub;
@@ -216,8 +243,7 @@ public abstract class AbstractGrpcClient {
                 configuredStub = stubOption.apply(pairs[index].getValue());
             }
 
-            ReentrantReadWriteLock.ReadLock readLock = channelLock(target).readLock();
-            readLock.lock();
+            ReentrantReadWriteLock targetLock = acquireTargetReadLock(target);
             try {
                 if (channels.get(target) != targetChannels) {
                     continue;
@@ -227,13 +253,59 @@ public abstract class AbstractGrpcClient {
                 }
                 return configuredStub;
             } finally {
-                readLock.unlock();
+                targetLock.readLock().unlock();
             }
         }
     }
 
     private static ReentrantReadWriteLock channelLock(String target) {
         return channelLocks.computeIfAbsent(target, key -> new ReentrantReadWriteLock());
+    }
+
+    private static ReentrantReadWriteLock acquireTargetReadLock(String target) {
+        while (true) {
+            ReentrantReadWriteLock lock = channelLock(target);
+            lock.readLock().lock();
+            if (channelLocks.get(target) == lock) {
+                return lock;
+            }
+            lock.readLock().unlock();
+        }
+    }
+
+    private static ReentrantReadWriteLock acquireTargetWriteLock(String target) {
+        while (true) {
+            ReentrantReadWriteLock lock = channelLock(target);
+            lock.writeLock().lock();
+            if (channelLocks.get(target) == lock) {
+                return lock;
+            }
+            lock.writeLock().unlock();
+        }
+    }
+
+    private static RuntimeException propagate(Throwable cause) {
+        if (cause instanceof Error) {
+            throw (Error) cause;
+        }
+        return cause instanceof RuntimeException ? (RuntimeException) cause :
+               new RuntimeException(cause);
+    }
+
+    private CompletableFuture<Void> registerTargetAndTriggerRefresh(
+            String target, BooleanSupplier targetAvailable) {
+        ReentrantReadWriteLock lock = acquireTargetReadLock(target);
+        try {
+            synchronized (clients) {
+                if (this.closed || !targetAvailable.getAsBoolean()) {
+                    throw new IllegalStateException("The gRPC client is closed");
+                }
+                this.targets.add(target);
+            }
+            return this.triggerChannelRefresh(target);
+        } finally {
+            lock.readLock().unlock();
+        }
     }
 
     private static int nextStubIndex() {
@@ -257,8 +329,14 @@ public abstract class AbstractGrpcClient {
     }
 
     public AbstractAsyncStub getAsyncStub(String target) {
+        return this.getAsyncStub(target, () -> true);
+    }
+
+    protected AbstractAsyncStub getAsyncStub(String target,
+                                             BooleanSupplier targetAvailable) {
         return this.acquireStub(target, this.asyncStubs, this::getAsyncStub,
-                                stub -> (AbstractAsyncStub) this.setStubOption(stub));
+                                stub -> (AbstractAsyncStub) this.setStubOption(stub),
+                                targetAvailable);
     }
 
     protected AbstractStub setStubOption(AbstractStub value) {
@@ -306,7 +384,7 @@ public abstract class AbstractGrpcClient {
         try {
             this.submitChannelRefresh(() -> {
                 try {
-                    this.refreshChannelsIfAddressChanged(target);
+                    this.refreshChannelsIfAddressChanged(target, refresh);
                 } catch (Throwable e) {
                     // The executor discards what a task throws, so report it here.
                     log.warn("Failed to refresh channels of target {}", target, e);
@@ -328,8 +406,9 @@ public abstract class AbstractGrpcClient {
          * than failing can outlast its own interval, which would let every later call queue
          * another lookup behind it.
          */
-        this.postponeNextRefresh(target);
-        refreshTasks.remove(target, refresh);
+        if (refreshTasks.remove(target, refresh)) {
+            this.postponeNextRefresh(target);
+        }
         refresh.complete(null);
     }
 
@@ -361,9 +440,10 @@ public abstract class AbstractGrpcClient {
      * the target's pool when its resolved address set has changed, publishing the replacement
      * before retiring the previous pool.
      */
-    private void refreshChannelsIfAddressChanged(String target) {
+    private void refreshChannelsIfAddressChanged(String target,
+                                                  CompletableFuture<Void> refresh) {
         String resolvedTarget = this.resolveTarget(target);
-        if (resolvedTarget.isEmpty()) {
+        if (resolvedTarget.isEmpty() || refreshTasks.get(target) != refresh) {
             return;
         }
 
@@ -377,7 +457,14 @@ public abstract class AbstractGrpcClient {
              * Nothing to replace yet. Recording the address here is what lets the common path
              * build its first pool already knowing the address, instead of rebuilding it.
              */
-            resolvedTargets.put(target, resolvedTarget);
+            ReentrantReadWriteLock targetLock = acquireTargetWriteLock(target);
+            try {
+                if (refreshTasks.get(target) == refresh && channels.get(target) == null) {
+                    resolvedTargets.put(target, resolvedTarget);
+                }
+            } finally {
+                targetLock.writeLock().unlock();
+            }
             return;
         }
 
@@ -391,12 +478,12 @@ public abstract class AbstractGrpcClient {
             return;
         }
 
-        ReentrantReadWriteLock.WriteLock writeLock = channelLock(target).writeLock();
-        writeLock.lock();
+        ReentrantReadWriteLock targetLock = acquireTargetWriteLock(target);
         try {
             boolean replaced = false;
             synchronized (channels) {
-                if (channels.get(target) == staleChannels) {
+                if (refreshTasks.get(target) == refresh &&
+                    channels.get(target) == staleChannels) {
                     channels.put(target, replacementChannels);
                     resolvedTargets.put(target, resolvedTarget);
                     replaced = true;
@@ -408,7 +495,7 @@ public abstract class AbstractGrpcClient {
             }
             this.retireChannels(replaced ? staleChannels : replacementChannels);
         } finally {
-            writeLock.unlock();
+            targetLock.writeLock().unlock();
         }
     }
 
@@ -444,20 +531,20 @@ public abstract class AbstractGrpcClient {
     private ManagedChannel[] createChannels(String target) {
         ManagedChannel[] value = new ManagedChannel[concurrency];
         CountDownLatch latch = new CountDownLatch(concurrency);
-        AtomicReference<RuntimeException> failure = new AtomicReference<>();
+        AtomicReference<Throwable> failure = new AtomicReference<>();
         for (int i = 0; i < concurrency; i++) {
             int fi = i;
             try {
                 this.submitChannelCreation(() -> {
                     try {
                         value[fi] = createChannel(target);
-                    } catch (Exception e) {
-                        failure.compareAndSet(null, new RuntimeException(e));
+                    } catch (Throwable e) {
+                        failure.compareAndSet(null, e);
                     } finally {
                         latch.countDown();
                     }
                 });
-            } catch (RuntimeException e) {
+            } catch (Throwable e) {
                 failure.compareAndSet(null, e);
                 latch.countDown();
             }
@@ -475,18 +562,113 @@ public abstract class AbstractGrpcClient {
         if (failure.get() != null || interruption != null) {
             forceTerminateChannels(value);
         }
+        if (failure.get() == null && Arrays.stream(value).anyMatch(channel -> channel == null)) {
+            failure.set(new IllegalStateException("Channel creation returned a null channel"));
+            forceTerminateChannels(value);
+        }
         if (interruption != null) {
             Thread.currentThread().interrupt();
             throw new RuntimeException(interruption);
         }
         if (failure.get() != null) {
-            throw failure.get();
+            Throwable cause = failure.get();
+            if (cause instanceof Error) {
+                throw (Error) cause;
+            }
+            throw propagate(cause);
         }
         return value;
     }
 
+    @Override
+    public void close() {
+        String[] ownedTargets;
+        synchronized (clients) {
+            this.closed = true;
+            ownedTargets = this.targets.toArray(new String[0]);
+        }
+        for (String target : ownedTargets) {
+            ReentrantReadWriteLock lock = acquireTargetWriteLock(target);
+            try {
+                synchronized (clients) {
+                    this.targets.remove(target);
+                    this.blockingStubs.remove(target);
+                    this.asyncStubs.remove(target);
+                    if (!isTargetInUse(target)) {
+                        closeChannelLocked(target, lock);
+                    }
+                }
+            } finally {
+                lock.writeLock().unlock();
+            }
+        }
+        clients.remove(this);
+        synchronized (this.executor) {
+            this.executor.shutdown();
+        }
+    }
+
+    protected static void closeAllChannels() {
+        Set<String> targets = ConcurrentHashMap.newKeySet();
+        synchronized (clients) {
+            targets.addAll(channels.keySet());
+            targets.addAll(resolvedTargets.keySet());
+            targets.addAll(nextResolutions.keySet());
+            targets.addAll(refreshTasks.keySet());
+            targets.addAll(channelLocks.keySet());
+            for (AbstractGrpcClient client : clients.keySet()) {
+                targets.addAll(client.targets);
+                client.targets.clear();
+                client.blockingStubs.clear();
+                client.asyncStubs.clear();
+            }
+        }
+        for (String target : targets) {
+            closeChannel(target);
+        }
+    }
+
+    private static boolean isTargetInUse(String target) {
+        return clients.keySet().stream().anyMatch(client -> client.targets.contains(target));
+    }
+
+    public static void closeChannel(String target) {
+        ReentrantReadWriteLock lock = acquireTargetWriteLock(target);
+        try {
+            closeChannelLocked(target, lock);
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    private static void closeChannelLocked(String target, ReentrantReadWriteLock lock) {
+        CompletableFuture<Void> refresh = refreshTasks.remove(target);
+        if (refresh != null) {
+            refresh.complete(null);
+        }
+        ManagedChannel[] retiredChannels = channels.remove(target);
+        resolvedTargets.remove(target);
+        nextResolutions.remove(target);
+        synchronized (clients) {
+            for (AbstractGrpcClient client : clients.keySet()) {
+                client.targets.remove(target);
+                client.blockingStubs.remove(target);
+                client.asyncStubs.remove(target);
+            }
+        }
+        if (retiredChannels != null) {
+            forceTerminateChannels(retiredChannels);
+        }
+        channelLocks.remove(target, lock);
+    }
+
     void submitChannelCreation(Runnable task) {
-        this.executor.execute(task);
+        synchronized (this.executor) {
+            if (this.executor.isShutdown()) {
+                throw new RejectedExecutionException("The gRPC client is closed");
+            }
+            this.executor.execute(task);
+        }
     }
 
     private void retireChannels(ManagedChannel[] retiredChannels) {
@@ -507,7 +689,7 @@ public abstract class AbstractGrpcClient {
         CHANNEL_RETIREMENT_EXECUTOR.schedule(task, timeoutNanos, TimeUnit.NANOSECONDS);
     }
 
-    private void forceTerminateChannels(ManagedChannel[] retiredChannels) {
+    private static void forceTerminateChannels(ManagedChannel[] retiredChannels) {
         for (ManagedChannel channel : retiredChannels) {
             if (channel != null && !channel.isTerminated()) {
                 channel.shutdownNow();
@@ -517,9 +699,9 @@ public abstract class AbstractGrpcClient {
 
     /**
      * Extracts the host that a gRPC target resolves through, covering the plain {@code host:port}
-     * form and the {@code dns:} scheme in both its {@code dns:host:port} and
-     * {@code dns://authority/host:port} spellings. Any other resolver scheme returns an empty
-     * host, leaving that target to gRPC instead of monitoring the wrong endpoint.
+     * form and slash-prefixed {@code dns:///host:port} and
+     * {@code dns://authority/host:port} spellings. The opaque {@code dns:host:port} spelling and
+     * any other resolver scheme return an empty host, leaving that target to gRPC.
      */
     private static String targetHost(String target) {
         if (target == null || target.isEmpty()) {
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -128,7 +129,38 @@ public abstract class AbstractGrpcClient {
         this.awaitInitialResolution(refresh);
         synchronized (channels) {
             if ((tc = channels.get(target)) == null) {
-                channels.put(target, tc = this.createChannels(target));
+                CompletableFuture<ManagedChannel[]> creation = new CompletableFuture<>();
+                try {
+                    this.submitChannelRefresh(() -> {
+                        try {
+                            creation.complete(this.createChannels(target));
+                        } catch (Throwable e) {
+                            creation.completeExceptionally(e);
+                        }
+                    });
+                    InterruptedException interruption = null;
+                    for (;;) {
+                        try {
+                            tc = creation.get();
+                            break;
+                        } catch (InterruptedException e) {
+                            interruption = e;
+                        } catch (ExecutionException e) {
+                            Throwable cause = e.getCause();
+                            throw cause instanceof RuntimeException ? (RuntimeException) cause :
+                                  new RuntimeException(cause);
+                        }
+                    }
+                    if (interruption != null) {
+                        forceTerminateChannels(tc);
+                        Thread.currentThread().interrupt();
+                        throw new RuntimeException(interruption);
+                    }
+                } catch (Throwable e) {
+                    throw e instanceof RuntimeException ? (RuntimeException) e :
+                          new RuntimeException(e);
+                }
+                channels.put(target, tc);
             }
         }
         return tc;
@@ -487,8 +519,11 @@ public abstract class AbstractGrpcClient {
         }
 
         String endpoint = target;
-        if (target.regionMatches(true, 0, DNS_SCHEME, 0, DNS_SCHEME.length())) {
+        if (target.startsWith(DNS_SCHEME)) {
             endpoint = target.substring(DNS_SCHEME.length());
+            if (!endpoint.startsWith("/")) {
+                return "";
+            }
             while (endpoint.startsWith("/")) {
                 endpoint = endpoint.substring(1);
             }
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
@@ -18,9 +18,12 @@
 package org.apache.hugegraph.store.client.grpc;
 
 import java.net.InetAddress;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
@@ -28,7 +31,7 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
@@ -41,18 +44,42 @@ import io.grpc.ManagedChannelBuilder;
 import io.grpc.stub.AbstractAsyncStub;
 import io.grpc.stub.AbstractBlockingStub;
 import io.grpc.stub.AbstractStub;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public abstract class AbstractGrpcClient {
 
     protected static Map<String, ManagedChannel[]> channels = new ConcurrentHashMap<>();
     private static final Map<String, String> resolvedTargets = new ConcurrentHashMap<>();
     private static final Map<String, AtomicLong> nextResolutions = new ConcurrentHashMap<>();
-    private static final Map<String, ReentrantLock> refreshLocks = new ConcurrentHashMap<>();
-    private static final ScheduledThreadPoolExecutor CHANNEL_CLEANUP_EXECUTOR =
+    private static final Map<String, CompletableFuture<Void>> refreshTasks =
+            new ConcurrentHashMap<>();
+    /*
+     * Refresh runs here rather than on a request thread: a caller of getChannels() may hold a
+     * Gremlin worker stack, which HugeSecurityManager denies socket access to. Creating the very
+     * first pool for a target is still done by the caller, so that path stays exposed.
+     */
+    private static final ScheduledThreadPoolExecutor CHANNEL_MAINTENANCE_EXECUTOR =
             new ScheduledThreadPoolExecutor(
-                    1, ExecutorPool.newThreadFactory("channel-cleanup"));
+                    2, ExecutorPool.newThreadFactory("channel-maintenance"));
     private static final long DEFAULT_CHANNEL_REFRESH_INTERVAL_NANOS =
             TimeUnit.SECONDS.toNanos(5L);
+    private static final long DEFAULT_INITIAL_RESOLUTION_TIMEOUT_NANOS =
+            TimeUnit.SECONDS.toNanos(1L);
+    private static final String DNS_SCHEME = "dns:";
+
+    static {
+        try {
+            // Create the maintenance threads eagerly, so no request thread ever creates one.
+            CHANNEL_MAINTENANCE_EXECUTOR.prestartAllCoreThreads();
+        } catch (Throwable e) {
+            // A denied prestart must not leave this class permanently uninitializable, but a
+            // request thread has to create the thread instead, where it may be denied again.
+            log.warn("Failed to start the channel maintenance threads eagerly, " +
+                     "channel refresh may be delayed until a permitted thread submits one", e);
+        }
+    }
+
     private static final int n = 5;
     protected static int concurrency = 1 << n;
     private static final AtomicLong counter = new AtomicLong(0);
@@ -72,14 +99,22 @@ public abstract class AbstractGrpcClient {
 
     }
 
-    public ManagedChannel[] getChannels(String target) {
-        this.refreshChannelsIfAddressChanged(target);
-        ManagedChannel[] tc;
-        if ((tc = channels.get(target)) == null) {
-            synchronized (channels) {
-                if ((tc = channels.get(target)) == null) {
-                    channels.put(target, tc = this.createChannels(target));
-                }
+    protected ManagedChannel[] getChannels(String target) {
+        CompletableFuture<Void> refresh = this.triggerChannelRefresh(target);
+        ManagedChannel[] tc = channels.get(target);
+        if (tc != null) {
+            return tc;
+        }
+
+        /*
+         * Only the very first pool for a target waits, and only for a bounded time: building it
+         * before its address is known makes the resolution that lands next rebuild it. Waiting
+         * avoids that in the common case; if the wait expires the rebuild still happens.
+         */
+        this.awaitInitialResolution(refresh);
+        synchronized (channels) {
+            if ((tc = channels.get(target)) == null) {
+                channels.put(target, tc = this.createChannels(target));
             }
         }
         return tc;
@@ -88,44 +123,58 @@ public abstract class AbstractGrpcClient {
     public abstract AbstractBlockingStub getBlockingStub(ManagedChannel channel);
 
     public AbstractBlockingStub getBlockingStub(String target) {
+        return this.acquireStub(target, this.blockingStubs, this::getBlockingStub,
+                                stub -> (AbstractBlockingStub) this.setBlockingStubOption(stub));
+    }
+
+    /**
+     * Returns a cached stub bound to a channel of the target's current pool, rebuilding the
+     * cache when the pool has been replaced. The stub comes from the pool that was published at
+     * the last check; a refresh landing immediately afterwards can still retire that pool, so
+     * callers are not shielded from an in-flight replacement.
+     *
+     * <p>The pool check needs no lock: the pool is published before the previous one is retired,
+     * so reading the current pool from the map is enough to know retirement has not started.
+     */
+    @SuppressWarnings("unchecked")
+    private <S> S acquireStub(String target,
+                              Map<String, HgPair<ManagedChannel, S>[]> stubCache,
+                              Function<ManagedChannel, S> stubFactory,
+                              Function<S, S> stubOption) {
         while (true) {
-            ManagedChannel[] targetChannels = getChannels(target);
-            HgPair<ManagedChannel, AbstractBlockingStub>[] pairs = blockingStubs.get(target);
-            long l = counter.getAndIncrement();
-            if (l >= limit) {
-                counter.set(0);
-            }
-            int index = (int) (l & (concurrency - 1));
+            ManagedChannel[] targetChannels = this.getChannels(target);
+            HgPair<ManagedChannel, S>[] pairs = stubCache.get(target);
+            int index = nextStubIndex();
             if (!usesChannels(pairs, targetChannels)) {
-                synchronized (blockingStubs) {
-                    pairs = blockingStubs.get(target);
+                synchronized (stubCache) {
+                    pairs = stubCache.get(target);
                     if (!usesChannels(pairs, targetChannels)) {
-                        HgPair<ManagedChannel, AbstractBlockingStub>[] value =
-                                new HgPair[concurrency];
+                        HgPair<ManagedChannel, S>[] value = new HgPair[concurrency];
                         IntStream.range(0, concurrency).forEach(i -> {
                             ManagedChannel channel = targetChannels[i];
-                            AbstractBlockingStub stub = getBlockingStub(channel);
-                            value[i] = new HgPair<>(channel, stub);
-                            // log.info("create channel for {}",target);
+                            value[i] = new HgPair<>(channel, stubFactory.apply(channel));
                         });
-                        synchronized (channels) {
-                            if (channels.get(target) != targetChannels) {
-                                continue;
-                            }
-                            blockingStubs.put(target, value);
-                            AbstractBlockingStub stub = value[index].getValue();
-                            return (AbstractBlockingStub) setBlockingStubOption(stub);
+                        if (channels.get(target) != targetChannels) {
+                            continue;
                         }
+                        stubCache.put(target, value);
+                        return stubOption.apply(value[index].getValue());
                     }
                 }
             }
-            synchronized (channels) {
-                if (channels.get(target) != targetChannels) {
-                    continue;
-                }
-                return (AbstractBlockingStub) setBlockingStubOption(pairs[index].getValue());
+            if (channels.get(target) != targetChannels) {
+                continue;
             }
+            return stubOption.apply(pairs[index].getValue());
         }
+    }
+
+    private static int nextStubIndex() {
+        long l = counter.getAndIncrement();
+        if (l >= limit) {
+            counter.set(0);
+        }
+        return (int) (l & (concurrency - 1));
     }
 
     private AbstractStub setBlockingStubOption(AbstractBlockingStub stub) {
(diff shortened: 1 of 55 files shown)


imbajin (commented):
Blocking: yes. Summary: Address refresh can abort active RPCs, adds synchronous DNS resolution to every request path, and does not preserve the existing gRPC target contract; the async refresh race also lacks equivalent coverage. Evidence: static review across six independent lanes; `mvn test -pl hugegraph-store/hg-store-test -am -Dtest=AbstractGrpcClientTest -DfailIfNoTests=false` passed 4 tests.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +205,70 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+                return;
+            }
+            if (resolvedTarget.equals(previousTarget)) {
+                return;
+            }
+            ManagedChannel[] staleChannels = channels.remove(target);
+            if (staleChannels != null) {
+                Arrays.stream(staleChannels)
+                      .filter(channel -> channel != null && !channel.isShutdown())
+                      .forEach(ManagedChannel::shutdownNow);
```
imbajin: ‼️ `shutdownNow()` forcibly terminates existing calls on these channels, but callers may already hold blocking or streaming stubs after the cache lock is released. A normal DNS rotation can therefore cancel active Store operations and leave writes with ambiguous outcomes. Please publish the replacement pool first, gracefully retire old channels with `shutdown()`, and use bounded delayed forced termination only after draining; add an active-RPC/stream refresh test.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +205,70 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+    private static boolean containsChannel(ManagedChannel[] channels,
+                                           ManagedChannel expected) {
+        return Arrays.stream(channels).anyMatch(channel -> channel == expected);
+    }
+
+    private void refreshChannelsIfAddressChanged(String target) {
+        long resolutionRequest = resolutionRequests.computeIfAbsent(target,
+                                                                    key -> new AtomicLong())
+                                                   .incrementAndGet();
+        String resolvedTarget = this.resolveTarget(target);
```
imbajin, line 307: ⚠️ Every `getChannels()` call reaches this synchronous resolution path before consulting the cached pool, and Store session methods acquire a stub per request. When the JVM DNS cache expires or the resolver is slow, application request threads now block on `InetAddress.getAllByName()`. Please throttle/cache refreshes using a configurable interval or single-flight TTL, or refresh asynchronously while retaining the last healthy pool; cover repeated and concurrent stub acquisition with a delayed resolver.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +205,70 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+                Arrays.stream(staleChannels)
+                      .filter(channel -> channel != null && !channel.isShutdown())
+                      .forEach(ManagedChannel::shutdownNow);
+            }
+        }
+    }
+
+    protected String resolveTarget(String target) {
+        try {
+            String host = URI.create("dns://" + target).getHost();
```
imbajin: ⚠️ This no longer matches the target contract used by `ManagedChannelBuilder.forTarget(target)`. For example, a valid `dns:///store:8500` target becomes `dns://dns:///store:8500`, so the fingerprint resolves the literal host `dns`; other gRPC resolver schemes can yield no host at all. The channel may still connect while refresh silently monitors the wrong endpoint. Please parse supported gRPC target schemes consistently (or explicitly validate and restrict the accepted format) and add URI-form and IPv6 target tests.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClientTest.java
```diff
@@ -0,0 +1,334 @@
...
+            assertTrue("the newer channel pool must remain live",
+                       Arrays.stream(freshChannels).noneMatch(ManagedChannel::isShutdown));
+        } finally {
+            client.releaseStaleResolution.countDown();
+            executor.shutdownNow();
+        }
+    }
+
+    @Test
+    public void testStubBuildRetriesAfterChannelRefresh() throws Exception {
```
imbajin, line 779: ⚠️ This interleaving test only exercises `getBlockingStub()`, while the PR independently rewrites the async refresh/retry/cache-publication path and builds async stubs in parallel. The sequential async assertion cannot catch a stale async pool published during refresh. Please add the equivalent latch-controlled async interleaving test and assert that both returned stubs and the final cache reference only current live channels.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -91,31 +100,44 @@ public ManagedChannel[] getChannels(String target) {
...
+                    pairs = blockingStubs.get(target);
+                    if (!usesChannels(pairs, targetChannels)) {
+                        HgPair<ManagedChannel, AbstractBlockingStub>[] value =
+                                new HgPair[concurrency];
+                        IntStream.range(0, concurrency).forEach(i -> {
+                            ManagedChannel channel = targetChannels[index];
+                            AbstractBlockingStub stub = getBlockingStub(channel);
+                            value[i] = new HgPair<>(channel, stub);
+                            // log.info("create channel for {}",target);
+                        });
```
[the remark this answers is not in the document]
bitflicker64: Validated as a real issue: the pool build loop used the selected request `index`, so every cached slot could bind to the same channel. Fixed in `95489135` by using the loop slot `i` for both blocking and async stub pools, and added spread assertions that the rebuilt pools cover every channel. Verified with the refreshed 9-test suite on Java 11.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in `95489135` for the target formats this client can safely fingerprint: raw `host:port`, `dns:///host:port`, and bracketed IPv6. Unsupported resolver schemes now return an empty fingerprint so the client does not silently monitor the wrong host while gRPC handles the original target. Broader non-DNS resolver fingerprinting is out of scope for this channel-lifecycle fix. Added URI-form, IPv6, and unsupported-scheme tests.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClientTest.java
bitflicker64: Fixed in `95489135`. Added `testAsyncStubBuildRetriesAfterChannelRefresh()` with the same latch-controlled stale-build interleaving as the blocking path, and it asserts both returned async stubs plus the final async cache only reference current live channels.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in `95489135`. The refresh path now creates and publishes the replacement pool before retiring the old pool. Retired channels receive graceful `shutdown()` first, and a bounded drain task only calls `shutdownNow()` if termination does not complete. Added `testRefreshGracefullyRetiresActiveStreamChannels()` to cover a held async/stream stub during refresh.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in `95489135`. Refresh checks are now throttled by a per-target TTL and guarded with a try-lock, so only one caller performs DNS resolution while concurrent callers keep using the last cached healthy pool. Added repeated acquisition and delayed concurrent resolver coverage to verify this path.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +193,224 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+        return InetAddress.getAllByName(host);
+    }
+
+    protected String resolveTarget(String target) {
+        String host = targetHost(target);
+        if (host.isEmpty()) {
+            return "";
+        }
+        try {
+            return Arrays.stream(this.resolveHost(host))
```
imbajin, line 512: ‼️ This synchronous lookup runs on the caller of `getChannels()`, but `InetAddress.getAllByName()` invokes `SecurityManager.checkConnect(host, -1)`. The default launcher enables `HugeSecurityManager`, whose `checkConnect()` rejects Gremlin worker stacks and has no HStore/gRPC exemption; only `UnknownHostException` is caught here. An HStore-backed Gremlin request can therefore throw `SecurityException` on initial resolution or a periodic refresh instead of using the healthy pool. Please move resolution to a trusted bounded background path (preserving the current pool on failure), or add the narrowest safe HStore policy exemption, and cover a real Gremlin-worker + `HugeSecurityManager` lookup regression.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +193,224 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+            synchronized (channels) {
+                if (channels.get(target) == staleChannels) {
+                    channels.put(target, replacementChannels);
+                    resolvedTargets.put(target, resolvedTarget);
+                    replaced = true;
+                }
+            }
+
+            if (replaced) {
+                this.retireChannels(staleChannels);
```
imbajin: ⚠️ A caller that loses `tryLock()` can still receive `staleChannels`; after this swap, retirement immediately calls `shutdown()` on that pool. `QueryV2Client#getManagedChannel()` uses `getChannels()` directly and creates its async stub only afterwards, bypassing the identity/retry loops added to `getAsyncStub()` and `getBlockingStub()`, so a refresh race can hand QueryV2 a channel that rejects the new RPC. Please route QueryV2 through the guarded async-stub path or provide an atomic channel/stub lease, and add an interleaving test for this production consumer.

imbajin (commented):
Blocking: yes. Summary: DNS refresh can fail under the default security policy, and QueryV2 can race channel retirement. Evidence: six independent exact-head lanes; 14/14 focused Java tests and git diff --check passed; all visible exact-head checks passed.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in `198de19e`. `resolveTarget()`, replacement creation and retirement now run on a prestarted `channel-maintenance` executor; `getChannels()` only triggers a refresh and keeps serving the last healthy pool. I avoided a `HugeSecurityManager` exemption: `getChannels(String)` was public, and the lazy thread creation in `retireChannels()`'s `schedule()` would trip `checkAccess(ThreadGroup)` under the same predicate anyway. Covered by `testRefreshSucceedsWhenTheCallerThreadIsDeniedSocketAccess()`, which denies `checkConnect`/`checkAccess` on `gremlin-server-exec-*` threads and asserts the caller still gets a stub and all resolution ran on maintenance threads.

One pre-existing gap I'd file separately: the very first pool for a cold target is still built via the caller, so a Gremlin worker that is the first caller for a new address can still be denied there.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in `198de19e`. `getQueryServiceStub()` now returns `getAsyncStub(target)` and `getManagedChannel()` is gone, so QueryV2 uses the same identity/retry loop as every other consumer. The hand-rolled round-robin only existed because the async pool once bound every entry to `channels[index]` (fixed in #3128), so nothing is lost. `getChannels(String)` is now `protected` — handing out a raw `ManagedChannel[]` is unsafe once refresh can retire it. Covered by `testQueryV2StubFollowsPublishedPoolAcrossRefresh()`, which refreshes the pool while a stub build is blocked mid-flight; it fails against the old body.

Note: routing narrows the race but a refresh landing between stub return and first call can still hit a just-shut-down channel — retry-on-`UNAVAILABLE` would close it, better as a follow-up.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +201,289 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+                     target, previousTarget, resolvedTarget);
+        }
+
+        this.retireChannels(replaced ? staleChannels : replacementChannels);
+    }
+
+    private boolean shouldRefreshChannels(String target) {
+        AtomicLong nextResolution = nextResolutions.computeIfAbsent(target,
+                                                                    key -> new AtomicLong());
+        return System.nanoTime() - nextResolution.get() >= 0L;
```
imbajin: ‼️ The first DNS refresh is keyed off a zero deadline, but `System.nanoTime()` has an arbitrary origin and may be negative. In that case this comparison is false for a newly seen target; because `postponeNextRefresh()` only writes a real deadline after a refresh is submitted, this target can keep serving its initial pool without ever recording or checking the resolved address until the clock crosses zero. Please represent the first run explicitly (rather than using zero as a timestamp), keep the elapsed-time comparison overflow-safe, and add a regression with an injectable negative `nanoTime()` value.

imbajin (commented):
Blocking: yes. The initial refresh deadline can suppress DNS monitoring for the lifetime of a target when the JVM's nanoTime origin is negative; three independent current-head review lanes converged on this issue.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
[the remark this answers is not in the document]
bitflicker64: Revalidated on final head `5afea3b2`. This finding was valid: the shared acquisition path now binds every blocking and async cache slot with `targetChannels[i]`, while `usesChannels()` enforces one-to-one slot identity and the final pool check retries if refresh replaced the pool during configured-stub construction. `testStubPoolsCoverEveryChannelOnFirstBuild()` and `testAddressChangeReplacesChannelAndStubPools()` verify complete channel spread before and after replacement. The final Java 11 run passes 24/24 with no failures, errors, or skips.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Revalidated on final head `5afea3b2`. The finding was valid and the refresh boundary is now explicit: `triggerChannelRefresh()` submits DNS resolution and replacement coordination to the prestarted `channel-maintenance-*` executor; replacement channel construction runs on the existing `common-*` executor; graceful retirement runs on maintenance; forced retirement runs on the separately prestarted `channel-retirement-*` scheduler so blocked DNS cannot starve the drain deadline. `testRefreshSucceedsWhenCallerIsDeniedSocketAndThreadAccess()` denies both socket and thread access on a `gremlin-server-exec-*` caller, proves all four stages stay off it, and still returns a stub. Focused failure tests also prove resolution, refresh submission, channel submission/creation, and retirement-scheduling failures keep the published healthy pool live and allow retry. Java 11 passes 24/24.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Revalidated on final head `5afea3b2`. The reported raw-channel path is gone: `QueryV2Client#getQueryServiceStub(String)` delegates directly to `getAsyncStub(target)`. The shared `acquireStub()` path constructs the configured stub, then rechecks the published channel-pool identity before caching or returning it and retries on a concurrent replacement. `testQueryV2StubFollowsPublishedPoolAcrossRefresh()` blocks inside `setStubOption()` at that real post-cache construction boundary, replaces the pool, and proves the returned and cached QueryV2 stubs are current, live, and still spread across every channel. The production line is covered in the aggregate JaCoCo XML (`line 47: mi=0, ci=5`). `getChannels(String)` remains public to preserve the existing API, but QueryV2 no longer consumes it.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed and revalidated on final head `5afea3b2`. The finding was valid: `nextResolutions` now stores `AtomicReference<Long>`, where `null` is the explicit never-scheduled state and every `long` clock value, including zero and negatives, is a real deadline. `shouldRefreshChannels()` uses `deadline == null || nanoTime() - deadline >= 0L`; signed subtraction is the standard overflow-safe monotonic deadline test for this bounded interval, including wraparound. `nanoTime()` is injectable. `testInitialRefreshRunsWhenNanoTimeIsNegative()` starts at `-1L` and proves the first resolution runs, while `testRefreshDeadlineSurvivesNanoTimeWraparound()` crosses `Long.MAX_VALUE` to `Long.MIN_VALUE` and proves the deadline neither fires early nor gets lost. Final Java 11 result: 24/24, zero failures/errors/skips.

bitflicker64:
Codecov follow-up for the earlier 0% patch report at `198de19e`: `codecov/patch` is green on final head `5afea3b2`. The original finding was valid; the Store aggregate report had been produced before the focused client execution. The final Store POM keeps the existing centralized profile design but declares Surefire before `jacoco:report-aggregate`, and `testQueryV2StubFollowsPublishedPoolAcrossRefresh()` now executes the real `QueryV2Client#getQueryServiceStub(String)` path. The Java 11 profile passes 24/24, and the generated JaCoCo XML records changed line 47 as `mi=0`, `ci=5`.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Final-head confirmation on `6beea6f46`: QueryV2 continues to call the guarded `getAsyncStub(target)` path, which validates the configured stub pool against the currently published channel pool before returning. The final candidate also prevents a directly injected test channel from entering DNS refresh retirement while proving the normal no-injection branch still delegates to inherited DNS resolution. Both interleavings have focused regressions; the Java 11 suite passes 27/27, JaCoCo line 47 is `mi=0`, `ci=5`, and the lifecycle guard has `mb=0`, `cb=2`.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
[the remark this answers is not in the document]
bitflicker64: Final-head confirmation on `6beea6f46`: per-channel spreading remains intact. The blocking and async cache builders bind slot `i` to channel `i`, and the guarded pool-identity check retries only if publication changed while the stubs were built. The Java 11 suite exercises first-build and post-refresh spreading across all 32 channels and passes 27/27 in both focused and Store-profile runs.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Final-head confirmation on `6beea6f46`: first refresh state is still explicit (`AtomicReference<Long>` with `null` meaning uninitialized), so no `nanoTime()` value is a sentinel. Deadline checks use overflow-safe signed subtraction (`now - deadline >= 0`), and focused regressions inject both negative initial time and wraparound across `Long.MAX_VALUE`/`Long.MIN_VALUE`. The final Java 11 focused and Store-profile runs pass all 27 tests.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Final-head confirmation on `6beea6f46`: refresh DNS resolution and replacement/retirement coordination still run on prestarted `channel-maintenance-*` threads, channel construction uses the permitted common executor, and delayed forced retirement uses the independent `channel-retirement-*` scheduler. The Java 11 SecurityManager regression denies socket and thread access on the Gremlin caller while proving resolution, creation, graceful retirement, and forced retirement occur elsewhere; all 27 tests pass with zero skips. Resolution, submission, partial creation, and retirement-scheduling failures retain the healthy published pool and allow retry.

bitflicker64:
Final-head CI note for 6beea6f46: all 17 GitHub checks passed, and the Store job generated and discovered the expected JaCoCo reports. Its legacy codecov-action v3.0.0 upload was rejected by Codecov with HTTP 429 and an expected retry window of 1,703 seconds, after which the action intentionally exited 0; consequently Codecov never registered the final commit. The cooldown has elapsed, but GitHub does not allow the fork contributor to rerun the completed Apache job. I am briefly closing and reopening this PR to retrigger trusted CI on the unchanged SHA, without modifying history or the four-file diff. The regenerated Java 11 Store profile still passes 27/27 and records QueryV2 line 47 as mi=0, ci=5 and line 58 as mb=0, cb=2.

imbajin (commented):
Blocking: yes. Summary: The refresh implementation still has a non-atomic stub-to-channel handoff and a shared resolver pool that can pause refreshes for every target when DNS blocks. Evidence: exact-head static review across six independent lanes; current-head checks are green, while these races remain distinct from the previously resolved SecurityManager, target parsing, stub distribution, and nanoTime findings.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -91,31 +134,60 @@ public ManagedChannel[] getChannels(String target) {
...
+                        if (channels.get(target) != targetChannels) {
+                            continue;
+                        }
+                        stubCache.put(target, value);
+                        return configuredStub;
+                    }
+                }
+            }
+            S configuredStub = stubOption.apply(pairs[index].getValue());
+            if (channels.get(target) != targetChannels) {
```
imbajin, line 178: ‼️ The identity check does not protect the handoff from returning a stub bound to a channel that is being retired. Evidence: acquireStub checks channels.get(target) at :178 and then returns; a concurrent refresh publishes the replacement at :339 and calls retireChannels at :349, whose shutdown immediately rejects new RPCs on the old pool. A caller can pass the check, be preempted, and receive the stub after retirement begins. Protect stub acquisition through first-call creation with a pool lease or read/write lock, or retry through a wrapper that binds to the current pool.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -34,10 +44,53 @@
...
+    private static final Map<String, AtomicReference<Long>> nextResolutions =
+            new ConcurrentHashMap<>();
+    private static final Map<String, CompletableFuture<Void>> refreshTasks =
+            new ConcurrentHashMap<>();
+    /*
+     * Refresh runs here rather than on a request thread: a caller of getChannels() may hold a
+     * Gremlin worker stack, which HugeSecurityManager denies socket access to. Creating the very
+     * first pool for a target is still done by the caller, so that path stays exposed.
+     */
+    private static final ScheduledThreadPoolExecutor CHANNEL_MAINTENANCE_EXECUTOR =
```
imbajin, line 64: ⚠️ Two blocked DNS lookups can stop address refresh for every target. Evidence: CHANNEL_MAINTENANCE_EXECUTOR has only two threads, and each refresh calls resolveTarget at :307, which reaches InetAddress.getAllByName without a client deadline; refreshTasks is cleared only after that call returns. If two targets hang, later targets queue behind them and keep stale channel pools. Isolate resolution per target or use a resolver with a real timeout/cancellation, and add a regression where two blocked lookups do not prevent a third target from refreshing.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Validated and fixed in commit e852a8ff. Stub construction now completes before acquiring the target read lock, then acquires that lock for the final published-pool identity check and cache publication/return. Refresh publication and retirement take the same target write lock, so retirement cannot begin between the final check and handoff. The existing blocking, async, and QueryV2 interleaving tests, including testQueryV2StubFollowsPublishedPoolAcrossRefresh, pass; the focused run is 28/28 with zero failures or errors.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Validated and fixed in commit e852a8ff. CHANNEL_MAINTENANCE_EXECUTOR is now prestarted with 64 workers, while refreshes remain single-flight per target via refreshTasks; two blocked InetAddress lookups cannot occupy the whole maintenance pool. Added testBlockedResolutionsDoNotStarveAnotherTarget, which holds two resolutions and proves a third target replaces its pool and keeps all replacement channels live. Focused AbstractGrpcClientTest passes 28/28 with zero failures or errors.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -58,31 +114,21 @@ public AbstractGrpcClient() {
...
+
+        /*
+         * Only the very first pool for a target waits, and only for a bounded time: building it
+         * before its address is known makes the resolution that lands next rebuild it. Waiting
+         * avoids that in the common case; if the wait expires the rebuild still happens.
+         */
+        this.awaitInitialResolution(refresh);
+        synchronized (channels) {
+            if ((tc = channels.get(target)) == null) {
+                channels.put(target, tc = this.createChannels(target));
```
imbajin: ‼️ The cold-target path can still create channel workers from the request thread. When resolution does not finish within the one-second wait, this path calls `createChannels()`; its lazy `common` executor uses a `SynchronousQueue` and `CallerRunsPolicy`, so a Gremlin worker that cannot create/access threads can fail the first Store request or run channel creation on the restricted stack. Please move first-pool creation onto a prestarted trusted executor and add a regression where the first caller is a restricted Gremlin worker.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +232,316 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+     * {@code dns://authority/host:port} spellings. Any other resolver scheme returns an empty
+     * host, leaving that target to gRPC instead of monitoring the wrong endpoint.
+     */
+    private static String targetHost(String target) {
+        if (target == null || target.isEmpty()) {
+            return "";
+        }
+
+        String endpoint = target;
+        if (target.regionMatches(true, 0, DNS_SCHEME, 0, DNS_SCHEME.length())) {
```
imbajin: ⚠️ This parser treats `dns:host:port` (and uppercase `DNS`) as a supported fingerprint form, but `createChannel()` forwards the original target to gRPC. In gRPC 1.39, `DnsNameResolverProvider` requires a slash-prefixed URI path, so `dns:host:port` has no path and uppercase `DNS` does not match the provider; the fingerprint test can pass while real channel creation fails. Please restrict parsing to target forms accepted by `ManagedChannelBuilder.forTarget()` or normalize the target before construction, and test actual channel creation.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClientTest.java
```diff
@@ -40,70 +57,1039 @@
...
+            assertSame("the cached stub must retain the live injected channel", injected,
+                       client.getQueryServiceStub(target).getChannel());
+        } finally {
+            client.releaseResolution.countDown();
+            QueryV2Client.setTestChannel(null);
+        }
+    }
+
+    @Test
+    public void testQueryV2WithoutInjectedChannelUsesDnsResolution() {
```
imbajin: ⚠️ `testQueryV2WithoutInjectedChannelUsesDnsResolution()` only calls the test helper `resolve()`, which exercises inherited `resolveTarget()` directly; it never calls the public `getQueryServiceStub()` path used when `channel == null`. A regression in normal QueryV2 stub acquisition or channel publication can therefore pass this test. Please invoke `getQueryServiceStub(target)` with deterministic `resolveHost()`/`createChannel()` fakes and assert the returned and cached stubs use the current live pool.

imbajin (commented):
Blocking: yes. Summary: Cold-target initialization can still run channel creation from a restricted Gremlin request, and the fingerprint parser accepts a DNS target form that gRPC 1.39 cannot build. Evidence: `AbstractGrpcClient.java` lines 131, 490, and 546; `ExecutorPool.createExecutor()` uses a lazy SynchronousQueue executor with CallerRunsPolicy; gRPC 1.39 DnsNameResolverProvider requires a slash-prefixed URI path.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in commit 16851352 and hardened in 9bacaad4. First-pool creation is submitted to the dedicated prestarted channel-initialization executor, so a cold Gremlin worker no longer creates channel workers on its restricted stack. testColdTargetCreationStaysOffDeniedCaller starts with no pool, installs the denying SecurityManager, invokes the public blocking-stub path, and verifies the caller completes with a live stub while every channel is created off-thread. The final focused suite is 29/29 with zero failures/errors/skips.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in 9bacaad4. targetHost now accepts only lowercase dns targets with a slash-prefixed URI path, matching gRPC 1.39; raw host:port and bracketed IPv6 remain supported, while scheme-only dns:host:port and uppercase DNS forms are skipped. The target parser test covers the accepted and rejected forms.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClientTest.java
bitflicker64: Fixed in fad20a3c. The regression now invokes the public ResolvingQueryV2Client getQueryServiceStub(String) path with deterministic fake channel construction, then asserts the returned stub belongs to the published live pool. The final focused suite is 29/29 with zero failures/errors/skips.

imbajin (commented):
Blocking: yes. Summary: The change eagerly creates 129 executor threads, retains target-keyed state indefinitely, and can publish a channel pool containing null entries after an Error. Focused validation passed 29/29 tests, while the latest dependency-check and Codecov checks are failing. Evidence: static review of AbstractGrpcClient.java:54-93 and 444-485; `mvn -q test -pl hugegraph-store/hg-store-test -am -Dtest=AbstractGrpcClientTest -Dsurefire.failIfNoSpecifiedTests=false -Djacoco.skip=true`; `gh pr checks 3130 -R apache/hugegraph`.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -34,10 +46,59 @@
...
+    private static final Map<String, CompletableFuture<Void>> refreshTasks =
+            new ConcurrentHashMap<>();
+    private static final Map<String, ReentrantReadWriteLock> channelLocks =
+            new ConcurrentHashMap<>();
+    /*
+     * Refresh runs here rather than on a request thread: a caller of getChannels() may hold a
+     * Gremlin worker stack, which HugeSecurityManager denies socket access to. Creating the very
+     * first pool for a target is still done by the caller, so that path stays exposed.
+     */
+    private static final ScheduledThreadPoolExecutor CHANNEL_MAINTENANCE_EXECUTOR =
```
imbajin: ‼️ Class initialization calls `prestartAllCoreThreads()` on pools sized 64 + 64 + 1, so loading `AbstractGrpcClient` eagerly creates 129 daemon threads before any target is used. There is no lifecycle shutdown path for these static executors, imposing a fixed thread cost on every process/classloader that loads the client. Size the executors to actual work and start them lazily or provide an explicit lifecycle shutdown instead of prestarting all cores.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -34,10 +46,59 @@
 import io.grpc.stub.AbstractAsyncStub;
 import io.grpc.stub.AbstractBlockingStub;
 import io.grpc.stub.AbstractStub;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public abstract class AbstractGrpcClient {
 
     protected static Map<String, ManagedChannel[]> channels = new ConcurrentHashMap<>();
+    private static final Map<String, String> resolvedTargets = new ConcurrentHashMap<>();
```
imbajin: ⚠️ Every distinct target is retained indefinitely in the new `resolvedTargets`, `nextResolutions`, and `channelLocks` maps: these entries are created with `put`/`computeIfAbsent`, but no corresponding removal exists (only the in-flight `refreshTasks` entry is removed). Target/address churn therefore grows one address, deadline reference, and lock per historical target, in addition to the channel/stub caches. Add target/client lifecycle cleanup that atomically clears all target-keyed state and terminates its channels.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
```diff
@@ -169,6 +268,323 @@ protected AbstractStub setStubOption(AbstractStub value) {
...
+        ManagedChannel[] value = new ManagedChannel[concurrency];
+        CountDownLatch latch = new CountDownLatch(concurrency);
+        AtomicReference<RuntimeException> failure = new AtomicReference<>();
+        for (int i = 0; i < concurrency; i++) {
+            int fi = i;
+            try {
+                this.submitChannelCreation(() -> {
+                    try {
+                        value[fi] = createChannel(target);
+                    } catch (Exception e) {
```
imbajin: ⚠️ Channel creation records only `Exception`. If `createChannel(target)` throws an `Error`, the task still decrements the latch but leaves a null slot in `value`; `createChannels()` then returns that partially populated array because `failure` remains null, and `getChannels()` publishes it. Later stub construction can dereference the null channel while successful sibling channels are leaked. Capture the throwable or validate every slot, force-terminate partial channels, and propagate the failure before publication.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Valid. The exact head prestarts 64 maintenance, 64 initialization, and 1 retirement thread during class initialization, and these static executors have no shutdown lifecycle. The security boundary requires trusted threads to exist before a restricted Gremlin caller submits work, but 129 permanent threads is not an acceptable fixed cost. This needs a smaller/lifecycle-aware executor design while preserving the restricted-caller guarantee.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Valid, with one scope clarification: the existing static channels map and per-instance stub caches already retain target keys, but resolvedTargets, nextResolutions, and channelLocks add more permanent state for every historical target. Only refreshTasks is removed today. A complete fix needs a target/client lifecycle operation that removes all associated metadata and retires the channel pool atomically, rather than clearing only one of the new maps.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Valid blocking finding. createChannels catches Exception around createChannel(target), so an Error still reaches finally and decrements the latch while leaving value[fi] null. With failure unchanged, the partially populated array can be returned and published; later stub construction can dereference the null entry and successful sibling channels are not cleaned up. The creation task must capture the throwable (while preserving fatal-error semantics as appropriate) or validate every slot before publication, terminate partial channels, and propagate the failure.

bitflicker64:
Current-head CI triage for fad20a3c: the dependency-check job did not execute dependency analysis; actions/setup-java failed while downloading JDK 11 after repeated socket hang up errors, so that failure is external setup/network noise. The Codecov failures are current and substantive as reported: patch coverage is 0% for two changed QueryV2Client lines and project coverage is 34.73%. The Store job itself passed, but its uploaded aggregate coverage did not demonstrate those changed lines to Codecov. The three new exact-head review findings were validated and answered inline; all three require code/lifecycle follow-up before the review can be considered clear.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in exact head `bdd8df11`. The trusted SecurityManager boundary is preserved with bounded prestarted pools: 4 maintenance threads, 4 cold-initialization threads, and 1 retirement thread (9 total rather than 129). The focused Java 11 `AbstractGrpcClientTest` suite now passes 37/37 with zero failures/errors, including executor footprint and restricted-caller coverage. Three independent reviews and the final Design Audit found no blocking issue on this head.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in exact head `bdd8df11`. `closeChannel(target)` and client `close()` now remove channel/stub pools plus all target-keyed resolution, deadline, refresh, and lock state; terminal Store notices invoke exact-node eviction, and stale sessions cannot recreate evicted state. Same-address replacements and duplicate stale notices are identity-safe. The Java 11 focused suite passes 37/37; independent lifecycle/concurrency review and the Design Audit found no blocker.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Fixed in exact head `bdd8df11`. Channel tasks capture `Throwable`, all submitted tasks converge, every slot is validated before publication, and partial channels are force-terminated on failure. Fatal `Error` is rethrown after cleanup, null pools cannot publish, creation submission/shutdown races cannot strand the latch, and interrupted callers regain interrupt status. The focused Java 11 suite passes 37/37, including `Error`, null-channel, partial-cleanup, submission, shutdown, and interrupt regressions; independent failure-path review and the Design Audit found no blocker.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Final exact-head confirmation on `bdd8df11`: both cold initialization and DNS refresh remain behind bounded prestarted trusted executors, so a denied Gremlin caller performs neither DNS/socket work nor thread creation. The focused Java 11 suite passes 37/37, including the real restricted-caller cold and refresh paths. The final concurrency review and Design Audit found no blocker.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Final exact-head confirmation on `bdd8df11`: QueryV2 still delegates through guarded `getAsyncStub(target)`, directly injected test channels are closed on replacement, and ordinary QueryV2 clients participate in inherited lifecycle cleanup. The focused Java 11 suite passes 37/37; JaCoCo records the changed injected-channel cleanup lines as covered (`mi=0`). No blocker remained in independent API/failure review or the Design Audit.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/grpc/AbstractGrpcClient.java
bitflicker64: Final exact-head confirmation on `bdd8df11`: first-run state remains explicit (`null` deadline), and deadline evaluation still uses overflow-safe signed subtraction. Negative-origin and `Long.MAX_VALUE`/`Long.MIN_VALUE` wraparound regressions remain in the 37/37 passing Java 11 focused suite. No blocker remained in the final reviews or Design Audit.

imbajin:
## Design direction to validate: keep Store identity at the DNS target layer

The current head (`bdd8df11`) shows that explicit address fingerprinting and pool replacement can be made defensive, but the added concurrency and lifecycle machinery suggests that the ownership boundary needs another look.

In Kubernetes, HugeGraph should use a stable logical target such as `store-1.<headless-service>:8500`, not the Pod IP behind it:

```mermaid
flowchart LR
    HG["HugeGraph Store client<br/>stable DNS target"]
    CH["gRPC ManagedChannel"]
    NR["gRPC NameResolver<br/>resolve + reconnect"]
    DNS["Kubernetes DNS"]
    OLD["old Store Pod<br/>10.0.0.8"]
    NEW["replacement Store Pod<br/>10.0.0.19"]

    HG --> CH --> NR --> DNS
    DNS -. "before replacement" .-> OLD
    DNS -- "after replacement" --> NEW
```

```text
Kubernetes       assigns the new Pod IP and updates DNS
Java DNS policy  permits a fresh answer (#3126)
gRPC             resolves the target and reconnects the transport
HugeGraph        owns Store identity, session state, retry, and shutdown
```

Under this model, HugeGraph does not need IP fingerprints, refresh deadlines, DNS worker pools, or a second connection-replacement state machine. It keeps the stable target and invalidates only the failed logical or transport state required for retry.

The validation plan is:

1. Start from merged #3126 and #3128 without this PR's DNS fingerprint and pool-replacement machinery.
2. Keep the same `ManagedChannel` for a stable Store DNS target.
3. Replace the Store Pod while keeping the DNS name and Store ID unchanged.
4. Keep the Server JVM alive and observe whether gRPC resolves the new IP after `UNAVAILABLE`.
5. If native recovery works, reduce this PR to the missing HugeGraph node, session, and retry lifecycle.
6. If it does not work, identify the exact failing boundary and add only the smallest proven invalidation mechanism.

```text
same Store DNS + same node ID + new Pod IP
    -> no Server restart
    -> bounded read/write recovery
    -> no data-integrity regression
```

Three independent tracks will validate this direction: gRPC and JDK source behavior, a minimal Java 11 runtime experiment, and a blind root-cause analysis.

imbajin:
## Validation update: native gRPC recovery works; refocus this PR on the HugeGraph lifecycle

Three independent tracks reached the same conclusion: gRPC and JDK source analysis, a Java 11 runtime experiment, and a blind root-cause analysis.

### Native gRPC recovery

The final Server distribution uses gRPC 1.47.0 for the behavior-driving api/core/netty/stub artifacts. Protobuf-related artifacts remain at 1.39.0, so the runtime is a mixed classpath.

The runtime experiment used Java 11.0.18, `networkaddress.cache.ttl=30`, one stable `dns:///store-lab:50051` target, and three distinct backend IPs:

```text
A  192.168.157.2
       |
       v
B  192.168.157.4
       |
       v
C  192.168.157.6
```

- All 32 original `ManagedChannel` instances recovered from A to B without channel replacement. The pool recovered about 33 seconds after the initial resolution.
- One original channel recovered from B to C in the same JVM about 31 seconds after the switch.

The relevant gRPC path is:

```text
old transport fails
  -> InternalSubchannel enters IDLE or TRANSIENT_FAILURE
  -> pick_first requests name-resolution refresh
  -> DnsNameResolver resolves after its TTL gate
  -> Subchannel receives the new address and reconnects
```

This shows that a Store IP change does not require HugeGraph to destroy and rebuild its channel pool. `resetConnectBackoff()` only shortens an active transport backoff; it does not bypass DNS TTL. `enterIdle()` rebuilds resolver and load-balancer state for the whole channel, so it should not be the default response to an RPC failure.

### HugeGraph lifecycle gap

The PR evicts the failed `HgStoreNode`, but `NodeTxExecutor.openNodeSession()` still caches `HgStoreSession` by node ID:

```text
transport failure
  -> evict current HgStoreNode
  -> retry obtains the current node with the same Store ID
  -> session cache returns the old session
  -> old session's node is no longer current
  -> retry cannot obtain a usable stub
```

The retry must replace a cached session when its node is no longer current. The regression needs the same Store ID and stable DNS target and must prove that the same transaction retry opens a current session.

### Refactor boundary

```mermaid
flowchart LR
    K8S["Kubernetes<br/>Pod IP + DNS"]
    JDK["Java DNS policy<br/>finite TTL from #3126"]
    GRPC["gRPC ManagedChannel<br/>resolve + reconnect"]
    HG["HugeGraph<br/>node + session + retry"]

    K8S --> JDK --> GRPC --> HG
```

The refactor should remove the external IP fingerprint, five-second polling, refresh executors, and whole-pool replacement. HugeGraph should keep the stable DNS target and fix only its node, session, retry, and explicit shutdown lifecycle.

The remaining integration test is a resource-bounded Kubernetes run on the exact `master + PR` result, recording the Store identity and IP change, Server process identity, first failure and recovery, and final data.

imbajin:
## K8s A/B result: the current PR cannot recover the in-flight retry

I ran a `1 PD + 1 Store + 1 Server` Kubernetes test. HugeGraph used the stable Store FQDN `pr3130-hugegraph-store-0.pr3130-hugegraph-store.pr-3130-hstore.svc:8500`, not a Pod IP. In each phase, the Store Pod received a new UID and IP while its PVC, Store ID, and FQDN stayed unchanged. The Server Pod and JVM did not restart.

| Phase | Store replacement | In-flight write | Later recovery | Server restart |
| --- | --- | --- | --- | --- |
| Base + #3126, without #3130 | IP `192.168.194.8 -> .10`; Store ID unchanged | HTTP 500 after 128.19s with `UNAVAILABLE` | HTTP 201 about 199s after deletion | No |
| Exact base + current PR (`4066d2ff`) | IP `192.168.194.10 -> .12`; Store ID unchanged | HTTP 500 after 38.39s with `The gRPC client is closed` | The next transaction succeeded immediately; then 10/10 writes passed | No |

The exact merge includes #3126 and uses `networkaddress.cache.ttl=30`. The behavior-driving gRPC core/API artifacts are 1.47.0; protobuf-related artifacts remain at 1.39.0.

```text
Store transport fails
  -> NOT_WORK evicts the current HgStoreNode and closes its target
  -> NodeTxExecutor retry reuses the session cached by node ID
  -> cached session still points to the evicted node
  -> node::isCurrent is false
  -> every remaining attempt sees "The gRPC client is closed"

next HTTP transaction
  -> creates a new transaction and session map
  -> obtains the current Store node and session
  -> succeeds immediately
```

This reproduces the source-level finding: the current PR invalidates the failed node and target, but the same transaction cannot replace its stale session.

Before merge, the PR needs to:

1. Replace a cached session when its `HgStoreNode` is no longer current.
2. Keep the stable Kubernetes FQDN as the Store target.
3. Let gRPC own DNS resolution and reconnect instead of maintaining IP fingerprints and rebuilding all channels.
4. Add a regression where the same Store ID and FQDN move to a new IP and the same transaction retry opens a current session.

Short client-side timeouts can leave Server-side requests running, so the final acceptance test must use a unique operation ID and distinguish an observed HTTP response from a later commit.

imbajin (commented):
Blocking: no. Summary: No new actionable findings were identified on the exact head after local review and focused Java 11 validation. Evidence: `mvn -q test -pl hugegraph-store/hg-store-test -am -Dtest=AbstractGrpcClientTest -Dsurefire.failIfNoSpecifiedTests=false -Djacoco.skip=true` (37/37 passed); `gh -R apache/hugegraph pr checks 3130` (17/17 passed).

imbajin:
## Final result: what changed and what the Kubernetes test proves

The final refactor is on head `18ca057856684ff09ab8ac46951d30de84c0aca7`.

### Before and after

The previous version assumed that HugeGraph had to detect every Store IP change itself. It added five-second DNS polling, address fingerprints, several maintenance executors, and full channel-pool replacement. Runtime testing showed that this duplicated work already owned by Kubernetes DNS, the JVM DNS policy from #3126, and gRPC name resolution.

The actual failure was in HugeGraph's retry state. After `UNAVAILABLE`, HugeGraph discarded the failed Store node, but the same transaction reused the session cached under the same Store ID. That session still belonged to the old node, so every retry failed even after gRPC could reach the replacement Store.

The final implementation keeps the stable Store FQDN and makes three focused changes:

- `UNAVAILABLE` discards only the exact failed node and target. `CANCELLED` and unrelated failures do not evict the node.
- The same transaction retry replaces a cached session when its Store node is no longer current.
- Blocking and async stubs are rebuilt only when their channel generation changes.

This reduced the PR from 2,295 to 486 added lines. Production additions fell from 732 to 140, and `AbstractGrpcClient` fell from 671 to 95 added lines. The deleted code implemented the duplicate DNS polling and connection-replacement state machine, not required recovery behavior.

### Exact Helm/Kubernetes acceptance

The exact `master + PR` image passed both the minimal `1 Server + 1 PD + 1 Store` test and a replicated `1 Server + 3 PD + 3 Store` test through the HugeGraph Helm chart.

| Scenario | Observed result |
| --- | --- |
| Minimal topology | Store Pod UID/IP changed while PVC, Store ID, and FQDN stayed stable; the in-flight request returned `201` after about 31 seconds without restarting Server |
| Rolling replacement | Store-1 rejoined with a new Pod UID/IP and the same PVC, Store ID, and FQDN; the tested 120-vertex transaction returned `201` |
| Leader Store failure | Store-0 was force deleted while leading 7/12 partitions; the replacement was still Pending when a 120-vertex transaction began; logs showed `UNAVAILABLE -> NOT_WORK -> retry`, and the request returned `201` after about 35 seconds |
| Follower Store failure | The same Store was force deleted again after it had zero Leaders; the tested 60-vertex transaction returned `201` immediately |
| Final cluster state | All 3 Stores were Up; all 12 partitions had 3 normal replicas; the final scan observed all 420 expected uniquely identified vertices |
| Recovery checks | A later write returned `201`, its delete returned `204`, and 10/10 scans returned `200`; logs contained no `gRPC client is closed` |
| Process and resources | The Server Pod UID, JVM PID, and process start time stayed unchanged with 0 restarts; every test container reported 0 OOM events |
| Cleanup | The dedicated namespace, PVCs, PVs, and image were removed; the pre-existing 3 PD + 3 Store cluster kept the same Pod/PVC identities and restart counts |

These results cover both leader movement and follower replacement while Kubernetes changes Pod IPs behind stable StatefulSet DNS names. They support the PR's intended Store reconnect and session-rebind path; they do not claim cross-process exactly-once commit behavior.

### Other verification

- Java 11 Store client suite: 9/9 passed.
- Compile, EditorConfig, Checkstyle, Apache RAT, and `git diff --check`: passed.
- Exact-merge Docker build: 38/38 modules passed.
- Exact-head GitHub CI: 17/17 checks passed.
- All independent review lanes passed, with 0 unresolved review threads.

### Future TODOs

These items are not blockers for this PR:

- **P1, data integrity:** define the outcome when a Store commits a write but the response is lost. Start with a failure-injection test using a non-idempotent operation, then choose either stable batch-ID deduplication or an explicit unknown-commit result.
- **P2, regression coverage:** automate the now-passing `1 + 3 + 3` leader/follower Store replacement scenario so future channel/session changes cannot regress it.

imbajin (approved):

PR #3134, opened 2026-08-02
feat(rocksdb): add optional ToplingDB integration

kenssa4eedfd:
## Purpose of the PR                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                         
  This PR introduces ToplingDB as an optional RocksDB-compatible storage engine                                                                                                                          
  for HugeGraph.                                                                                                                                                                                         
                                                                                                                                                                                                         
  ToplingDB is explicitly enabled through configuration. Standard RocksDB remains                                                                                                                        
  the default, so existing deployments and data paths are unaffected unless                                                                                                                              
  `rocksdb.provider=topling` is configured.                                                                                                                                                              
                                                                                                                                                                                                         
  ## Main Changes                                                                                                                                                                                        
                                                                                                                                                                                                         
  - Add explicit storage engine selection through `rocksdb.provider`.                                                                                                                                    
  - Integrate ToplingDB with HugeGraph Server, PD, and Store.                                                                                                                                            
  - Use ToplingDB Easy Migrate YAML configuration while continuing to use the                                                                                                                            
    standard RocksDB Java API.                                                                                                                                                                           
  - Add component-specific configurations for Server, PD, and Store, including                                                                                                                           
    separate HTTP ports and JRaft column-family settings.                                                                                                                                                
  - Add startup scripts that:                                                                                                                                                                            
    - Select and preload the ToplingDB JNI library.                                                                                                                                                      
    - Extract required native libraries and Web UI resources.                                                                                                                                            
    - Configure `LD_PRELOAD`, `LD_LIBRARY_PATH`, and                                                                                                                                                     
      `TOPLINGDB_EASY_MIGRATE_CONF`.                                                                                                                                                                     
    - Validate the operating system and architecture.                                                                                                                                                    
    - Verify downloaded dependencies using SHA-256.                                                                                                                                                      
    - Handle jemalloc and Ubuntu 24.04 `libaio` compatibility.                                                                                                                                           
  - Align the RocksDB JNI version across Server, PD, and Store.                                                                                                                                          
  - Update CI workflows to install and test the ToplingDB runtime dependencies.                                                                                                                          
  - Update third-party dependency metadata, LICENSE information, operational                                                                                                                             
    documentation, troubleshooting guides, and security notes.                                                                                                                                           
                                                                                                                                                                                                         
  ## Compatibility                                                                                                                                                                                       
                                                                                                                                                                                                         
  - Standard RocksDB remains the default provider.                                                                                                                                                       
  - Existing RocksDB data and APIs remain compatible.                                                                                                                                                    
  - ToplingDB is enabled only when `rocksdb.provider=topling` is configured.                                                                                                                             
  - No public HugeGraph API changes are introduced.                                                                                                                                                      
  - The bundled ToplingDB runtime currently targets Linux x86_64.                                                                                                                                        
                                                                                                                                                                                                         
  ## Verifying these changes                                                                                                                                                                             
                                                                                                                                                                                                         
    - Build all modules:                                                                                                                                                                                 
      ```bash                                                                                                                                                                                            
      mvn clean compile -Dmaven.javadoc.skip=true                                                                                                                                                        
      ```                                                                                                                                                                                                
    - Run HugeGraph Server tests with the RocksDB backend:                                                                                                                                               
      ```bash                                                                                                                                                                                            
      mvn test -pl hugegraph-server/hugegraph-test -am \                                                                                                                                                 
        -P core-test,rocksdb                                                                                                                                                                             
      ```                                                                                                                                                                                                
    - Run PD tests:                                                                                                                                                                                      
      ```bash                                                                                                                                                                                            
      mvn install -pl hugegraph-struct -am -DskipTests                                                                                                                                                   
      mvn test -pl hugegraph-pd/hg-pd-test -am                                                                                                                                                           
      ```                                                                                                                                                                                                
    - Run Store tests:                                                                                                                                                                                   
      ```bash                                                                                                                                                                                            
      mvn test -pl hugegraph-store/hg-store-test -am                                                                                                                                                     
      ```                                                                                                                                                                                                
    - Verify Server, PD, and Store startup in both standard RocksDB mode and                                                                                                                             
      ToplingDB mode.                                                                                                                                                                                    
                                                                                                                                                                                                         
  - [x] Dependencies                                                                                                                                                                                     
  - [x] Modify configurations                                                                                                                                                                            
  - [ ] The public API                                                                                                                                                                                   
  - [x] Other affects: startup scripts, native library loading, and CI workflows                                                                                                                         
  - [ ] Nope                                                                                                                                                                                             
                                                                                                                                                                                                         
  ## Documentation Status                                                                                                                                                                                
                                                                                                                                                                                                         
  - [ ] `Doc - TODO`                                                                                                                                                                                     
  - [x] `Doc - Done`                                                                                                                                                                                     
  - [ ] `Doc - No Need`


hugegraph-server/hugegraph-dist/pom.xml
```diff
@@ -112,6 +112,9 @@
             <resource>
                 <directory>${basedir}/src/assembly/static</directory>
                 <filtering>true</filtering>
+                <excludes>
+                    <exclude>lib/topling/**</exclude>
```
bitflicker64, line 116: ‼️ This exclusion exists to stop Maven resource filtering from corrupting a binary that should not be in the tree. `src/assembly/static/lib/topling/rocksdbjni-8.10.2-20260725.141011-1.jar` is 9.5 MB (`size: 9522363` from the contents API) and it is a timestamped SNAPSHOT: unreproducible, and it lands in the source release.

You already added the repository it comes from, three files over in `.github/configs/settings.xml`:

```xml
<url>https://maven.pkg.github.com/hugegraph/toplingdb</url>
```

Declare it as an optional/profile-scoped dependency, or `curl` it in `prepare-topling.sh` from that same URL. Then the JAR, this `<excludes>`, the matching `<exclude>lib/topling/**</exclude>` in `assembly.xml`, the 202-line `LICENSE-rocksdbjni-8.10.2-SNAPSHOT.txt`, and the `ls -1 "$lib_dir"/rocksdbjni*.jar | sort -V | tail -n1` pick in `common-topling.sh` all go away in one move.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphIndexTransaction.java
```diff
@@ -857,6 +857,18 @@ private MatchedIndex collectMatchedIndex(SchemaLabel schemaLabel,
 
     private ConditionQuery constructSearchQuery(ConditionQuery query, MatchedIndex index) {
         ConditionQuery newQuery = query;
+        ConditionQuery filterQuery = query;
+        Query rootQuery = query.rootOriginQuery();
```
bitflicker64, line 861: ‼️ This is a search-index filter fix - checking an `IN` condition against the root query instead of the flattened `EQ` sub-query. It has nothing to do with `rocksdb.provider`.

Same for the transaction-leak set that rides along with it: `HugeFactory.closeCurrentThreadTransactions`, `StandardHugeGraph.closeCurrentThreadTransaction`, `OltpTraverser.close()`, the `cleanup` hook on `HugeGraphAuthProxy.ContextTask`, the two new `HugeFactoryAuthProxy` reflection registrations, and the new `HugeFactoryTest` / `ContextTaskTest` / `OltpTraverserTest`. That is roughly 400 lines, applied unconditionally to every provider, inside a PR whose description only claims an opt-in storage engine and states "No public HugeGraph API changes".

Split them into their own PR - they will get a review that is actually about them, and this one shrinks to the ToplingDB change. If any of it is a ToplingDB prerequisite, name which and why in the description; right now nothing in the PR explains it.

hugegraph-pd/hg-pd-dist/src/assembly/static/conf/rocksdb_pd.yaml
```diff
@@ -0,0 +1,162 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# common parameters
+http:
+  # normally parent path of db path
+  document_root: ./library/rocksdb_resource
+  listening_ports: '127.0.0.1:2012'
```
bitflicker64, line 21: ⚠️ This file, `hugegraph-store/.../conf/rocksdb_store.yaml` and `hugegraph-server/.../conf/toplingdb.yaml` are byte-identical apart from this one line:

```
$ diff rocksdb_pd.yaml toplingdb.yaml
22c22
<   listening_ports: '127.0.0.1:2012'
---
>   listening_ports: '127.0.0.1:2011'
```

(`rocksdb_store.yaml` differs only by `2013`.) That is 486 new lines to vary a port three ways. The description says "component-specific configurations for Server, PD, and Store, including separate HTTP ports and JRaft column-family settings" - the column-family settings are identical too.

Ship one `toplingdb.yaml`. `preload-topling.sh` already computes the component and exports `TOPLINGDB_EASY_MIGRATE_CONF`; have it export the port alongside, or template the single file through the resource filtering the assembly already applies to `static/`. 324 lines deleted.

While you are in there: lines 24-30 are ToplingDB's upstream sample placeholders, shipped verbatim as production config in all three copies -

```yaml
setenv:
  StrSimpleEnvNameNotOverwrite: StringValue
  OverwriteThisEnv:
    value: force overwrite this env by overwrite true
```

hugegraph-server/hugegraph-dist/src/assembly/static/bin/common-topling.sh
```diff
@@ -0,0 +1,288 @@
...
+           [ -e /usr/lib/x86_64-linux-gnu/libaio.so.1t64 ]; then
+            mkdir -p "$dest_dir"
+            ln -sfn /usr/lib/x86_64-linux-gnu/libaio.so.1t64 \
+                    "$dest_dir/libaio.so.1"
+            echo "Prepared component-local libaio.so.1 compatibility link"
+        fi
+    fi
+}
+
+function download_and_verify() {
```
bitflicker64, line 128: ⚠️ `download_and_verify` already exists at `hugegraph-store/hg-store-dist/src/assembly/static/bin/util.sh:310`, and the new `<file>` entry in `hg-store-dist/src/assembly/descriptor/server-assembly.xml` drops `common-topling.sh` into that same `bin/`. So on Store this definition silently overrides the one `start-hugegraph-store.sh` sourced ~30 lines earlier, and the two disagree on whether argument 3 is an MD5 or a SHA-256.

The copy is also the weaker of the two. The existing one downloads to `mktemp` and `mv -f`, with a comment saying why:

```bash
# mktemp, not $$: a PID is shared by concurrent background subshells.
```

This one does `rm -f "$filepath"` and then `curl -o "$filepath"` - the exact non-atomic shape that comment was added to remove.

Source `util.sh` and teach the existing function a sha256 mode, instead of shipping a second implementation next to it. Same for `download_and_setup_jemalloc` below: `start-hugegraph-store.sh:42-64` already does the arch-select, the download and the `LD_PRELOAD`, and your version short-circuits on `LD_PRELOAD` anyway, so on Store it is 40 lines that never run.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/verify-rocksdb-provider.sh
```diff
@@ -0,0 +1,148 @@
...
+if [ "$#" -lt 3 ] || [ "$#" -gt 4 ]; then
+    echo "Usage: $0 <server|pd|store> <rocksdb|topling> <data-paths> [enforce]" >&2
+    exit 2
+fi
+
+COMPONENT="$1"
+PROVIDER="$2"
+DATA_PATHS="$3"
+ENFORCE="${4:-false}"
+MARKER_NAME=".hugegraph-rocksdb-provider"
```
bitflicker64, line 30: ⚠️ Keep the marker - refusing to open Topling data as plain RocksDB is the right guard. It is the 148 lines around it I would cut.

`flock` plus `/proc/self/fd` pinning plus two `stat -Lc '%d:%i'` TOCTOU re-checks defends against someone racing the data directory. That is not the failure being prevented; the failure is an operator flipping one line in a config file. And it buys Linux-plus-GNU-coreutils-only for a feature that is already Linux x86_64 only in a different way.

More to the point, `grep -rn verify-rocksdb-provider.sh` finds callers in exactly three places: the Server, PD and Store `docker-entrypoint.sh`. A tarball install - `bin/start-hugegraph-pd.sh`, `init-store.sh`, the test harness - never runs it, so the guard is absent precisely where hand-edited config files live.

`RocksDBStore.open()` already reads `RocksDBOptions.PROVIDER` and is the one place the DB is actually opened. Write and compare the marker file under `this.dataPath` right there, next to the existing `PROVIDER` read. A few lines, every launch path covered, and this script plus the `<file>` entries added to both `server-assembly.xml` descriptors go away.

bitflicker64 (commented):
Blocking: no. Summary: The ToplingDB switch itself is small - one `rocksdb.provider` option and a `truncate()` branch - but it ships wrapped in ~400 lines of unrelated transaction and search-index bug fixes, a 9.5 MB timestamped SNAPSHOT JAR committed into the source tree, three byte-identical 162-line YAML files, and a second copy of `download_and_verify`; the single biggest simplification is deleting the vendored JAR in favour of the Maven repository this PR already adds to `.github/configs/settings.xml`. Evidence: net diff for head 18635f6 via `gh api repos/apache/hugegraph/pulls/3134` (108 files, +6577/-207); `diff` of `rocksdb_pd.yaml`, `rocksdb_store.yaml` and `toplingdb.yaml` reports one differing line each (`listening_ports`); `grep -n 'download_and_verify' hugegraph-store/hg-store-dist/src/assembly/static/bin/util.sh` -> line 310, with the jemalloc arch-select already at `start-hugegraph-store.sh:42-64`; `gh api repos/apache/hugegraph/contents/.../lib/topling/rocksdbjni-8.10.2-20260725.141011-1.jar?ref=18635f6` -> `size: 9522363`; `grep -rn verify-rocksdb-provider.sh` shows callers only in the three `docker-entrypoint.sh` files.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/util.sh
```diff
@@ -24,6 +24,75 @@ function command_available() {
...
+    fi
+    dir=$(cd -P "$(dirname "$path")" 2>/dev/null && pwd) || return 1
+    printf '%s/%s\n' "$dir" "$(basename "$path")"
+}
+
+# Print JAR paths below a component's lib directory except anything whose
+# canonical target lives below its optional Topling subtree. `find -P` keeps
+# directory aliases from being traversed, while direct file aliases are
+# resolved and emitted canonically after the exclusion check.
+function find_standard_lib_jars() {
```
bitflicker64, line 59: ⚠️ This helper and `canonical_path` above it (69 lines, including a hand-rolled `realpath` with a 40-level symlink depth limit) exist to do one thing: keep `lib/topling/` out of the Server classpath scan. `find` already does that with `-prune`:

```bash
CP=$(find -L "$LIB" -path "$LIB/topling" -prune -o -name 'log4j-slf4j-impl*.jar' -print | sort | tr '\n' ':')
CP="$CP":$(find -L "$LIB" -path "$LIB/topling" -prune -o -name 'hugegraph*.jar' -print | sort | tr '\n' ':')
CP="$CP":$(find -L "$LIB" -path "$LIB/topling" -prune -o -name '*.jar' \! -name 'hugegraph*' \! -name 'log4j-slf4j-impl*.jar' -print | sort | tr '\n' ':')
```

Same in `init-store.sh`. The canonical-path check guards against a symlink elsewhere in `lib/` pointing into `lib/topling/`; `lib/` is the operator's own install directory, not input from outside, so that case does not need code. Even smaller: ship the Topling JAR outside `lib/` (for example `topling/` next to it), and the original `find` lines stay untouched.

hugegraph-pd/hg-pd-dist/src/assembly/static/bin/util.sh
```diff
@@ -394,3 +394,21 @@ function kill_process_and_wait() {
     kill_process "$process_name" "$pid"
     wait_for_shutdown "$process_name" "$pid" "$timeout_s"
 }
+
+# Find HugeGraph server directory in parent path using prefix glob.
+# Usage: find_hugegraph_server_dir "/path/to/parent"
+# Returns: first matching directory path or empty string
+function find_hugegraph_server_dir() {
```
bitflicker64, line 401: ‼️ `find_hugegraph_server_dir` has no callers. `git grep find_hugegraph_server_dir f2f3003` returns only this definition and its copy at `hugegraph-store/hg-store-dist/src/assembly/static/bin/util.sh:441`. Delete both (36 lines); add it back in the PR that first calls it.

hugegraph-server/hugegraph-dist/src/assembly/static/bin/common-topling.sh
```diff
@@ -0,0 +1,288 @@
...
+    while [[ -h "$SOURCE" ]]; do
+        local DIR
+        DIR="$(cd -P "$(dirname "$SOURCE")" && pwd)"
+        SOURCE="$(readlink "$SOURCE")"
+        [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
+    done
+    cd -P "$(dirname "$SOURCE")" && pwd
+}
+
+function extract_so_with_jar() {
```
bitflicker64, line 36: ⚠️ `extract_so_with_jar` and `extract_html_css_from_jar` are the same 35 lines twice, and most of each is checks that cannot fail here: `prepare_toplingdb` just found `$jar_file` with `ls`, so the existence check is dead, and `unzip` takes a relative path, so the `realpath`/`readlink -f` branch is not needed. What they do is:

```bash
unzip -j -o "$jar_file" '*.so' -d "$dest_dir" >/dev/null
```

The HTML/CSS half feeds the Topling HTTP monitor, which is off in all three shipped YAMLs (`auto_start_http: false`, line 22), and the caller already treats its failure as a warning. Drop it, or if you want it ready: `unzip -j -o "$jar_file" '*.html' '*.css' -d "$dest_dir/rocksdb_resource" >/dev/null || true`. Keep one `command -v unzip` check in `prepare_toplingdb` if you want the friendly message. The unused `abs_path` at the top of this file can go too; `prepare-topling.sh` computes `BIN` itself.

.specs/hugegraph-server/ToplingDB/task.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Lifecycle and Testing
+
+- [x] Use normal RocksDB Java close paths during graceful shutdown.
+- [x] Rely on native `MaybeForgetCF` and `MaybeForgetDB`; do not add Java-side
+  repository cleanup.
+- [x] Verify JAR/native mapping without constructing a plugin repository.
+- [x] Run every ToplingDB DB/CF functional test with Easy Migrate configuration
+  enabled.
+- [ ] Keep general Store shutdown timeout and unrelated thread-exit fixes in
```
bitflicker64, line 31: ‼️ `.specs/` is a new top-level directory holding this PR's working notes: a requirements list, a design doc, and a checkbox task list with an open item. Nothing else in the repo uses `.specs/`, and the design content (architecture, component scope, config contract, open/close lifecycle, rollback) is already in `docs/toplingdb/toplingdb.md` and `toplingdb-development.md` from this same PR. Delete the three files (239 lines). If a line in `design.md` is missing from `docs/toplingdb/`, move it there.

bitflicker64 (commented):
Blocking: no. Summary: The five simplification comments from the earlier review at 18635f6 still apply at f2f3003 (the new head is a master merge) and are not repeated here. Of what is left, the biggest cut is the 69-line `canonical_path` / `find_standard_lib_jars` pair in the Server `util.sh`, which exists only to keep `lib/topling/` off the classpath and can be one `-prune` per `find`. After that: delete the unused `find_hugegraph_server_dir` from PD and Store, collapse the two JAR extractors in `common-topling.sh` into one `unzip` line, and drop the `.specs/` planning files. Evidence: `git diff 83ef9f3...f2f3003` (109 files, +6593/-196); `git grep find_hugegraph_server_dir f2f3003` returns only the two definitions; `find -L "$LIB" -path "$LIB/topling" -prune -o -name '*.jar' ! -name 'hugegraph*' ! -name 'log4j-slf4j-impl*.jar' -print` checked against a fixture `lib/` with a `topling/` subdir and returns the same set as the current helper for plain files; `auto_start_http: false` at line 22 of `toplingdb.yaml`, `rocksdb_pd.yaml` and `rocksdb_store.yaml`.

Issue #3137, opened 2026-08-03
[Feature] Orchestrate graph creation through PD: new graphs are not consistently available across Server replicas ("Could not rebind [g]")


bitflicker64:
### Status (2026-08-13)

**Phase 1 (minimal patch) — merged.** [#3138](https://github.com/apache/hugegraph/pull/3138)
landed on `master` on 2026-08-12. The Server that handles `CreateGraph` now
waits for its own local Gremlin binding before returning HTTP 200.

**This issue stays open** for Phases 2–3 below: cluster-wide readiness
([#3139](https://github.com/apache/hugegraph/pull/3139)) and moving graph
creation ownership into PD so Servers become pure converging replicas.

Load-balanced multi-Server deployments can still see `Could not rebind` on
replicas that have not yet converged. The Helm chart documents mitigations in
PR [#3132](https://github.com/apache/hugegraph/pull/3132).

### Summary

In distributed mode (PD + HStore), creating a graph returns 200 before the
graph is usable everywhere, and in the worst case before it is usable
anywhere. The creating Server registers its own Gremlin binding through an
event it does not await, and every other Server replica converges
independently through a PD metadata watch plus a local backend open. There is
no upper bound on that window and no signal when it closes. Any deployment
that load-balances several Servers turns the window into a user-visible 400:
a query lands on a replica that has not yet bound the graph and fails with
TinkerPop's "Could not rebind" error.

I propose fixing this in three phases: make the creating replica synchronous
(a one-line parity fix), give clients a way to await cluster-wide
availability, and finally move graph creation into PD so Servers become pure
converging replicas of PD state. The last step is the same direction as
#3118 / #3119, which make Server-side init-store skippable in distributed
mode because the storage side already owns its metadata.

### Reproducible symptom

Setup: current master images, PD + HStore, two or more Servers behind one
load-balanced endpoint (in Kubernetes, a ClusterIP Service; any round-robin
proxy reproduces it). Timeline:

```
t0  client -> LB -> Server A : POST create graph "graph_164" (DEFAULT space)
t1  Server A : opens backend, writes config to PD meta, puts the graph in
               its local map, fires GRAPH_CREATE without awaiting it,
               returns 200
t2  client -> LB -> Server B : POST /gremlin with
               aliases {"g": "__g_DEFAULT-graph_164"}
t3  Server B : PD watch not yet delivered, or delivered but the backend
               open is still running; the Gremlin global bindings contain
[...]

PR #3138, opened 2026-08-03
fix(server): wait for GRAPH_CREATE event when creating graph on PD path

bitflicker64:
## Purpose

This is Phase 1 of #3137. It closes the local race on the Server that handles graph creation; cluster-wide readiness remains a follow-up.

In distributed mode (PD + HStore), graph creation previously returned HTTP 200 after scheduling `GRAPH_CREATE`, but before the embedded Gremlin Server had necessarily registered the graph and its `TraversalSource`. An immediate Gremlin or Cypher request to the same Server could therefore fail with:

```text
Could not rebind [g] to [__g_<name>]
```

## Before and after

```text
Before: create graph -> schedule GRAPH_CREATE -> publish/return 200 -> bind later
After:  create graph -> bind locally and verify listeners -> publish -> return 200
```

After this PR, HTTP 200 guarantees that the **creating Server** has completed its local REST and Gremlin registrations, including the `__g_<space>-<graph>` traversal binding.

## Changes

- Add synchronous `EventHub.notifySync()` dispatch while preserving the existing asynchronous `notify()` behavior for other callers.
- Report attempted and successful listeners from the same notification snapshot, avoiding false create results if listeners change concurrently.
- Run graph-create listeners and the standalone local drop listener synchronously because they define the request's local success boundary.
- Fail graph creation when any registered `GRAPH_CREATE` listener fails.
- On local binding failure, clear the local graph and remove partial REST/Gremlin registrations; make Gremlin removal safe when only part of the binding was installed.
- Publish the graph to PD metadata only after local binding succeeds.
- Carry graph identity through REST and Gremlin cleanup, and serialize local Gremlin registration/removal so a stale rollback cannot remove a replacement binding.

## Scope: intentionally not handled here

- **Other Server replicas may still be loading the graph after 200.** Requests routed to another replica can still need retry or sticky routing. Cluster-wide `LOADING / READY / FAILED` reporting is Phase 2 of #3137.
- Graph creation ownership is not moved into PD; that architectural change remains Phase 3 of #3137.
- Failures after local binding during PD metadata publication or `prepareSchema` are not made transactional here; they belong to the Phase 3 PD creation state machine.
- The distributed graph-drop path is unchanged.
- Full serialization of concurrent create/drop operations for the same graph name is not introduced; that belongs to the broader graph lifecycle state-machine work.
- Requests racing with Server shutdown are outside this local create-readiness guarantee.
- Listener deadlines and cancellation are unchanged; lifecycle listeners are expected to remain local and non-blocking.

## Verification

- `EventHubTest`: 21 tests passed on JDK 11, including synchronous dispatch, listener-failure counting, and snapshot mutation.
- `hugegraph-api` and all required reactor modules compiled successfully on JDK 11.
- The original distributed reproduction remains: create a graph and immediately query it on the creating Server. That Server must no longer return 200 before its Gremlin binding exists.

- [ ] Dependencies
- [ ] Modify configurations
- [ ] Public API contract
- [x] Internal graph lifecycle handling

## Documentation Status

- [x] Doc - No Need

### DIFF
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
@@ -1353,7 +1353,7 @@ public final class GraphManager {
         }
 
         // Let gremlin server and rest server context add graph
-        this.eventHub.notify(Events.GRAPH_CREATE, graph);
+        this.notifyAndWaitEvent(Events.GRAPH_CREATE, graph);
 
         if (init) {
             String schema = propConfig.getString(
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
@@ -34,8 +34,10 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -151,6 +153,13 @@ public final class GraphManager {
     public static final String DELIMITER = "-";
     public static final String NAMESPACE_CREATE = "namespace_create";
     private static final Logger LOG = Log.logger(GraphManager.class);
+    /*
+     * The graph create/drop listeners only do in-memory registrations (put the
+     * graph into the rest server context and into the gremlin server bindings),
+     * so they finish in microseconds on a healthy server. The bound is only a
+     * guard against a stuck or starved event worker.
+     */
+    private static final long EVENT_WAIT_TIMEOUT = 30L;
     private KvStore kvStore;
 
     private final String cluster;
@@ -1193,17 +1202,19 @@ public final class GraphManager {
 
             // Init graph and start it
             graph.create(this.graphsDir, this.globalNodeRoleInfo);
+
+            // Let gremlin server and rest server add graph to context
+            this.notifyAndWaitEvent(Events.GRAPH_CREATE, graph);
         } catch (Throwable e) {
             LOG.error("Failed to create graph '{}' due to: {}",
                       name, e.getMessage(), e);
             if (graph != null) {
+                // The create event may have added the graph to the context
+                this.graphs.remove(graph.spaceGraphName());
                 this.dropGraphLocal(graph);
             }
             throw e;
         }
-
-        // Let gremlin server and rest server add graph to context
-        this.notifyAndWaitEvent(Events.GRAPH_CREATE, graph);
 
         return graph;
     }
@@ -1342,18 +1353,36 @@ public final class GraphManager {
         graph.updateTime(timeStamp);
 
         String graphName = spaceGraphName(graphSpace, name);
+        this.graphs.put(graphName, graph);
+
+        /*
+         * Let gremlin server and rest server context add graph before the
+         * graph is published, so that a failed local binding can't leave the
+         * graph behind in meta for the other servers to converge on
+         */
+        try {
+            this.notifyAndWaitEvent(Events.GRAPH_CREATE, graph);
+        } catch (Throwable e) {
+            this.graphs.remove(graphName);
+            try {
+                graph.close();
+            } catch (Exception e1) {
+                if (graph instanceof StandardHugeGraph) {
+                    ((StandardHugeGraph) graph).clearSchedulerAndLock();
+                }
+            }
+            HugeFactory.remove(graph);
+            throw e;
+        }
+
         if (init) {
             this.creatingGraphs.add(graphName);
             this.metaManager.addGraphConfig(graphSpace, name, configs);
             this.metaManager.notifyGraphAdd(graphSpace, name);
         }
-        this.graphs.put(graphName, graph);
         if (!grpcThread) {
             this.metaManager.updateGraphSpaceConfig(graphSpace, gs);
         }
-
-        // Let gremlin server and rest server context add graph
-        this.notifyAndWaitEvent(Events.GRAPH_CREATE, graph);
 
         if (init) {
             String schema = propConfig.getString(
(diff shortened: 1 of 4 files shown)


imbajin (commented):
Blocking: yes. Summary: The PD create path now waits for GRAPH_CREATE listeners, but listener failures remain non-fatal and the current head has a failing codecov/project check. Evidence: EventHub.notify() catches listener Throwable at hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/event/EventHub.java:200-205; final gate reports codecov/project FAILURE.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -1353,7 +1353,7 @@ public HugeGraph createGraph(String graphSpace, String name, String creator,
         }
 
         // Let gremlin server and rest server context add graph
-        this.eventHub.notify(Events.GRAPH_CREATE, graph);
+        this.notifyAndWaitEvent(Events.GRAPH_CREATE, graph);
```
imbajin: ⚠️ This wait only guarantees that the listener task completed, not that graph registration succeeded. EventHub.notify() catches every listener Throwable at hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/event/EventHub.java:200-205, while notifyAndWaitEvent() catches future.get() failures at GraphManager.java:1777-1780. A failing ContextGremlinServer.injectGraph() can therefore leave the new graph absent from the Gremlin graph/traversal/global bindings while this PD create still returns success. Please propagate GRAPH_CREATE registration failures or verify the required bindings before responding.

imbajin: ⚠️ Making this path synchronous also introduces an unbounded request wait: notifyAndWaitEvent() calls future.get() without a timeout at GraphManager.java:1777, and its catch(Throwable) swallows InterruptedException without restoring the interrupt status. A hung listener or shutdown can now hold the create request indefinitely, or return before registration after interruption. Please use a bounded wait and handle InterruptedException explicitly.

bitflicker64: Right, fixed in 9a4ac02e. `EventHub.notify` resolves with the count of listeners that returned normally, so `notifyAndWaitEvent` now compares that against the registered listener count and fails the create when one is missing.

Rolling back `addGraphConfig` / `notifyGraphAdd` would let this replica's binding failure delete a graph the others bound fine, so the graph is instead bound locally before it is published to meta; the failure path unregisters and closes it, like a failed backend init.

The count is an inference, not the real exception. Happy to switch to verifying the bindings directly, or to having `EventHub` surface listener failures, if you prefer either.

bitflicker64: Agreed, fixed in 9a4ac02e: bounded `future.get(30s)`, and `InterruptedException` is caught on its own and restores the interrupt status before failing. The drop path gets the same bounded wait but stays lenient, since the data is already gone when the event fires and TinkerPop's `removeGraph` throws for a graph the Gremlin server never bound.

30s is a constant rather than an option, as `ServerOptions` has nothing comparable; happy to promote it if you want it tunable.

One call I would rather leave to you: a timeout currently fails the create, so a merely slow listener breaks a create that actually worked. Treating a timeout as unknown (log loudly, do not fail) is the alternative.

bitflicker64:
Both points addressed in 9a4ac02e, details in the inline replies.

Two things left out on purpose:

- The PD drop path (`GraphManager.java:2153`) still fires `GRAPH_DROP` without waiting, and passes the graph after `close()`. Same class of bug, pre-existing, on the drop side. Glad to fold it in if you want create and drop symmetric here.
- Creation is still publish-then-finish overall, since `prepareSchema` runs after the config is in meta. That is the rollback risk in #3137 that phase 3 is meant to close.

There is no automated coverage for the new failure semantics: nothing exercises `notifyAndWaitEvent` today, and the race itself needs a multi-server deployment. A unit test registering a deliberately throwing listener on a scratch `EventHub` would cover the count check and the cleanup cheaply, if you want it here.

`codecov/project` looks like a base-report artifact: it reports +11 files and +1150 lines against base, which this diff cannot produce, while `codecov/patch` passes.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -1771,10 +1800,60 @@ private void listenMetaChanges() {
...
-        Future<?> future = this.eventHub.notify(event, graph);
+        String graphName = graph.spaceGraphName();
+        // Listeners of ANY_EVENT are notified too, so they count as expected
+        int expected = this.eventHub.listeners(event).size() +
+                       this.eventHub.listeners(EventHub.ANY_EVENT).size();
+        int notified;
         try {
-            future.get();
+            Future<Integer> future = this.eventHub.notify(event, graph);
+            notified = future.get(EVENT_WAIT_TIMEOUT, TimeUnit.SECONDS);
```
imbajin: ‼️ The timeout bounds the caller's wait, but not the event task's side effects. `Future.get(timeout)` can throw while the `EventHub` worker remains queued or running; the caller then closes/removes the graph, and a late `injectGraph()` can register that closed graph and its TraversalSource afterward. Please make cancellation and cleanup deterministic so no listener can mutate bindings after rollback begins.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -1342,19 +1353,37 @@ public HugeGraph createGraph(String graphSpace, String name, String creator,
...
+
+        /*
+         * Let gremlin server and rest server context add graph before the
+         * graph is published, so that a failed local binding can't leave the
+         * graph behind in meta for the other servers to converge on
+         */
+        try {
+            this.notifyAndWaitEvent(Events.GRAPH_CREATE, graph);
+        } catch (Throwable e) {
+            this.graphs.remove(graphName);
```
imbajin: ‼️ This rollback removes only `GraphManager.graphs` and closes the graph. `ContextGremlinServer.injectGraph()` updates the Gremlin graph, TraversalSource, and global bindings in separate steps, so a later step can throw after earlier bindings were installed; the listener-count check detects the failure, but those partial bindings remain. Please explicitly undo the Gremlin-side registrations (or make injection atomic) before closing the graph; the local create failure path needs the same guarantee.

imbajin (commented):
Blocking: yes. The Phase 1 boundary is appropriate: this PR only needs to guarantee that the creating Server completes its local graph and TraversalSource bindings before returning 200; cross-replica readiness remains in #3137. The happy path now meets that scope, but the two local rollback races below can still leave stale bindings after a failed create. Codecov is intentionally ignored here.

imbajin (dismissed):
Re-reviewed current head `5569400b2` against the Phase 1 scope. No blocking findings.

- Graph creation now completes and validates local REST/Gremlin registration before publishing or returning success.
- Failed creation rolls back partial bindings and local caches; Gremlin registration/removal is synchronized and identity-guarded.
- Existing asynchronous `EventHub.notify()` behavior remains unchanged for other callers.
- The PR description clearly separates the creating-server guarantee from cross-replica readiness, PD lifecycle state, distributed drop, and full same-name lifecycle serialization.

Verification: `EventHubTest` passes 20/20 on JDK 11, and the `hugegraph-api` reactor compiles successfully on JDK 11. The remaining lack of a direct GraphManager/Gremlin integration test is a non-blocking coverage gap.

imbajin (approved):
Re-reviewed current head `92997d893` with six independent adversarial lanes. No in-scope blocking findings.

The listener outcome check now derives attempted and successful counts from the same COW snapshot, while the existing asynchronous EventHub API retains its behavior. Local binding failure rollback, identity guards, and synchronized Gremlin registration/removal remain sound.

Verification: EventHubTest passes 21/21 on JDK 11 and the hugegraph-api reactor compiles successfully. Residual risks are explicitly scoped: no direct GraphManager/Gremlin integration test, no listener deadline, and post-bind PD metadata transactions remain follow-up work.

bitflicker64:
@copilot-pull-request-reviewer[bot] Validated the final head `92997d893`. The synchronous path counts attempted and successful listeners from the same dispatch iterator; failed local bindings are rolled back through synchronous, identity-guarded Gremlin cleanup before the graph is removed and closed. The existing asynchronous `notify()` API remains unchanged. The review-comments API exposes no Copilot inline findings to address individually despite the summary reporting five. `EventHubTest` passes on Java 11 with `-DskipCommonsTests=false`; the intentional throwing-listener warning is covered by that test.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -1342,19 +1350,38 @@ public HugeGraph createGraph(String graphSpace, String name, String creator,
         graph.updateTime(timeStamp);
 
         String graphName = spaceGraphName(graphSpace, name);
+        this.graphs.put(graphName, graph);
```
imbajin, line 1353: ‼️ `graphs.put()` and the synchronous GRAPH_CREATE notification run before `addGraphConfig`, `notifyGraphAdd`, `updateGraphSpaceConfig`, and schema initialization. If any later step throws, this method exits without removing the local graph/bindings, closing and unregistering the graph, or compensating metadata already written, so the failed request can leave a graph that blocks retry and can be published to other servers. Please extend one rollback boundary across the publish/schema phase, with best-effort local binding and HugeFactory cleanup plus compensation for partially written PD metadata.

imbajin (commented):
Blocking: yes. The local GRAPH_CREATE synchronization is sound, but PD graph creation still publishes local state before later metadata and schema steps that can fail without a unified rollback; the current head also has a failing codecov/project check. Evidence: GraphManager.java:1353-1393; gh pr checks -R apache/hugegraph 3138.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
imbajin: @bitflicker64 Confirmed. This is a real follow-up gap: after local GRAPH_CREATE succeeds, a failure in PD metadata publication or schema initialization can still leave local graph/bindings and partially written metadata behind.

For the current PR, we are keeping this outside the minimal Phase 1 boundary (the creating Server's local bindings must be ready before 200). Please track it under [#3137](https://github.com/apache/hugegraph/issues/3137) as the Phase 3 graph-creation state-machine work, including compensating metadata cleanup and one rollback boundary across publication/schema initialization.

Marking this as a follow-up for now; I am leaving the thread unresolved until that work is addressed.

PR #3140, opened 2026-08-06
fix(hstore): preserve range-index ordering across partitions

### LINKED ISSUE #3053: [Bug]  HStore range-index scans don't guarantee global ordering or stable paging
### Bug Type (问题类型)

logic (逻辑设计问题)

### Before submit

### Environment (环境信息)

- Server Version: master / PR #2994 based branch
- Backend: HStore
- OS: GitHub Actions hstore CI

While investigating the hstore CI failure in #2994, we found an existing latent issue in the HStore range-index scan path.

For range-index queries with `limit`, `offset`, or paging, HugeGraph's upper layer assumes that backend range scan results are returned in global range-index-key order. It also assumes that the returned `PageState.position()` can be reused as a HugeGraph range cursor.

However, HStore's multi-node/tablet scan path can return entries in backend iterator order instead of globally sorted key order. The page state is also an internal storage cursor, not necessarily a HugeGraph range-index key. This can make range-index queries return unstable ordering or skip valid entries when paging is involved.

One concrete failure exposed by #2994 was:

```java
graph.traversal().V().hasLabel("person")
     .has("birth", P.between(date2013, date2016))
     .limit(2)
     .toList();
```
The expected range-index order is:
2013 -> 2014 -> 2015

But the HStore scan returned entries like:
2014 -> 2013 -> 2015

Then limit(2) selected the wrong first two entries.

Another paging-related failure showed that after the first page, the page position was an HStore internal cursor. Reusing it as the range scan start could skip valid range-index entries.

In #2994 we added a narrow workaround in GraphIndexTransaction: for HStore range-index queries whose visible result depends on limit, offset, or paging, the index layer reads the matched range-index entries, sorts them by range-index value, and slices them at the HugeGraph layer. Unbounded range-index scans still use the original streaming path to avoid disturbing count, joint-index, and cleanup paths.

This workaround fixes the immediate user-visible correctness issue, but the lower-level contract is still unclear.

### Expected behavior
HStore range scans should have a clear and reliable contract:

If HugeGraph range-index scan semantics require ordered results, HStore should return globally sorted entries across node/tablet iterators.
PageState.position() should have a well-defined meaning. It should be clear whether it is a backend-internal cursor or a HugeGraph key cursor.
Range-index paging should not skip valid entries or depend on accidental backend iterator order.

### Possible fix direction
A more complete fix should probably be handled in the HStore store-client / scan iterator layer:

- define whether IdRangeQuery results must be globally ordered by key;
- merge multiple node/tablet iterators by key order when serving ordered range scans;
- separate backend-internal page cursor semantics from HugeGraph range-key cursor semantics;
- add HStore-specific regression tests for:
  - range index + limit;
  - range index + offset;
  - range index + paging across multiple pages;
  - cross-node/tablet range scans;
  - count / joint-index / left-index cleanup paths to avoid regressions.

### Related context
This was exposed during #2994, but it does not seem to be caused by the query-condition refactoring itself. The PR only made the latent HStore issue visible in CI.

```javascript

```

```javascript

```

contrueCT:
## Purpose of the PR

- close #3053

HStore range-index queries with a limit, offset, or page cursor require globally ordered backend keys. The previous multi-partition path exposed partition iteration order and an internal storage cursor, which could return the wrong limited slice or skip entries on continuation.

This PR adds an explicit ordered range-scan path while leaving unbounded count, joint-index, and cleanup scans on the existing path.

## Typical Gremlin examples

These are the queries this change is meant to make predictable on HStore:

```gremlin
g.V().has('date', gte('2013')).limit(2)
g.V().has('age', gte(30)).range(0, 10)
g.V().has('score', gt(80)).limit(20)
```

Before this PR, the same logical query could return a slice in partition-local order. For example, when `2013` and `2014` live on different partitions, `limit(2)` could see `2014` before `2013`.

After this PR, ordered range scans are merged by full key across partitions, so the client sees one global order end to end.

## Main Changes

```mermaid
flowchart LR
    subgraph Store1["Store 1"]
        P1["Partition 1"] --> M1["Local K-way merge"]
        P2["Partition 2"] --> M1
    end
    subgraph Store2["Store 2"]
        P3["Partition 3"] --> M2["Local K-way merge"]
        P4["Partition 4"] --> M2
    end
    M1 --> G["Client global K-way merge"]
    M2 --> G
    G --> R["Ordered range-index results"]
```

- Preserve the next unread physical index key as the HugeGraph page cursor and preserve ordered index IDs through the following element lookup.
- Add `ORDER_BY_KEY` to the HStore scan protocol. Each Store merges its local leader-partition iterators, and the client globally merges one stream per Store instead of opening one stream per partition.
- Use fixed 64-entry pages, request later pages only when a Store's current page is exhausted, and bound concurrent first-page initialization to eight streams.
- Keep legacy scan ordering and cursor behavior unchanged for requests that do not explicitly select the ordered path. Client and Store nodes must be upgraded together before using ordered range scans.

## Verifying these changes

    - `ClientSuiteTest`: 13 tests passed.
    - `ServerSuiteTest`: 6 tests passed.
    - `HstoreSessionsImplTest` and `HstoreTableTest`: 8 tests passed.
    - `GraphIndexTransactionTest`, `IdHolderTest`, and `QueryResultsTest`: 4 tests passed.
    - `git diff --check` passed.

- [ ]  Modify configurations
- [x]  The public API
- [x]  Other affects: extends the HStore scan protocol; coordinated client and Store upgrades are required for the ordered path.
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO`
- [ ]  `Doc - Done`
- [x]  `Doc - No Need`


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/page/QueryList.java
```diff
@@ -258,7 +258,7 @@ private QueryResults<R> each(IdHolder holder) {
                     return null;
                 }
 
-                return this.queryByIndexIds(ids);
+                return this.queryByIndexIds(ids, holder.keepOrder());
```
imbajin, line 262: ⚠️ important — `keepOrder` is propagated to each batch `IdQuery`, but `QueryResults.keepInputOrderIfNeeded()` snapshots `queryIds()` before the lazy `FlatMapperIterator` has materialized later batches. When the index result exceeds `QUERY_BATCH`, `ids` contains only the first batch; after the map is filled, `map.size() > ids.size()` falls back to `map.keySet()`, which is backend fetch order. The new ordering path can therefore return multi-batch range-index results out of input order. Please accumulate all batch IDs before ordering and add a regression test with more than `QUERY_BATCH` IDs.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxSessionProxy.java
```diff
@@ -748,6 +807,44 @@ private List<NodeTkv> toNodeTkvList(String table, HgOwnerKey startKey, HgOwnerKe
...
+                                             .build())
+                            .setMethod(ScanMethod.RANGE)
+                            .setTable(nodeTkv.getTable())
+                            .setStart(toByteString(nodeTkv.getKey().getKey()))
+                            .setEnd(toByteString(nodeTkv.getEndKey().getKey()))
+                            .setLimit(scanLimit)
+                            .setCode(nodeTkv.getKey().getKeyCode())
+                            .setScanType(scanType)
+                            .setPageSize(ORDERED_SCAN_PAGE_SIZE)
+                            .setOrderType(ScanOrderType.ORDER_BY_KEY)
```
imbajin, line 839: ⚠️ important — This sends `ORDER_BY_KEY` as an optional protobuf field, but an older Store silently ignores unknown field 15 and processes the request via the legacy all-partition scan, which concatenates partitions rather than sorting by key. `OrderedKvIterator` then heap-merges an unsorted source and can return incorrect global order/limit/page results during rolling upgrades. Please negotiate support or fail closed with an operation old nodes reject, and add a mixed-version test.

imbajin (commented):
Blocking: yes. Summary: The ordered range-index path has a multi-batch ordering defect and is unsafe during mixed-version Store upgrades. Evidence: static review of QueryList/QueryResults and NodeTxSessionProxy/store_stream_meta.proto at head b45fdede3451c6f3115b46fe370f90c43e307f1c.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/QueryResults.java
```diff
@@ -102,12 +102,12 @@ public <T extends Idfiable> Iterator<T> keepInputOrderIfNeeded(
             return origin;
         }
         Collection<Id> ids;
-        if (!this.mustSortByInputIds() || this.paging() ||
+        if (!this.mustSortByInputIds() ||
```
imbajin: ‼️ critical — Removing the paging short-circuit makes `keepInputOrderIfNeeded()` call `fillMap(origin, map)` on the lazy page iterator. The new HStore paging path sets `mustSortByInputIds`, so an unbounded multi-page query is now fully materialized before returning, which can exhaust all pages and cause large memory use/OOM. Preserve lazy page boundaries while ordering each page, and add an unbounded multi-page regression test.

hugegraph-store/hg-store-node/src/main/java/org/apache/hugegraph/store/node/grpc/ScanUtil.java
```diff
@@ -71,7 +74,14 @@ static ScanIterator getIterator(ScanStreamReq request, HgStoreWrapperEx wrapper)
                 iter = wrapper.scanPrefix(graph, partition, table, prefix, scanType, query);
                 break;
             case RANGE:
-                iter = wrapper.scan(graph, partition, table, start, end, scanType, query);
+                if (partition == SCAN_ALL_PARTITIONS_ID &&
```
imbajin: ⚠️ important — The `ORDER_BY_KEY` routing is implemented only for `ScanStreamReq`. One-shot scans call `getIterator(toSq(request), wrapper)`, but `ScanQuery` drops `orderType` and its RANGE branch always invokes the legacy partition scan. An ordered one-shot request can therefore return partition-concatenated results and apply its limit incorrectly. Preserve the order flag through `ScanQuery` or route one-shot requests through the request overload, with an interleaved-partition test.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/OrderedMultiPartitionIterator.java
```diff
@@ -0,0 +1,215 @@
...
+        if (this.currentPartitionId == null) {
+            return EMPTY_BYTES;
+        }
+        return ByteBuffer.allocate(Integer.BYTES)
+                         .putInt(this.currentPartitionId)
+                         .array();
+    }
+
+    @Override
+    public void seek(byte[] position) {
```
imbajin, line 134: ⚠️ important — `ScanUtil.getIterator(ScanStreamReq)` unconditionally calls `iter.seek(request.getPosition())`, but this new iterator throws for every non-empty position. A continuation request carrying the previous ordered-scan cursor therefore fails instead of resuming the next page. Implement a per-partition resume strategy or reject this mode before opening the scan and document the contract; add a non-empty-cursor test.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/OrderedMultiPartitionIterator.java
```diff
@@ -0,0 +1,215 @@
...
+                this.closeSource(source);
+            }
+        } catch (RuntimeException | Error e) {
+            this.closeAfterFailure(e);
+            throw e;
+        }
+        return (T) current;
+    }
+
+    @Override
```
imbajin, line 123: ⚠️ important — This iterator does not override `ScanIterator.count()`, whose default is `0`; `FilterIterator` and `SelectIterator` delegate to it, while the existing `MultiPartitionIterator` sums source counts. Ordered scans therefore report zero to count callers. Implement consistent count semantics and add a count regression test.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/OrderedKvIterator.java
```diff
@@ -0,0 +1,254 @@
...
+        this.initialized = true;
+        List<Callable<SourceEntry>> tasks =
+                new ArrayList<>(this.iterators.size());
+        for (int i = 0; i < this.iterators.size(); i++) {
+            int source = i;
+            tasks.add(() -> this.firstEntry(source));
+        }
+        try {
+            List<Future<SourceEntry>> futures =
+                    this.initializer.invokeAll(tasks);
```
imbajin: ⚠️ important — `invokeAll(tasks)` waits for every first-entry task before observing any `ExecutionException`. If one Store fails quickly while another is slow or blocked, the failure cannot close the other streams until the slowest task returns. Use completion/future handling that cancels unfinished tasks on the first failure and closes all iterators in a `finally` path.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/OrderedKvIterator.java
```diff
@@ -0,0 +1,254 @@
...
+
+import org.apache.hugegraph.store.HgKvEntry;
+import org.apache.hugegraph.store.HgKvIterator;
+import org.apache.hugegraph.store.client.util.ExecutorPool;
+import org.apache.hugegraph.store.client.util.HgStoreClientConst;
+
+final class OrderedKvIterator implements HgKvIterator<HgKvEntry> {
+
+    private static final int INITIALIZE_THREADS = 8;
+    private static final ExecutorService INITIALIZER =
```
imbajin, line 42: ⚠️ important — All ordered scans share one fixed eight-thread initializer. Because each scan uses `invokeAll()` for every Store, slow partitions from eight concurrent queries can occupy the global pool and block unrelated queries behind them. Add bounded cancellation/timeouts and isolation or use an existing query-scoped executor; cover a slow-query concurrency case.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandler.java
```diff
@@ -69,6 +69,9 @@ void doPut(String graph, int code, String table, byte[] key, byte[] value) throw
     ScanIterator scan(String graph, int code, String table, byte[] start,
                       byte[] end, int scanType) throws HgStoreException;
 
+    ScanIterator scanOrdered(String graph, String table, byte[] start,
```
imbajin: ⚠️ important — Adding a new abstract method to the public `BusinessHandler` interface (and the matching abstract `scanOrdered` method on public `HstoreSessions.Session`) is a source and binary compatibility break for external implementations, plugins, and test doubles. Prefer a default capability method or a separate optional interface with a guarded fallback, and document the rollout requirement.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/NodeTxSessionProxyTest.java
```diff
@@ -0,0 +1,398 @@
...
+        Assert.assertEquals(10, first.getKey().getKeyCode());
+        Assert.assertEquals(20, first.getEndKey().getKeyCode());
+        Assert.assertEquals(30, second.getKey().getKeyCode());
+        Assert.assertEquals(40, second.getEndKey().getKeyCode());
+        Assert.assertEquals(7, first.getKey().getSerialNo());
+        Assert.assertEquals(8, first.getEndKey().getSerialNo());
+    }
+
+    @Test
+    public void testScanIteratorOrderedUsesOneStreamPerStoreLazily()
```
imbajin, line 97: ⚠️ important — The new tests use fake iterators and do not exercise the cross-module contract: two or more Stores/partitions, more than the 64-entry page, empty and partially failing sources, continuation cursors, and early close. Add an integration-level regression covering interleaved keys across multiple pages and the final cursor/limit behavior.

imbajin (commented):
Blocking: yes. Summary: The ordered range-index path still has correctness, cursor, resource, and compatibility defects beyond the existing review coverage. Evidence: exact-head static review of QueryResults, ScanUtil, OrderedKvIterator, OrderedMultiPartitionIterator, and public scan contracts at b45fdede3451.

imbajin:
A few potential follow-up performance directions to keep in mind after this PR lands. These are not required for the correctness fix in this PR, and each item should be validated by benchmark before we decide whether it is worth implementing.

Suggested benchmark coverage first:
- small `limit` on many partitions
- medium `limit` with sparse hits
- large `limit` / page cursor continuation
- the same query shape on 1, 4, 16, and 64 partitions
- warm-cache and cold-cache runs, especially when network latency is visible

1. Partition pruning / fan-out reduction

Idea: avoid opening streams for partitions that cannot possibly match the range. In practice, this is like asking only the shelves that may contain the target book, instead of checking every shelf.

```text
+---------------- ordered range query ----------------+
|                                                     |
|      partition metadata / range boundary check       |
|                                                     |
+-----------+-----------+-----------+-----------+-----+
            |           |           |           |
          match       skip        match       skip
            |                       |
       +----v----+             +----v----+
       |  P1     |             |  P3     |
       | stream  |             | stream  |
       +---------+             +---------+
```

Why it may help: it directly reduces RPC count, merge width, and client heap pressure. This is likely the highest-ROI direction if benchmark shows the current fan-out is material.

2. Adaptive page size

Idea: replace the fixed 64-entry page with a size chosen from `limit`, partition count, and observed latency. This is like choosing the bucket size based on how much water is needed and how far the well is.

```text
+---------------- page-size policy ----------------+
| query shape                  | possible choice    |
+------------------------------+--------------------+
| small limit                  | smaller first page |
| many partitions              | moderate page      |
| high network round-trip cost | larger page        |
+------------------------------+--------------------+
```

Why it may help: it can reduce wasted reads for small limits and reduce round trips for larger or higher-latency scans. This is mostly a tuning policy, so it is a good candidate only after baseline numbers show the fixed size is suboptimal.

3. Prefetch / pipeline

Idea: start fetching the next page before the current page is fully consumed. The goal is to hide network wait behind result consumption.

```text
+---------------- time ---------------->

consume page N:      [====================]
fetch page N + 1:              [====================]
consume page N + 1:                       [====================]

effect: network wait is partially overlapped with iteration
```

Why it may help: it can improve tail latency when network RTT dominates. The tradeoff is more state: cancellation, timeout, backpressure, and cleanup paths all need to stay correct.

4. Deeper merge pushdown

Idea: let Store side merge more aggressively so the client receives fewer ordered streams. This is like letting each Store produce one already-merged local stream before the client performs the final merge.

```text
+---------------- Store A ----------------+
|  P1 stream   P2 stream   P3 stream       |
|      \          |          /             |
|       +---------v---------+              |
|       |  local ordered    |              |
|       |  merge in Store   |              |
|       +---------+---------+              |
+-----------------+------------------------+
                  |
                  v
+---------------- Client -----------------+
|        final merge across Stores         |
+------------------------------------------+
```

Why it may help: it reduces client-side merge pressure and can lower the number of streams the client manages. It is also the most structural option because it tends to touch cursor semantics, paging state, and protocol/versioning.

Suggested follow-up order:

1. benchmark the current baseline
2. evaluate partition pruning
3. evaluate adaptive page size
4. evaluate prefetch / pipeline
5. consider deeper pushdown only if benchmark data justifies the added protocol and state complexity

My recommendation is to keep this PR focused on correctness and treat the items above as benchmark-driven follow-up candidates, not as changes to add into this PR.

imbajin:
⚠️ **Important — keep ordered-scan initialization fail-fast after the eight-worker pool is saturated**

`OrderedKvIterator.initialize()` previously submitted every Store first-entry task before reading any completion. The shared initializer uses eight workers, a `SynchronousQueue`, and `CallerRunsPolicy`; once all workers were busy, the next `submit()` could execute `firstEntry()` on the query/coordinator thread. If that Store blocked, an earlier Store failure could already be queued but remain unobserved, delaying cancellation and stream cleanup.

**Before**
```text
submit Store 1..8 -> worker pool
submit Store 9    -> coordinator runs it inline and blocks
Store 2 fails     -> failure waits unobserved in completion queue
```

**After**
```text
submit at most 8 -> wait for one completion
completion fails -> cancel and close immediately
completion succeeds -> submit the next Store and keep the window full
```

ordered-scan-sliding-window

**Sliding-window data structures**

```mermaid
flowchart LR
    P["iterators<br/>nextSource points to the next pending Store"]

    subgraph W["Sliding window: inFlight <= 8"]
        E["INITIALIZER<br/>8 worker threads"]
        F["futures<br/>submitted async tasks for cancellation"]
    end

    C["CompletionService<br/>completed Future queue"]
    H["PriorityQueue of SourceEntry<br/>first entry from each Store"]
    X["Failure cleanup<br/>cancel futures and close all sources"]

    P -->|"submit: nextSource++, inFlight++"| E
    E -.->|"record Future"| F
    E -->|"task completes"| C
    C -->|"success: addFirst(), inFlight--"| H
    C -->|"one slot is free: submit next Store"| P
    C -->|"failure"| X
```

For example, with 40 Stores:

```text
initial
  inFlight = [S1 ... S8]       pending = [S9 ... S40]

after S3 completes
  merge heap += first(S3)
  inFlight = [S1 S2 S4 ... S8 S9]
  pending  = [S10 ... S40]
```

The window remains full while work is available, but the coordinator always drains a completion before opening the next slot. A failure therefore reaches `take().get()` before a later Store can block the coordinator.

Implemented in `b6644a908`: if the shared pool rejects work, caller-thread fallback is used only when this query has no in-flight task whose failure could be hidden. The regression coverage includes the 9-Store failure case and both rejection paths.

contrueCT:
Follow-up in [`3941b1c91`](https://github.com/apache/hugegraph/pull/3140/commits/3941b1c910cbf264724be0c289bed5f7498a88a1) makes two small initialization improvements on top of the bounded-concurrency change:

- **Multiple Stores:** after waiting for one completion, drain every completion already queued before refilling the 8-source window. This makes an already-completed failure visible before another Store stream (for example, `S9`) is started.
- **Exactly one Store:** initialize its first entry inline and bypass executor/future dispatch, since the caller must wait for that entry anyway. The previous interruption contract is preserved: restore the interrupt flag, report the same initialization error, and close the source.

The ordered merge, concurrency bound, cancellation, and cleanup behavior remain unchanged.

Ordered scan initialization follow-up

imbajin (approved):

Issue #3090, opened 2026-07-10
Track HStore core-test exclusions and property codec mismatch


imbajin:
## Problem

PR #3008 exposed 22 historically excluded HStore core-test cases after the Maven `backend` property was propagated to every Surefire execution. Scoping that property back to the API-test execution fixes the PR regression, but it does not resolve the underlying HStore defects or the legacy test guards.

The exclusions were introduced by `d36e05ece8554fee73ba332147560930b3eba234` as temporary HStore skips. Their conditions use `Assume.assumeTrue(isHstore)` despite saying `skip this test for hstore`. Before PR #3008, the core-test JVM did not receive the backend property, so these tests were skipped because the value was null. A global property exposed 21 failures/errors and reduced skipped tests from 42 to 20.

Historical evidence:

- Passing baseline: https://github.com/apache/hugegraph/actions/runs/29094884977
- First exposing run: https://github.com/apache/hugegraph/actions/runs/29097183435
- Latest repeated failure: https://github.com/apache/hugegraph/actions/runs/29105710720

## Confirmed technical issues

1. HStore operator-sinking property decoding expects a self-describing byte containing `(cardinality << 6) | dataType`, while the Server binary writer stores the raw property value. Store-side filtering can therefore decode the first value byte as invalid `Cardinality code 0` or `DataType code 43`.
2. Core-test schema cleanup uses asynchronous deletion with a fixed 100 ms wait, allowing schema tasks and cache-clear events to cross test boundaries.
3. `HstoreSessionsImpl.clear()` does not remove the graph from `infoInitializedGraph`, so delete and recreate lifecycle state is incomplete.
4. HStore feature flags claim support for several OLAP, scan, range, and paging capabilities whose excluded tests currently fail.
5. RamTable is explicitly unsupported by HStore; those tests should remain clearly and correctly excluded.

## Required work

- [ ] Define an HStore-specific, versioned property codec for operator sinking without changing the RocksDB binary format.
- [ ] Define backward-read or migration behavior for existing HStore data.
- [ ] Make schema cleanup wait for task completion and verify schema/index/cache convergence.
- [ ] Fix the HStore graph clear and recreate lifecycle, including `infoInitializedGraph`.
- [ ] Correct the legacy HStore test guards and align them with feature flags.
- [ ] Re-enable supported OLAP, scan, range, paging, and multi-graph tests.
- [ ] Keep only explicitly unsupported tests, such as RamTable, skipped with an accurate reason.

[...]

PR #3141, opened 2026-08-07
fix(hstore): invalidate graph registration after clear

ghost:
Fixes the HStore clear/recreate lifecycle defect described in #3090.

This focused change invalidates the local `infoInitializedGraph` marker after a successful store-side graph deletion. The marker is cleared in a `finally` block, so a later `open()` can re-register the same graph even when the PD deletion call reports an error.

A regression test repeats the initialization-state add/clear transition 50 times to guard against stale JVM lifecycle state.

Validation:
- `git diff --check`
- Java 11 syntax parsing for both modified sources
- patch apply and reverse-apply checks against the supplied base snapshot

Maven tests were not run because Maven is unavailable in the execution image.

### DIFF
--- a/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreSessionsImpl.java
+++ b/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreSessionsImpl.java
@@ -184,7 +184,13 @@ public class HstoreSessionsImpl extends HstoreSessions {
             hgStoreClient.getPdClient().delGraph(this.graphName);
         } catch (PDException ignored) {
 
+        } finally {
+            clearInitializedGraph(this.graphName);
         }
+    }
+
+    private static void clearInitializedGraph(String graphName) {
+        infoInitializedGraph.remove(graphName);
     }
 
     @Override
--- a/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreSessionsImpl.java
+++ b/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreSessionsImpl.java
@@ -74,6 +74,7 @@ public class HstoreSessionsImpl extends HstoreSessions {
     private static volatile Boolean initializedNode = Boolean.FALSE;
     private static volatile PDClient defaultPdClient;
     private static volatile HgStoreClient hgStoreClient;
+    private final GraphStoreClient graphStoreClient;
     private final HugeConfig config;
     private final HstoreSession session;
     private final Map<String, Integer> tables;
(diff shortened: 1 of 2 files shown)


hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreSessionsImplTest.java
```diff
@@ -36,4 +39,30 @@ public void testProductionClassDoesNotReferenceTestAssert() throws IOException {
...
+        Method clearInitializedGraph =
+                HstoreSessionsImpl.class.getDeclaredMethod(
+                        "clearInitializedGraph", String.class);
+        clearInitializedGraph.setAccessible(true);
+
+        String graphName = "hugegraph/hstore-clear-test";
+        try {
+            for (int i = 0; i < 50; i++) {
+                Assert.assertTrue(initializedGraphs.add(graphName));
+                clearInitializedGraph.invoke(null, graphName);
```
imbajin: ⚠️ This test invokes private `clearInitializedGraph()` via reflection instead of exercising `HstoreSessionsImpl.clear()`. It would still pass if the `finally` call at `HstoreSessionsImpl.java:188` were removed or misplaced, and it does not verify the reported clear/recreate lifecycle. Please cover the public clear path (ideally the repeated clear/recreate sequence) and assert that a subsequent open re-registers the graph, or use controllable fakes to verify that behavior.

imbajin (commented):
Blocking: no. Summary: The production change clears the stale registration marker after graph deletion, but the added regression test only exercises the private helper and does not validate the public clear/recreate lifecycle. Evidence: Static review of HstoreSessionsImpl.clear() and HstoreSessionsImplTest.testClearInitializedGraphStateRepeatedly(); no CI checks reported for this exact head.

hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreSessionsImplTest.java
ulofiai: Thanks for catching this. You're right—the current test only exercises the private helper and would not fail if clear() stopped invoking it. I'll replace it with a public open → clear → open lifecycle test using controllable fakes and verify that the graph is re-registered after clear.

ghost:
Updated in e0536098. The regression test now exercises the public open() -> clear() -> open() lifecycle for 50 iterations using controllable store and PD fakes, and verifies that the graph is re-registered after every clear. Removing the marker invalidation from clear() now makes the test fail.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreSessionsImpl.java
```diff
@@ -181,20 +195,27 @@ public void truncateTable(String table) {
     public void clear() {
         this.session.deleteGraph();
         try {
-            hgStoreClient.getPdClient().delGraph(this.graphName);
+            this.graphStoreClient.delGraph(this.graphName);
         } catch (PDException ignored) {
 
+        } finally {
+            clearInitializedGraph(this.graphName);
```
imbajin, line 202: ⚠️ The production-level clear/reopen path is still untested and can leave the graph unregistered. `HstoreStore.open()` returns through `sessions.useSession()` when its session pool is still open, while `BackendSessionPool.useSession()` only reattaches/detects the existing session and never calls `HstoreSessionsImpl.open()` or `setGraph()`. Thus `clear(true)` followed by `store.open(config)` can delete the PD graph and then skip recreating it; the new test calls `HstoreSessionsImpl.open()` directly and does not cover this path. Please make the store-level reopen re-register the graph and add a `HstoreStore` clear → open regression test.

imbajin (commented):
Blocking: yes. Summary: The marker is cleared, but the actual HstoreStore clear/reopen path can still skip graph re-registration. Evidence: exact-head HstoreStore.open() at lines 216-223, BackendSessionPool.useSession() at lines 73-81, and the passing HstoreSessionsImpl-only regression test; latest-head CI checks are all successful.

Issue #3145, opened 2026-08-08
[Proposal] Java compatibility fixtures and PoC selection matrix for the Rust modernization groundwork


dpol1:
Follow-up to the discussion in #3110. @imbajin confirmed the testing and validation groundwork can start before any Rust decision, so this issue tracks the Java-only part: a selection matrix for the first PoC candidate and golden compatibility fixtures for the codec classes.

**Scope**

Three deliverables, all plain Java against the current build:

1. A selection matrix for PoC candidates (first cut below, to be refined in this thread).
2. A golden fixture corpus for `LongEncoding` plus a replay test that runs it against the current implementation.
3. Short written notes on the behavioral contract the fixtures lock down (edge cases, error behavior, invariants).

**Non-goals**

No Rust, no JNI, no Cargo, no packaging changes, no new Maven modules, no runtime wiring. Those wait for the component decision, a mentor, and a design review with contributors who have deeper Rust experience, as discussed in #3110.

**Selection matrix**

Criteria: an existing test oracle, enough call sites to matter, work per call big enough to survive a JNI crossing, and, following the maintainer feedback in #3110, keeping the Java implementation alive has to stay cheap: graceful fallback is the maintainer-preferred direction, which this matrix adopts as a selection criterion.

| Candidate | Oracle | Fallback cost | Notes |
|---|---|---|---|
| `LongEncoding` | `LongEncodingTest` (684 lines) | low, pure functions | output feeds storage keys, so the contract needs locking either way |
| `NumericUtil` | `NumericUtilTest` (475 lines) | low | coupled to `LongEncoding.encodeNumber`, candidate for the same corpus later |
| `BytesBuffer` | `BytesBufferTest` (1351 lines) | high | persisted format, 30+ callers, a fallback would have to stay byte-identical; the store's read path structurally parses server-encoded values via the struct copy, so the shared format surface spans ids, vints and property encoding (thread) |
| `IntSet` / `IntMap` | `IntSetTest`, `IntMapTest` | high | concurrent structures on `Unsafe`, a live double implementation risks divergence |

With the fallback criterion the concurrent structures and the serializer drop out of round one. `LongEncoding` is the first fixture target. That is a choice about where to write fixtures first, not a decision that it becomes the Rust PoC.

**Fixture shape**

[...]

Issue #3110, opened 2026-07-24
[Roadmap] Incremental Rust modernization for HugeGraph core modules


imbajin:
## Goal

Explore an **incremental Rust implementation** for selected HugeGraph Server, HStore, and PD components, especially performance- or memory-sensitive paths where Rust can provide measurable safety and efficiency benefits.

This is a parent tracking issue, not a commitment to rewrite the whole server. The current implementation remains the compatibility baseline, and Java/Rust components may coexist behind stable process, protocol, or storage boundaries.

## Roadmap at a glance

```text
Toolchain foundations
        │
        v
┌─ This repository ─────────────────┐
│  Server core -> HStore -> PD      │
└───────────────────────────────────┘
        │
        v
Computer kernels -> Vermeer/OLAP

Each arrow is a review gate, not an automatic rewrite.
```

## Proposed priority

1. **Server core**
   - Begin with bounded leaf components, memory management, codecs, or performance-sensitive data structures rather than the whole request path.
   - Preserve REST/Gremlin behavior, authentication, observability, and plugin contracts.
2. **HStore**
   - Evaluate storage/replication paths only after workload baselines, on-disk and wire compatibility, failure recovery, and rolling-upgrade requirements are explicit.
3. **PD**
   - Evaluate scheduling and metadata components after their distributed-system invariants and compatibility contract are covered by repeatable tests.

The sequence may change based on benchmark evidence and maintainer consensus.

## Delivery model

```text
proposal/RFC -> isolated component -> differential/fault tests
             -> benchmark -> opt-in deployment -> staged adoption
```

## Planning guardrails (required)

Before implementation starts, each child issue **must** define:

- scope, owner/mentor, and non-goals;
- API, protocol, storage-format, and operational compatibility;
- differential, concurrency, fault-injection, and upgrade/rollback tests;
- reproducible latency, throughput, memory, and recovery benchmarks;
- security/dependency checks and supported platforms;
- observability, deployment, coexistence, migration, and rollback plans;
- explicit acceptance criteria and user/operator/contributor documentation.

No large rewrite PR should start before its design and acceptance criteria are reviewed. Correctness, recoverability, and compatibility take precedence over headline benchmark numbers. Small, reviewable PRs and stable interoperability boundaries are required; an unplanned rewrite PR may be closed or asked to return to the RFC/design stage.

## Candidate work items

- [ ] Select the first small server-side proof of concept
- [ ] Document stable Java/Rust interoperability boundaries
[...]

PR #3146, opened 2026-08-08
test(commons): add golden fixture corpus for LongEncoding

dpol1:
### Visual summary

LongEncoding compatibility corpus
## Purpose of the PR

- part of #3145: delivers the golden corpus and replay tests; the selection matrix and compatibility inventory continue in the issue (context: #3110)

`LongEncoding` encodes ids and sortable keys, so its exact behavior is a compatibility contract, including the intentional overflow wrapping around `Long.MIN_VALUE` and inputs the decoder accepts but the encoder never produces. Today that contract exists only implicitly in the implementation. This PR pins it in a committed, implementation-independent corpus that any port or refactor can be checked against line by line.

## Main Changes

- `LongEncodingGoldenGenerator` (test scope): produces the corpus by running the current implementation and recording, for each input, the exact output or the exception class. Deterministic: fixed value sets plus a fixed random seed, duplicate lines dropped.
- `longencoding-golden.txt` (test resource): 3940 cases, plain ASCII, tab separated with `\uXXXX` escapes. Covers full-range round trips including `Long.MIN_VALUE`/`MAX_VALUE`, non-canonical decode inputs (`00` decodes to MIN while encode produces `010`), arbitrary UTF-16 decode input including surrogates, the error taxonomy (`IllegalArgumentException` vs `NumberFormatException` vs NPE), and length limits.
- `LongEncodingGoldenTest`: two checks. It replays every corpus line against the implementation, and it regenerates the corpus in memory to assert the committed file matches the generator output exactly, so generator and resource cannot drift apart.
- Both tests registered in `UnitTestSuite` so commons CI executes them.

Review guide: the review surface is the generator (case selection, ~370 lines), the corpus format documented in the file header, and the two-way check mechanics. The corpus file is generated output verified by `testCorpusMatchesGenerator` and does not need line-by-line review.

Contract boundaries: six operations (`encodeSortable`/`decodeSortable`, `encodeB64`/`decodeB64`, `encodeSignedB64`/`decodeSignedB64`). Exact output strings and decoded values are contract; the exception class is contract, the exception message deliberately is not. No production code changes.

## Verifying these changes

    - `mvn test -pl hugegraph-commons/hugegraph-common -Dtest=UnitTestSuite -DskipCommonsTests=false` on JDK 11 (the commons CI toolchain): 345 tests, the only errors are 2 environment NPEs in `MachineInfoTest` that reproduce identically on a clean master checkout in the same environment.
    - Tamper check: altering one byte of the corpus fails both new tests; regenerating restores green.
    - `apache-rat:check`, `checkstyle:check` and `git diff --check` are clean.

- [x]  Nope

## Documentation Status

- [x]  `Doc - No Need` (user docs unaffected; the technical contract is documented in #3145, the corpus file header, and this PR)


dpol1:
@imbajin seeing your interest in my proposal wdyt here?

imbajin:
@dpol1 Sorry I haven't replied to your proposal directly sooner. We've been focused on getting 1.8.0 released as soon as possible, and we're looking to take the Rust work forward in the 1.9–2.0 timeframe.

As the release work wraps up, we'll discuss the proposal together with the other PMC members and follow up here and in the related issues. A Rust refactor touches quite a few areas, so we do need some additional time to research the options and confirm the scope and compatibility requirements. Thanks for the proposal and the groundwork in this PR.

The community is very interested in your proposal and supportive of your involvement and this direction. As the release progresses, we'll follow up as soon as possible to exchange ideas and discuss the next steps, including through IM or Discord for quicker back-and-forth.

imbajin:
Completed six independent review lanes and pushed a small portability fix in c6cf10be65fe89ea05a1f9f950ce7e0322d38688: pin the golden corpus to LF in `.gitattributes`. With `core.autocrlf=true`, checkout previously converted it to CRLF and broke the byte-exact generator comparison; the fixed checkout now matches the committed corpus byte for byte.

Java 11 validation passed: 19 focused tests and all 345 commons `UnitTestSuite` tests, with no failures or skips. The final three-line fix received an independent check. No production code or corpus content changed.

Review score: **9.0/10**. No remaining actionable code findings within the six-operation contract; final-head CI is running, so approval is pending those results.

imbajin (approved):
+1, approved. @dpol1, thank you for the compatibility groundwork in this PR.

The repository's updated [AGENTS.md](https://github.com/apache/hugegraph/blob/master/AGENTS.md) and [.editorconfig](https://github.com/apache/hugegraph/blob/master/.editorconfig) specify a **120-column maximum** for Java. I have adjusted the two new Java files in `ed67283c2` accordingly, keeping multiline arguments aligned. The aim is to make the code easier for both people and LLMs to read and work on; 120 is an upper limit, not a target every line must reach. The formatting change preserves the Java tokens and corpus bytes; all 19 focused tests and the full clean compile passed.

> If useful, a separate formatting-only PR can apply this convention more broadly across the repository, starting with the core code and keeping functional changes separate.

We very much welcome your continued work on testing, correctness, compatibility fixtures, and the foundations for a future Rust refactor in #3145. This groundwork is important to the community regardless of which components eventually move to Rust. We would be glad to see you build deeper expertise and take an ongoing ownership role in this area, with other contributors participating in the design and reviews.

dpol1:
Thanks for the review, the LF pin, and the formatting pass, @imbajin. I accept ongoing ownership of the testing, compatibility, and fixture work in #3145 and will keep it in small PRs. #3145 describes the next step.

PR #3153, opened 2026-08-12
fix(store): allocate graph id before batch writes

### LINKED ISSUE #3095: [Bug] GraphSpace, graph and store isolation do not work with HStore in HugeGraph 1.7.0
### Bug Type (问题类型)

logic (逻辑设计问题)

### Before submit

### Environment (环境信息)

- Server Version: 1.7.0 (Core 1.7.0, REST API 0.71.0.0, Gremlin 3.5.1)
- Backend: HStore, one PD + one Store + one Server
- Docker images: hugegraph/server:1.7.0, hugegraph/pd:1.7.0, hugegraph/store:1.7.0
- Deployment: Docker Compose on a Linux server through 1Panel
- Authentication: enabled
- Server config: usePD=true; pd.peers=hugegraph-hstore-pd:8686
- Graph config: backend=hstore; serializer=binary; task.scheduler_type=distributed
- Store config: each graph uses a unique store name
- Data directories: reproduced after rebuilding with completely fresh PD and Store bind directories

## Description

GraphSpace isolation does not work in this HStore deployment.

Two graphs are created in two different GraphSpaces. The GraphSpace names, graph names, and store names are all different. However, a vertex inserted only into graph A in GraphSpace A can also be queried from graph B in GraphSpace B.

This reproduces through direct HugeGraph REST API calls, without any application code, SDK cache, or reverse proxy. GraphSpace and schema metadata appear separate, but the underlying graph data is shared or reused.

## Steps to reproduce

1. Start Server, PD, and Store with completely fresh data directories.
2. Confirm that only the DEFAULT GraphSpace exists.
3. Create two GraphSpaces with auth=true: curl_iso_a_<unique_id> and curl_iso_b_<unique_id>.
4. Create a different graph and store in each GraphSpace: graph_a_<unique_id> and graph_b_<unique_id>.
5. Create equivalent property keys and vertex labels in both graphs.
6. Insert one uniquely named marker vertex only into graph A.
7. Query all vertices from graph A.
8. Query all vertices from graph B.

All GraphSpace, graph, store, property-key, label, and marker names are newly generated for every test run.

## Expected behavior

Graph A returns the inserted marker. Graph B must return:

```json
{"vertices":[]}
```

Graphs in different GraphSpaces with different graph and store names must not share data.

## Actual behavior

Graph B unexpectedly returns the vertex inserted only into graph A.

A second test used completely new GraphSpace, graph, store, property-key, vertex-label, and marker names. The newly created graph also returned data from the first test run, and the older vertex was exposed using the new schema label.

Therefore, GraphSpace, graph, and store isolation all appear ineffective with HStore 1.7.0. This prevents multi-tenant isolation and can expose one tenant's data to another.

This appears separate from #2912, which addresses an exception during dynamic graph creation.

```javascript
// Inserted only into graph A:
POST /graphspaces/{space_a}/graphs/{graph_a}/graph/vertices

{
  "id": "1:only_a_19077_27057",
  "label": "iso_person_19077_27057"
}

// Query graph A:
GET /graphspaces/{space_a}/graphs/{graph_a}/graph/vertices

{
  "vertices": [
    {
      "id": "1:only_a_19077_27057",
      "label": "iso_person_19077_27057"
    }
  ]
}

// Query graph B (unexpectedly returns graph A data):
GET /graphspaces/{space_b}/graphs/{graph_b}/graph/vertices

{
  "vertices": [
    {
      "id": "1:only_a_19077_27057",
      "label": "iso_person_19077_27057"
    }
  ]
}

// Second run with completely different names:
{
  "vertices": [
    {
      "id": "1:only_a_19077_27057",
      "label": "iso_person_23806_7305"
    },
    {
      "id": "1:only_a_23806_7305",
      "label": "iso_person_23806_7305"
    }
  ]
}
```

```javascript

```

contrueCT:
## Overview

> [!IMPORTANT]
> HStore uses a **2-byte GraphId as the graph-isolation boundary**. Before this fix, a new graph whose first write used batch PUT or MERGE could encode data with the reserved missing GraphId `0xFFFE`. Multiple graphs could then share the same physical RocksDB key range.

<p align="center">
  HStore GraphId isolation failure and fix
</p>

Closes #3095. This PR extracts the focused fix from https://github.com/hugegraph/hugegraph/pull/163.

## Physical key layout

```text
┌──────────────────┬──────────────────────┬──────────────────┐
│ GraphId · 2 bytes │ Logical key          │ Key code · 2 bytes │
└──────────────────┴──────────────────────┴──────────────────┘
```

The GraphId prefix is what keeps data from different graphs in separate physical ranges:

| Graph | Expected physical key | Buggy first batch write |
|---|---|---|
| Graph A | `0001 │ shared-key │ code` | `FFFE │ shared-key │ code` |
| Graph B | `0002 │ shared-key │ code` | `FFFE │ shared-key │ code` |

`65534` (`0xFFFE`) is reserved for a missing GraphId mapping. The old batch path called `getKey()`, which returned this sentinel instead of allocating an ID.

## Trigger conditions

The collision requires **all three conditions**:

| 1. New graph | 2. No mapping | 3. First write is batched |
|---|---|---|
| The graph has not written data in this partition | No GraphId exists for the graph in the target partition | The first operation enters batch PUT or MERGE |

Once multiple graphs meet these conditions in the same partition, they share the `FFFE` prefix.

## Data impact

| Operation | Result when prefixes collide |
|---|---|
| **PUT** | The same logical key can overwrite another graph's value |
| **MERGE** | Counters from different graphs can be combined |
| **TRUNCATE** | Deleting the shared physical range can remove another graph's data |

> [!NOTE]
> RocksDB accepts these keys as valid, so the corruption is silent: requests can succeed without exceptions, failed health checks, or Raft errors.

<details>
<summary><strong>Why this can remain hidden in a long-running deployment</strong></summary>

- Normal single-write paths allocate a GraphId before encoding the key.
- Long-running graphs usually already have a persisted mapping and never re-enter the first-write state.
- The failure needs multiple new graphs in the same partition; PUT/MERGE collisions are most visible when their logical keys overlap.
- Graph truncate is uncommon, and the storage layer cannot distinguish an unintended shared prefix from a valid key range.

</details>

## Fix

| Before | After |
|---|---|
| Batch PUT/MERGE used `getKey()` | Batch PUT/MERGE uses `getKeyOrCreate()` |
| Missing mapping encoded as `FFFE` | A real GraphId is allocated before key encoding |
| Truncate could race with in-flight encoded keys | A graph/partition lifecycle lock fences transaction commit or rollback from truncate |
| Batch failures could retain transaction resources | All three batch call sites roll back and release the session and lifecycle lock |

The first-allocation path also checks whether the vertex table exists before scanning it. This avoids upgrading the column-family read lock held by the prepared batch into a write path that creates the missing table.

## Regression coverage

`BatchGraphIsolationTest` exercises the real `BusinessHandler.doBatch()` and RocksDB path:

- distinct PUT values for the same logical key in two graphs;
- target-graph deletion and neighboring-graph preservation after truncate;
- isolated MERGE counters;
- first batch write on an empty partition without creating the vertex table;
- rollback of failed transfer batches;
- truncate waiting for an in-flight batch transaction.

```bash
mvn test -pl hugegraph-store/hg-store-test -am \
    -P store-core-test -Djacoco.skip=true -ntp
```

## Scope

This focused change keeps the physical-key format, public APIs, dependencies, configuration, and CI workflow unchanged. It does **not** repair already affected data or include broader GraphId lifecycle, ID-reuse, or historical-data migration work.

## PR checklist

- [x] Regression coverage added and verified
- [x] No dependency or configuration changes
- [x] No public API changes
- [x] Documentation update not required

### DIFF
--- a/hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
+++ b/hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
@@ -1577,7 +1577,7 @@ public class BusinessHandlerImpl implements BusinessHandler {
         public TxBuilder put(int code, String table, byte[] key, byte[] value) throws
                                                                                HgStoreException {
             try {
-                byte[] targetKey = keyCreator.getKey(this.partId, graph, code, key);
+                byte[] targetKey = keyCreator.getKeyOrCreate(this.partId, graph, code, key);
                 this.op.put(table, targetKey, value);
             } catch (DBStoreException e) {
                 throw new HgStoreException(HgStoreException.EC_RKDB_DOPUT_FAIL, e.toString());
@@ -1642,7 +1642,7 @@ public class BusinessHandlerImpl implements BusinessHandler {
                                                                                  HgStoreException {
 
             try {
-                byte[] targetKey = keyCreator.getKey(this.partId, graph, code, key);
+                byte[] targetKey = keyCreator.getKeyOrCreate(this.partId, graph, code, key);
                 op.merge(table, targetKey, value);
             } catch (DBStoreException e) {
                 throw new HgStoreException(HgStoreException.EC_RKDB_DOMERGE_FAIL, e.toString());
--- a/hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
+++ b/hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
@@ -38,6 +38,9 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -121,6 +124,8 @@ public class BusinessHandlerImpl implements BusinessHandler {
 
     private static final Map<String, HugeGraphSupplier> GRAPH_SUPPLIER_CACHE =
             new ConcurrentHashMap<>();
+    private static final int GRAPH_LOCK_STRIPES = 1024;
+    private static final ReadWriteLock[] GRAPH_LOCKS = createGraphLocks();
     private static final int batchSize = 10000;
     private static Long indexDataSize = 50 * 1024L;
     private static final RocksDBFactory factory = RocksDBFactory.getInstance();
@@ -170,6 +175,20 @@ public class BusinessHandlerImpl implements BusinessHandler {
 
             }
         });
+    }
+
+    private static ReadWriteLock[] createGraphLocks() {
+        ReadWriteLock[] locks = new ReadWriteLock[GRAPH_LOCK_STRIPES];
+        for (int i = 0; i < locks.length; i++) {
+            locks[i] = new ReentrantReadWriteLock(true);
+        }
+        return locks;
+    }
+
+    private static ReadWriteLock graphLock(String graph, int partId) {
+        int hash = 31 * partId + graph.hashCode();
+        hash ^= hash >>> 16;
+        return GRAPH_LOCKS[hash & (GRAPH_LOCK_STRIPES - 1)];
     }
 
     public static HugeConfig initRocksdb(Map<String, Object> rocksdbConfig,
@@ -1014,11 +1033,17 @@ public class BusinessHandlerImpl implements BusinessHandler {
     public void truncate(String graphName, int partId) throws HgStoreException {
         // Each partition corresponds to a rocksdb instance, so the rocksdb instance name is
         // rocksdb + partId
-        try (RocksDBSession dbSession = getSession(graphName, partId)) {
-            dbSession.sessionOp().deleteRange(keyCreator.getStartKey(partId, graphName),
-                                              keyCreator.getEndKey(partId, graphName));
-            // Release map ID
-            keyCreator.delGraphId(partId, graphName);
+        Lock lifecycleLock = graphLock(graphName, partId).writeLock();
+        lifecycleLock.lock();
+        try {
+            try (RocksDBSession dbSession = getSession(graphName, partId)) {
+                dbSession.sessionOp().deleteRange(keyCreator.getStartKey(partId, graphName),
+                                                  keyCreator.getEndKey(partId, graphName));
+                // Release map ID
+                keyCreator.delGraphId(partId, graphName);
+            }
+        } finally {
+            lifecycleLock.unlock();
         }
     }
 
@@ -1287,7 +1312,7 @@ public class BusinessHandlerImpl implements BusinessHandler {
 
     @Override
     public TxBuilder txBuilder(String graph, int partId) throws HgStoreException {
-        return new TxBuilderImpl(graph, partId, getSession(graph, partId));
+        return new TxBuilderImpl(graph, partId);
     }
 
     @Override
@@ -1564,13 +1589,35 @@ public class BusinessHandlerImpl implements BusinessHandler {
         private final int partId;
         private final RocksDBSession dbSession;
         private final SessionOperator op;
+        private final Lock lifecycleLock;
+        private boolean completed;
 
-        private TxBuilderImpl(String graph, int partId, RocksDBSession dbSession) {
+        private TxBuilderImpl(String graph, int partId) {
             this.graph = graph;
             this.partId = partId;
-            this.dbSession = dbSession;
-            this.op = this.dbSession.sessionOp();
-            this.op.prepare();
+            this.lifecycleLock = graphLock(graph, partId).readLock();
+            this.lifecycleLock.lock();
+
+            RocksDBSession session = null;
+            SessionOperator operator = null;
+            try {
+                session = getSession(graph, partId);
+                operator = session.sessionOp();
+                operator.prepare();
+            } catch (RuntimeException | Error e) {
+                try {
+                    if (session != null) {
+                        session.close();
+                    }
+                } catch (Throwable closeError) {
+                    e.addSuppressed(closeError);
+                } finally {
+                    this.lifecycleLock.unlock();
+                }
+                throw e;
+            }
+            this.dbSession = session;
+            this.op = operator;
         }
 
         @Override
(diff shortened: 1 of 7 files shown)


hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
```diff
@@ -1577,7 +1577,7 @@ private TxBuilderImpl(String graph, int partId, RocksDBSession dbSession) {
         public TxBuilder put(int code, String table, byte[] key, byte[] value) throws
                                                                                HgStoreException {
             try {
-                byte[] targetKey = keyCreator.getKey(this.partId, graph, code, key);
+                byte[] targetKey = keyCreator.getKeyOrCreate(this.partId, graph, code, key);
```
imbajin, line 1627: ‼️ Critical — `getKeyOrCreate()` adds an exception path after `TxBuilderImpl` has already called `op.prepare()`. ID exhaustion or a failed `checkCount` scan is propagated as `HgStoreException`, but this method catches only `DBStoreException`; `DataManagerImpl.write()` and `DefaultDataMover.doWriteData()` also leave the builder without rollback. Because `prepare()` holds `cfHandleReadLock`, a failed first batch can retain the lock and session reference and block later table/partition operations. Please make every failed `put`/`merge` roll back and close the session, or perform allocation before preparing the transaction, and cover the failure path in a test.

imbajin (commented):
Blocking: yes. Summary: The new batch graph-ID path can leave RocksDB transaction locks held on allocation errors, persist IDs for batches that never commit, and reuse an ID while an in-flight batch still owns encoded keys. The added regression test also exercises an invalid partition instead of the empty-table transaction path. Evidence: exact-head static review; mvn test -pl hugegraph-store/hg-store-test -am -P store-core-test -Djacoco.skip=true -ntp passed 3 tests locally; latest-head CI completed, with Codecov upload 429 treated as non-blocking.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
```diff
@@ -1642,7 +1642,7 @@ public TxBuilder merge(int code, String table, byte[] key, byte[] value) throws
                                                                                  HgStoreException {
 
             try {
-                byte[] targetKey = keyCreator.getKey(this.partId, graph, code, key);
+                byte[] targetKey = keyCreator.getKeyOrCreate(this.partId, graph, code, key);
```
imbajin, line 1692: ⚠️ Important — This new allocation happens before the enclosing batch commits: `getGraphIdOrCreate()` persists and flushes the graph mapping/ID slot, while `BusinessHandler.doBatch()` only rolls back the data `WriteBatch`. If a later entry in the same PUT/MERGE batch fails, the graph ID remains consumed even though no data was committed, and repeated failed batches can eventually exhaust the ID space. Please make ID allocation transactional with the batch or release newly allocated IDs on rollback, and add a failing-batch regression test.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
imbajin, line 1627: ‼️ Critical — The encoded ID is not fenced against `truncate()`/ID reuse until `build().commit()` returns. `truncate()` deletes the graph range and calls `delGraphId()`, after which another graph may receive the released ID while this batch still holds keys encoded with it; a delayed commit can then write under the new graph's prefix. Please coordinate ID release/reuse with in-flight `TxBuilder` transactions (or bind allocation, commit, and release atomically) and add a deterministic truncate/reuse concurrency test.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/core/BatchGraphIsolationTest.java
```diff
@@ -0,0 +1,186 @@
...
+        RaftRocksdbOptions.initRocksdbGlobalConfig(rocksdbConfig);
+        BusinessHandlerImpl.initRocksdb(rocksdbConfig, null);
+
+        HgStoreEngineOptions options = new HgStoreEngineOptions();
+        options.setDataPath(databasePath.toString());
+        options.setRaftPath(databasePath.toString());
+
+        HgStoreEngineOptions.FakePdOptions fakePdOptions =
+                new HgStoreEngineOptions.FakePdOptions();
+        fakePdOptions.setPartitionCount(1);
```
imbajin: ⚠️ Important — The empty-table/deadlock assertion is not running on a valid configured partition. `partitionCount` is 1 and `hasPartition()` only accepts partition 0, but `EMPTY_PARTITION_ID` is 1; the test therefore creates an extra `00001` RocksDB and calls `getGraphIdOrCreate()` directly without the `TxBuilderImpl.op.prepare()` lock used by the batch path. Please make partition 1 valid and exercise a first PUT/MERGE batch while `g+v` is absent, with a timeout assertion, so this test covers the production deadlock regression.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
contrueCT: Fixed in 2f2765f8. All three TxBuilder call sites now create one Tx before queuing entries and roll it back on any put, merge, or commit failure; rollback failures are attached as suppressed exceptions so the original failure is preserved. TxBuilder rollback closes the RocksDB session and releases the graph lifecycle lock. BatchGraphIsolationTest now covers the failed DataManagerImpl and DefaultDataMover write paths.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
contrueCT: I kept GraphId allocation durable in this focused change. getGraphIdOrCreate already has durable mapping semantics on the non-batch doPut path, and deleting a newly created mapping from one batch rollback is unsafe when another concurrent batch for the same graph may already have observed that ID. Repeated failures for the same graph reuse its mapping; distinct failed graph names can still consume mappings, matching the existing non-batch lifecycle. A transactional or reference-counted GraphId reservation needs a separate cross-path lifecycle change rather than a rollback-only deletion here.

hugegraph-store/hg-store-core/src/main/java/org/apache/hugegraph/store/business/BusinessHandlerImpl.java
contrueCT: Fixed in 2f2765f8 with a striped graph-and-partition lifecycle read/write lock. TxBuilder holds the read lock from prepare through commit or rollback, while truncate holds the write lock across both deleteRange and delGraphId. The new deterministic regression test starts truncate after a batch has queued an encoded key, verifies truncate remains blocked, then commits and verifies truncate can finish before the next graph is written.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/core/BatchGraphIsolationTest.java
contrueCT: Fixed in 2f2765f8. The fake PD now configures two valid partitions and hasPartition accepts partition 1. The regression test sends a real doBatch PUT with code 32768 to partition 1 while g+v is absent, uses a 5-second JUnit timeout, and reads the value back through the normal graph path.

imbajin (approved):
TODO: check the diff carefully (inconsistent)

Issue #3151, opened 2026-08-12
[Bug] Server graph metadata can permanently diverge from PD when graph watch events are missed


contrueCT:
### Bug Type

server / metadata synchronization / distributed consistency

### Summary

In distributed mode with PD enabled, PD graph metadata is the durable source of truth, while each HugeGraph Server maintains a local derived state:

- local `GraphManager` graph instances
- embedded Gremlin Server graph bindings
- traversal source bindings such as `__g_<graphspace>-<graph>`

Currently, Server relies primarily on PD KV watch events such as `GRAPH/ADD` to keep this local state synchronized.

However, the current watch mechanism is not replayable, and graph event handling has no reconciliation fallback.

As a result, a Server can permanently miss a graph that already exists in PD.

This is related to #3137 and complements #3138.

#3138 guarantees that the **Server handling CreateGraph** has completed its own local graph/Gremlin registration before publishing the graph to PD.

The remaining problem is that **other Server replicas are still converging asynchronously from PD metadata**, and there is currently no mechanism guaranteeing that a replica eventually catches up if an incremental notification is lost or local processing fails.

---

### Current synchronization flow

```mermaid
flowchart TD
    A[Create graph on Server A] --> B[Server A opens graph]
    B --> C[Server A registers local Gremlin binding]
    C --> D[Persist GRAPH_CONF in PD]
    D --> E[Write GRAPH/ADD event]

    E --> F[Server B PD watch]
    F --> G[graphAddHandler]
    G --> H[Read graph config from PD]
    H --> I[Create local graph]
    I --> J[Register Gremlin binding]

    G -. callback failure .-> X[Event lost]
    E -. watch disconnected .-> Y[Event missed]
    X --> Z[Server B permanently lacks graph]
    Y --> Z
```

The desired invariant should be:

```text
If a graph exists in durable PD graph metadata,
every eligible Server should eventually have the corresponding
local graph instance and Gremlin binding.
```

Currently that invariant is not guaranteed.

---

#### 1. Graph event handler failure is not retried

`GraphManager.graphAddHandler()` loads the graph from PD and calls:

```java
createGraph(graphSpace, graphName, creator, config, false)
```

If graph construction fails, the exception is eventually caught by `ConsumerWrapper`, which only logs:

```java
LOG.error("Listener exception occurred.", e);
```

There is no retry, nack, or later reconciliation.

Therefore a transient local error can result in:

```text
PD contains graph A
    |
[...]

PR #3157, opened 2026-08-15
fix(pd): keep KvClient watches alive after reconnect failures

### LINKED ISSUE #3152: [Bug] PD KvClient watch can permanently stop after reconnect failure
### Bug Type

PD client / reliability / metadata synchronization

### Summary

`KvClient` automatically attempts to reconnect PD KV watches after a gRPC error or PD leader change.

However, the current recovery path performs only one reconnect attempt.

If that attempt also fails, the watch permanently stops.

`onCompleted()` currently has no recovery behavior either.

Since `PdMetaDriver.listen()` and `listenPrefix()` both depend on `KvClient`, this affects all metadata consumers using PD KV watches, including:

- graph lifecycle events
- graph space events
- service metadata events
- cache invalidation events
- other MetaManager listeners

This issue is related to #3137 and to the Server graph reconciliation issue: improving reconnect reliability reduces missed events, while reconciliation is still required because the current watch protocol has no replay mechanism.

---

### Current behavior

`KvClient` handles a watch error approximately as follows:

```mermaid
flowchart TD
    A[Watch running] --> B[gRPC error / leader changed]
    B --> C[clientId = 0]
    C --> D[Call listen again]

    D -->|success| E[Watch restored]
    D -->|PDException| F[Log warning]
    F --> G[Sleep 1 second]
    G --> H[Return]
    H --> I[Watch permanently stopped]
```

The relevant wrapper currently behaves like:

```java
try {
    listen(key, consumer);
} catch (PDException e) {
    log.warn(...);
    Thread.sleep(1000);
}
```

There is no later retry after the sleep.

Similarly:

```java
@Override
public void onCompleted() {
}
```

does not recreate the watch.

---

### Failure example

```mermaid
sequenceDiagram
    participant S as Server
    participant C as KvClient
    participant P as PD

    S->>C: listen(metadata key)
    C->>P: open watch
    P-->>C: Started

    P--xC: connection lost

    C->>P: reconnect attempt
    P--xC: PD still unavailable

    C->>C: log + sleep(1s)
    Note over C: reconnect logic exits

    Note over S,C: this metadata watch is now permanently gone
```

A temporary PD outage can therefore produce a long-lived Server that appears healthy but no longer receives future metadata changes.

---

### Missed events cannot currently be replayed

Fixing reconnect is important, but reconnect alone is not sufficient to guarantee metadata convergence.

The current protocol contains no revision/offset:

```protobuf
message WatchRequest {
  WatchState state = 2;
  string key = 3;
  int64 clientId = 4;
}
```

The PD service registers the current observer and pushes events directly when KV mutations happen.

There is no persisted event stream used to replay events emitted during disconnection.

```mermaid
sequenceDiagram
    participant C as KvClient
    participant P as PD

    C->>P: watch key
    P-->>C: Started

    P--xC: disconnected

    Note over P: event A
    Note over P: event B

    C->>P: reconnect
    P-->>C: Started

    Note over C: A and B are not replayed
```

Therefore this issue should remain focused on **watch connection reliability**, while stateful consumers should use reconciliation against durable PD state when correctness requires recovery of missed events.

---

### Relationship with graph metadata synchronization

Graph lifecycle synchronization currently uses:

```text
durable state:
GRAPH_CONF/<graph>

incremental notification:
EVENT/GRAPH/ADD
```

The event key is not an append-only log. Each new graph-add notification overwrites the same event key with the latest graph name.

Therefore missed graph-add events cannot be reconstructed from the event key itself.

```mermaid
flowchart LR
    A[PD GRAPH_CONF] -->|durable source of truth| R[Graph reconciliation]
    B[PD KV watch] -->|low-latency notification| S[Server]

    B -. reconnect reliability .-> C[This issue]
    A -. eventual convergence .-> D[Server graph reconciliation issue]

    C --> S
    D --> S
```

The two fixes are complementary:

- this issue: keep future watch delivery alive after transient failures;
- graph reconciliation: recover durable state changes missed while the watch was unavailable.

---

### Proposed solution

Keep the initial scope small:

1. Continue reconnecting after `onError()` until success or client close.
2. Continue reconnecting after `Leader_Changed`.
3. Recover from unexpected `onCompleted()`.
4. Add bounded retry delay/backoff to avoid tight reconnect loops.
5. Ensure only one reconnect loop exists for each watch.
6. Stop reconnecting cleanly when `KvClient.close()` is called.
7. Add deterministic tests for repeated failures followed by successful recovery.

Example target behavior:

```mermaid
flowchart TD
    A[Watch disconnected] --> B{Client closed?}

    B -->|yes| C[Stop]
    B -->|no| D[Reconnect]

    D -->|success| E[Resume watch]
    D -->|failure| F[Backoff]
    F --> B
```

---

### Out of scope

This issue does not propose:

- adding a durable watch event log;
- adding revisions/offsets to the watch protocol;
- guaranteeing replay of events emitted during downtime;
- changing Server request handling.

Those would be larger architectural changes.

For graph metadata, durable graph configs can already serve as the source of truth, so Server-side reconciliation is a smaller way to guarantee eventual convergence.

---

### Acceptance criteria

- One failed reconnect attempt does not permanently terminate a watch.
- Multiple consecutive connection failures are retried and eventually recover when PD becomes available.
- PD leader changes restore existing watches.
- Unexpected stream completion triggers recovery.
- `close()` prevents further reconnect attempts.
- Reconnect behavior has deterministic test coverage.
- No claim is made that events emitted during the disconnected window are replayed.

---

### Work status

I am currently working on this and plan to submit a focused PR for the reconnect lifecycle and regression tests.

Related Server graph reconciliation issue: #3151

### Visual summary

PD watch reconnect

contrueCT:
## Purpose of the PR

- close #3152

`KvClient` previously retried a failed watch only once. If that reconnect also failed, the
watch stopped permanently, and unexpected stream completion did not trigger recovery.

The reconnect path also exposed a lifecycle issue: `AbstractClient.resetStub()` invoked the
virtual `close()` method, so transport initialization on a `KvClient` dispatched to
`KvClient.close()` and marked the whole client closed.

## Reconnect Lifecycle

PD KV watch reconnect lifecycle before and after

## Main Changes

- Separate `AbstractClient` transport cleanup from the overridable client lifecycle close.
- Track each exact-key or prefix watch as an independent subscription with its current observer.
- Retry `onError`, `Leader_Changed`, and unexpected `onCompleted` with a fixed delay until the
  watch recovers or the client closes.
- Deduplicate reconnect scheduling for the same observer and ignore stale observer callbacks.
- Stop the reconnect executor and invalidate active observers when `KvClient.close()` is called.
- Add deterministic regression coverage for repeated failures followed by recovery, leader
  changes, completion, deduplication, stale observers, prefix semantics, and close behavior.

This PR does not add event replay, protocol revisions, PD Server changes, or the Server-side
reconciliation tracked by #3151. Events emitted while a watch is disconnected are still not
replayed.

## Verifying these changes

    - `mvn -q -o test -pl hugegraph-pd/hg-pd-test -am -Dtest=KvClientTest -DfailIfNoTests=false -Drat.skip=true -Djacoco.skip=true` (12 tests, 0 failures/errors)
    - `mvn -q -o package -pl hugegraph-pd -am -DskipTests -Dmaven.javadoc.skip=true -Drat.skip=true -Djacoco.skip=true`
    - `mvn -q -o -f hugegraph-pd/hg-pd-client/pom.xml apache-rat:check`
    - `mvn -q -o -f hugegraph-pd/hg-pd-test/pom.xml apache-rat:check`

The full local PD suite was also attempted. The common suite passed 83/83 and the core suite
passed 89/91 with 2 skipped. The client suite could not complete without a local PD service at
`127.0.0.1:8686` (4 failures and 5 errors reported connection refused), which prevented the rest
suite from executing in that Maven run.

Repository-wide RAT is not a valid signal in this checkout because pre-existing ignored
`.upgrade-artifacts` files produce 68 unrelated unapproved-license entries; RAT passes for both
changed modules.

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Other affects: PD KV watch reconnect lifecycle
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO`
- [ ]  `Doc - Done`
- [x]  `Doc - No Need`

### Visual summary

PD watch reconnect


hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,49 +186,122 @@ public void onNext(WatchResponse value) {
...
         try {
-            WatchRequest k =
-                    WatchRequest.newBuilder().setClientId(clientId.get()).setKey(prefix).build();
-            streamingCall(KvServiceGrpc.getWatchPrefixMethod(), k, observer, 1);
+            if (subscription.prefix) {
+                streamingCall(KvServiceGrpc.getWatchPrefixMethod(), request, observer, 1);
+            } else {
+                streamingCall(KvServiceGrpc.getWatchMethod(), request, observer, 1);
+            }
+            return true;
```
imbajin, line 337: ‼️ Critical. `startWatch()` treats a normal return from `streamingCall()` as a successful watch, but `AbstractClient.streamingCall()` catches the terminal `StatusRuntimeException` after its retry limit and returns without invoking the observer (AbstractClient.java:252-266). The observer remains installed and the semaphore can remain held, so `reconnect()` never schedules another attempt; the new `TestKvClient.streamingCall()` test double also bypasses this production path by throwing `PDException` directly. Propagate the final transport failure or return an explicit failure, clean the observer/semaphore state, and add a test that exercises the production streaming path.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,49 +186,122 @@ public void onNext(WatchResponse value) {
...
         }
     }
 
+    private void requestReconnect(WatchSubscription subscription,
+                                  StreamObserver<WatchResponse> sourceObserver) {
+        if (closed.get() ||
+            !subscription.observer.compareAndSet(sourceObserver, null)) {
+            return;
+        }
+        clientId.set(0L);
```
imbajin: ‼️ Critical. A single watch failure resets the shared `clientId` to zero, but the same field is used by `lock`, `unlock`, and `keepAlive` (KvClient.java:357-416). The server stores locks under that ID and `unlock` only deletes a lock when `owned(key, clientId)` matches (KvService.java:230-273); after this watch reconnect obtains a new ID, an existing lock remains under the old ID and cannot be unlocked or renewed, expiring only at its TTL. Separate watch identity from lock identity, or coordinate reconnects so the lock session ID is never reset.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,49 +186,122 @@ public void onNext(WatchResponse value) {
 
             @Override
             public void onError(Throwable t) {
```
imbajin, line 237: ⚠️ Important. `onError(Throwable t)` discards the status and unconditionally schedules another watch; `onCompleted()` follows the same path. A terminal authentication/validation failure therefore becomes an unbounded one-second RPC retry with no way for the caller to observe or stop the subscription. Classify retryable transport/leader-transition errors separately, and terminate/report non-retryable failures instead of rescheduling them forever.

hugegraph-pd/hg-pd-test/src/main/java/org/apache/hugegraph/pd/client/KvClientTest.java
```diff
@@ -67,6 +97,189 @@ public void testCreateBlockingStub() {
...
+
+    @Test
+    public void testStaleObserverDoesNotScheduleReconnect() throws Exception {
+        try (WatchTestContext context = newWatchTestContext()) {
+            context.client.listen("key", response -> { });
+            StreamObserver<WatchResponse> staleObserver = context.client.call(0).observer;
+            staleObserver.onError(new RuntimeException("disconnected"));
+            context.runNextReconnect();
+
+            staleObserver.onCompleted();
```
imbajin, line 969: ⚠️ Important. `testStaleObserverDoesNotScheduleReconnect()` only calls `onCompleted()` on the stale observer; it never sends a stale `Started` response containing an event after the new observer is installed. A regression removing the identity check in `KvClient.java:162-164` would still pass this test and could duplicate business callbacks. Send an event through the stale observer and assert the consumer is untouched, while verifying the current observer still delivers events.

imbajin (commented):
Blocking: yes. Summary: The reconnect implementation has critical identity and failure-propagation risks that can silently stop watches and invalidate existing locks; permanent callback errors also retry without termination. Evidence: static review of KvClient.java and AbstractClient.java at head 772b3a529ce719c5a16eb16f99e160573e33cc8e, with exact server lock/watch handling cross-checked.

imbajin (commented):
Blocking: yes. Summary: The current head addresses the previously reported watch reconnect, lock identity, terminal-error retry, and stale-observer issues; focused validation passes, but coverage checks remain failed and independent review evidence is incomplete. Evidence: `mvn -pl hugegraph-pd/hg-pd-test -am -Dtest=KvClientTest -DfailIfNoTests=false test` (18/18 passed), `git diff --check` passed; `codecov/project` and `codecov/patch` failed.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +200,144 @@ public void onNext(WatchResponse value) {
...
+    }
+
+    private void reconnect(WatchSubscription subscription) {
+        subscription.reconnectScheduled.set(false);
+        if (closed.get()) {
+            return;
+        }
+        try {
+            startWatch(subscription);
+        } catch (PDException e) {
```
bitflicker64: ⚠️ `reconnect()` catches only `PDException`, so an unchecked failure ends the retry chain permanently.

`reconnect()` clears `subscription.reconnectScheduled` at the top and then calls `startWatch` guarded only by `catch (PDException e)`. Any `RuntimeException` escapes into the `ScheduledExecutorService`, where it is swallowed into the (discarded) `ScheduledFuture`. At that point the flag is already `false`, `subscription.observer` is null, and nothing reschedules — the subscription is left permanently dead, which is the exact failure mode this PR removes.

This is reachable through the production stub path: `startWatch` → `streamingCall` → `AbstractClient.getStub()` → `resetStub()`. `resetStub()` assigns `leaderHost` from the members response (`AbstractClient.java:150`) *before* it assigns `proxy.setStub(...)` (`AbstractClient.java:156`). If stub creation throws for every host, `resetStub()` returns a non-empty `leaderHost` while `proxy.getStub()` is still null, and `getStub()` then evaluates `setAsyncParams(null, config)` → NPE, which is unchecked and outside the `try` in `streamingCall`.

No test covers it either: `TestKvClient.streamingCall` only ever throws `PDException` (`KvClientTest.java:752`).

Requested change: catch `Throwable` (or at least `RuntimeException`) here and route it through the same `scheduleReconnect(subscription)` path — or move the rescheduling into a `finally` — and add a test whose stubbed `streamingCall` throws an unchecked exception, asserting a further reconnect is still scheduled.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +200,144 @@ public void onNext(WatchResponse value) {
...
+        }
+
+        StreamObserver<WatchResponse> observer = getObserver(subscription);
+        subscription.observer.set(observer);
+        if (closed.get()) {
+            subscription.observer.compareAndSet(observer, null);
+            return false;
+        }
+
+        acquire(watchClientId, watchSemaphore);
```
bitflicker64: ⚠️ All reconnects share one thread that can block here indefinitely, so one stalled stream stops every watch from recovering.

`acquire(watchClientId, watchSemaphore)` performs an untimed `semaphore.acquire()` whenever `watchClientId == 0` (`KvClient.java:340-352`). The permit only comes back when some watch receives `Starting` (`KvClient.java:185`), fails (`KvClient.java:286`), or is stopped (`KvClient.java:301`). Reconnects all run on the single-thread executor created at `KvClient.java:85-89`.

That combination is exercised by an ordinary leader change: `KvWatchSubject.notifyClientChangeLeader()` sends `Leader_Changed` to every observer and then calls `removeClient(...)`, which completes each stream, so all subscriptions call `requestReconnect` (resetting `watchClientId` to 0) and queue on that one thread. The first task re-issues its stream and returns; the second parks in `acquire` until the first stream answers. Since `resetStub()` builds channels with `ManagedChannelBuilder.forTarget(host).usePlaintext()` and no keepalive (`AbstractClient.java:142`), a half-open connection that never delivers `Starting` and never errors parks that thread for good, and no subscription in the client can reconnect.

Requested change: replace the untimed `acquire` with `tryAcquire(timeout, unit)` and treat a timeout as a failed attempt that reschedules, and/or give the reconnect executor more than one thread so a stalled subscription cannot block the others.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +200,144 @@ public void onNext(WatchResponse value) {
...
+
+    private void stopWatch(WatchSubscription subscription,
+                           StreamObserver<WatchResponse> sourceObserver,
+                           Throwable throwable) {
+        if (!subscription.observer.compareAndSet(sourceObserver, null)) {
+            return;
+        }
+        release(watchSemaphore);
+        subscriptions.remove(subscription);
+        log.error("Watch for key {} stopped after a non-retryable error: {}",
```
bitflicker64, line 483: 🧹 A permanently stopped watch is invisible to the caller.

`stopWatch()` drops the subscription from `subscriptions` and records a log line, but the consumer registered through `listen`/`listenPrefix` is never told. Both production callers — `PdMetaDriver.listen`/`listenPrefix` (`hugegraph-server/hugegraph-core/.../meta/PdMetaDriver.java:110-126`) and `SchemaDriver.listen` (`hugegraph-struct/.../SchemaDriver.java:199-205`) — only see the initial `PDException`, so after a `CANCELLED`, `UNAUTHENTICATED` or `PERMISSION_DENIED` stream error the application keeps running as if it were still subscribed and the only signal is a log line.

Requested change: give the caller a way to observe termination — for example an optional error/termination callback on `listen`, or a handle returned from `listen` whose state can be queried — so a permanently stopped watch is detectable without scraping logs.

bitflicker64 (commented):
Blocking: no. Summary: Splitting the lock and watch client IDs, tracking each watch as an independent subscription, and propagating the terminal streaming failure are all solid, and the regression suite is genuinely deterministic; two gaps remain in the reconnect loop itself, plus one visibility nit. Evidence: read of KvClient.java, AbstractClient.java and KvClientTest.java at 3c2f4dfe7a57aaa85992113a0134a99bb200090d; repo greps for streamingCall/listen callers; PD server watch path in KvWatchSubject.notifyClientChangeLeader and KvServiceGrpcImpl.clientWatch; gh -R apache/hugegraph pr checks 3157 (17 pass, only codecov/patch and codecov/project fail).

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +200,170 @@ public void onNext(WatchResponse value) {
...
     }
 
-    private void acquire() {
+    private void requestReconnect(WatchSubscription subscription,
+                                  StreamObserver<WatchResponse> sourceObserver) {
+        if (closed.get() ||
+            !subscription.observer.compareAndSet(sourceObserver, null)) {
+            return;
+        }
+        watchClientId.set(0L);
```
imbajin: ‼️ Critical. `requestReconnect()` resets the observer and watch ID but leaves the cached async stub intact. After `Leader_Changed` or an asynchronous `UNAVAILABLE`, `reconnect()` calls `getStub()` and reuses the same channel; the server-side `onError` callback never enters `streamingCall()`'s synchronous peer retry. A leader move can therefore retry the old follower forever. Please invalidate/rotate the async stub before scheduling the reconnect and cover this with a real two-peer leader-failover test.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -56,13 +60,38 @@
 @Slf4j
 public class KvClient<T extends WatchResponse> extends AbstractClient implements Closeable {
 
-    private AtomicLong clientId = new AtomicLong(0);
-    private Semaphore semaphore = new Semaphore(1);
-    private AtomicBoolean closed = new AtomicBoolean(false);
-    private Set<StreamObserver> observers = ConcurrentHashMap.newKeySet();
+    private static final long RECONNECT_DELAY_MS = 1000L;
+    private static final Set<Status.Code> NON_RETRYABLE_WATCH_ERRORS =
+            Set.of(Status.Code.CANCELLED,
```
imbajin: ⚠️ Important. `CANCELLED` is treated as permanently non-retryable, but `AbstractClient.resetStub()` closes the shared channel with `shutdownNow()`. That internal refresh can deliver `CANCELLED` to unrelated live watches, which then enter `stopWatch()` and are removed forever even though `KvClient` is still open. Please distinguish explicit client close from internal channel replacement, or otherwise make these cancellations recoverable, and add a multi-subscription channel-reset test.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -127,7 +156,7 @@ public TTLResponse putTTL(String key, String value, long ttl) throws PDException
 
     private void onEvent(WatchResponse value, Consumer<T> consumer) {
         log.debug("receive message for {},event Count:{}", value, value.getEventsCount());
-        clientId.compareAndSet(0L, value.getClientId());
+        watchClientId.compareAndSet(0L, value.getClientId());
```
imbajin: ‼️ Critical. The global watch ID can be restored by another live subscription while a failed subscription is waiting to reconnect. For example, A clears the ID after disconnecting, B receives an event and writes the old ID back here, then A reconnects with that old ID. The server registers observers with `putIfAbsent(key@clientId, observer)`, so A's stale server observer can win and the new stream appears started but never receives later events. Please use a per-subscription/generation ID or atomically replace the stale server observer, and test the one-failed/one-still-active sequence.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -237,7 +377,7 @@ private void acquire() {
         }
     }
 
-    private void release() {
+    private void release(Semaphore semaphore) {
```
imbajin, line 597: ⚠️ Important. Permit ownership is still racy: this check-then-release is not atomic, and `tryAcquire()` can return true for a nonzero client ID without acquiring a permit while close/failure paths still call `release()`. Concurrent `Starting`, error, or close callbacks can therefore raise the permit count above one, allowing multiple reconnects to issue `clientId=0` streams and record inconsistent server IDs. Please model acquisition ownership explicitly (or use an atomic state machine) and assert that the effective permit count never exceeds one under concurrent callbacks.

imbajin (commented):
Blocking: yes. The latest update fixes the reconnect executor parking reported in comment 3885719895, and the focused KvClientTest passes 19/19, but four current-head races can still silently strand watches. The hstore check is red on VertexCoreTest.testQueryByJointIndexesWithSearchAndTwoRangeIndexesAndWithin; exact base 7bb624b99 fails the same assertion, so it appears baseline-related, but visible CI remains a merge gate.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +200,170 @@ public void onNext(WatchResponse value) {
...
+            subscription.observer.compareAndSet(observer, null);
+            return false;
+        }
+        if (closed.get()) {
+            subscription.observer.compareAndSet(observer, null);
+            release(watchSemaphore);
+            return false;
+        }
+
+        WatchRequest request = WatchRequest.newBuilder()
```
imbajin, line 307: ⚠️ Important. `WatchRequest` is built after the subscription and semaphore are installed, but outside the `try` that releases them. A null key (or another unchecked builder failure) therefore leaves the observer registered and the permit held; a later valid `listen` can block indefinitely on the watch semaphore. Build the request inside the guarded cleanup path and add an invalid-input test that verifies no permit or subscription is leaked.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +200,170 @@ public void onNext(WatchResponse value) {
...
+    private void reconnect(WatchSubscription subscription) {
+        subscription.reconnectScheduled.set(false);
+        if (closed.get()) {
+            return;
+        }
+        try {
+            if (!startWatch(subscription, false)) {
+                scheduleReconnect(subscription);
+            }
+        } catch (PDException e) {
```
imbajin: ⚠️ Important. After peer retries are exhausted, `AbstractClient.streamingCall()` wraps the final gRPC status as `PDException`. `reconnect()` catches every `PDException` and schedules another attempt, so permanent statuses such as permission or authentication failures are retried forever without classification; `NON_RETRYABLE_WATCH_ERRORS` only protects the async callback path. Preserve the gRPC status and stop/report non-retryable synchronous failures.

imbajin (commented):
Blocking: yes. The focused KvClientTest passes 19/19, but four current-head lifecycle and failure-handling gaps can still strand watches. The hstore check is red on VertexCoreTest.testQueryByJointIndexesWithSearchAndTwoRangeIndexesAndWithin; exact base 7bb624b99 fails the same assertion, so it appears baseline-related, but visible CI remains a merge gate.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +200,170 @@ public void onNext(WatchResponse value) {
...
         try {
-            WatchRequest k =
-                    WatchRequest.newBuilder().setClientId(clientId.get()).setKey(prefix).build();
-            streamingCall(KvServiceGrpc.getWatchPrefixMethod(), k, observer, 1);
+            if (subscription.prefix) {
+                streamingCall(KvServiceGrpc.getWatchPrefixMethod(), request, observer, 1);
+            } else {
+                streamingCall(KvServiceGrpc.getWatchMethod(), request, observer, 1);
+            }
+            return true;
```
imbajin, line 337: ⚠️ Important. `startWatch()` returns true immediately after `asyncServerStreamingCall()` returns, before any `Starting` frame or other readiness signal. If the connection is blackholed or the server never sends the first frame, the subscription remains installed and the semaphore stays held forever, so reconnect cannot make progress. Add a first-frame deadline/watchdog that cleans up and reschedules the subscription when readiness is not observed.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -56,13 +60,38 @@
...
+    private final AtomicLong lockClientId = new AtomicLong(0);
+    private final AtomicLong watchClientId = new AtomicLong(0);
+    private final Semaphore lockSemaphore = new Semaphore(1);
+    private final Semaphore watchSemaphore = new Semaphore(1);
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+    private final Set<WatchSubscription> subscriptions = ConcurrentHashMap.newKeySet();
+    private final ScheduledExecutorService reconnectExecutor;
 
     public KvClient(PDConfig pdConfig) {
+        this(pdConfig, Executors.newSingleThreadScheduledExecutor(runnable -> {
```
imbajin: ⚠️ Important. This executor is created for every `KvClient`, but it is shut down only by `KvClient.close()`. `SchemaDriver.destroy()` clears its caches and singleton without closing the client that owns these reconnect tasks, so old watches can continue after destroy and invoke callbacks against destroyed state. Make the driver lifecycle close the client, or give the client a destroy-safe ownership contract, and cover destroy with an active-watch test.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +199,286 @@ public void onNext(WatchResponse value) {
 
             @Override
             public void onError(Throwable t) {
-                release();
-                if (!closed.get()) {
-                    clientId.set(0);
-                    listenWrapper.accept(key, consumer);
+                if (isRetryableWatchError(t)) {
```
imbajin, line 238: ‼️ Critical. `shouldRotateWatchTransport()` only handles `UNAVAILABLE`, but the server reports a non-leader watch through `responseObserver.onError(new PDException(-1, msg))` without a redirect status. This reaches the retry path as an `UNKNOWN`-style error, so the cached stub is retained and a watch that lands on a follower can reconnect to that same follower forever. Return an explicit redirectable status from the server or invalidate the transport for this response so leader discovery runs again, and add a two-peer regression test for this path.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +199,286 @@ public void onNext(WatchResponse value) {
...
+                return;
+            }
+            ScheduledFuture<?> previous = subscription.startTimeout.getAndSet(timeout);
+            if (previous != null) {
+                previous.cancel(false);
+            }
+        }
+    }
+
+    private void onStartTimeout(WatchSubscription subscription,
```
imbajin, line 398: ⚠️ Important. `onStartTimeout()` checks `firstFrameReceived` under the subscription lock, releases it, and then calls `requestReconnect()`, which does not recheck that flag. If the `Starting` frame arrives in this window, the timeout still clears the now-healthy observer and schedules a duplicate watch. Make the timeout check and observer invalidation one atomic transition, or recheck `firstFrameReceived` inside `requestReconnect()`.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
```diff
@@ -111,11 +117,14 @@ public static void init(PDConfig pdConfig, int cacheSize, long expiration) {
     }
 
     public static void destroy() {
-        SchemaDriver instance = INSTANCE.get();
+        SchemaDriver instance = INSTANCE.getAndSet(null);
```
imbajin: ⚠️ Important. `INSTANCE` is cleared before the old client and caches finish closing. A concurrent `SchemaDriver.init()` can observe `null` and publish a second driver while the first driver's watches, reconnect executor, and cache cleanup are still active, producing overlapping metadata listeners. Serialize init/destroy with one lifecycle lock, or clear `INSTANCE` only after cleanup completes.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
```diff
@@ -84,7 +85,12 @@ public class SchemaDriver {
 
     private SchemaDriver(PDConfig pdConfig, int cacheSize,
                          long expiration) {
-        this.client = new KvClient<>(pdConfig);
+        this(new KvClient<>(pdConfig), cacheSize, expiration);
```
imbajin, line 92: ⚠️ Important. This constructor now owns a `KvClient` with reconnect resources and starts four watches before construction completes. If a later `listenMetaChanges()` call fails after an earlier watch has started, the constructor exits without closing that client or its executor because `INSTANCE` is assigned only after construction succeeds. Add failure cleanup around listener initialization, or make client ownership explicit and close it on partial startup.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -56,13 +62,37 @@
...
+                   Status.Code.UNIMPLEMENTED,
+                   Status.Code.DATA_LOSS,
+                   Status.Code.UNAUTHENTICATED);
+
+    private final AtomicLong lockClientId = new AtomicLong(0);
+    private final Semaphore lockSemaphore = new Semaphore(1);
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+    private final Set<WatchSubscription> subscriptions = ConcurrentHashMap.newKeySet();
+    private final ScheduledExecutorService reconnectExecutor;
+    private long transportGeneration;
```
imbajin: 🧹 Design suggestion. The reconnect behavior is now spread across per-subscription atomics, nested locks, a shared `transportGeneration`, a scheduler, and a first-frame watchdog. Before adding another guard for each race, consider one explicit per-watch state/attempt object with an atomic `ACTIVE -> RECONNECTING -> STOPPED` transition, the observer/client ID, and the cancellable timeout; keep transport refresh behind one small helper and make the server's follower response an explicit redirectable status. Then timeout, error, leader change, and close can share the same transition instead of accumulating independent flags and patches.

imbajin (commented):
Blocking: yes. Summary: the latest head fixes several earlier global-state and lifecycle issues, but four production correctness gaps remain. Evidence: exact head 4202f266c217bc1d7c99027bda1a3539326567e9, CI is green, and the focused tests pass 33/33; a separate design comment proposes consolidating the watch state machine instead of adding more guards.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -56,13 +64,37 @@
...
 
     public KvClient(PDConfig pdConfig) {
+        this(pdConfig, Executors.newSingleThreadScheduledExecutor(runnable -> {
```
imbajin: ⚠️ Important. This single-thread scheduler runs `reconnect()` synchronously, and `startWatch()` calls `streamingCall()` while holding the client monitor. After a watch invalidates the async stub, `getStub()` enters `AbstractClient.resetStub()`, where the blocking `getMembers()` call has the default 60-second deadline and is repeated for every configured host. One blackholed peer can therefore occupy the only scheduler thread for up to `hostCount * grpcTimeOut`; other subscriptions cannot run their reconnect or first-frame watchdog, and `KvClient.close()` waits for the same monitor. Move transport discovery off the shared scheduler/monitor or otherwise isolate and bound reconnect work, and add a multi-watch test with one unreachable peer and another watch recovering.

imbajin (commented):
Blocking: yes. Summary: The single-thread reconnect scheduler can serialize all watches behind a 60-second peer-discovery attempt, delaying recovery and close. Evidence: exact head 0a77c55ac6fa0baaf0e97d98eb0d15a1c28a1531; PDConfig defaults grpcTimeOut to 60000 ms; reconnectExecutor is single-threaded; startWatch synchronously reaches AbstractClient.resetStub() and blocking getMembers().

hugegraph-struct/src/test/java/org/apache/hugegraph/SchemaDriverTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+            Assert.assertTrue(client.closed);
+            Assert.assertNull(SchemaDriver.getInstance());
+        } finally {
+            instanceReference().set(null);
+            client.close();
+        }
+    }
+
+    @Test(timeout = 3000L)
```
bitflicker64: ‼️ These two new tests fail CI at this head, so the PR is currently red on code it adds.

`build-server-macos-rocksdb (macos-15-intel)`, job `99501401894`:

```
[ERROR] Tests run: 3, Failures: 1, Errors: 1 - in org.apache.hugegraph.SchemaDriverTest
[ERROR]   SchemaDriverTest.testDestroyClosesOwnedKvClient:46 expected null, but was:<org.apache.hugegraph.SchemaDriver@29176cc1>
[ERROR]   SchemaDriverTest.testDestroyKeepsInstanceUntilResourcesAreClosed:55 » TestTimedOut
[INFO] BUILD FAILURE
```

The file is new here, so this is not pre-existing, and the same class passes on the ubuntu `pd` and `struct` runners at this head, so it is timing dependent rather than deterministic.

The two failures look like one problem. All three tests mutate the `SchemaDriver.INSTANCE` static by reflection (lines 40, 48, 57-58, 92), and `@Test(timeout = 3000L)` only interrupts the test thread; JUnit's `FailOnTimeout` does not stop it. The reported frame is line 55, `new BlockingCloseKvClient()`, so first-touch loading of the `KvClient` and gRPC stub classes plausibly blew the 3 s budget on a slow Intel runner. The abandoned thread would then still publish a driver into `INSTANCE` and start `destroyThread`, which holds the `static synchronized destroy()` monitor. That fits both symptoms, including the sibling test's 2.587 s runtime, but I inferred the interleaving from the logs rather than reproducing it.

Requested change: snapshot and restore `INSTANCE` in `@Before`/`@After` instead of inside each test body, and replace the wall clock `@Test(timeout = 3000L)` and `await(1L, TimeUnit.SECONDS)` handshake with latches the test fully controls, so a slow runner cannot leave a thread mutating global state after the timeout fires.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -180,50 +201,307 @@ public void onNext(WatchResponse value) {
...
+        WatchRequest request = WatchRequest.newBuilder()
+                                           .setClientId(subscription.clientId.get())
+                                           .setKey(subscription.key)
+                                           .build();
+        StreamObserver<WatchResponse> observer = getObserver(subscription);
         try {
-            WatchRequest k =
-                    WatchRequest.newBuilder().setClientId(clientId.get()).setKey(prefix).build();
-            streamingCall(KvServiceGrpc.getWatchPrefixMethod(), k, observer, 1);
+            synchronized (this) {
```
bitflicker64: ⚠️ The client monitor is held across channel teardown, so gRPC callbacks for the other watches block on it.

`streamingCall()` runs inside this `synchronized (this)` (lines 292 and 294). It reaches `getStub()`, and when `proxy.getStub()` is null that calls `AbstractClient.resetStub()`, then `closeConnections()`, then `closeChannel()`, whose loop is unbounded (`AbstractClient.java:292`):

```java
while (channel != null &&
       !channel.shutdownNow().awaitTermination(100, TimeUnit.MILLISECONDS)) {
    continue;
}
```

The stub is null on exactly the common reconnect path, since `requestReconnect()` calls `invalidateAttemptStub()` (line 379) for `Leader_Changed` and `UNAVAILABLE`. Shutting that channel down makes gRPC deliver `onError` to the other live watches sharing it, and each of those callbacks runs `requestReconnect()` and then `invalidateAttemptStub()`, which is `synchronized` on the same client (line 384). They block on the monitor this thread is holding.

The executor is single threaded (`pd-kv-watch-reconnect`, lines 88 and 89), and both `reconnect()` (line 462) and the `onStartTimeout()` watchdog (line 324) are scheduled on it, so one in-flight attempt stalls every other subscription's reconnect and its start timeout.

I am not claiming a proven deadlock: whether the stall becomes permanent depends on whether gRPC channel termination waits on a blocked listener, which I could not exercise here. The contention itself follows from the source.

Requested change: issue `streamingCall()` outside the client monitor, keeping only the observer and `attemptGeneration` bookkeeping inside it, so gRPC callback threads never contend with a thread that is inside `resetStub()`.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
```diff
@@ -106,16 +121,26 @@ public static void init(PDConfig pdConfig, int cacheSize, long expiration) {
                     "allowed to be initialized again", instance.caches.limit(),
                     instance.caches.expiration(), instance.client);
         }
-        INSTANCE.compareAndSet(null, new SchemaDriver(pdConfig, cacheSize,
-                                                      expiration));
+        INSTANCE.set(new SchemaDriver(pdConfig, cacheSize, expiration));
     }
 
-    public static void destroy() {
+    public static synchronized void destroy() {
```
bitflicker64: ⚠️ `destroy()` can pin the `SchemaDriver.class` monitor and block `init()` indefinitely.

`init()` (line 115) and `destroy()` share that class monitor, and `destroy()` now blocks inside `closeResources()`, then `this.client.close()` (line 140), and finally `AbstractClient.closeChannel()`, which spins until the channel terminates:

```java
while (channel != null &&
       !channel.shutdownNow().awaitTermination(100, TimeUnit.MILLISECONDS)) {
    continue;
}
```

A channel that does not terminate holds the monitor for good, and every later `init()` blocks on it.

The exposure is new. Before this PR `destroy()` was not synchronized and did not close the client; it only cancelled the cache cleaner and cleared `INSTANCE`, all non-blocking. Serialising the double-init race is the right call, but it now puts an unbounded network shutdown under the same lock. This PR's macos CI run is consistent with that: `testDestroyClosesOwnedKvClient` spent 2.587 s before reaching its assertion, which fits waiting on this monitor.

Requested change: keep only the `INSTANCE` transition under the monitor and move the blocking client close outside it. Capping the await inside `closeChannel()` would help too, but that is pre-existing code and better as a follow-up.

bitflicker64 (commented):
Blocking: yes. Summary: The `AbstractClient` transport/lifecycle split and the per-subscription `WatchSubscription` model are the right fixes, and the new watch tests genuinely run and pass, but the `SchemaDriverTest` added here fails CI at this head, and two new lifecycle paths hold a monitor across an unbounded channel shutdown loop. Evidence: exact head `0a77c55ac6fa0baaf0e97d98eb0d15a1c28a1531` read via `gh api -X GET repos/apache/hugegraph/pulls/3157/files --paginate` and `gh api -X GET "repos/apache/hugegraph/contents/<path>?ref=0a77c55a" -H 'Accept: application/vnd.github.v3.raw'`; CI via `gh api -X GET repos/apache/hugegraph/commits/0a77c55a/check-runs` (22 success, 2 failure) with logs from `gh api -X GET repos/apache/hugegraph/actions/jobs/99501401894/logs`; `PDClientSuiteTest` runs 45 tests on #3164 and 77 here, matching the 3 to 35 growth in `KvClientTest`.

Three smaller notes, not worth inline comments. The description says `-Dtest=KvClientTest` ran "12 tests, 0 failures/errors", but this head has 35 `@Test` methods in that class, so that figure looks like it came from the first of the six commits on the branch. "Main Changes" does not mention the `hugegraph-struct` and `SchemaDriver` work, which is two of the five changed files. And the `hstore` failure on `VertexCoreTest.testQueryByJointIndexesWithSearchAndTwoRangeIndexesAndWithin` is joint and range index behaviour with no path to the PD watch client or `SchemaDriver`, and that job passes on #3164, so I do not attribute it to this PR, though I did not reproduce it.

I did not restate the review comments already open at this head on watch readiness, unconditional reconnect, `shouldRotateWatchTransport` coverage, semaphore ownership, `WatchRequest` placement, stopped watch visibility, or the stale `Started` test gap. The three below are distinct from those. No local build backs these findings: fetching the head into a checkout was not available in this session, so this is static review of the exact head sources plus CI logs.

bitflicker64 (commented):
Blocking: yes. Summary: the per-subscription watch model, the `AbstractClient` transport/lifecycle split and the 40-test regression suite are the right fixes for #3152 and all pass locally, but `KvClient.stubResetTimeoutMillis()` applies the 5 s watch-start budget to every blocking KV and lock call, cutting the honoured `pd.grpc_timeout` from a measured 21.3 s to a pinned 5.0 s on the same peers; `SchemaDriver.destroy()` also publishes the driver while its `KvClient` is already closed. Evidence: exact head f99b6bdddf7554760f2920f719238bb1811bcdd7 vs merge base 7bb624b99; `mvn -o -pl hugegraph-pd/hg-pd-test -am test -Dtest=KvClientTest` 40/40 and `mvn -o -pl hugegraph-struct -am test -Dtest=SchemaDriverTest` 5/5 pass; upstream CI green 24/24; timing probe over three unreachable peers with identical `PDConfig` gives `KvClient.get()` 4984/5213/5212 ms at head against 21328/35403/6511 ms at base, while `MetaClient.getGraphs()` on the same peers stays at 25071 ms.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/KvClient.java
```diff
@@ -75,6 +131,16 @@ protected AbstractBlockingStub createBlockingStub() {
         return KvServiceGrpc.newBlockingStub(channel);
     }
 
+    @Override
+    protected long stubResetTimeoutMillis() {
+        return Math.max(1L, Math.min(config.getGrpcTimeOut(), WATCH_START_TIMEOUT_MS));
```
bitflicker64, line 136: ⚠️ Important. The watch-start budget is applied client-wide, so it also shrinks the PD reconnect budget for every blocking KV and lock call.

`stubResetTimeoutMillis()` is consumed by `AbstractClient.resetStub()` as the total deadline for the whole peer loop (AbstractClient.java:156-159), and `resetStub()` is reached from `getBlockingStub()` (AbstractClient.java:107-116) — not just from the watch path. That method backs `put`, `get`, `delete`, `deletePrefix`, `scanPrefix`, `putTTL`, `keepTTLAlive`, `lock`, `lockWithoutReentrant`, `isLocked`, `unlock` and `keepAlive`. With the default `PDConfig.grpcTimeOut = 60000` (PDConfig.java:34) this override returns 5000, and the derived per-peer deadline becomes `min(60000, remaining / remainingHosts)` — about 1.7 s each across three PD peers instead of 60 s.

Measured against three unreachable peers (`10.255.255.1-3:8686`), same `PDConfig`, same `getBlockingStub()` path:

| build | `KvClient.get()` | `MetaClient.getGraphs()` |
|---|---|---|
| base `7bb624b99` | 21328 / 35403 / 6511 ms | 37053 ms |
| head `f99b6bdd` | 4984 / 5213 / 5212 ms | 25071 ms |

At this head `KvClient` pins to ~5.0 s on every run — exactly `WATCH_START_TIMEOUT_MS` — while a sibling client on identical configuration keeps its multi-peer budget. So a configured `pd.grpc_timeout` of 60 s is silently honoured for `MetaClient` but capped at 5 s for `KvClient`. During a slow PD leader election or a GC pause longer than the per-peer slice, `lock()` and `keepAlive()` now fail with `PD_UNREACHABLE` where they previously waited. The same cap applies to store-node startup config reads (`PdConfigureListener.onApplicationEvent` calls `scanPrefix`/`put`) and to `PdMetaDriver`/`SchemaDriver`.

Requested change: scope the 5 s readiness budget to the watch-start path only — for example thread it through the async stub reset that `startWatch()` triggers, or gate the override on the async stub — and leave `getBlockingStub()` on the inherited `grpcTimeOut * hostCount` budget so `pd.grpc_timeout` keeps governing lock and KV traffic.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
```diff
@@ -98,24 +117,73 @@ public static void init(PDConfig pdConfig) {
...
     public static void destroy() {
-        SchemaDriver instance = INSTANCE.get();
-        if (instance != null) {
-            instance.caches.cancelScheduleCacheClean();
-            instance.caches.destroyAll();
-            INSTANCE.set(null);
+        SchemaDriver instance = null;
+        CountDownLatch completion;
+        boolean closeResources = false;
+        synchronized (LIFECYCLE_LOCK) {
```
bitflicker64, line 137: ⚠️ Important. `destroy()` keeps `INSTANCE` published while the driver's `KvClient` is being closed, so callers can be handed a driver whose PD client is already shut down.

`INSTANCE` is only cleared in the `finally` at line 158-161, after `instance.closeResources()` (line 156) has already run `client.close()`. The new `testDestroyKeepsInstanceUntilResourcesAreClosed` asserts exactly this window. Meanwhile `SchemaGraph.schemaDriverInit()` (SchemaGraph.java:56-64) checks `SchemaDriver.getInstance() == null`, sees the still-published instance, skips `init()`, and returns it — and `SchemaGraph`'s constructor immediately calls `loadConfig()` -> `graphConfig()` -> `client.get()` on that driver.

By then `KvClient.close()` has set `closed = true`, shut down both reconnect executors and called `shutdownNow()` on the channel, so `AbstractClient.resetStub()` breaks out on its first `isShutdown()` check (AbstractClient.java:159) and `getBlockingStub()` throws `PD_UNREACHABLE`. On the base commit `destroy()` never closed the client, so a driver observed in this window still worked. `init()` is unusable in the window too: it takes `LIFECYCLE_LOCK`, sees the non-null `INSTANCE`, and throws `NotAllowException` (asserted by `testInitDoesNotWaitForDestroyCleanup`).

This is the mirror image of the earlier review note about clearing `INSTANCE` too early — the fix moved the hazard rather than removing it, because publication and usability are still two separate states.

Requested change: make the instance unreachable before its resources are torn down while still preventing a premature re-`init()` — for example clear `INSTANCE` inside the first `LIFECYCLE_LOCK` block and let the `destroying` flag alone block `init()` until cleanup finishes, so `getInstance()` never returns a driver with a closed client. Please add a test that calls `getInstance()` during the cleanup window and asserts the returned driver is either absent or still usable.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
```diff
@@ -249,39 +326,70 @@ protected <ReqT, RespT> KVPair<Boolean, RespT> concurrentBlockingUnaryCall(
...
+                                  "RPC streaming call failed", e);
+        }
+    }
+
+    protected <ReqT, RespT> void streamingCall(MethodDescriptor<ReqT, RespT> method, ReqT request,
+                                               StreamObserver<RespT> responseObserver,
+                                               int retry,
+                                               Consumer<Channel> attemptConsumer)
+            throws PDException {
+        Consumer<Channel> previous = this.streamingAttemptConsumer.get();
```
bitflicker64, line 362: 🧹 Minor. Routing `attemptConsumer` through a `ThreadLocal` makes it invisible to the documented 4-arg override point, which silently disables transport rotation.

This overload stores the consumer in `streamingAttemptConsumer` (line 63) and relies on the 4-arg `streamingCall` reading it back (line 335-338). A subclass that overrides the 4-arg method — which `testLegacyStreamingCallOverrideRemainsDispatchTarget` (KvClientTest.java:893, `LegacyStreamingOverrideKvClient` at KvClientTest.java:1502-1518) deliberately establishes as supported — never invokes the consumer.

`KvClient.recordAttemptChannel()` then never runs, so `subscription.attemptChannel` stays null. On `Leader_Changed`, `requestReconnect(..., rotateTransport = true)` copies that null into `reconnectChannel` (KvClient.java:429-432), `reconnect()` skips `invalidateAttemptStub()` (KvClient.java:556-558), the cached async stub survives, and the watch re-attaches to the same follower — the exact failure this PR is fixing, reintroduced for any subclass that keeps the legacy override.

The coupling is also implicit: the two calls only work because they happen on the same thread, and the value is keyed on a per-instance `ThreadLocal` whose entry is only reclaimed when the `finally` runs or the client is collected.

Requested change: pass the consumer explicitly rather than through thread state — have both public overloads delegate to one private implementation that takes `Consumer<Channel> attemptConsumer` as a parameter (null for the legacy signature), and drop the `ThreadLocal` field. That keeps the 4-arg override working as a dispatch target while making it obvious at the call site that an overrider must forward the consumer.

SebastianGruza:
Live before/after for this branch (head `f99b6bd`) on a real cluster: PD + 3 store nodes, two HugeGraph servers built from the same tree sharing the PD in PD-meta mode (`usePD=true`), server A on :8080 and server B on :8082. Server distributions built from master `98477f0` (before) and from this head (after); PD and stores unchanged.

**Observable that lazy loading cannot compensate:** in PD-meta mode `GraphManager.graph()` loads an unknown graph from PD meta on first access, so "B sees a graph created on A" proves nothing about the watch. Removal is different — B keeps a graph it has already constructed until the `GRAPH/REMOVE` event arrives. So each probe is: A creates `g` → B constructs it on first access (HTTP 200) → A drops `g` → poll B's `GET …/graphs/g/schema/propertykeys` until it stops answering 200. Between probes PD is stopped for 30 s (SIGTERM via `stop-hugegraph-pd.sh`, then `kill -9` of the port owner) and restarted; the stores stayed up throughout.

| Step | master `98477f0` | this branch |
|---|---|---|
| baseline: create → drop | B dropped `g1` after **1 s** (`Accept graph add signal` seen once) | B dropped `g1` after **1 s** |
| PD down 30 s, back, settle 15 s → create → drop | B **still serves the dropped `g2` after 60 s** (HTTP 200); no add signal in B's log → watch dead | B dropped `g2` after **1 s**; add signal seen |
| second PD outage → create → drop | B **still serves `g3` after 60 s**; watch dead | B dropped `g3` after **1 s**; add signal seen |

B's log on this branch during each outage: `Failed to reconnect watch for key HUGEGRAPH/hg/EVENT/GRAPH/{ADD,REMOVE,UPDATE,CLEAR,SCHEMA/CLEAR}` about once per second per subscription, then `AbstractClient connect to host = … success` and five fresh `set watch client id` lines within a second of PD listening again. On master B logs nothing after the first `set watch client id` — the single re-`listen()` from `onError` never comes back once PD is down.

So the failure described in #3152 reproduces on master with a plain 30 s PD restart, and this branch recovers every metadata watch after it, twice in a row.

Two small remarks, neither blocking:
- The retry path logs a WARN **with a full stack trace** for every attempt: with the 4 graph-event subscriptions plus one schema-cache-clear subscription per open graph, server B wrote ~210 `Failed to reconnect watch for key …` warnings (each with a trace) per 30 s outage. A one-line WARN, or the trace only on the first failure of a subscription, would keep this readable during a longer PD outage.
- Unrelated to this PR but hit while setting the test up: with `usePD=true` the server does not start on a JVM whose default locale uses a decimal comma (`GraphSpace.info()` does `Float.parseFloat(String.format("%.2f", …))`, `pl_PL` → `"0,00"` → `NumberFormatException`). I'll file it separately.

Scripts and full logs (both sides, both variants): https://github.com/SebastianGruza/hugegraph-oracle-suite — `cluster/pd_watch_exp2.sh`, `results/pd_watch_3152_*.log`. Happy to re-run on the next head — ping me.

imbajin (commented):
Pushed eeb7041e8 to this branch to fix the two confirmed P1 issues: blocking KV/lock discovery keeps its configured timeout, and SchemaDriver is unpublished before resource cleanup while concurrent init remains rejected. Both regression checks failed before the fix and pass now.

Five independent review lanes, including an adversarial review, found no remaining P0/P1 blocker after the fixes; the final incremental diff was independently checked. Validation: KvClientTest 41/41, SchemaDriverTest 5/5, editorconfig:format, full clean compile and diff check passed. Review score: 8.5/10; merge readiness still depends on CI for this new head.

The legacy four-argument streamingCall override concern in discussion_r3901065273 remains open as a compatibility follow-up. There is no production override in this repository; replacing both overloads with a private delegate would bypass the existing override dispatch, so I left that contract change for discussion. This PR restores watch reconnection; missed-event reconciliation remains tracked separately in #3151.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
```diff
@@ -98,24 +117,75 @@ public static void init(PDConfig pdConfig) {
...
-        SchemaDriver instance = INSTANCE.get();
-        if (instance != null) {
-            throw new NotAllowException(
-                    "The SchemaDriver [cacheSize=%s, expiration=%s, " +
-                    "client=%s] has already been initialized and is not " +
-                    "allowed to be initialized again", instance.caches.limit(),
-                    instance.caches.expiration(), instance.client);
+        synchronized (LIFECYCLE_LOCK) {
+            if (destroying) {
+                throw new NotAllowException("The SchemaDriver is being destroyed");
```
bitflicker64, line 122: ⚠️ Important. `init()` now throws for the whole duration of a `destroy()`, and the only production caller has no handling for it.

`destroy()` unpublishes `INSTANCE` first (line 145) and keeps `destroying = true` until `closeResources()` finishes, so this guard rejects every `init()` made in that window. The only production caller is `SchemaGraph.schemaDriverInit()` (`hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaGraph.java:56-65`), which calls `init()` precisely when `getInstance()` returns `null`:

```java
if (SchemaDriver.getInstance() == null) {
    synchronized (SchemaDriver.class) {
        if (SchemaDriver.getInstance() == null) {
            SchemaDriver.init(this.pdConfig);
        }
    }
}
return SchemaDriver.getInstance();
```

On `master`, `destroy()` kept `INSTANCE` published until cleanup finished, so a `new SchemaGraph(...)` racing with a destroy saw a non-null instance, skipped `init()` and returned normally. At this head the same race propagates `NotAllowException("The SchemaDriver is being destroyed")` out of the `SchemaGraph` constructor. The line-64 `return SchemaDriver.getInstance()` can also hand back `null` when a destroy lands just after a successful `init()`, and `loadConfig()` dereferences it immediately.

Requested change: have `init()` return the instance it published (or add a `getOrInit`) so the caller uses that reference instead of re-reading `INSTANCE`, and make the `destroying` window wait for completion — or be retried by `SchemaGraph` — rather than surfacing as an exception to graph construction.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
```diff
@@ -98,24 +117,75 @@ public static void init(PDConfig pdConfig) {
...
+            }
+            SchemaDriver instance = INSTANCE.get();
+            if (instance != null) {
+                throw new NotAllowException(
+                        "The SchemaDriver [cacheSize=%s, expiration=%s, " +
+                        "client=%s] has already been initialized and is not " +
+                        "allowed to be initialized again", instance.caches.limit(),
+                        instance.caches.expiration(), instance.client);
+            }
+            INSTANCE.set(new SchemaDriver(pdConfig, cacheSize, expiration));
```
bitflicker64, line 132: ⚠️ Important. The driver is constructed while `LIFECYCLE_LOCK` is held, so `destroy()` now blocks on PD reachability.

`new SchemaDriver(...)` runs inside this `synchronized (LIFECYCLE_LOCK)` block, and the constructor calls `listenMetaChanges()`, which registers four watches. Each `KvClient.listen()` reaches `AbstractClient.resetStub()` and a blocking `getMembers()` per peer, budgeted by `KvClient.asyncStubResetTimeoutMillis()` = `min(grpcTimeOut, 5000)` ms — roughly 5 s per watch against a slow or unreachable PD, up to ~20 s for all four. `destroy()` acquires the same lock at line 140 and waits that whole time before it can even begin unpublishing.

On `master` the constructor ran outside any lock (`INSTANCE.compareAndSet(null, new SchemaDriver(...))`) and `destroy()` took no lock at all, so shutdown was never gated on PD reachability. This is the mirror of the "destroy pins the monitor" problem already fixed earlier in this PR.

Requested change: reserve publication under the lock with an `initializing` flag, construct the driver outside the lock, then re-acquire it to publish — so no network I/O happens while the lifecycle lock is held.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
```diff
@@ -98,79 +104,154 @@ public static <T extends AbstractStub> T setAsyncParams(T stub, PDConfig config)
                                new Authentication(config.getUserName(), config.getAuthority()));
     }
 
-    protected AbstractBlockingStub getBlockingStub() throws PDException {
+    protected synchronized AbstractBlockingStub getBlockingStub() throws PDException {
```
bitflicker64, line 107: ⚠️ Important. Making `getBlockingStub()`/`getStub()` fully `synchronized` removes the lock-free cached-stub path, so healthy unary calls now queue behind a reconnect.

On `master` both methods used double-checked locking and returned without touching the client monitor whenever the stub was already cached:

```java
protected AbstractBlockingStub getBlockingStub() throws PDException {
    if (proxy.getBlockingStub() == null) {
        synchronized (this) { ... resetStub(); }
    }
    return setBlockingParams(proxy.getBlockingStub(), config);
}
```

At this head every call acquires the monitor, and `resetStub()` runs under that same monitor while making blocking `getMembers()` calls bounded by `stubResetTimeoutMillis()` (= `grpcTimeOut * hostCount`, 180 s at the 60 s default with three peers) or 5 s for `KvClient`'s async stub. So while one watch reconnect sits in `getStub()` → `resetStub()`, every unrelated `put`/`get`/`delete`/`scanPrefix`/`lock`/`unlock`/`keepAlive` on the same client blocks — even though its blocking stub is still valid. `PdMetaDriver` uses one `KvClient` for both the watches and all unary metadata operations (`hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/meta/PdMetaDriver.java:66-209`), so this is the server's hot metadata path.

Requested change: keep the double-checked fast path — publish the stub through a `volatile` field or an `AtomicReference` on `AbstractClientStubProxy` so the read is safe — and enter the monitor only when a reset is actually needed.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
```diff
@@ -98,79 +104,154 @@ public static <T extends AbstractStub> T setAsyncParams(T stub, PDConfig config)
...
+            String host = resetStub(asyncStubResetTimeoutMillis());
+            if (host.isEmpty()) {
+                throw new PDException(Pdpb.ErrorType.PD_UNREACHABLE_VALUE,
+                                      "PD unreachable, pd.peers=" + config.getServerHost());
             }
         }
         return setAsyncParams(proxy.getStub(), config);
     }
 
+    protected synchronized void invalidateAsyncStub() {
```
bitflicker64: 🧹 Minor. This no-arg `invalidateAsyncStub()` has no caller.

`git grep invalidateAsyncStub` at this head matches only the declaration here, the `Channel`-scoped overload at line 133, and the two uses of that overload (`AbstractClient.java:351` and `KvClient.java:439`). Nothing in `hg-pd-client`, `hg-pd-test`, `hugegraph-struct` or the server calls this form.

Requested change: drop it, or document why it is an intended extension point — it is new `protected` surface on a client base class, so leaving it unused invites a subclass to call the unconditional variant and clear a stub another thread just installed.

bitflicker64 (commented):
Blocking: no. Summary: the `AbstractClient` transport/lifecycle split and the per-subscription watch model are the right fix for #3152, CI is green at this head, and the earlier P0/P1 reports are addressed; three regressions introduced by the hardening itself remain open (two in the new `SchemaDriver` lifecycle, one in `AbstractClient` locking) plus one piece of dead protected API. Evidence: full exact-head diff at eeb7041e89c522c2c2420e5adcb83f8fa22745da compared against `origin/master` baselines of `AbstractClient`, `KvClient` and `SchemaDriver`; call-site reads of `SchemaGraph.schemaDriverInit()`, `PdMetaDriver`, `AbstractClientStubProxy` and `PDException`; `gh pr checks 3157` reports 24/24 green.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
imbajin: Removed the unused overload in 3124faebc. Channel-scoped invalidation remains the only supported path; targeted tests and the full reactor compile pass.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
imbajin: A four-argument override that delegates to super retains the attempt consumer; the ThreadLocal is restored/removed in finally. No production override replacing this transport exists in the repository. Routing both overloads directly to a private delegate would bypass the legacy virtual override, contradicting the compatibility test. Custom transport/channel tracking remains an extension-contract discussion, not a demonstrated regression in this PR.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
imbajin: The destroy-window rejection is real and explicitly tested. However, concurrent graph creation during global driver destruction has no established supported contract or production destroy caller in this repository; the getInstance() re-read/null race also exists on the base. Keeping this as a lifecycle-contract discussion: adding getOrInit or waiting/retry semantics needs a maintainer decision and should not be folded into watch recovery without one.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
imbajin: Construction does hold the lifecycle lock. With unreachable PD, though, the first synchronous registration failure aborts the constructor, so four failed registrations / ~20 s is not the actual path. Serialization prevents publication during teardown; replacing it with an initializing/cancellation protocol changes lifecycle semantics. With no production destroy caller in this repository, I am retaining this as a design discussion rather than making that broader change here.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
imbajin: The getter-only fix is insufficient: once resetStub() closes the shared channel, the base unary retry path also waits on the monitor. However, the new watch-triggered rotation adds a real tradeoff: an old leader can still forward unary RPCs to the new leader (KvServiceGrpcImpl.get), while watch reconnect closes that usable channel and may next probe an unresponsive peer. Unary traffic then waits for discovery. Keeping this open as a shared-transport availability discussion; preserving a usable unary channel during watch discovery needs a broader lifecycle decision, not just an unlocked getter.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
imbajin: Closing for this PR: retain the existing virtual dispatch and super-delegating override behavior. No affected production replacement override has been identified; custom transport extensions can be addressed when a concrete downstream contract requires them. No private-delegate rewrite is needed here.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
imbajin: Closing with the accepted lifecycle boundary: callers coordinate initialization, use and destruction. This PR does not guarantee successful graph construction concurrent with global driver destruction; the existing fail-fast behavior remains. No getOrInit/wait/retry redesign is needed for watch recovery.

hugegraph-struct/src/main/java/org/apache/hugegraph/SchemaDriver.java
imbajin: Accepted for this PR: initialization remains serialized with destruction, including its PD discovery work. Callers coordinate lifecycle transitions; adding an initialization/cancellation state machine would broaden this fix without a current production requirement. Resolving on that scope decision.

hugegraph-pd/hg-pd-client/src/main/java/org/apache/hugegraph/pd/client/AbstractClient.java
imbajin: Accepted tradeoff for this PR: ordinary KV/lock requests may experience transient delay while watch recovery rotates the shared transport. Recovery-time isolation is not a requirement here. Transport separation can be considered independently if that requirement changes; an unlocked getter alone would not provide it. Resolving with this explicit limitation, not as a claim that the delay was eliminated.

imbajin (approved):
+1. Score: 9/10 for the agreed watch-recovery scope. Five independent review lanes, including an adversarial review, and final whole-diff verification are complete. KvClientTest (41) and SchemaDriverTest (5), formatting, full clean compile and all 22 current-head CI checks pass. No remaining blockers within this scope.

Accepted boundaries: shared-transport recovery may briefly delay ordinary KV/lock requests; callers coordinate SchemaDriver initialization and destruction; existing streaming override dispatch is preserved. The remaining threads are resolved on these explicit scope decisions, not on a claim that the tradeoffs were eliminated.

Next: #3151 for missed-event graph metadata reconciliation and discussion #3205 for schema cache reconciliation. This PR restores future notifications for #3152; it does not replay missed events or resolve those convergence workstreams.

PR #3159, opened 2026-08-16
refactor(server): adapt REST API for hubble & enhance k8s mode

imbajin:
## Closeout status

- [x] Publish REST API `0.72` through source properties, the API JAR manifest
  and the runtime version endpoint.
- [x] Support explicit GraphSpace-wide `ALL_GRAPHS` roles for existing and
  future graphs.
- [x] Prepare the audit limiter during login so the first authenticated write
  succeeds without a warm-up request.
- [x] Apply concrete-graph `ALL_GRAPHS` checks and reject mixed-GraphSpace role
  disclosure.
- [x] Keep trusted metadata callbacks behind an internal elevation boundary and
  restore every auth, request GraphSpace and task context.
- [x] Remove the incomplete general Gremlin mutation classifier and implicit
  request-time legacy observer migration/cleanup.
- [x] Preserve the HStore schema-metadata clear bugfix and its required remote
  cache invalidation.
- [x] Restore only `docker/hbase/**` to the existing Docker CI trigger.
- [x] Rebase onto current ASF `master` without dropping the existing bugfixes.
- [ ] Finish current-head CI and exact-image Kubernetes/Hubble acceptance.
- [ ] Obtain upstream approval.

## Before → After

| Area | Before | After |
| --- | --- | --- |
| API capability | Current master and 1.7 both reported API `0.71` | Current master reports API `0.72` |
| GraphSpace roles | Default roles were not recognized consistently for concrete graphs | Explicit `ALL_GRAPHS` roles cover existing and future graphs |
| Role disclosure | A manager shortcut could expose a mixed-space grant | The shortcut accepts only one GraphSpace |
| Trusted callbacks | Elevation depended on ambient/thread context | A private callback boundary saves, clears and restores all contexts |
| Gremlin permissions | An incomplete mutation classifier changed the established contract | The existing GREMLIN EXECUTE contract remains |
| Legacy roles | Request paths scanned and mutated per-graph observer bindings | API `0.72` uses explicit `ALL_GRAPHS`; migration is a separate TODO |
| HStore deletion | Backend clear could leave PD schema metadata behind | Schema metadata and remote caches are cleared before provider data |

## Final-head evidence

- Head: `6851b1cd49f1770f6f27152e468a6e58c933af79`
- Base synced to ASF master: `c6853e70c6fd10e441d66a0633aca60640d0c7dd`
- Focused reactor tests: 44 passed, 0 failed
- Standard `UnitTestSuite`: 661 passed, 0 failed, 1 skipped
- Reactor build, Checkstyle/RAT and `git diff --check`: passed
- Three independent re-reviews completed; the broader deletion-lifecycle race
  is explicitly tracked in the consolidated TODO rather than expanded here
- Current-head GitHub checks and exact-image runtime acceptance: in progress

Approval and merge remain upstream review gates, not claims of this closeout.


imbajin:
### Why this branch contains PD Raft changes

The PD change in `3d9d9544` is independent of the REST API `0.72` bump. It was identified while preparing the real 3-PD distributed matrix, where PD peers use stable hostnames but container/Pod IPs may change after restart or membership updates.

Before:

- `IpAuthHandler` resolved peer hostnames only at startup or an explicit peer-list update;
- the same hostname resolving to a new Pod IP could leave the Raft allowlist stale and reject a valid peer;
- peer changes could update Raft membership without a reliably synchronized allowlist;
- a fully unresolved/empty result followed the old allow-all path;
- resolver, singleton, RPC-server and failed-start cleanup were incomplete.

After:

- configured/runtime peer addresses are normalized and bounded before use;
- DNS resolution has timeout, concurrency and stale-entry limits, with background refresh;
- successful membership changes refresh the active allowlist;
- startup/shutdown failures clean the resolver, handler and RPC resources;
- unit, service-update and Raft integration tests cover these boundaries.

Most of the apparent size is defensive behavior and tests: this commit is `+1375/-226`, including about 700 added test lines and relocation/expansion of the old `IpAuthHandlerTest`.

Scope note: this is not required to distinguish API `0.71` from `0.72`, nor by the Hubble permission capability itself. It should be reviewed as a standalone PD reliability/security change (and preferably split from the API-version PR). Passing tests do not replace the pending final 3+3+3 runtime validation.
### Scope freeze

The Hubble dependency is limited to the API `0.72` contract and the GraphSpace-wide default-role behavior. The PD Raft DNS/IP work is independent and must not grow further except for a blocker reproduced by the required 3+3+3 runtime. It should remain separately reviewable/splittable from the Hubble compatibility change.

imbajin (commented):
Blocking: yes. Summary: No actionable code defect was independently confirmed at this head. Approval remains blocked by the current codecov/project failure. Evidence: Java 11 targeted tests passed: PD auth/raft 20/20, PDService raft update 6/6, GraphSpace API 13/13; current checks show codecov/project FAILURE.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
```diff
@@ -2461,4 +2518,36 @@ public String toString() {
...
+        Set<HugePermission> permissions = EnumSet.noneOf(HugePermission.class);
+        collectTraversalPermissions(traversal, permissions);
+        return permissions;
+    }
+
+    private static void collectTraversalPermissions(
+                        Traversal.Admin<?, ?> traversal,
+                        Set<HugePermission> permissions) {
+        for (Step<?, ?> step : traversal.getSteps()) {
+            if (step instanceof AddVertexStartStep ||
```
imbajin: ‼️ Critical: This condition lists every step treated as a mutation, but omits TinkerPop 3.7.6's `MergeVertexStep` and `MergeEdgeStep`. A merge traversal with `onMatch` can update an existing vertex/edge through `property(...)` while the new pre-check sees no WRITE permission; the existing element iterator returns raw graph elements, so the later property call is not protected by the proxy's direct `addVertexProperty`/`addEdgeProperty` methods. Please include both merge step classes in the WRITE set, recurse through their child traversals, and add execute-only regression tests for both create and onMatch paths.

imbajin (commented):
Blocking: yes. Summary: The new Gremlin mutation pre-check misses TinkerPop merge steps, so execute-only users can reach mutation paths without a required WRITE check. Evidence: `traversalPermissions()` only recognizes AddVertexStartStep, AddVertexStep, AddEdgeStartStep, AddEdgeStep, AddPropertyStep, and DropStep at HugeGraphAuthProxy.java:2533-2541; TinkerPop 3.7.6 also has MergeVertexStep and MergeEdgeStep, whose on-match path can call Vertex.property directly. Please classify both merge steps as WRITE (including nested children) and add execute-only regression tests for mergeV/mergeE create and onMatch cases.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
imbajin: Fixed in 8c69441c. `MergeVertexStep` and `MergeEdgeStep` are now classified as WRITE steps (including provider subclasses via class-hierarchy matching), while the existing `TraversalParent` walk recursively inspects all local/global children. The class-name guard is intentional because this branch still compiles against TinkerPop 3.5.1, so importing 3.7.x merge classes would force an unrelated dependency upgrade. Added strategy-level execute-only regressions for vertex/edge create and onMatch shapes plus a merge-child recursion case. Verified with `mvn -ntp -P unit-test -pl hugegraph-server/hugegraph-test -am -Dtest=HugeGraphAuthProxyTest -Dsurefire.failIfNoSpecifiedTests=false test` (27 tests, all passing).

imbajin: Follow-up hardening in 17615dc0 scopes compatibility matching to the exact TinkerPop `org.apache.tinkerpop.gremlin.process.traversal.step.map.MergeVertexStep` / `MergeEdgeStep` class names while still walking superclasses for provider implementations. An unrelated step with the same simple name is now explicitly verified as non-WRITE; exact 3.7 class-name compatibility and recursive child traversal coverage remain. `HugeGraphAuthProxyTest`: 25 tests passed; targeted API/test Checkstyle also passed.

imbajin: Correctness-test follow-up in 8d03d9ca: added documented test-only fixtures under the exact TinkerPop merge FQCNs because this branch is pinned to 3.5.1. Vertex/edge create and onMatch shapes now flow through `traversalPermissions()` and `TraversalStrategyProxy.apply()`, proving execute-only users are rejected at WRITE precheck. The exact-package merge child test verifies recursive WRITE+DELETE collection, while the unrelated same-simple-name fixture remains non-WRITE. `HugeGraphAuthProxyTest`: 28 tests passed; targeted API/test Checkstyle passed.

imbajin: Final fixture-packaging hardening in 2ed446b3: moved the exact-package TinkerPop 3.7 compatibility fixtures from `src/main/java` to `src/test/java`, added `target/test-classes` only to the unit-test execution classpath, and made the main-source regression test construct fixtures reflectively. The four vertex/edge create/onMatch strategy tests, exact-package child recursion test, and external same-simple-name negative test remain active. Clean focused run: 28 tests passed; targeted API/test Checkstyle passed. After a clean build, exact merge fixtures exist only in `target/test-classes`; exact `org/apache/tinkerpop/.../Merge(Vertex|Edge)Step.class` entries are absent from `target/classes` and the main JAR input/archive.

imbajin: Version-bound fixture selection finalized in b94347b5. The compatibility sources now live under `src/test-tinkerpop-3.5.1/java`, and build-helper adds only `src/test-tinkerpop-${tinkerpop.version}/java`. Current 3.5.1 runs retain all 28 focused tests; an effective-POM check with `-Dtinkerpop.version=3.7.6` selects `src/test-tinkerpop-3.7.6/java`, which is absent, so the 3.5.1 fixtures cannot silently shadow real 3.7 classes. Targeted Checkstyle passed; exact merge fixture counts remain target/classes=0, main JAR=0, target/test-classes=2.

imbajin: Final API-compatibility adjustment in f38b4ad6: the reflective test factory now prefers the official TinkerPop 3.7.6 merge constructor `(Traversal.Admin, boolean)` with `isStart=true`, and onMatch/child setup uses `addChildOption(Merge.onMatch, Traversal.Admin)` when the real `Merge` enum exists. Only the version-bound 3.5.1 fixture path falls back to its test-only `addChild` API; those fixtures now expose the official two-argument constructor, so current tests exercise constructor selection deterministically. Focused result remains 28/28; targeted API/test Checkstyle passed; main JAR exact merge fixture count remains 0.

imbajin: Correction: this Server branch is pinned to TinkerPop 3.5.1, where MergeVertexStep and MergeEdgeStep do not exist. Hubble does not depend on those steps; its mergeV/mergeE entries are editor syntax hints only. The forward-looking 3.7 guard and test fixtures were therefore speculative scope and have been fully reverted in 6f78df53. The current tree is identical to 28641f54. Merge-step authorization should be handled with the real classes and runtime evidence in a dedicated TinkerPop upgrade change.

imbajin (commented):
Blocking: yes. Summary: The current head has authorization correctness and isolation regressions in GraphSpace role checks, Gremlin mutation enforcement, and callback context cleanup. Evidence: exact-head static review plus six independent read-only lanes; latest checks pass for builds/Docker/CodeQL, while dependency-review is cancelled and Codecov failures are non-blocking.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -203,20 +205,25 @@ public String checkDefaultRole(@Context GraphManager manager,
...
+                           StringUtils.isNotEmpty(graph);
         if (hasGraph) {
             validGraph(manager, name, graph);
         }
 
         boolean result;
         if (hasGraph) {
-            result = authManager.isDefaultRole(name, graph, user,
-                                               defaultRole);
+            result = authManager.isDefaultRole(name, graph, user, defaultRole);
```
imbajin: ‼️ Critical: A space-wide OBSERVER is persisted through `createSpaceDefaultRole()` under the `ALL_GRAPHS` marker, but this graph-specific branch checks only `isDefaultRole(name, graph, user, defaultRole)`. A user with the new space-wide role is therefore reported as `false` when the same check includes a concrete `graph` parameter; the parallel `ManagerAPI` path has the same omission. Please treat the `ALL_GRAPHS` role as covering the requested graph, and add a regression test for both endpoints.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeAuthenticator.java
```diff
@@ -359,8 +363,15 @@ public static boolean match(Object role, RolePermission grant,
                 }
             }
 
-            RolePermission rolePerm = RolePermission.fromJson(role);
-            return rolePerm.contains(grant);
+            RolePermission grantedRole = RolePermission.fromJson(grant);
+            RolePerm rolePerm = RolePerm.fromJson(role);
+            if (resourceObject != null &&
```
imbajin, line 369: ⚠️ Important: This new authorization shortcut returns `true` when the operator is a SPACE manager for the current graph space and the target grant merely contains an entry for that space. `UserAPI.role()` then serializes the complete `rolePermission(user)` object, so a `space-a` manager can read the `space-b` entries of a multi-space grant. Please return a graph-space-scoped projection (or reject mixed grants) before allowing this path, and add a cross-space response-isolation test.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
imbajin: ‼️ Critical: The classifier only recognizes concrete `Add*`, `AddPropertyStep`, and `DropStep` instances. A Groovy lambda such as `g.V().sideEffect { it.get().property('k', 'v') }` contains none of these steps, while `vertices()` returns raw graph elements whose `property()` mutates the backend directly. An EXECUTE-only user can therefore bypass the new WRITE check. Please reject or sandbox lambda mutations, or route element mutations through an authorized proxy; add an execute-only strategy-level regression that calls `apply()` rather than only reflecting over the classifier.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -2434,19 +2428,14 @@ public static ConsumerWrapper wrap(Consumer consumer) {
 
         @Override
         public void accept(T t) {
-            boolean grpcThread = false;
             try {
-                grpcThread = Thread.currentThread().getName().contains("grpc");
-                if (grpcThread) {
-                    HugeGraphAuthProxy.setAdmin();
+                if (Thread.currentThread().getName().contains("grpc")) {
```
imbajin: ‼️ Critical: The previous gRPC listener wrapper cleared authentication state in `finally`; this path now scopes only `AuthContext` through `runAsAdmin()`. `HugeGraphAuthProxy.CONTEXTS` is an `InheritableThreadLocal`, and `REQUEST_GRAPH_SPACE` is not cleared here, so a reused or inherited listener thread can retain a prior request identity/graph space after the callback or an exception. Please clear or save/restore all auth thread-locals at this listener boundary in `finally`, and add a thread-reuse/exception regression test.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
```diff
@@ -2327,7 +2377,9 @@ public TraversalStrategiesProxy(TraversalStrategies strategies) {
 
         @Override
         public List<TraversalStrategy<?>> toList() {
```
imbajin, line 2392: ⚠️ Important: TinkerPop's `TraversalStrategies.toList()` contract requires an immutable list, and both the interface default and `DefaultTraversalStrategies` wrap their result with `Collections.unmodifiableList`. This override returns a mutable `ArrayList`, allowing callers to add/remove strategy proxies and changing the public API contract. Please return `Collections.unmodifiableList(proxies)` and add an immutability regression test.

.github/workflows/docker-build-ci.yml
```diff
@@ -24,10 +24,17 @@ on:
...
   pull_request:
     paths:
-      - '**/Dockerfile*'
+      - '.github/workflows/docker-build-ci.yml'
       - '.dockerignore'
-      - 'hugegraph-server/hugegraph-dist/docker/**'
-      - 'hugegraph-server/hugegraph-dist/src/assembly/static/bin/util.sh'
+      - '.mvn/**'
+      - 'pom.xml'
+      - 'hugegraph-commons/**'
```
imbajin, line 31: ⚠️ Important: Removing the old `**/Dockerfile*` filter leaves the repository's `docker/hbase/**` tree outside every `pull_request.paths` entry, although it still contains a Dockerfile, entrypoint, and HBase configuration. Changes to that image will no longer trigger Docker Build CI and can merge without any image validation. Please retain `docker/hbase/**` in the trigger or add a dedicated HBase build job with checks appropriate to that image.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -259,16 +266,19 @@ public void deleteDefaultRole(@Context GraphManager manager,
...
-        E.checkArgument(!hasGraph || StringUtils.isNotEmpty(graph),
-                        "Must set a graph for observer");
+        boolean hasGraph = defaultRole.equals(HugeDefaultRole.OBSERVER) && StringUtils.isNotEmpty(graph);
         if (hasGraph) {
             validGraph(manager, name, graph);
         }
         if (hasGraph) {
             authManager.deleteDefaultRole(name, user, defaultRole, graph);
         } else {
             authManager.deleteDefaultRole(name, user, defaultRole);
```
imbajin, line 265: ⚠️ Important: Space-wide OBSERVER deletion removes the `ALL_GRAPHS` role first and then performs independent per-graph deletes. If any later metadata delete fails, the request returns an error while the global role is already gone and some legacy graph-level grants remain active, leaving authorization state partially migrated. Please make this migration transactional or explicitly idempotent/resumable with compensation, and cover an injected mid-loop failure.

looksaw2:
Superseded by hugegraph/hugegraph#198. The reviewed fixes and remaining compatibility acceptance are now tracked in the organization repository; please close this ASF PR.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
```diff
@@ -2414,6 +2466,11 @@ public void apply(Traversal.Admin<?, ?> traversal) {
              */
             String caller = Thread.currentThread().getName();
             if (!caller.contains(TraversalStrategiesProxy.REST_WORKER)) {
+                for (HugePermission permission :
```
imbajin: ‼️ Critical
Blocking: yes. Summary: The new traversal pre-check maps data mutations to `ResourceType.GREMLIN` WRITE/DELETE, so existing roles with data-resource WRITE/DELETE plus GREMLIN EXECUTE are rejected even though the public Gremlin contract only grants `gremlin_execute`. Evidence: lines 2469-2475 call `verifyNamePermission` with `ResourceType.GREMLIN`, while the existing role fixtures grant WRITE on VERTEX and EXECUTE on GREMLIN; resource matching is type-specific. Please check the actual graph data resource or introduce and migrate explicit GREMLIN WRITE/DELETE permissions with compatibility tests.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
```diff
@@ -178,13 +204,27 @@ public static void setRequestGraphSpace(String graphSpace) {
         REQUEST_GRAPH_SPACE.set(graphSpace);
     }
 
-    public static Context setAdmin() {
-        Context old = getContext();
-        AuthContext.useAdmin();
-        return old;
+    public static void runAsAdmin(Runnable runnable) {
```
imbajin, line 198: ‼️ Critical
Blocking: yes. Summary: This new public admin closure can be invoked by untrusted Gremlin reflection unless it is explicitly hidden, and it changes the caller seen by every operation in the closure to `User.ADMIN`. Evidence: `runAsAdmin()` sets `AuthContext` to `User.ADMIN` at lines 207-211, `getContext()` gives that context priority at lines 221-225, while `HugeFactoryAuthProxy.registerPrivateActions()` filters `getContext`/`resetContext` but does not filter `runAsAdmin`. Please make this trusted bridge non-public or add it to the script method denylist, and add an unauthorized Gremlin regression test.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -259,16 +266,19 @@ public void deleteDefaultRole(@Context GraphManager manager,
...
+            if (defaultRole.equals(HugeDefaultRole.OBSERVER)) {
+                for (String currentGraph : manager.graphs(name)) {
```
imbajin: ⚠️ Important
Blocking: yes. Summary: Space-wide OBSERVER cleanup only visits graphs currently returned by `manager.graphs(name)`, so a legacy `<graph>_observer` role for a graph deleted earlier is never removed and can become active again if that graph name is recreated. Evidence: this loop is the only legacy-role cleanup at lines 276-280, while `GraphManager.dropGraph()` removes graph config before clearing the backend and does not remove the graph-scoped auth metadata. Please enumerate legacy observer roles from auth metadata or clean them during graph deletion, and cover delete/recreate with a user and group.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
```diff
@@ -519,6 +519,9 @@ public void clearBackend() {
 
         LockUtil.lock(this.spaceGraphName(), LockUtil.GRAPH_LOCK);
         try {
+            if (this.isHstore()) {
+                ((CachedSchemaTransactionV2) this.schemaTransaction()).clear();
```
imbajin, line 527: ‼️ Critical
Blocking: yes. Summary: The new HStore schema clear is protected only by `GRAPH_LOCK`, but schema writes use independent per-type/per-id locks, so a concurrent schema write can recreate metadata after `super.clear()` and leave it inconsistent with the subsequently cleared store. Evidence: `StandardHugeGraph.clearBackend()` calls `CachedSchemaTransactionV2.clear()` at line 523, while `SchemaTransactionV2.saveSchema()` and `removeSchema()` acquire only `hugeType2Group(schema.type())` locks. Please coordinate clear with the same graph-wide/distributed schema-write lock and add a concurrent clear/write regression.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
```diff
@@ -467,6 +467,7 @@ public void clear() {
         // Clear schema info firstly
         super.clear();
         this.clearCache(false);
+        this.notifySchemaCacheClear();
```
imbajin, line 470: ⚠️ Important
Blocking: no. Summary: HStore clear now broadcasts `schema-cache-clear` directly, and `storeProvider.clear()` immediately emits `STORE_CLEAR`, whose listener calls `clearCache(true)` and broadcasts the same event again under the default `task.sync_deletion=false`. Evidence: the new `notifySchemaCacheClear()` at line 470 follows `clearCache(false)`, while the existing STORE_CLEAR listener uses `notify = !STORE_INIT` and `clearCache(true)`. Please make one layer own the broadcast or add a clear-operation guard, and verify the event is emitted once.

.github/workflows/docker-build-ci.yml
```diff
@@ -78,3 +86,34 @@ jobs:
...
+            's|.*<Implementation-Version>\([^<]*\)</Implementation-Version>.*|\1|p' \
+            hugegraph-server/hugegraph-api/pom.xml)
+          ACTUAL_MANIFEST=$(unzip -p "$API_JAR" META-INF/MANIFEST.MF |
+            sed -n 's/^Implementation-Version: *//p' | tr -d '\r')
+          EXPECTED_PROPERTY=$(sed -n 's/^ApiVersion=//p' \
+            hugegraph-commons/hugegraph-common/src/main/resources/version.properties)
+          ACTUAL_PROPERTY=$(unzip -p "$COMMON_JAR" version.properties |
+            sed -n 's/^ApiVersion=//p' | tr -d '\r')
+
+          [[ "$ACTUAL_MANIFEST" == "$EXPECTED_MANIFEST" ]] || {
```
imbajin, line 113: ⚠️ Important
Blocking: no. Summary: This check validates the manifest and `version.properties` against separate source literals, but it does not verify the version returned by the runtime endpoint or that the two API version representations are equivalent. Evidence: `ApiVersion.VERSION` prefers the API JAR manifest and `VersionAPI` returns it, so this head can pass with manifest `0.72.0.0` and property `0.72`; lines 112-118 compare each artifact independently and never exercise `/versions`. Please compare normalized canonical values and assert the packaged runtime API response as well.

imbajin (commented):
Blocking: yes. Summary: The selected head still has unresolved authorization and lifecycle correctness risks, and the workflow's exact-head runs are waiting for maintainer approval. Evidence: current-head checks show seven `action_required` workflow runs; the inline findings below are independent of that external gate.

imbajin:
TODO follow-ups intentionally kept out of this minimal API 0.72 closeout:

1. Design an explicit, versioned migration/cleanup path for legacy
   `<graph>_observer` bindings, including deleted-and-recreated graph names.
   API 0.72 now uses explicit `ALL_GRAPHS`; request-time graph scans and hidden
   role mutation were removed rather than extended.
   Context: https://github.com/apache/hugegraph/pull/3159#discussion_r3838617173
2. Coordinate graph deletion with concurrent data/schema mutations through one
   documented graph-wide lifecycle gate, including lock ordering and a
   concurrency regression. This spans the existing `storeProvider.clear()`
   boundary and should not be approximated by a local HStore-only lock.
   Context: https://github.com/apache/hugegraph/pull/3159#discussion_r3838617174
3. Define exact-once schema-cache notification ownership for HStore graph clear
   without losing the direct broadcast required when
   `task.sync_deletion=true`. The final code preserves correctness and may emit
   a bounded duplicate under the default asynchronous-deletion path.
   Context: https://github.com/apache/hugegraph/pull/3159#discussion_r3838617176

These are explicit follow-ups, not silently implemented compatibility behavior
in this PR. The final head keeps the HStore metadata/cache clear bugfix, restores
the required direct remote invalidation, and leaves broader lifecycle/event
redesign for a separately scoped change.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
imbajin: Fixed on `6851b1cd`: concrete-graph checks now recognize the explicit `ALL_GRAPHS` role in `GraphSpaceAPI`, `ManagerAPI`, and `StandardAuthManagerV2`. Focused endpoint coverage is included.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeAuthenticator.java
imbajin: Fixed on `6851b1cd`: the GraphSpace-manager shortcut now accepts only a single-space grant and rejects mixed-space role disclosure. Cross-space regressions are covered.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
imbajin: Removed on `6851b1cd`: the incomplete traversal mutation classifier and pre-check are gone. This PR keeps the established GREMLIN EXECUTE contract instead of introducing a partial WRITE/DELETE model.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
imbajin: Fixed on `6851b1cd`: the internal admin callback saves, clears, and restores `AuthContext`, proxy contexts, request GraphSpace, and `TaskManager` context, including the exceptional path. Focused tests cover restoration.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
imbajin: Fixed on `6851b1cd`: `toList()` materializes `TraversalStrategyProxy` wrappers and returns an unmodifiable list. Tests verify wrapper preservation and mutation rejection.

.github/workflows/docker-build-ci.yml
imbajin: Fixed on `6851b1cd`: `docker/hbase/**` is restored to the existing Docker build path filter without adding a separate workflow.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
imbajin: Removed on `6851b1cd`: API 0.72 no longer performs implicit per-graph legacy observer cleanup. Explicit `ALL_GRAPHS` is the current contract; versioned legacy migration is recorded in the consolidated TODO.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
imbajin: Removed on `6851b1cd`: the general Gremlin mutation pre-check is deleted, so existing GREMLIN EXECUTE permissions are not silently remapped to WRITE/DELETE.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeGraphAuthProxy.java
imbajin: Fixed on `6851b1cd`: `runAsAdmin` is filtered from Gremlin reflection and is used only behind a private internal callback wrapper. All ambient contexts are restored in `finally`.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
imbajin: The implicit scan/migration path was removed from API 0.72. A versioned legacy-role migration covering deleted/recreated graph names is now explicit in the consolidated TODO: https://github.com/apache/hugegraph/pull/3159#issuecomment-5457674257

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
imbajin: Recorded as a separately scoped lifecycle TODO: deletion must be coordinated with concurrent schema/data mutations through one documented graph-wide gate, not a local HStore-only lock. See https://github.com/apache/hugegraph/pull/3159#issuecomment-5457674257. The current head preserves the focused HStore metadata-clear bugfix.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedSchemaTransactionV2.java
imbajin: The direct notification is intentionally retained: when `task.sync_deletion=true`, the later `STORE_CLEAR` path suppresses its broadcast, so removing the direct call can leave remote schema caches stale. Exact-once ownership without losing that correctness guarantee is recorded in https://github.com/apache/hugegraph/pull/3159#issuecomment-5457674257.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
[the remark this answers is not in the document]
imbajin: Fixed on `6851b1cd`: the HStore path now checks that the schema transaction is `CachedSchemaTransactionV2` before the cast; failure stops provider clear. The ordering, failure, and unexpected-type paths are covered.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -203,20 +205,25 @@ public String checkDefaultRole(@Context GraphManager manager,
...
-                                               defaultRole);
+            result = authManager.isDefaultRole(name, user, defaultRole);
+            if (!result && defaultRole.equals(HugeDefaultRole.OBSERVER)) {
+                for (String currentGraph : manager.graphs(name)) {
+                    if (authManager.isDefaultRole(name, currentGraph, user, defaultRole)) {
+                        result = true;
+                        break;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
imbajin: Removed on `6851b1cd`: the request-time O(number of graphs) legacy observer fallback no longer exists.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/ManagerAPI.java
```diff
@@ -287,20 +287,24 @@ public String checkDefaultRole(@Context GraphManager manager,
...
-                                               defaultRole);
+            result = authManager.isDefaultRole(graphSpace, user, defaultRole);
+            if (!result && defaultRole.equals(HugeDefaultRole.OBSERVER)) {
+                for (String currentGraph : manager.graphs(graphSpace)) {
+                    if (authManager.isDefaultRole(graphSpace, currentGraph, user, defaultRole)) {
+                        result = true;
+                        break;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
imbajin: Removed on `6851b1cd` with the same legacy observer compatibility rollback; API 0.72 uses the explicit `ALL_GRAPHS` role.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/api/space/SchemaTemplateAPITest.java
```diff
@@ -0,0 +1,81 @@
...
+    private static final String GRAPHSPACE = "space";
+    private static final String CREATOR = "creator";
+
+    @Test
+    public void testCreatorCanManageTemplate() {
+        Supplier<AuthManager> authManager =
+                Mockito.mock(Supplier.class);
+
+        Assert.assertTrue(canManage(authManager, CREATOR));
+        Mockito.verifyZeroInteractions(authManager);
```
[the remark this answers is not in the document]
imbajin: Fixed on `6851b1cd`: the test now uses `verifyNoInteractions`.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/auth/HugeAuthenticator.java
```diff
@@ -359,8 +363,15 @@ public static boolean match(Object role, RolePermission grant,
...
+                !RolePermission.isAdmin(grantedRole) &&
+                grantedRole.roles().containsKey(resourceObject.graphSpace()) &&
+                rolePerm.matchSpace(resourceObject.graphSpace(), "space")) {
+                return true;
+            }
+            return RolePermission.fromJson(role).contains(grantedRole);
```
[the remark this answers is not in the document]
imbajin: Fixed on `6851b1cd`: `RolePermission` is parsed once and reused for the single-space manager decision.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -2434,19 +2428,14 @@ public static ConsumerWrapper wrap(Consumer consumer) {
...
+                    HugeGraphAuthProxy.runAsAdmin(() -> this.consumer.accept(t));
+                } else {
+                    this.consumer.accept(t);
                 }
```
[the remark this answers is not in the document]
imbajin: Fixed on `6851b1cd`: thread-name privilege inference was removed. The private metadata callback wrapper now elevates explicitly and restores every context.

.github/workflows/docker-build-ci.yml
imbajin: Verified on the exact final-head HStore image. Pod imageID is `docker-pullable://hugegraph/server@sha256:4111ad8291604de449ff74ec649b3b66eb21cd48c47418e2fc0bc0b4c4c3ce2e`, built with revision label `6851b1cd49f1770f6f27152e468a6e58c933af79`. Helm test called `/versions` and received `"api":"0.72.0.0"`, then `/graphs` returned `hugegraph`; the hook completed successfully.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
```diff
@@ -519,6 +519,13 @@ public void clearBackend() {
 
         LockUtil.lock(this.spaceGraphName(), LockUtil.GRAPH_LOCK);
         try {
+            if (this.isHstore()) {
+                E.checkState(this.schemaTransaction() instanceof
+                             CachedSchemaTransactionV2,
+                             "The HStore schema transaction must be %s",
+                             CachedSchemaTransactionV2.class.getSimpleName());
+                ((CachedSchemaTransactionV2) this.schemaTransaction()).clear();
+            }
```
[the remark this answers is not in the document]
imbajin: No code change. StandardHugeGraph.schemaTransaction() delegates to the current thread-local Txs object: Transaction.schemaTransaction() returns getOrNewTransaction().schemaTx, so both calls are the same instance throughout this locked synchronous block. The instanceof guard already produces the clearer HStore invariant failure; introducing a local here is a style-only cleanup and is intentionally omitted from this minimal Server closeout.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/space/GraphSpaceAPI.java
```diff
@@ -203,20 +200,19 @@ public String checkDefaultRole(@Context GraphManager manager,
...
         if (hasGraph) {
             validGraph(manager, name, graph);
         }
 
         boolean result;
         if (hasGraph) {
             result = authManager.isDefaultRole(name, graph, user,
-                                               defaultRole);
+                                               defaultRole) ||
+                     authManager.isDefaultRole(name, user, defaultRole);
```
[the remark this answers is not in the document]
imbajin: No code change. The fallback is redundant only for StandardAuthManagerV2. GraphSpaceAPI is backend-neutral and StandardAuthManager.isDefaultRole(graphSpace, graph, ...) checks only the concrete graph marker; it does not consult ALL_GRAPHS. The explicit second check is therefore required to preserve API 0.72 ALL_GRAPHS behavior outside the V2/HStore implementation. Focused GraphSpace API coverage passes on the final head.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/auth/ManagerAPI.java
```diff
@@ -287,20 +287,18 @@ public String checkDefaultRole(@Context GraphManager manager,
...
         if (hasGraph) {
             validGraph(manager, graphSpace, graph);
         }
 
         boolean result;
         if (hasGraph) {
             result = authManager.isDefaultRole(graphSpace, graph, user,
-                                               defaultRole);
+                                               defaultRole) ||
+                     authManager.isDefaultRole(graphSpace, user, defaultRole);
```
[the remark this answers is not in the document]
imbajin: No code change. ManagerAPI supports both StandardAuthManager and StandardAuthManagerV2. Only the V2 graph overload folds in ALL_GRAPHS; the standard manager checks the concrete graph marker alone. Keeping the explicit GraphSpace-wide check preserves backend-neutral API 0.72 behavior and the focused Manager/GraphSpace role tests pass.

imbajin:
Final dependency-review rerun note: attempt 3 completed with the same stale all-dependencies-added snapshot and failed on existing vulnerable/low-score packages. The PR changes no dependency coordinates; its only pom.xml delta changes the API JAR Implementation-Version from 0.71.0.0 to 0.72.0.0. The repository-specific dependency-check job passes, all build/storage/Docker jobs pass, and Codecov patch/project pass. This dependency-review result is classified as a base-snapshot failure rather than a current-head dependency regression; no product or dependency suppression change is made.

MrJs133 (approved):

PR #3161, opened 2026-08-17
fix(ci): preserve PD and Store coverage aggregation

### LINKED ISSUE #3160: [Bug] JaCoCo coverage is overwritten across PD/Store CI test profiles
### Bug Type

CI / Test Coverage

### Environment

Current `master`

### Expected behavior

The PD and Store CI workflows run multiple test suites in separate Maven invocations:

```text
PD: pd-common-test, pd-core-test, pd-client-test, pd-rest-test
Store: store-common-test, store-client-test, store-core-test,
       store-rocksdb-test, store-server-test, store-raftcore-test
```

The Codecov upload should represent the union of every suite executed by the job. For
example, coverage produced by `KvClientTest` during `pd-client-test` must still be present
after `pd-rest-test` finishes.

### Actual behavior

Coverage from earlier suites can be lost or omitted before upload:

1. The workflows invoke Maven separately for each profile while reusing module-local JaCoCo
   execution data.
2. `hg-pd-test` and `hg-store-test` regenerate `report-aggregate` during each test invocation
   instead of producing one final report.
3. The PD job runs `mvn clean package` between coverage-producing suites, deleting previously
   collected `target` data.
4. Aggregate reports are generated below the test modules, but the workflow points Codecov at
   the repository-level `target/site/jacoco` path.
5. Store's aggregate project does not directly depend on `hg-store-rocksdb`, so the RocksDB
   bundle is not guaranteed to appear in the aggregate report.

A recent example is PR #3157: the focused `KvClientTest` suite passes and exercises the changed
reconnect paths, but Codecov reports 0% patch coverage for the changed `KvClient` /
`AbstractClient` lines.

### Java 17 compatibility

PD and Store currently use JaCoCo 0.8.4, which officially supports class files only through
Java 12. HugeGraph is moving toward Java 17, while JaCoCo 0.8.8 is the first release with
official Java 17/18 support and is already used by `hugegraph-server/hugegraph-test`.

The coverage fix should therefore update the PD/Store JaCoCo configuration to 0.8.8 as part of
making this pipeline usable on Java 17.

### Proposed solution

1. Move destructive packaging/cleaning before all coverage-producing suites.
2. Append execution data across the separate suite invocations.
3. Assign a deterministic JaCoCo session ID to every suite.
4. Generate one `report-aggregate` during a final `verify` invocation.
5. Upload the exact module-local `jacoco.xml` file.
6. Validate that the report contains every expected suite session, covered instructions, and
   expected module groups.
7. Add `hg-store-rocksdb` as a direct dependency of the Store aggregate project.
8. Upgrade PD/Store JaCoCo from 0.8.4 to 0.8.8 for Java 17 compatibility.

### Scope

This issue covers repository-level PD/Store coverage collection, deterministic
aggregation/upload, and the minimum JaCoCo upgrade needed for Java 17 compatibility.

It does not require production-code changes, artificial coverage increases, Codecov
flags/carry-forward configuration, or a broader Codecov Action modernization.

contrueCT:
## Purpose of the PR

- close #3160

PD and Store CI run several Maven test profiles in separate invocations, but
each invocation previously replaced JaCoCo execution data and generated a
partial report. PD also ran `mvn clean package` after two coverage-producing
test profiles, deleting their data before upload. As a result, Codecov could
receive only the last profile's coverage or an imprecisely selected report.

The existing JaCoCo 0.8.4 configuration also cannot instrument Java 17 class
files, which blocks the project's planned Java 17 migration.

## Main Changes

- Move the PD clean/package step before all coverage-producing test profiles.
- Upgrade PD and Store JaCoCo plugins from 0.8.4 to 0.8.8 and append execution
  data across Maven invocations.
- Assign a deterministic JaCoCo session ID to every selected PD/Store test
  profile and verify the complete expected session set before upload.
- Require at least one non-skipped Surefire test for every selected suite.
  Exclude the existing Store core/server placeholder suites from the coverage
  gate until they contain real tests.
- Generate one aggregate XML report during `verify` and upload that exact file
  to Codecov.
- Parse the aggregate JaCoCo XML structurally, rejecting malformed XML and
  comment-only evidence while accepting valid attribute reordering.
- Require non-zero instruction coverage for the instrumented PD common/client/
  core groups and Store common/client/RocksDB groups. Keep zero-class and
  externally executed aggregate dependencies as presence-only checks.
- Include `hg-store-rocksdb` as a direct report-aggregate dependency only in
  the `jacoco` profile so normal Store test dependency resolution is unchanged.
- Run EditorConfig before tests create RocksDB runtime files, then skip only
  that already-completed check during the final aggregate-only `verify`.
- Add validator and contract tests for malformed arguments, missing reports,
  zero-test and all-skipped suites, malformed and spoofed XML, reordered
  attributes, uncovered reports and individual required groups, missing
  sessions/modules, Maven lifecycle bindings, dependency scope, exact workflow
  inputs, and timeout behavior without GNU coreutils.

## Verifying these changes

    - `hugegraph-server/hugegraph-dist/src/assembly/travis/test-check-jacoco-report.sh`
    - Parse `.github/workflows/pd-store-ci.yml` and all changed POMs.
    - `mvn -q apache-rat:check -N -ntp`
    - `mvn editorconfig:check -pl hugegraph-server/hugegraph-dist -am -ntp`
    - `mvn editorconfig:check -pl hugegraph-pd/hg-pd-test -am -ntp`
    - `mvn editorconfig:check -pl hugegraph-store/hg-store-test -am -ntp`
    - `mvn verify -pl hugegraph-pd/hg-pd-test -am -P jacoco -DskipTests -Deditorconfig.skip=true -ntp`
    - `mvn verify -pl hugegraph-store/hg-store-test -am -P jacoco -DskipTests -Deditorconfig.skip=true -ntp`
    - Verify the normal Store dependency tree still resolves PowerMock's
      `org.javassist:javassist:3.24.0-GA`.
    - Java 17 smoke: JaCoCo 0.8.4 rejects class-file major version 61, while
      0.8.8 instruments the same class and writes execution data successfully.

The complete four-suite PD/four-suite Store workflow and the external Codecov
upload are left to GitHub Actions because the local environment uses JDK 17
while this workflow currently installs JDK 11.

- [ ] Dependencies ([add/update license](https://hugegraph.apache.org/docs/contribution-guidelines/contribute/#321-check-licenses) info & [regenerate_known_dependencies.sh](../install-dist/scripts/dependency/regenerate_known_dependencies.sh))
- [x] Modify configurations
- [ ] The public API
- [x] Other affects (PD/Store CI coverage collection and upload)
- [ ] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`

### DIFF
--- a/.github/workflows/pd-store-ci.yml
+++ b/.github/workflows/pd-store-ci.yml
@@ -100,6 +100,10 @@ jobs:
         run: |
           mvn clean package -U -Dmaven.javadoc.skip=true -Dmaven.test.skip=true -ntp --fail-at-end
 
+      - name: Check source formatting
+        run: |
+          mvn editorconfig:check -pl hugegraph-pd/hg-pd-test -am -ntp
+
       # These tests do not require starting a PD instance. Run them after the
       # clean package so their JaCoCo data survives until the final report.
       - name: Run common test
@@ -154,7 +158,8 @@ jobs:
 
       - name: Generate aggregate coverage report
         run: |
-          mvn verify -pl hugegraph-pd/hg-pd-test -am -P jacoco -DskipTests -ntp
+          mvn verify -pl hugegraph-pd/hg-pd-test -am -P jacoco \
+            -DskipTests -Deditorconfig.skip=true -ntp
 
       - name: Validate aggregate coverage report
         run: |
@@ -209,6 +214,10 @@ jobs:
         # todo remove --fail-at-end after test
         run: |
           mvn clean package -U -Dmaven.javadoc.skip=true -Dmaven.test.skip=true -ntp --fail-at-end
+
+      - name: Check source formatting
+        run: |
+          mvn editorconfig:check -pl hugegraph-store/hg-store-test -am -ntp
 
       - name: Check startup test prerequisites (Store)
         id: store-preflight
@@ -279,7 +288,8 @@ jobs:
 
       - name: Generate aggregate coverage report
         run: |
-          mvn verify -pl hugegraph-store/hg-store-test -am -P jacoco -DskipTests -ntp
+          mvn verify -pl hugegraph-store/hg-store-test -am -P jacoco \
+            -DskipTests -Deditorconfig.skip=true -ntp
 
       - name: Validate aggregate coverage report
         run: |
--- a/.github/workflows/pd-store-ci.yml
+++ b/.github/workflows/pd-store-ci.yml
@@ -33,6 +33,9 @@ jobs:
         with:
           fetch-depth: 5
 
+      - name: Run JaCoCo report validator tests
+        run: hugegraph-server/hugegraph-dist/src/assembly/travis/test-check-jacoco-report.sh
+
       - name: Use staged maven repo settings
         run: |
           cp $HOME/.m2/settings.xml /tmp/settings.xml || true
@@ -65,7 +68,7 @@ jobs:
       USE_STAGE: 'false' # Whether to include the stage repository.
       # TODO: remove outdated env
       TRAVIS_DIR: hugegraph-server/hugegraph-dist/src/assembly/travis
-      REPORT_DIR: target/site/jacoco
+      REPORT_FILE: hugegraph-pd/hg-pd-test/target/site/jacoco/jacoco.xml
 
     steps:
       - name: Install JDK 11
@@ -92,20 +95,22 @@ jobs:
           cp $HOME/.m2/settings.xml /tmp/settings.xml
           mv -vf .github/configs/settings.xml $HOME/.m2/settings.xml
 
-      - name: Run common test
-        run: |
-          mvn test -pl hugegraph-pd/hg-pd-test -am -P pd-common-test
-
-      - name: Run core test
-        run: |
-          mvn test -pl hugegraph-pd/hg-pd-test -am -P pd-core-test
-
-      # The above tests do not require starting a PD instance.
-
       - name: Package
         # todo remove --fail-at-end after test
         run: |
           mvn clean package -U -Dmaven.javadoc.skip=true -Dmaven.test.skip=true -ntp --fail-at-end
+
+      # These tests do not require starting a PD instance. Run them after the
+      # clean package so their JaCoCo data survives until the final report.
+      - name: Run common test
+        run: |
+          mvn test -pl hugegraph-pd/hg-pd-test -am \
+            -P pd-common-test -Djacoco.sessionId=pd-common-test
+
+      - name: Run core test
+        run: |
+          mvn test -pl hugegraph-pd/hg-pd-test -am \
+            -P pd-core-test -Djacoco.sessionId=pd-core-test
 
       - name: Check startup test prerequisites (PD)
         id: pd-preflight
@@ -139,16 +144,32 @@ jobs:
 
       - name: Run client test
         run: |
-          mvn test -pl hugegraph-pd/hg-pd-test -am -P pd-client-test
+          mvn test -pl hugegraph-pd/hg-pd-test -am \
+            -P pd-client-test -Djacoco.sessionId=pd-client-test
 
       - name: Run rest test
         run: |
-          mvn test -pl hugegraph-pd/hg-pd-test -am -P pd-rest-test
+          mvn test -pl hugegraph-pd/hg-pd-test -am \
+            -P pd-rest-test -Djacoco.sessionId=pd-rest-test
+
+      - name: Generate aggregate coverage report
+        run: |
+          mvn verify -pl hugegraph-pd/hg-pd-test -am -P jacoco -DskipTests -ntp
+
+      - name: Validate aggregate coverage report
+        run: |
+          $TRAVIS_DIR/check-jacoco-report.sh \
+            --require-session pd-common-test \
+            --require-session pd-core-test \
+            --require-session pd-client-test \
+            --require-session pd-rest-test \
+            "$REPORT_FILE" \
+            hg-pd-grpc hg-pd-common hg-pd-client hg-pd-core hg-pd-service hg-pd-dist
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3.0.0
         with:
-          file: ${{ env.REPORT_DIR }}/*.xml
+          files: ${{ env.REPORT_FILE }}
 
   store:
     needs: struct
@@ -157,7 +178,7 @@ jobs:
       USE_STAGE: 'false' # Whether to include the stage repository.
       # TODO: remove outdated env
       TRAVIS_DIR: hugegraph-server/hugegraph-dist/src/assembly/travis
-      REPORT_DIR: target/site/jacoco
+      REPORT_FILE: hugegraph-store/hg-store-test/target/site/jacoco/jacoco.xml
 
     steps:
       - name: Install JDK 11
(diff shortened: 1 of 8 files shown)


hugegraph-server/hugegraph-dist/src/assembly/travis/check-jacoco-report.sh
```diff
@@ -0,0 +1,60 @@
...
+    exit 1
+fi
+
+if ! grep -Eq '<counter type="INSTRUCTION" missed="[0-9]+" covered="[1-9][0-9]*"' \
+     "${REPORT_FILE}"; then
+    echo "ERROR: JaCoCo report has no covered instructions: ${REPORT_FILE}" >&2
+    exit 1
+fi
+
+for session in "${REQUIRED_SESSIONS[@]}"; do
```
imbajin: ⚠️ Important: This only proves that some reactor module emitted `<sessioninfo id="...">`. Because the workflow uses `-am` and assigns the same `-Djacoco.sessionId` to the whole reactor, a target profile that matches zero tests can still pass when an upstream module emits the same session and another module contributes covered instructions. Please assert a non-zero test count for the selected PD/Store suite, or use suite/module-specific execution data, before allowing the upload.

hugegraph-server/hugegraph-dist/src/assembly/travis/check-jacoco-report.sh
```diff
@@ -0,0 +1,60 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -uo pipefail
+
+REQUIRED_SESSIONS=()
+while [[ "${1:-}" == "--require-session" ]]; do
+    REQUIRED_SESSIONS+=("${2:-}")
+    shift 2 || true
```
imbajin: ⚠️ Important: If the command ends with `--require-session`, `${2:-}` is accepted as an empty session and `shift 2` fails; `|| true` leaves `$1` unchanged, so this `while` loop never terminates. Please reject a missing or empty value before shifting and exit nonzero for malformed input.

hugegraph-server/hugegraph-dist/src/assembly/travis/check-jacoco-report.sh
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+for session in "${REQUIRED_SESSIONS[@]}"; do
+    if ! grep -Fq "<sessioninfo id=\"${session}\" " "${REPORT_FILE}"; then
+        echo "ERROR: missing JaCoCo session '${session}' in ${REPORT_FILE}" >&2
+        exit 1
+    fi
+done
+
+for module in "$@"; do
```
imbajin: ⚠️ Important: The validator accepts a report with a covered counter and required session even when no module arguments are supplied, because this loop is skipped for an empty `"$@"`. A future workflow edit that drops the module list would therefore still upload an incomplete aggregate. Please require at least one module and add a contract test for the missing-module case, ideally asserting the exact PD/Store module lists.

imbajin (commented):
Blocking: no. Summary: The aggregate report and Maven lifecycle checks pass, but the validator can accept incomplete suite/module coverage and can hang on malformed arguments. Evidence: Exact-head PD/Store aggregate builds and validator contract tests passed; current-head CI Codecov uploads returned HTTP 429 and exited 0.

.github/workflows/pd-store-ci.yml
```diff
@@ -228,32 +268,52 @@ jobs:
...
+          mvn test -pl hugegraph-store/hg-store-test -am \
+            -P store-rocksdb-test -Djacoco.sessionId=store-rocksdb-test
 
       - name: Run raft-core test
         run: |
-          mvn test -pl hugegraph-store/hg-store-test -am -P store-raftcore-test
+          mvn test -pl hugegraph-store/hg-store-test -am \
+            -P store-raftcore-test -Djacoco.sessionId=store-raftcore-test
+
+      - name: Generate aggregate coverage report
```
imbajin, line 292: ⚠️ Important

Blocking: yes. The Store job no longer runs the existing `store-core-test` or `store-server-test` suites, even though both executions remain in `hg-store-test/pom.xml` and the aggregate report still claims coverage for `hg-store-core` and `hg-store-node`. This makes the CI gate a partial test run while presenting a full Store report. Restore both profiles and require their reports/sessions, or explicitly remove them from the covered scope with a documented replacement.

imbajin (commented):
Blocking: yes. Summary: The Store CI job drops two existing test suites while still publishing an aggregate coverage report, so the coverage gate no longer represents the full Store test scope. Evidence: exact-head .github/workflows/pd-store-ci.yml patch and hg-store-test/pom.xml suite definitions.

imbajin (commented):
Blocking: yes. Summary: The coverage workflow now exercises the intended PD/Store profiles and produces aggregate reports, but the validator still permits false-positive coverage evidence in two paths. Evidence: exact-head CI is green, while the findings below concern the validator's acceptance criteria and local contract portability.

.github/workflows/pd-store-ci.yml
```diff
@@ -228,32 +268,69 @@ jobs:
...
+
+      - name: Validate aggregate coverage report
+        run: |
+          # CoreSuiteTest and ServerSuiteTest are existing placeholders with zero tests.
+          $TRAVIS_DIR/check-jacoco-report.sh \
+            --require-test-report \
+            "$TEST_REPORT_DIR/TEST-org.apache.hugegraph.store.common.CommonSuiteTest.xml" \
+            --require-test-report \
+            "$TEST_REPORT_DIR/TEST-org.apache.hugegraph.store.client.ClientSuiteTest.xml" \
+            --require-suite-report \
```
imbajin: ⚠️ Important
Blocking: yes. Summary: The Store coverage gate accepts the Core and Server suites even when both execute zero tests, so it can publish a full-looking aggregate without exercising those modules. Evidence: exact-head `pd-store-ci.yml:312-317` uses `--require-suite-report`, and the exact-head Store job reports `CoreSuiteTest` and `ServerSuiteTest` with `Tests run: 0`; `CoreSuiteTest` remains an empty `@Suite.SuiteClasses({})`. Please either add real tests and use `--require-test-report`, or remove these placeholder suites from the required coverage scope and validate the remaining modules explicitly.

hugegraph-server/hugegraph-dist/src/assembly/travis/check-jacoco-report.sh
```diff
@@ -0,0 +1,145 @@
...
+        echo "ERROR: unable to parse Surefire report: ${test_report}" >&2
+        return 1
+    fi
+    if [[ "${require_tests}" == "true" ]] && (( test_count <= 0 )); then
+        echo "ERROR: Surefire report has no tests: ${test_report}" >&2
+        return 1
+    fi
+}
+
+for suite_report in "${REQUIRED_SUITE_REPORTS[@]}"; do
```
imbajin: ⚠️ Important
Blocking: no. Summary: The validator exits before normal validation on macOS's bundled Bash 3.2 whenever no suite reports are supplied, which is the normal PD path. Evidence: with `set -uo pipefail`, iterating `"${REQUIRED_SUITE_REPORTS[@]}"` after `REQUIRED_SUITE_REPORTS=()` returns `REQUIRED_SUITE_REPORTS[@]: unbound variable` under `/bin/bash` 3.2; the PD workflow does not pass `--require-suite-report`. Please guard the loop with an empty-array check and run the contract harness on Bash 3.2 or provide a portable fallback.

hugegraph-server/hugegraph-dist/src/assembly/travis/check-jacoco-report.sh
```diff
@@ -0,0 +1,145 @@
...
+    validate_test_report "${suite_report}" false || exit 1
+done
+
+for test_report in "${REQUIRED_TEST_REPORTS[@]}"; do
+    validate_test_report "${test_report}" true || exit 1
+done
+
+if ! grep -Eq '<counter type="INSTRUCTION" missed="[0-9]+" covered="[1-9][0-9]*"' \
```
imbajin: ‼️ Critical
Blocking: yes. Summary: The validator searches raw JaCoCo XML text instead of parsing the report, so commented or truncated snippets can satisfy the session, module, and covered-instruction checks, while a valid report with reordered attributes can be rejected. Evidence: `check-jacoco-report.sh:125-143` uses `grep` for all three element types; a report containing the required tags only inside XML comments still matches these expressions. Please parse the XML elements and attributes, reject malformed/comment-only reports, and add reordered-attribute and commented-snippet contract cases.

hugegraph-server/hugegraph-dist/src/assembly/travis/check-jacoco-report.sh
```diff
@@ -0,0 +1,200 @@
...
+
+    local test_count
+    if ! test_count=$(python3 - "${test_report}" <<'PY'
+import sys
+import xml.etree.ElementTree as ET
+
+root = ET.parse(sys.argv[1]).getroot()
+if root.tag.rsplit("}", 1)[-1] != "testsuite" or "tests" not in root.attrib:
+    raise ValueError("not a Surefire testsuite report")
+test_count = int(root.attrib["tests"])
```
imbajin, line 101: ⚠️ Important

Blocking: yes. This only validates that the Surefire `tests` attribute is positive; it ignores `skipped`, so a report such as `tests=2 skipped=2` is accepted even though the suite executed no tests. Because the aggregate/session/module checks can still pass from other reactor output, CI may upload coverage for an all-skipped suite. Please require at least one non-skipped test (for example `tests - skipped > 0`) and add an all-skipped contract case.

hugegraph-server/hugegraph-dist/src/assembly/travis/test-check-jacoco-report.sh
```diff
@@ -0,0 +1,438 @@
...
+    CASE_OUTPUT=$("${VALIDATOR}" "$@" 2>&1)
+    CASE_RC=$?
+}
+
+run_report_case() {
+    run_case --require-test-report "${TMP_DIR}/tests.xml" "$@"
+}
+
+run_case_with_timeout() {
+    CASE_OUTPUT=$(timeout 2 "${VALIDATOR}" "$@" 2>&1)
```
imbajin: ⚠️ Important

Blocking: no. The malformed-argument timeout case invokes GNU-only `timeout`, but the workflow does not install a dependency and this script is also runnable locally. On stock macOS Bash, `timeout` is absent (only `gtimeout` may be available), so the harness exits 127 instead of checking the validator's termination. Please use a portable Python subprocess timeout or explicitly detect/install `gtimeout`.

imbajin (commented):
Blocking: yes. Summary: The current coverage gate still accepts an all-skipped suite, and the new validator contract harness is not portable to stock macOS. Evidence: exact-head validator/workflow diff, Surefire parsing at check-jacoco-report.sh:98-110, and the contract harness timeout invocation.

imbajin (approved):
It's important to point out the critical CI issue to ensure the test coverage & code quality

PR #3167, opened 2026-08-23
fix(server): close shortest path edge iterators

### LINKED ISSUE #3166: [Bug] ShortestPathTraverser does not close edge iterators on early return
### Bug Type (问题类型)

logic / resource lifecycle

### Before submit

- [x] I searched the existing issues, pull requests, and FAQ and did not find a report for this exact `ShortestPathTraverser` iterator leak.

### Environment (环境信息)

- Server source: current `apache/hugegraph` `master` (`431f6e6b431014feb5b50d27ea67180b4610f10b`)
- Backend: HStore, one PD + one Store + one Server on one machine
- OS: Ubuntu 24.04, 16 vCPU, about 78 GiB RAM, swap disabled during the reproduction
- Data size: 41,652,230 vertices and 1,468,365,182 directed edges (SNAP Twitter-2010)
- Workload: REST `shortestpath`, concurrency 12, 180-second warmup + 600-second measurement

Every edge iterator returned by `edgesOfVertex()` should be closed on normal completion, early return, and exception paths. For HStore, closing the iterator is also the signal that allows the remote scanner stream to be cancelled and cleaned up.

`ShortestPathTraverser.Traverser` does not currently guarantee that ownership contract:

1. `forward()` returns immediately when it finds the requested path without closing the current edge iterator.
2. `backward()` has the equivalent early-return path.
3. `superNode()` passes an edge iterator to `IteratorUtils.count()` without closing it explicitly.

In the HStore reproduction, successful requests left Store threads waiting in the scanner delivery path (`HgChannel.send`). The blocked frames accumulated from 11 shortly after launch to 38, 75, 108, and finally 113. The measured phase completed 5,817 of 5,818 requests; one request returned HTTP 500 after approximately 61 seconds. Swap remained zero, so the failure was not caused by swapping.

As a diagnostic control, reducing the Store client-take wait caused abandoned scanner senders to be reaped instead of accumulating, but it also produced 107 timeout warnings and reduced throughput. This confirms the scanner lifecycle path; increasing or reducing timeouts is not a fix.

A focused regression test uses close-tracking edge iterators while keeping the path assertions unchanged. On unpatched `master`, all three path results are correct and all three close assertions fail:

```text
Tests run: 3, Failures: 3
```

The proposed fix retains the original iterator returned by `edgesOfVertex()` and closes it with `CloseableIterator.closeIterator()` in `finally` blocks. The original iterator must be retained because `skipSuperNodeIfNeeded()` may consume and replace it. `superNode()` receives the same explicit `finally` cleanup. Query semantics, limits, and timeouts do not change.

With the patch, the focused test passes 3/3. A full-graph HStore validation then completed all 1,516 warmup and 5,791 measured requests with zero errors, zero swap, stable service PIDs, and zero `HgChannel.send` frames at every sampled point from 0 through 600 seconds.

The problem is independent of a particular vertex or edge value. It is triggered when bounded bidirectional shortest-path traversal finds a path before the current backend iterator has been exhausted.

```groovy
schema.vertexLabel("user")
      .useCustomizeNumberId()
      .ifNotExist()
      .create()
schema.edgeLabel("link")
      .sourceLabel("user")
      .targetLabel("user")
      .ifNotExist()
      .create()
```

contrueCT:
## Purpose of the PR

- close #3166

`ShortestPathTraverser` could return after finding a path without closing the current backend edge iterator. With HStore, the abandoned iterator could leave its remote scanner active and Store threads waiting to deliver pages that the client no longer consumed.

The same ownership gap also existed in the backward search and in the iterator used by the super-node probe.

## Main Changes

- Retain the original iterator returned by `edgesOfVertex()` in both `forward()` and `backward()`.
- Close that source iterator in `finally`, including early-return and exception paths. The source iterator is closed directly because `skipSuperNodeIfNeeded()` may consume and replace it.
- Close the iterator used by `superNode()` after counting it.
- Add focused regression coverage for forward match, backward match, and super-node probe cleanup, and register it in `UnitTestSuite`.

The change does not modify traversal semantics, degree/capacity limits, request timeouts, HStore code, configuration, or public APIs.

## Verifying these changes

    - On unpatched `apache/master`, the new test preserves the expected path results and fails all three iterator-close assertions: `Tests run: 3, Failures: 3`.
    - On this branch, `ShortestPathTraverserTest` passes: `OK (3 tests)`.
    - `MAVEN_OPTS='-Xms64m -Xmx256m' mvn clean compile -pl hugegraph-server/hugegraph-test -am -Dmaven.javadoc.skip=true` completes with `BUILD SUCCESS` for all 21 affected reactor modules.
    - `mvn editorconfig:format` completes with `BUILD SUCCESS`, and `git diff --check` reports no errors.

Additional HStore runtime validation of the same production patch used one PD, one Store, and one Server with the full SNAP Twitter-2010 graph (41,652,230 vertices and 1,468,365,182 directed edges). The bounded shortest-path gate completed 1,516/1,516 warmup and 5,791/5,791 measured requests with zero errors, zero swap, stable service PIDs, and zero blocked `HgChannel.send` frames at all sampled points through 600 seconds.

- [ ]  Modify configurations
- [ ]  The public API
- [x]  Nope

## Documentation Status

- [ ]  `Doc - TODO`
- [ ]  `Doc - Done`
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
@@ -31,7 +31,6 @@ import org.apache.hugegraph.traversal.algorithm.steps.EdgeStep;
 import org.apache.hugegraph.type.define.Directions;
 import org.apache.hugegraph.util.E;
 import org.apache.tinkerpop.gremlin.structure.Edge;
-import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
 
 import com.google.common.collect.ImmutableList;
@@ -187,6 +186,7 @@ public class ShortestPathTraverser extends HugeTraverser {
 
                 Iterator<Edge> sourceEdges = edgesOfVertex(
                         source, this.direction, this.labels, degree);
+                Throwable failure = null;
                 try {
                     Iterator<Edge> edges = skipSuperNodeIfNeeded(
                             sourceEdges, this.degree, this.skipDegree);
@@ -212,8 +212,11 @@ public class ShortestPathTraverser extends HugeTraverser {
                             return paths;
                         }
                     }
+                } catch (RuntimeException | Error e) {
+                    failure = e;
+                    throw e;
                 } finally {
-                    CloseableIterator.closeIterator(sourceEdges);
+                    closeIterator(sourceEdges, failure);
                 }
             }
 
@@ -237,6 +240,7 @@ public class ShortestPathTraverser extends HugeTraverser {
 
                 Iterator<Edge> sourceEdges = edgesOfVertex(
                         source, opposite, this.labels, degree);
+                Throwable failure = null;
                 try {
                     Iterator<Edge> edges = skipSuperNodeIfNeeded(
                             sourceEdges, this.degree, this.skipDegree);
@@ -261,8 +265,11 @@ public class ShortestPathTraverser extends HugeTraverser {
                             return results;
                         }
                     }
+                } catch (RuntimeException | Error e) {
+                    failure = e;
+                    throw e;
                 } finally {
-                    CloseableIterator.closeIterator(sourceEdges);
+                    closeIterator(sourceEdges, failure);
                 }
             }
 
@@ -278,10 +285,14 @@ public class ShortestPathTraverser extends HugeTraverser {
             }
             Iterator<Edge> edges = edgesOfVertex(vertex, direction,
                                                  this.labels, this.skipDegree);
+            Throwable failure = null;
             try {
                 return IteratorUtils.count(edges) >= this.skipDegree;
+            } catch (RuntimeException | Error e) {
+                failure = e;
+                throw e;
             } finally {
-                CloseableIterator.closeIterator(edges);
+                closeIterator(edges, failure);
             }
         }
 
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
@@ -31,6 +31,7 @@ import org.apache.hugegraph.traversal.algorithm.steps.EdgeStep;
 import org.apache.hugegraph.type.define.Directions;
 import org.apache.hugegraph.util.E;
 import org.apache.tinkerpop.gremlin.structure.Edge;
+import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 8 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
```diff
@@ -184,33 +185,36 @@ public PathSet forward(boolean all) {
             while (this.pathResults.hasNextKey()) {
                 Id source = this.pathResults.nextKey();
 
-                Iterator<Edge> edges = edgesOfVertex(source, this.direction,
-                                                     this.labels, degree);
-                edges = skipSuperNodeIfNeeded(edges, this.degree,
-                                              this.skipDegree);
+                Iterator<Edge> sourceEdges = edgesOfVertex(
```
imbajin, line 187: ⚠️ Important

Blocking: yes. `edgesOfVertex(...)` is called before the `try`. For multiple labels, `HugeTraverser` builds an `ExtendableIterator` and opens one backend iterator per label; if a later label query throws, earlier closeable HStore iterators are already owned by that local aggregator but are lost before this `finally` is installed. The same acquisition pattern is used by `backward()`. Please make multi-label iterator acquisition close all previously opened children on failure (or provide an exception-safe acquisition API) before relying on this cleanup block.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
```diff
@@ -184,33 +185,36 @@ public PathSet forward(boolean all) {
...
-                        return paths;
+                        if (paths.isEmpty()) {
+                            continue;
+                        }
+                        results.addAll(paths);
+                        if (!all) {
+                            return paths;
+                        }
                     }
+                } finally {
```
imbajin, line 218: ⚠️ Important

Blocking: no. `CloseableIterator.closeIterator()` wraps an `AutoCloseable.close()` failure in a `RuntimeException`. In these `finally` blocks, that cleanup exception replaces an exception already thrown by `hasNext()`, `next()`, or path matching, so a backend query failure can be reported as a close failure; the same applies to the backward and super-node cleanup blocks. Please preserve the primary exception and attach a close failure as suppressed, propagating it directly only when no primary exception exists.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/traversal/ShortestPathTraverserTest.java
```diff
@@ -0,0 +1,170 @@
...
+                            path.vertices());
+        Assert.assertTrue(forwardEdges.closed());
+        Assert.assertTrue(backwardEdges.closed());
+    }
+
+    @Test
+    public void testCloseEdgesWhenCheckingSuperNode() {
+        Id source = IdGenerator.of(1L);
+        Id target = IdGenerator.of(2L);
+        TrackingIterator sourceEdges = edges(edgeTo(target));
```
imbajin, line 84: ⚠️ Important

Blocking: no. This test supplies one source edge while `skipDegree=2`, so `skipSuperNodeIfNeeded()` never reaches its threshold branch that returns the empty iterator; it also uses a raw `TrackingIterator`, empty labels, and only `shortestPath(..., all=false)`. The added tests therefore do not prove cleanup for the actual super-node skip, wrapper chain, `allShortestPaths()`, or iterator exception paths. Please add threshold, wrapped-backend, full-traversal, and throwing-iterator cases with close assertions.

imbajin (commented):
Blocking: yes. Summary: The new cleanup covers iterators after acquisition, but multi-label acquisition can still leak already-opened scanners, and the regression tests do not exercise the threshold, wrapper, or exception paths. Evidence: exact-head ShortestPathTraverser/HugeTraverser control flow, TinkerPop CloseableIterator behavior, and the added ShortestPathTraverserTest.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
contrueCT: Fixed in 74e87019. Both multi-label overloads now use one exception-safe acquisition helper that tracks every opened child iterator. If a later label query fails, all previously opened iterators are closed; the query failure remains primary and any close failures are suppressed. The normal ExtendableIterator close path now also attempts every child even when one close fails. Added Map and List acquisition-failure tests plus a multi-label wrapper close-failure test.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
contrueCT: Fixed in 74e87019. The forward, backward, and super-node paths now retain the traversal/query failure and attach a close failure as suppressed; a close failure is propagated directly only when there is no primary failure. Added focused tests for all three primary-plus-close-failure paths and the close-only wrapper path.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/traversal/ShortestPathTraverserTest.java
contrueCT: Fixed in 74e87019. The coverage now includes the actual skipDegree threshold, a target that reaches the super-node threshold, the real label ExtendableIterator plus LimitIterator wrapper chain, allShortestPaths(), multi-label acquisition failures, close-only failures, and throwing iterators for forward, backward, and super-node paths. ShortestPathTraverserTest passes all 13 tests locally; ExtendableIteratorTest passes all 13 tests.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/algorithm/ShortestPathTraverser.java
```diff
@@ -184,33 +184,40 @@ public PathSet forward(boolean all) {
...
+                        results.addAll(paths);
+                        if (!all) {
+                            return paths;
+                        }
                     }
+                } catch (RuntimeException | Error e) {
+                    failure = e;
+                    throw e;
+                } finally {
+                    closeIterator(sourceEdges, failure);
```
imbajin, line 219: ⚠️ Important

Blocking: yes. When a finite multi-label iterator reaches its limit, `LimitIterator.fetch()` already closes the wrapped `ExtendableIterator`. This `finally` block then closes the same wrapper again, and `ExtendableIterator.close()` retains its children without a closed guard, so the current backend iterator receives a second `close()` call.

`AutoCloseable` does not require idempotent close semantics, so the second call can trigger duplicate remote cleanup or turn an otherwise successful traversal into a cleanup failure. Please make the wrapper/origin close path idempotent and add a regression asserting that every child is closed exactly once.

imbajin (commented):
Blocking: yes. The cleanup direction is correct, but finite multi-label iteration can close the same backend iterator twice; please make cleanup idempotent and cover the exactly-once path before approval.

imbajin (approved):
Good catch (set line to 120 in PR 3159)

PR #3169, opened 2026-08-24
fix(ci): modernize Codecov uploads

### LINKED ISSUE #3168: [Bug] Legacy Codecov uploads are silently dropped by tokenless rate limits
### Bug Type

CI / Test Coverage

### Environment

Current `master` and pull requests from forks.

### Expected behavior

Every workflow that produces coverage should upload its report reliably to the same Codecov commit. A missing upload should not make the commit appear to have lost coverage from modules whose tests completed successfully.

### Actual behavior

The repository still uses `codecov/codecov-action@v3` in Commons, PD/Store/HStore, and Server CI. Pull requests from forks cannot access `CODECOV_TOKEN`, so these jobs use the legacy anonymous uploader.

The legacy uploader shares a global tokenless rate limit. When that limit is exceeded, Codecov returns HTTP 429, but the uploader exits with status 0 by default. GitHub Actions therefore shows the producing job as green even though the report never reached Codecov.

PR #3167 demonstrated the failure at commit `74e8701`: only one of the expected upload sessions was accepted, causing an apparent 4.92 percentage-point project coverage drop. PR #3161 reproduced the same behavior at commit `a63e48e9`: PD and HStore uploads returned 429 while other uploads happened to succeed, leaving `codecov/project` with an incomplete report set.

### Evidence

- PR #3167 analysis: https://github.com/apache/hugegraph/pull/3167#issuecomment-5392946433
- PR #3161 PD/Store/HStore run: https://github.com/apache/hugegraph/actions/runs/32705078797
- Codecov documents that tokenless uploads from legacy uploaders share a global rate limit and recommends upgrading the CI action: https://docs.codecov.com/docs/codecov-tokens#legacy-upload-methods

### Proposed solution

1. Upgrade every repository coverage upload from `codecov/codecov-action@v3` to `@v5` or newer.
2. Replace the deprecated `file` input with `files`.
3. Pass the existing `CODECOV_TOKEN` to trusted push/non-fork runs; fork pull requests will use the modern action's fork-aware tokenless flow because repository secrets are intentionally unavailable there.
4. Add a lightweight configuration contract test covering all Codecov upload steps so workflows cannot drift back to the legacy uploader or omit the trusted-run token input.
5. Keep the current non-blocking uploader policy in this change. Whether an external Codecov outage should fail otherwise healthy test jobs is a separate gate-policy decision.

### Scope

This issue covers repository-wide Codecov uploader modernization and authentication routing only. It does not change JaCoCo collection, coverage thresholds, flags/carry-forward behavior, or production code. The PD/Store JaCoCo aggregation work remains tracked independently by #3160 and #3161.

contrueCT:
## Purpose of the PR

- close #3168

HugeGraph still uses `codecov/codecov-action@v3` for Commons,
PD/Store/HStore, and Server coverage uploads. Fork pull requests cannot access
`CODECOV_TOKEN`, so v3 falls back to the legacy anonymous uploader. That
uploader shares a global rate limit: HTTP 429 failures were observed on #3167
and #3161, while the uploader still exited with status 0 and left Codecov with
only a subset of the expected reports.

This PR modernizes the upload boundary only. It does not change JaCoCo
collection, aggregation, coverage thresholds, or production code. The
PD/Store report aggregation work remains independent in #3160 and #3161.

## Main Changes

- Upgrade all five repository coverage uploads from
  `codecov/codecov-action@v3` to `@v5`.
- Pass the existing `CODECOV_TOKEN` for trusted push and same-repository runs.
  On fork pull requests the secret remains unavailable by design, and v5 uses
  its fork-aware tokenless upload path.
- Replace the deprecated `file` input with `files` in every upload step.
- Add a repository-wide configuration contract that rejects legacy Codecov
  actions, missing `files` inputs, missing trusted-run token inputs, removed
  expected uploads, and newly added legacy upload steps.
- Keep Codecov transport errors non-blocking in this PR; this change does not
  add `fail_ci_if_error: true`.

## Verifying these changes

    - `bash hugegraph-server/hugegraph-dist/src/assembly/travis/test-codecov-upload-config.sh`
    - Parse the three changed workflows with SnakeYAML 1.33.
    - `mvn -q apache-rat:check -N -ntp`
    - `mvn editorconfig:check -pl hugegraph-server/hugegraph-dist -am -ntp`
    - `git diff --check`
    - The contract failed against the pre-change v3 configuration and passed
      after all five upload steps were migrated.
    - Mutation checks verified that it rejects a legacy upload in a newly
      added workflow and does not mistake `env.token` for `with.token`.

The external upload result is intentionally left to this PR's GitHub Actions
run, which exercises the v5 fork-aware tokenless path that cannot be reproduced
as a local Codecov upload.

- [x]  Modify configurations
- [ ]  The public API
- [x]  Other affects (Codecov coverage upload transport and authentication)
- [ ]  Nope

## Documentation Status

- [ ]  `Doc - TODO`
- [ ]  `Doc - Done`
- [x]  `Doc - No Need`

### DIFF
--- a/.github/workflows/commons-ci.yml
+++ b/.github/workflows/commons-ci.yml
@@ -58,6 +58,9 @@ jobs:
           mvn test -pl hugegraph-commons/hugegraph-rpc -Dtest=UnitTestSuite -DskipCommonsTests=false
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3.0.0
+        uses: codecov/codecov-action@v5
         with:
-          file: target/jacoco.xml
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: hugegraph-commons/target/jacoco.xml
+          disable_search: true
+          fail_ci_if_error: false
--- a/.github/workflows/pd-store-ci.yml
+++ b/.github/workflows/pd-store-ci.yml
@@ -36,6 +36,9 @@ jobs:
       - name: Run JaCoCo report validator tests
         run: hugegraph-server/hugegraph-dist/src/assembly/travis/test-check-jacoco-report.sh
 
+      - name: Run Codecov upload configuration tests
+        run: hugegraph-server/hugegraph-dist/src/assembly/travis/test-codecov-upload-config.sh
+
       - name: Use staged maven repo settings
         run: |
           cp $HOME/.m2/settings.xml /tmp/settings.xml || true
@@ -184,9 +187,12 @@ jobs:
             hg-pd-grpc hg-pd-common hg-pd-client hg-pd-core hg-pd-service hg-pd-dist
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3.0.0
+        uses: codecov/codecov-action@v5
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           files: ${{ env.REPORT_FILE }}
+          disable_search: true
+          fail_ci_if_error: false
 
   store:
     needs: struct
@@ -316,9 +322,12 @@ jobs:
             hg-store-grpc hg-store-common hg-store-client hg-store-rocksdb
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3.0.0
+        uses: codecov/codecov-action@v5
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           files: ${{ env.REPORT_FILE }}
+          disable_search: true
+          fail_ci_if_error: false
 
   hstore:
     needs: struct
@@ -387,6 +396,9 @@ jobs:
           $TRAVIS_DIR/run-tinkerpop-test.sh $BACKEND tinkerpop
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3.0.0
+        uses: codecov/codecov-action@v5
         with:
-          file: ${{ env.REPORT_DIR }}/*.xml
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ${{ env.REPORT_DIR }}/*.xml
+          disable_search: true
+          fail_ci_if_error: false
--- a/hugegraph-server/hugegraph-dist/src/assembly/travis/test-codecov-upload-config.sh
+++ b/hugegraph-server/hugegraph-dist/src/assembly/travis/test-codecov-upload-config.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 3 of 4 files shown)


.github/workflows/pd-store-ci.yml
```diff
@@ -146,9 +149,10 @@ jobs:
           mvn test -pl hugegraph-pd/hg-pd-test -am -P pd-rest-test
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3.0.0
+        uses: codecov/codecov-action@v5
         with:
-          file: ${{ env.REPORT_DIR }}/*.xml
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ${{ env.REPORT_DIR }}/*.xml
```
imbajin: ‼️ Blocking: yes. Summary: This PD/Store input must be rebased with #3161's aggregate report path. Evidence: #3161 now defines and validates `${{ env.REPORT_FILE }}` under the PD/Store test modules, while exact-head jobs 97408972447 and 97408972425 logged `not_found_files: ["target/site/jacoco/*.xml"]` and then discovered reports from multiple modules. Resolve the conflict by using `${{ env.REPORT_FILE }}` here and set `disable_search: true`; retain #3161's JaCoCo validator step as well.

.github/workflows/commons-ci.yml
```diff
@@ -58,6 +58,7 @@ jobs:
           mvn test -pl hugegraph-commons/hugegraph-rpc -Dtest=UnitTestSuite -DskipCommonsTests=false
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3.0.0
+        uses: codecov/codecov-action@v5
         with:
-          file: target/jacoco.xml
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: target/jacoco.xml
```
imbajin: ‼️ Blocking: yes. Summary: The explicit Commons report path is incorrect and the default search masks the failure. Evidence: exact-head job 97406468118 reported `not_found_files: ["target/jacoco.xml"]` and then uploaded four reports, including PD, Store, and Server XMLs. Use `hugegraph-commons/target/jacoco.xml` and set `disable_search: true` so this job uploads only the Commons report.

hugegraph-server/hugegraph-dist/src/assembly/travis/test-codecov-upload-config.sh
```diff
@@ -0,0 +1,182 @@
...
+
+    for line_number, version, inputs in uploads:
+        version_match = version_pattern.match(version)
+        if version_match is None or int(version_match.group("major")) < 5:
+            errors.append(
+                f"{relative_path}:{line_number}: Codecov action {version} "
+                "uses the legacy uploader"
+            )
+
+        if not inputs.get("files"):
```
imbajin: ⚠️ Blocking: no. Summary: The new contract test does not guard the report-selection behavior this change depends on. Evidence: it checks only action count/version, `files`, and token; the exact-head CI still passed the configuration test while Codecov reported missing explicit files and auto-discovered unrelated reports. Extend the contract to assert each workflow's intended report path and `disable_search: true`, while keeping #3161's report-content validation separate.

imbajin (commented):
Blocking: yes. Summary: The Codecov action upgrade is directionally correct, but this head is not merge-ready after #3161 because it still uses incorrect or over-broad report inputs. Evidence: exact-head jobs 97408972447, 97408972425, and 97406468118 reported missing explicit files and then uploaded unrelated reports; the PR is also currently conflicting with master.

imbajin (approved):
Verified exact head 424cbc9. All required CI checks pass, and Codecov accepted all 7 expected v5 upload sessions (Commons, PD, Store, HStore, and 3 Server backends); each resolved one explicit report and completed without HTTP 429, missing-file, or upload errors. The red project status (37.44% vs 51.16%) is caused by comparing the complete 7-session head report with an incomplete 3-session base report, not by a missing head upload. Patch coverage is unaffected. Coverage thresholds and JaCoCo aggregation remain outside #3168's scope.

PR #3171, opened 2026-08-25
perf(docker): share reactor build across images

imbajin:
## Summary

- add `docker/bake.hcl` so BuildKit can solve the PD, Store, HStore Server, and standalone Server images together
- keep the Maven build stage identical across the four Dockerfiles and run it once on the native build platform
- build amd64 and arm64 runtime images from the shared Java artifacts
- use one complete shared build cache and separate lightweight runtime caches
- preserve direct Dockerfile builds and the serial fallback used by older source revisions

Companion workflow PR: https://github.com/hugegraph/actions/pull/27

## Before → after

```mermaid
flowchart LR
  B[Four image builds] --> C[Four Maven reactor builds]
  A[One Bake graph] --> S[One shared Maven build]
  S --> P[PD runtime]
  S --> T[Store runtime]
  S --> H[HStore Server runtime]
  S --> R[Standalone Server runtime]
```

The portable Java build runs once on `BUILDPLATFORM`. Only the runtime stages fan out for amd64 and arm64, so the expensive Maven work stays native instead of running again through QEMU.

## Benchmark

Both measurements used cold caches, the same runner class, the same four images, the same functional checks, and a real Docker Hub publish. Queue time is excluded.

| Cold publish | Build step | Complete job | Change |
| --- | ---: | ---: | ---: |
| Serial image builds | 27m06s | 29m32s | baseline |
| Shared Bake graph | 11m10s | 13m49s | **58.8% faster build, 53.2% faster job** |

## Design

| Area | Implementation |
| --- | --- |
| Shared work | The four Dockerfiles expose the same native Maven stage, which becomes one reusable LLB node |
| Multi-platform build | `BUILDPLATFORM` keeps Maven native; runtime stages use the correct target-platform JRE and only the ARM runtime work uses QEMU |
| Cache layout | The shared build cache uses `mode=max`; each runtime image writes its own `mode=min` cache |
| Local validation | Both platform variants are loaded into Docker's containerd image store under the final tag before publication |
| Compatibility | Each Dockerfile remains directly buildable, and source revisions without Bake keep the serial workflow path |

CI checks that the shared stages remain identical and validates the evaluated Bake targets, platforms, outputs, and cache layout. The published test images contain both linux/amd64 and linux/arm64 variants. Functional Compose, Gremlin, and standalone checks run against the local amd64 variants before any image is pushed.

Repository housekeeping in this PR moves the build and contribution guides under `docs/`, updates their references, removes the tracked IntelliJ project metadata, and limits HStore test noise without changing production task lifecycle logging.

Only temporary test tags were published. `latest` was not changed.


.github/workflows/docker-build-ci.yml
```diff
@@ -25,11 +25,71 @@ on:
...
+                exit
+              }
+              in_build {
+                print
+                seen = 1
+              }
+            ' "$1"
+          }
+
+          reference="hugegraph-pd/Dockerfile"
```
bitflicker64, line 68: If the stage header ever changes casing or name (`as build` is valid Dockerfile syntax), `extract_build_stage` returns empty for every file and the diffs pass vacuously, so the guard silently stops guarding. Reproduced this: rewriting `AS build` to `as build` in all four files makes the check pass even with divergent `mvn` lines. A partial rename is still caught. Cheap fix: assert the reference extraction is non-empty first.

```suggestion
          reference="hugegraph-pd/Dockerfile"
          [ -n "$(extract_build_stage "$reference")" ] || {
            echo "ERROR: no 'AS build' stage found in $reference; guard would pass vacuously"
            exit 1
          }
```

bitflicker64 (approved):
Independently re-verified the core claims. The `build` stage is byte-identical across all four Dockerfiles at c522d73 (no pre-FROM ARGs, single root `.dockerignore`), `SOURCE_REVISION` reaches every Bake target and expands in the cache mount id at runtime, and only the `build-cache` target writes the shared cache. The timing table matches raw job-step timestamps from the linked hugegraph/actions runs: 1754s -> 330s (dry-run) / 501s (first publish) on the same runner class, with the Maven stage fully rebuilt in both candidate runs. Those runs built 89a2586b, whose Dockerfiles and `docker-bake.hcl` are byte-identical to this head; the strengthened CI guard is covered by the Docker Build CI run on the head itself. Ran `buildx bake --print` locally: the jq assertion fails on platform or output mutations, and the awk stage-diff catches a divergent `mvn` line. One hardening nit inline; non-blocking.

Two observations, no action needed in this PR:
- The per-module `buildcache-*` refs are read but no longer refreshed on the Bake path, so runtime-stage layers will slowly go stale and eventually rebuild each run. If that ever shows in timings, runtime targets could export a cache gated on `EXPORT_CACHE`.
- Local builds share `id=hugegraph-maven-local` across revisions, and `mvn install` writes SNAPSHOTs into that shared `.m2`. A full-reactor build resolves modules from the reactor, so the practical risk is low; worth a doc line at most.

docker/bake.hcl
```diff
@@ -0,0 +1,116 @@
...
+  cache-to = EXPORT_CACHE ? [
+    "type=registry,ref=hugegraph/hugegraph:shared-${CACHE_CHANNEL},mode=max",
+  ] : []
+}
+
+target "pd" {
+  inherits   = ["_common"]
+  dockerfile = "hugegraph-pd/Dockerfile"
+  tags       = ["hugegraph/pd:${IMAGE_TAG}"]
+  output     = ["type=docker"]
```
imbajin, line 71: ‼️ critical — `_common` supplies both `linux/amd64` and `linux/arm64`, but this target fixes the output to `type=docker` (and the other three runtime targets do the same). Buildx's Docker exporter only writes a single-platform image; with these two platforms the target cannot export the required manifest list, so the default Bake group cannot build the claimed multi-platform images. Please use a multi-platform-capable `registry`/`image` exporter for the publishing graph, or split out a single-platform local target, and make the CI guard validate the actual export contract.

imbajin (commented):
Blocking: yes. Summary: The default multi-platform Bake graph configures every runtime target to use the single-platform Docker exporter, so it cannot produce the claimed amd64/arm64 images. Evidence: `docker buildx bake --print` at c522d73 expands the runtime targets with platforms=[linux/amd64,linux/arm64] and output type=docker; the Docker exporter supports single-platform results.

.github/workflows/docker-build-ci.yml
imbajin: Fixed in 9fd092a: the CI now fails when the reference build-stage extraction is empty. The mutation check also confirms that changing all headers to `as build` no longer passes vacuously. The final-head `docker-bake-check` passed: https://github.com/apache/hugegraph/actions/runs/33267786454/job/99140828935

docker/bake.hcl
imbajin: Verified on the final heads with Docker's containerd image store. `type=docker` loaded both linux/amd64 and linux/arm64 variants for each final tag, the workflow tested those exact local tags, and only then pushed them. Run/job: https://github.com/hugegraph/actions/actions/runs/33268703532/job/99143223821. Post-push inspection confirms both platforms in all four indexes.

MrJs133 (approved):

PR #2347, opened 2023-11-10
fix(api): refactor/downgrade record logic for slow log

imbajin:
## Purpose of the PR

add some TODOs & assign @SunnyBoy-WYH to address it

- subtask of #2346 (fix #2327)

## Verifying these changes

- [ ] Trivial rework / code cleanup without any test coverage. (No Need)
- [ ] Already covered by existing tests, such as *(please modify tests here)*.
- [x] Need tests and can be verified as follows:
    - xxx

## Does this PR potentially affect the following parts?

- [ ]  Nope
- [ ]  Dependencies (add/update license info) 
- [ ]  Modify configurations
- [ ]  The public API
- [x]  Other affects (slow log logic)

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

PR #2327, opened 2023-10-16
feat(api): support recording slow query log

SunnyBoy-WYH:
## Purpose of the PR

subtask of https://github.com/apache/incubator-hugegraph/issues/2325 

## Main Changes

1. change the log4j.xml to record slow query log to single log file.
2. record log in AccessLogFilter.java
3. add some config items about threshold and enabled.

image

## Verifying these changes

we can run test , and we can see:

image

## Does this PR potentially affect the following parts?

- [ ]  Nope
- [ ]  Dependencies (add/update license info) 
- [x]  Modify configurations
- [ ]  The public API
- [x]  Other affects (typed here)

## Documentation Status

- [ ]  `Doc - TODO`

PR #3177, opened 2026-08-29
fix(server): record bounded request body and client IP in slow query log

### LINKED ISSUE #2468: [Bug] server slow log, support loader import & client IP

bitflicker64:
## Purpose

Replaces #2466 by @SunnyBoy-WYH (the head branch lives on a personal fork and now conflicts with master). Fixes #2468.

The slow query log has printed `body=null` since #2347 commented out the body capture added in #2327, because reading the body broke gzip batch imports from hugegraph-loader. The TODO left in `AccessLogFilter` also asked for the client IP.

## Changes

- `AccessLogFilter` now also runs as a (post-matching) request filter and keeps a bounded preview of the request body for POST/PUT requests on slow-log paths (`needRecordLog`). It reads at most `log.slow_query_body_limit` bytes (+1 to detect truncation) and replays them in front of the untouched remainder of the entity stream with a `SequenceInputStream`, so the resource still receives the whole body and per-request memory is bounded by the limit, not by the body size.
- Nothing is read when the slow query log is off (`log.slow_query_threshold=0`), when the limit is 0, for GET/DELETE, or for paths that are never slow-logged. Loader batch imports (`.../vertices/batch`, `.../edges/batch`) are not slow-log paths, so their entity stream is never touched, gzip or not.
- When the matched resource method carries `@Decompress` (`DecompressInterceptor` runs later as a ReaderInterceptor), the entity is not read and the log shows `<encoded>`. The decision is based on the resource, not on the `Content-Encoding` header, so a client cannot opt out of body recording by sending a header, and no header value is echoed into the log.
- The preview is decoded with a `CharsetDecoder`, so a multi-byte character cut by the limit is dropped rather than logged as U+FFFD. A truncated preview ends with `...`.
- Path, query and body are CR/LF-escaped when the line is written, so one request is one log line. Before this change `%0A` in a query string produced a raw line break in `slow_query.log`.
- Client IP comes from the Grizzly `Request` peer address via `Provider<Request>`, the same pattern `AuthenticationFilter` uses, with `<unknown_ip>` as fallback. No DNS lookup on the log path.
- New option `log.slow_query_body_limit` (bytes, default 512, 0 disables body recording, max 1 MiB). This answers the open question in #2466 about making the 512 limit configurable and gives operators a switch when gremlin/cypher bodies must not reach the log. The option description and `rest-server.properties` say that the prefix is written as-is and may contain sensitive literals. Body recording stays on by default, matching #2327, #2347 and #2466; note that the same file already receives full GET query strings (`/gremlin?gremlin=...`) today.
- Log line: `[Slow Query] ip=..., execTime=...ms, method=..., path=..., query=..., body=...`
- The unused `PathFilter.REQUEST_PARAMS_JSON` constant is removed; `PathFilter` is otherwise untouched.

## Review threads from #2466

- javeme (naming, cut large bodies, PUT/DELETE, local variables, `<unknown_ip>`): bodies are cut at the configured limit, PUT is handled, the log call uses local variables, `<unknown_ip>` is the placeholder. DELETE is left out on purpose: every `@DELETE` endpoint takes path or query parameters only, so there is no body to record, and Copilot asked to drop it as well.
- javeme (GET and `REQUEST_PARAMS_JSON`): the original stored `getPathParameters()` from a `@PreMatching` filter, before matching happens, so it was always empty. GET requests log `query=` and the path, which carry the parameters already.
- Copilot and VGalaxies (High, memory): the body is no longer read in full before truncation. Reads are bounded and gated.
- Copilot and VGalaxies (Medium, client IP): `getRemoteAddr()` replaces `InetAddress.getByName(uri.getHost())`, which resolved the server host name and did a blocking DNS lookup on every slow query.
- VGalaxies (forwarded headers): not added. Trusting `X-Forwarded-For` without a trusted-proxy list lets any client choose the logged IP. That needs its own option and is left for a follow-up.

## Verification

- `AccessLogFilterTest` (new, 18 cases, added to `UnitTestSuite`): slow-log path selection, bounded capture and full replay, truncation mark, exact-limit body, multi-byte cut, CR/LF kept in the preview, empty body, `Content-Encoding` header ignored, skips (log off, limit 0, GET/DELETE, batch import on a `@Decompress` resource, `@Decompress` resource on a slow-log path), and the log line (client IP, GET query, one line with CR/LF in path/query/body, `<unknown_ip>`, fast requests, log off). The test swaps the module's log4j2 logger for the class under test and restores it, so it passes when run alone as well as inside the suite.
- `mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test` on JDK 11 (Linux x86_64): `UnitTestSuite` 679 tests, 0 failures, 0 errors, 1 skipped (pre-existing `RocksDBSessionTest.testMergeWithStringList`). `AccessLogFilterTest` also passes when run alone and when a single method is run alone.
- Manual run of the packaged server (rocksdb, `log.slow_query_threshold=1`, `log.slow_query_body_limit=64`):
  - gzip batch POST and PUT to `/graphs/hugegraph/graph/vertices/batch` return 201/200 (loader path, stream untouched, not logged)
  - a 5000 byte gremlin body (Content-Length and chunked) returns the value computed from its last bytes, the log shows the first 64 bytes plus `...`
  - a CJK body cut mid-character logs cleanly, GET logs `query=limit=2, body=null`, cypher logs the statement, a pretty-printed body with CR/LF is one log line, a query string with `%0A` is one log line
  - a request from another host logs its peer address (`ip=100.123.70.122`), local ones log `ip=127.0.0.1`
  - a `Content-Encoding: gzip` header on a plain `/gremlin` body neither hides the body nor changes the result
  - `log.slow_query_body_limit=0` logs `body=null`, `log.slow_query_threshold=0` logs nothing, no exceptions from the filter

## Notes

- `path=` in the line still comes from the existing `normalizePath()` used for metric names, which prints for example `graphspaces/graphspace/graphs/hugegraph/graph/vertices`. Unchanged here.
- `log.slow_query_body_limit` is listed in `rest-server.properties`; hugegraph-doc needs a line for it.

### DIFF
--- a/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
+++ b/hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
@@ -23,30 +23,43 @@ import static org.apache.hugegraph.metrics.MetricsUtil.METRICS_PATH_RESPONSE_TIM
 import static org.apache.hugegraph.metrics.MetricsUtil.METRICS_PATH_SUCCESS_COUNTER;
 import static org.apache.hugegraph.metrics.MetricsUtil.METRICS_PATH_TOTAL_COUNTER;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.SequenceInputStream;
+import java.lang.reflect.Method;
 import java.net.URI;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CodingErrorAction;
 import java.util.Map;
 
+import org.apache.hugegraph.api.API;
+import org.apache.hugegraph.api.filter.DecompressInterceptor.Decompress;
 import org.apache.hugegraph.config.HugeConfig;
 import org.apache.hugegraph.config.ServerOptions;
 import org.apache.hugegraph.core.GraphManager;
 import org.apache.hugegraph.metrics.MetricsUtil;
 import org.apache.hugegraph.util.Log;
+import org.glassfish.grizzly.http.server.Request;
 import org.slf4j.Logger;
 
 import jakarta.inject.Singleton;
 import jakarta.ws.rs.HttpMethod;
 import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
 import jakarta.ws.rs.container.ContainerResponseContext;
 import jakarta.ws.rs.container.ContainerResponseFilter;
+import jakarta.ws.rs.container.ResourceInfo;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.ext.Provider;
 
-// TODO: should add test for this class
 @Provider
 @Singleton
-public class AccessLogFilter implements ContainerResponseFilter {
+public class AccessLogFilter implements ContainerRequestFilter, ContainerResponseFilter {
 
     private static final Logger LOG = Log.logger(AccessLogFilter.class);
 
@@ -55,14 +68,27 @@ public class AccessLogFilter implements ContainerResponseFilter {
     private static final String GREMLIN = "gremlin";
     private static final String CYPHER = "cypher";
 
+    // Request property holding the bounded request body preview for the slow query log
+    public static final String REQUEST_BODY = "request_body";
+    public static final String UNKNOWN_IP = "<unknown_ip>";
+
+    private static final String TRUNCATED_MARK = "...";
+    private static final String ENCODED_BODY = "<encoded>";
+    private static final Charset CHARSET = Charset.forName(API.CHARSET);
+
     @Context
     private jakarta.inject.Provider<HugeConfig> configProvider;
 
     @Context
     private jakarta.inject.Provider<GraphManager> managerProvider;
 
+    @Context
+    private jakarta.inject.Provider<Request> requestProvider;
+
+    @Context
+    private ResourceInfo resourceInfo;
+
     public static boolean needRecordLog(ContainerRequestContext context) {
-        // TODO: add test for 'path' result ('/gremlin' or 'gremlin')
         String path = context.getUriInfo().getPath();
 
         // GraphsAPI/CypherAPI/Job GremlinAPI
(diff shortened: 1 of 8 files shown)


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
```diff
@@ -161,6 +226,48 @@ public void filter(ContainerRequestContext requestContext,
...
+
+    private boolean decodesEntity() {
+        Method method = this.resourceInfo == null ? null : this.resourceInfo.getResourceMethod();
+        return method != null && method.isAnnotationPresent(Decompress.class);
+    }
+
+    private static String preview(byte[] bytes, int length, int limit) {
+        boolean truncated = length > limit;
+        int size = Math.min(length, limit);
+        CharsetDecoder decoder = CHARSET.newDecoder()
```
imbajin: ⚠️ Important: The preview decoder is hard-coded to API.CHARSET (UTF-8), while the String message provider used by GremlinAPI and CypherAPI honors the request media type charset. A valid request such as application/json; charset=UTF-16 can therefore be decoded correctly by the resource but logged as replacement or garbled text here. Derive the preview charset from requestContext.getMediaType() with UTF-8 fallback, and add non-UTF-8 coverage, so the slow-query body matches the executed query.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AccessLogFilter.java
```diff
@@ -100,6 +126,43 @@ private static String normalizePath(ContainerRequestContext requestContext) {
...
+    /**
+     * Keep a bounded preview of the request body for the slow query log.
+     * Only the first {@link ServerOptions#SLOW_QUERY_LOG_BODY_LIMIT} bytes are
+     * read, and they are replayed in front of the untouched remainder of the
+     * entity stream, so the resource method still receives the whole body.
+     *
+     * @param requestContext requestContext
+     */
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
```
imbajin, line 137: ⚠️ Important: This global request filter has no explicit priority, as does the dynamically registered RedirectFilter. On @RedirectMasterRole jobs/gremlin requests, RedirectFilter may run first, call abortWith(), and stop the remaining request-filter chain before this code records REQUEST_BODY; the response filter then logs body=null for a slow redirect. Assign an explicit priority that guarantees capture before redirect, or preserve the preview through the redirect, and cover this path with an integration test.

imbajin (commented):
Blocking: no. Summary: Two important correctness gaps remain in the request-body preview path: declared non-UTF-8 request charsets are decoded incorrectly, and redirected job requests can skip capture because the request-filter ordering is unspecified. Evidence: exact head 0260a43c; AccessLogFilterTest passed 18/18 locally; all reported latest-head GitHub checks passed.

imbajin (approved):

PR #3178, opened 2026-08-30
fix(server): reject vertex-property metadata during addV

### LINKED ISSUE #3165: [Bug] HugePrimaryKeyStrategy silently converts vertex-property metadata into vertex properties
### Bug Type (问题类型)

gremlin (unexpected result)

### Before submit

- [x] I searched the existing issues and did not find the same problem.

### Environment (环境信息)

- HugeGraph Server: current `apache/hugegraph` `master` at [`431f6e6b`](https://github.com/apache/hugegraph/commit/431f6e6b431014feb5b50d27ea67180b4610f10b)
- TinkerPop: 3.5.1
- Backend: backend-independent; the incorrect rewrite happens in a traversal strategy before the vertex is written

### Expected and actual behavior (期望与实际表现)

`HugePrimaryKeyStrategy` folds a single-cardinality `AddPropertyStep` immediately following `addV()` into the vertex-creation step.

For this traversal:

```groovy
g.addV("person").
  property("name", "marko", "country", "cn")
```

TinkerPop treats `country=cn` as a meta-property attached to the `name` vertex property. It does not treat `country` as a sibling property of the vertex.

HugeGraph currently reports `supportsMetaProperties() == false`, and `HugeVertex.property(...)` explicitly throws `VertexProperty.Exceptions.metaPropertiesNotSupported()` when meta-properties are passed. The expected behavior is therefore an explicit rejection.

The optimization changes that behavior. `HugePrimaryKeyStrategy` copies every `AddPropertyStep` parameter other than `T.key` and `T.value` into the `addV()` configuration, then removes the original `AddPropertyStep`. If `country` is also a valid schema property, HugeGraph can silently create an ordinary vertex property `country=cn` instead of rejecting the unsupported meta-property.

This is a data-semantics problem because an unsupported construct is accepted with a different meaning.

### Minimal reproduction

Create a schema in which both keys are valid ordinary vertex properties:

```groovy
schema = graph.schema()
schema.propertyKey("name").asText().create()
schema.propertyKey("country").asText().create()
schema.vertexLabel("person").
       properties("name", "country").
       primaryKeys("name").
       create()
```

Run the traversal and inspect the resulting vertex:

```groovy
g.addV("person").
  property("name", "marko", "country", "cn").
  iterate()

g.V().hasLabel("person").has("name", "marko").valueMap(true)
```

Expected:

- The write is rejected with `metaPropertiesNotSupported()`.
- `country=cn` is never stored as an ordinary vertex property.

Actual on the affected strategy path:

- The `AddPropertyStep` is removed.
- `country=cn` is folded into the vertex-creation step as an ordinary vertex property.

A strategy-level reproduction shows the rewritten traversal as:

```text
AddVertexStartStep({country=[cn], label=[person], name=[marko]})
```

### Source analysis

- [TinkerPop 3.5.1 `AddPropertyStep`](https://github.com/apache/tinkerpop/blob/3.5.1/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AddPropertyStep.java#L93-L99) separates `T.key` and `T.value` from the remaining vertex-property key/value pairs, then passes the remaining pairs to `Vertex.property(...)` as meta-properties.
- [HugeGraph declares that meta-properties are unsupported](https://github.com/apache/hugegraph/blob/431f6e6b431014feb5b50d27ea67180b4610f10b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/structure/HugeFeatures.java#L261-L265).
- [`HugeVertex.property(...)` rejects them](https://github.com/apache/hugegraph/blob/431f6e6b431014feb5b50d27ea67180b4610f10b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/structure/HugeVertex.java#L436-L444).
- [`HugePrimaryKeyStrategy` changes the remaining pairs into `addV()` parameters](https://github.com/apache/hugegraph/blob/431f6e6b431014feb5b50d27ea67180b4610f10b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java#L75-L98).

The strategy was introduced in [`4b74ab6c`](https://github.com/apache/hugegraph/commit/4b74ab6cec4a8357dd6a8feef4e3c9ea616fa52f) in March 2023. The same behavior is still present on the current public `master` branch. It was noticed while reviewing a TinkerPop 3.8.1 adaptation, but it is not introduced by that upgrade.

### Suggested fix and acceptance criteria

When an `AddPropertyStep` contains vertex-property metadata, `HugePrimaryKeyStrategy` should not fold that step into `addV()`. Leaving the step in place preserves the existing explicit rejection from `HugeVertex.property(...)`.

A regression test should verify that:

1. The traversal above throws the unsupported-meta-property exception.
2. The metadata key is never written as an ordinary vertex property.
3. Normal `addV().property(key, value)` folding still works.
4. Both `AddVertexStartStep` and mid-traversal `AddVertexStep` paths preserve the same behavior.

1fanwang:
## Purpose of the PR

- Closes https://github.com/apache/hugegraph/issues/3165

`addV().property(key, value, metaKey, metaValue)` accepts unsupported
meta-properties instead of rejecting the traversal.

## Main Changes

Fold only `T.key` and `T.value` into `addV`, then leave metadata-bearing
`AddPropertyStep` instances in the traversal. If the branch executes,
`HugeVertex.property()` rejects the metadata. Unselected `coalesce()` and
`choose()` branches stay lazy, primary-key labels receive their key before
`addV` executes, and `T.id` keeps its specific unsupported-ID error.

## Verifying these changes

| Scenario | Result |
|---|---|
| Upstream behavior | Executed metadata traversals complete without the expected error |
| Final behavior | Executed branches reject at runtime; unreachable branches are unchanged |
| Memory-backed core suite | Passed on JDK 11 |

<details>
<summary>Raw logs</summary>

```text
$ RED=/tmp/hg-base
$ TEST=hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/PrimaryKeyStrategyCoreTest.java
$ git worktree add --detach "$RED" 90035b6e6339572cca2fe1bafd5242b1c300e0ea
$ git archive cf4f1598608b074579a778ca0a5e810ff29d6ebc "$TEST" | tar -x -C "$RED"
$ cd "$RED"
$ JAVA_HOME=$(/usr/libexec/java_home -v 11) mvn test \
    -pl hugegraph-server/hugegraph-test -am -P core-test,memory \
    -Dtest=PrimaryKeyStrategyCoreTest -DfailIfNoTests=false -Drevision=1.7.0
Tests run: 8, Failures: 5, Errors: 0, Skipped: 0
Start, mid-traversal, and primary-key cases:
  No exception was thrown(expected java.lang.UnsupportedOperationException)
Unfolded-step case: assertion failed
T.id: Bad exception type java.lang.IllegalArgumentException
BUILD FAILURE

$ GREEN=/tmp/hg-green
$ git worktree add --detach "$GREEN" cf4f1598608b074579a778ca0a5e810ff29d6ebc
$ cd "$GREEN"
$ JAVA_HOME=$(/usr/libexec/java_home -v 11) mvn test \
    -pl hugegraph-server/hugegraph-test -am -P core-test,memory \
    -Dtest=PrimaryKeyStrategyCoreTest -DfailIfNoTests=false -Drevision=1.7.0
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

</details>

- [x] Nope

## Documentation Status

- [x] `Doc - No Need`

### DIFF
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java
@@ -30,6 +30,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.step.map.AddVertexStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect.AddPropertyStep;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.AbstractTraversalStrategy;
 import org.apache.tinkerpop.gremlin.structure.T;
+import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty.Cardinality;
 
 public class HugePrimaryKeyStrategy
@@ -76,7 +77,6 @@ public class HugePrimaryKeyStrategy
                 || propertyStep.getCardinality() == null) {
 
                 Object[] kvs = new Object[2];
-                List<Object> kvList = new LinkedList<>();
 
                 propertyStep.getParameters().getRaw().forEach((k, v) -> {
                     if (T.key.equals(k)) {
@@ -84,16 +84,11 @@ public class HugePrimaryKeyStrategy
                     } else if (T.value.equals(k)) {
                         kvs[1] = v.get(0);
                     } else {
-                        kvList.add(k.toString());
-                        kvList.add(v.get(0));
+                        throw VertexProperty.Exceptions.metaPropertiesNotSupported();
                     }
                 });
 
                 curAddStep.configure(kvs);
-
-                if (!kvList.isEmpty()) {
-                    curAddStep.configure(kvList.toArray(new Object[0]));
-                }
 
                 removeSteps.add(step);
             } else {
--- a/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java
+++ b/hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java
@@ -19,6 +19,7 @@ package org.apache.hugegraph.traversal.optimize;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.tinkerpop.gremlin.process.traversal.Step;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
@@ -30,7 +31,6 @@ import org.apache.tinkerpop.gremlin.process.traversal.step.map.AddVertexStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect.AddPropertyStep;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.AbstractTraversalStrategy;
 import org.apache.tinkerpop.gremlin.structure.T;
-import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty.Cardinality;
 
 public class HugePrimaryKeyStrategy
@@ -77,20 +77,33 @@ public class HugePrimaryKeyStrategy
                 || propertyStep.getCardinality() == null) {
 
                 Object[] kvs = new Object[2];
+                boolean extraParams = false;
 
-                propertyStep.getParameters().getRaw().forEach((k, v) -> {
-                    if (T.key.equals(k)) {
-                        kvs[0] = v.get(0);
-                    } else if (T.value.equals(k)) {
-                        kvs[1] = v.get(0);
+                for (Map.Entry<Object, List<Object>> param :
+                     propertyStep.getParameters().getRaw().entrySet()) {
+                    Object paramKey = param.getKey();
+                    if (T.key.equals(paramKey)) {
+                        kvs[0] = param.getValue().get(0);
+                    } else if (T.value.equals(paramKey)) {
+                        kvs[1] = param.getValue().get(0);
                     } else {
-                        throw VertexProperty.Exceptions.metaPropertiesNotSupported();
+                        extraParams = true;
                     }
-                });
+                }
 
                 curAddStep.configure(kvs);
 
-                removeSteps.add(step);
+                if (extraParams) {
+                    /*
+                     * Rejecting here would run for every child traversal, so an
+                     * unreachable coalesce()/choose() branch would fail too. Leave
+                     * the step for HugeVertex.property() to reject when it really
+                     * runs, and stop folding across it.
+                     */
+                    curAddStep = null;
+                } else {
+                    removeSteps.add(step);
+                }
             } else {
                 curAddStep = null;
             }
--- a/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/PrimaryKeyStrategyCoreTest.java
+++ b/hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/PrimaryKeyStrategyCoreTest.java
@@ -20,7 +20,9 @@ package org.apache.hugegraph.core;
 import org.apache.hugegraph.schema.SchemaManager;
 import org.apache.hugegraph.testutil.Assert;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
 import org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect.AddPropertyStep;
+import org.apache.tinkerpop.gremlin.structure.T;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.junit.Test;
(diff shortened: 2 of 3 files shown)


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java
```diff
@@ -76,25 +77,19 @@ public void apply(Traversal.Admin<?, ?> traversal) {
...
                 propertyStep.getParameters().getRaw().forEach((k, v) -> {
                     if (T.key.equals(k)) {
                         kvs[0] = v.get(0);
                     } else if (T.value.equals(k)) {
                         kvs[1] = v.get(0);
                     } else {
-                        kvList.add(k.toString());
-                        kvList.add(v.get(0));
+                        throw VertexProperty.Exceptions.metaPropertiesNotSupported();
                     }
```
imbajin, line 91: ⚠️ Throwing from the strategy makes this validation eager for every child traversal. TinkerPop applies strategies to `coalesce()`/`choose()` children before branch selection, so an unreachable fallback such as `coalesce(unfold(), addV(...).property(..., metaKey, metaValue))` now fails even when `unfold()` succeeds. Please leave this `AddPropertyStep` unfolded when extra parameters are present and let `HugeVertex.property()` reject it only if that branch executes; this also keeps validation behavior centralized.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/PrimaryKeyStrategyCoreTest.java
```diff
@@ -0,0 +1,100 @@
...
+    public void testFoldsSingleCardinalityProperties() {
+        this.initSchema();
+
+        GraphTraversal<Vertex, Vertex> traversal = graph().traversal()
+                                                       .addV(LABEL)
+                                                       .property(
+                                                         VertexProperty.Cardinality.single,
+                                                         "name", "marko")
+                                                       .property(
+                                                         VertexProperty.Cardinality.single,
```
imbajin: 🧹 Several expressions in this test wrap well before the project's 120-column limit. For example, each metadata `property(...)` call is about 99 columns as one line, each cardinality call is about 116, the helper signature is about 84, and the exception text is about 87. Please use the available width so these short expressions remain compact and easier to scan.

imbajin (commented):
The core fix is directionally correct, but the eager validation changes lazy branch behavior. I also left one readability note about wrapping well before the project's 120-column limit.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java
```diff
@@ -76,26 +77,33 @@ public void apply(Traversal.Admin<?, ?> traversal) {
...
+                        kvs[1] = param.getValue().get(0);
                     } else {
-                        kvList.add(k.toString());
-                        kvList.add(v.get(0));
+                        extraParams = true;
                     }
-                });
+                }
 
                 curAddStep.configure(kvs);
```
imbajin, line 94: ⚠️ This still gives the failing `AddPropertyStep` a side effect before it throws. On the exact head, after `addV(...).property("country", "cn", "since", "2024").next()` throws, catching the exception and committing the transaction persists `country=cn` because it was copied into `addV()` here. A non-folded `AddPropertyStep` rejects before setting that property. Please ensure a rejected step cannot commit its own key/value, and add a catch-and-commit regression; the explicit rollback in the new test currently masks this difference.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/traversal/optimize/HugePrimaryKeyStrategy.java
```diff
@@ -76,26 +77,33 @@ public void apply(Traversal.Admin<?, ?> traversal) {
...
-                        kvs[0] = v.get(0);
-                    } else if (T.value.equals(k)) {
-                        kvs[1] = v.get(0);
+                boolean extraParams = false;
+
+                for (Map.Entry<Object, List<Object>> param :
+                     propertyStep.getParameters().getRaw().entrySet()) {
+                    Object paramKey = param.getKey();
+                    if (T.key.equals(paramKey)) {
+                        kvs[0] = param.getValue().get(0);
```
imbajin, line 86: ⚠️ Traversal-valued parameters are now owned by two steps: the same child traversal remains on the retained `AddPropertyStep` and is also configured into `AddVertexStep`. With `val = __.addV("audit").constant("marko")`, `addV("p").property("name", val, "since", "2024")` creates two `audit` vertices before raising the metadata error on the exact head. Please leave traversal-valued pairs unfolded, or otherwise ensure each child traversal is attached and evaluated exactly once.

imbajin (commented):
The previous review concerns are addressed, but retaining the metadata step while copying its key/value into `addV()` still changes runtime side effects in two concrete cases.

imbajin (approved):
+1. This PR now fixes the reported scalar meta-property rewrite with a focused change while preserving lazy branch behavior and the normal folding path. A traversal-valued key/value combined with unsupported metadata may still be evaluated twice before the required error, but that is a nested invalid-input corner case and is explicitly non-blocking for this minimal fix.

PR #3182, opened 2026-09-01
fix(server): filter index results before input ordering

### LINKED ISSUE #3180: [Bug] HStore combined range/search index test intermittently drops results
> [!IMPORTANT]
> **Highest-priority related change to investigate: #3140**
>
> PR #3140 changed the HStore cross-partition range-index scan, merge, and ordering paths most closely related to this failure. Use merge commit `2ae73704cdc5988127613c919ee7bb23d2568be7` as the first bisect boundary. This is a strong correlation for investigation, not yet proof of root cause.

### Problem

The HStore core-test lane intermittently fails:

```text
VertexCoreTest.testQueryByJointIndexesWithSearchAndTwoRangeIndexesAndWithin
expected:<3> but was:<1>
```

The query combines:

- a range index on `type`;
- a range index with `P.within(...)` on `confirmType`;
- a search index with `Text.contains("诚信")` on `name`.

Four vertices are inserted and three should match, but HStore sometimes returns only one. The same test passes on other runs, so the behavior is intermittent.

### Repeated master evidence

The same assertion failed on multiple independent `master` pushes:

- [run 33256851831](https://github.com/apache/hugegraph/actions/runs/33256851831), head `321ba4dadad7bd37b8c659cdec34b056912d34b8`
- [run 33302137810](https://github.com/apache/hugegraph/actions/runs/33302137810), head `90035b6e6339572cca2fe1bafd5242b1c300e0ea`
- [run 33323387889](https://github.com/apache/hugegraph/actions/runs/33323387889), head `7bb624b99dc12c948cd62736ae26583a05195da1`
- [run 33324047387](https://github.com/apache/hugegraph/actions/runs/33324047387), head `8e167c7b1ac84194be3336426dea5a2448269016`

The merge run for #3140 and later master runs also passed, which reinforces that this is intermittent and explains why the related change's own CI did not catch it.

### Most relevant recent change

[#3140: preserve range-index ordering across partitions](https://github.com/apache/hugegraph/pull/3140) was merged on 2026-08-26 as `2ae73704cdc5988127613c919ee7bb23d2568be7`.

It changed the paths most likely to influence this result:

- HStore multi-partition scan and merge;
- ordered iterator behavior;
- range-index result ordering;
- query result combination and paging;
- Store client and server scan streams.

Its HStore CI passed, so the issue may require repeated execution, a specific partition layout, or a timing-sensitive stream/merge condition.

### Test history

The failing regression test was introduced on 2022-04-19 by:

- [#1803: fix query by within + Text.contains](https://github.com/apache/hugegraph/pull/1803)
- commit `1f5d502e0e98bab6c8037cff9eeacc5dc2063715`

The test predates the current HStore ordering implementation and has since moved with repository/module restructuring.

### Reproduction

The CI lane runs:

```bash
mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,hstore
```

The focused target is:

```text
VertexCoreTest#testQueryByJointIndexesWithSearchAndTwoRangeIndexesAndWithin
```

Because the failure is intermittent, run the focused target repeatedly and retain the first failing backend logs and test artifacts.

### Suggested investigation

1. Compare repeated focused runs immediately before and after `2ae73704c`.
2. Log the candidate vertex IDs returned by each range/search index before intersection.
3. Log each partition's scan results, stream completion, and final ordered merge.
4. Confirm whether one partition/result stream is dropped, terminated early, or consumed out of order.
5. Compare the failing HStore query plan and result IDs with a passing backend.

### Expected behavior

The first query consistently returns three vertices on every backend and every run.

contrueCT:
## Purpose of the PR

- close #3180
- Fix HStore queries that combine flattened `within()` conditions with search or other
  post-filtered indexes and can drop valid vertices or edges.

### Root cause

HStore does not sort these backend results by input IDs. HugeGraph therefore restores input
order with `InputOrderIterator`, which may prefetch the next flattened subquery. That prefetch
updates the shared origin query's `resultsFilter`. Previously, record matching happened after
input-order restoration, so rows from subquery A could be checked with subquery B's filter.

HStore query filter ordering before and after the fix

### CI follow-up

Count-query fallbacks wrap the original `ConditionQuery` in an `IdQuery`. A transaction-cache
hit could therefore return index candidates before the original condition was applied. The
cache eligibility check now follows the origin-query chain, and both vertex and edge caches are
bypassed whenever any query in that chain still needs post-filtering.

## Main Changes

- Apply index-result matching to each backend subquery before restoring input order.
- Keep undefined-label diagnostics and hidden/deleting-label filtering at the public graph-query
  boundary.
- Skip vertex and edge transaction caches for queries whose origin chain still requires
  post-filtering, including search indexes, `INDEX_FILTER`, `SORT_KEYS`, and `INDEX` queries
  without a label.
- Make the existing vertex regression deterministic by committing before querying, and add an
  equivalent edge regression plus cache-classification coverage.

## Verifying these changes

    - `mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test`
      - 682 tests, 0 failures, 0 errors, 1 skipped
    - Focused cache-classification unit test after the CI follow-up
      - 1 test, 0 failures, 0 errors
    - Focused `core-test,hstore` regression set
      - 4 tests, 0 failures, 0 errors on a freshly initialized PD/Store
    - Focused `core-test,memory` regression set
      - 4 tests, 0 failures, 0 errors
    - Focused `core-test,rocksdb` regression set
      - 4 tests, 0 failures, 0 errors
    - `mvn clean compile -Dmaven.javadoc.skip=true`
      - all 38 reactor modules compiled successfully
    - `mvn editorconfig:format`
      - no files required formatting

- [ ] Dependencies ([add/update license](https://hugegraph.apache.org/docs/contribution-guidelines/contribute/#321-check-licenses) info & [regenerate_known_dependencies.sh](../install-dist/scripts/dependency/regenerate_known_dependencies.sh))
- [ ] Modify configurations
- [ ] The public API
- [x] Other affects: internal index-result filtering order and transaction-cache eligibility
- [ ] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`


contrueCT:
Addressed the follow-up review in `1019e617`:

- Treat `PRIMARY_KEY` and vertex labeled `INDEX` queries as requiring post-filter cache bypass, while preserving the edge labeled-index fast path.
- Add a warm-cache primary-key regression test and focused query-classification coverage.
- Strengthen the mixed search/range regressions with concrete vertex and edge assertions.

Validation:
- CachedGraphTransaction focused unit test: 1/1 passed
- RocksDB focused core tests: 4/4 passed
- HStore focused core tests: 4/4 passed
- Full clean compile: 38/38 modules passed

bitflicker64 (commented):
Blocking: no. Summary: The reordering is sound: moving index-candidate matching upstream of `keepInputOrderIfNeeded` means each record is tested while the shared `resultsFilter` still belongs to its own flattened sub-query, and the new cache guard compensates for `filterUnmatchedRecords` no longer running in `queryVertices`/`queryEdges`. Below are one hardening item, one question about a behavior change that looks incidental, and three maintenance items. Evidence: reviewed at head via `git diff $(git merge-base FETCH_HEAD origin/master) FETCH_HEAD`; `queryNeedsPostFilter` checked against every branch of `rightResultFromIndexQuery` (GraphTransaction.java:1915-1979); CI green at head apart from the `codecov/project` coverage threshold.

One finding has no line to anchor to, so it goes here. `.github/images/issue-3180-query-filter-order.png` adds 1,033,655 bytes to the repository. `origin/master` has no image files anywhere in the tree and no `.github/images/` directory, so this introduces both a new directory convention and the repository's first binary asset. The PR body already loads the diagram from `raw.githubusercontent.com` on a fork SHA, so the description renders whether or not the file lands in `apache/hugegraph`. Please upload it through GitHub's own attachment upload on the PR description and drop the file from the diff.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -1870,10 +1872,10 @@ private void removeLeftIndexIfNeeded(Map<Id, HugeVertex> vertices) {
         }
     }
 
-    private <T extends HugeElement> Iterator<T> filterUnmatchedRecords(
+    private <T extends HugeElement> Iterator<T> filterInvalidRecords(
```
bitflicker64, line 1885: ⚠️ `HugeFactoryAuthProxy.java:273` registers the method being renamed here, `"filterUnmatchedRecords"`, with `Reflection.registerMethodsToFilter`. The new sibling `filterInvalidRecords` is not registered, so it remains enumerable via `getDeclaredMethods()` while the method it was split out of is hidden. This affects enumeration only: `filterInvalidRecords` is still private and access control is unchanged.

Please add `"filterInvalidRecords"` to the `Reflection.registerMethodsToFilter(GraphTransaction.class, ...)` list next to `"filterUnmatchedRecords"`.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -441,6 +449,29 @@ protected Iterator<HugeEdge> queryEdgesFromBackend(Query query) {
         return new ExtendableIterator<>(edges.iterator(), rs);
     }
 
+    private static boolean queryNeedsPostFilter(Query query) {
```
bitflicker64: 🧹 This predicate has to track `GraphTransaction.rightResultFromIndexQuery` (GraphTransaction.java:1915-1979), which decides whether a record is actually dropped, but nothing links the two.

The pairing is safe today. `rightResultFromIndexQuery` returns true early for a non-ConditionQuery whose immediate origin is also not a ConditionQuery (1920-1925), for edge + LABEL + `conditions().size() == 1`, and for edge + LABEL + `optimized() == INDEX`; otherwise it falls through to `optimized() == NONE || cq.test(elem)` at 1948. `queryNeedsPostFilter` allows caching only for the INDEX and NONE shapes, so it is strictly more conservative. The risk is drift: the rule lives in two classes in two shapes, and editing one would silently let the caches serve unmatched records.

Please derive both from a single helper. Note the refactor also has to move `CachedGraphTransactionTest.testQueryNeedsPostFilter`, which calls `Whitebox.invokeStatic(CachedGraphTransaction.class, ..., "queryNeedsPostFilter", ...)`.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -861,6 +861,7 @@ protected Iterator<HugeVertex> queryVerticesFromBackend(Query query) {
         Iterator<HugeVertex> vertices = new MapperIterator<>(entries,
                                                              this::parseEntry);
         vertices = this.filterExpiredResultFromBackend(query, vertices);
+        vertices = this.filterUnmatchedRecords(vertices, query);
```
bitflicker64, line 864: 🧹 On the non-paging edge branch the filter runs in `queryEdgesFromBackendInternal(cq)` with `cq` already flattened, and re-flattening yields exactly one sub-query (`ConditionQueryFlatten.flattenRelations` returns `ImmutableList.of(cq)`), so `cq` owns its `resultsFilter`. Paging and non-ConditionQuery edge queries skip that flatten and go straight to `queryEdgesFromBackendInternal` at line 1080.

The vertex path has no outer flatten at all: `this.query(query)` flattens internally and every sub-query pushes its filter onto this same shared `query` through `ConditionQuery.updateResultsFilter` (ConditionQuery.java:708-727), called from `QueryList.java:223`.

It is correct today only because `FilterIterator.fetch()` tests each element immediately and `WrappedIterator.hasNext()` returns the buffered element, so the advance into the next sub-query always follows the test. Please state that invariant here, or bind the filter to the sub-query the records came from.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -420,6 +424,10 @@ protected Iterator<HugeEdge> queryEdgesFromBackend(Query query) {
         }
 
         Iterator<HugeEdge> rs = super.queryEdgesFromBackend(query);
+        if (queryNeedsPostFilter(query)) {
```
bitflicker64, line 424: 🧹 This reads as dead code: the same call ran at line 402 on the same object, and the comment does not say why the answer can change.

It is not dead. `ConditionQuery.optimized()` propagates up the `originQuery` chain (ConditionQuery.java:681-696) and `ConditionQuery.copy()` sets `originQuery(this)` (line 577), so the flattened children point back at this `query`. `super.queryEdgesFromBackend(query)` reaches `GraphIndexTransaction.queryIndex` (line 383), which sets `INDEX` at line 402, and `INDEX_FILTER` can be set at line 609. So `query.optimized()` can move off `NONE` between the two checks.

Please expand the comment to name that, for example: `super.queryEdgesFromBackend() may promote query.optimized() via origin-chain propagation, so re-check before caching`.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/core/VertexCoreTest.java
```diff
@@ -5424,7 +5443,7 @@ public void testQueryByJointIndexesWithSearchAndTwoRangeIndexesAndWithin() {
         graph.addVertex(T.label, "test", "name", "诚信文明",
                         "confirmType", 3, "type", 1, "kid", 3);
 
-        this.mayCommitTx();
+        this.commitTx();
```
bitflicker64, line 5447: 🧹 Making the regression deterministic is the right call, since the bug only reproduces once the rows are in the backend store. But `mayCommitTx()` is a coin flip (`BaseCoreTest.java:141-146`), so this test used to exercise the uncommitted-transaction path roughly half the time, and that coverage is now gone rather than merely made reliable.

The uncommitted path is not incidental to this change: `ConditionQuery.test` deliberately skips `resultsFilter` for fresh elements (`ConditionQuery.java:625`), so it is a different branch of the code this PR touches.

Please keep the deterministic committed assertions and restore the other half, either by repeating the three query blocks after a `mayCommitTx()` or by adding a sibling test that queries before committing.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -845,7 +845,7 @@ public Iterator<Vertex> queryVertices(Query query) {
         query.resetActualOffset();
 
         Iterator<HugeVertex> results = this.queryVerticesFromBackend(query);
-        results = this.filterUnmatchedRecords(results, query);
+        results = this.filterInvalidRecords(results, query);
```
bitflicker64: ⚠️ The split also reverses the two predicates relative to each other, which looks incidental rather than intended.

Before this PR a single filter ran the undefined-label warning, then hidden, then deleting-label, then `rightResultFromIndexQuery`. Now `filterUnmatchedRecords` runs upstream (864 and 1104) and `filterInvalidRecords` runs downstream (here and 1015). Two consequences:

- Hidden and deleting-label elements now reach `rightResultFromIndexQuery`, so `this.indexTx.asyncRemoveIndexLeft(cq, elem)` (1939, 1955, 1972) can be scheduled for elements whose schema label is already being deleted. Previously they were dropped before reaching that point.
- The "Left record is found" warning (1882-1884) is now unreachable on index queries for any record `filterUnmatchedRecords` already dropped, which is the left-index case that warning exists for.

Please confirm the reordering is intended, and if it is, either restore the warning for left records on index queries or note why it is no longer needed.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -803,7 +804,8 @@ protected Iterator<Vertex> queryVerticesByIds(Object[] vertexIds, boolean adjace
         if (!query.empty()) {
             // Query from backend store
             query.mustSortByInput(false);
-            Iterator<HugeVertex> it = this.queryVerticesFromBackend(query);
+            Iterator<HugeVertex> it =
+                    this.queryValidVerticesFromBackend(query);
```
imbajin: ‼️ `queryVerticesByIds()` now uses `queryValidVerticesFromBackend()` at this call site, and that wrapper applies `invalidRecord()` with the default `showHidden=false`. This method is also used by `queryTaskInfos(id)` (`StandardTaskScheduler.findTask()`), while task records use the hidden label `~task` (`HugeTask.P.TASK = Hidden.hide("task")`). Consequently internal task lookup returns null: exact-head `EdgeCoreTest#testQueryByTextContainsProperty` fails on the first index-label `.create()` with `NotFoundException: Can't find task with id '1'`, while the base commit passes; the CI server/hstore failures show the same missing-task pattern. Keep task/server ID lookups on the raw or an explicit `showHidden(true)` path, and add a regression for waiting on a schema task.

imbajin (commented):
Blocking: yes. Summary: The new valid-record wrapper hides internal task vertices from task lookup, so schema/index operations cannot wait for their asynchronous tasks. Evidence: at exact head 88ba98c, the local EdgeCoreTest#testQueryByTextContainsProperty fails at the first index-label create with NotFoundException: Can't find task with id '1', while the base commit passes; the latest server and hstore CI failures show the same missing-task pattern.

bitflicker64 (commented):
Blocking: yes. Summary: The new `queryValid*FromBackend()` wrappers apply `showHidden` and `showDeleting` filtering to `Query` objects the transaction builds for itself, which never carried those flags, so the regression already reported for task lookup also lands on auth relationship lookup by edge id and on adjacent-edge collection during vertex deletion. Evidence: CI at this head fails on memory, rocksdb, hbase, hstore and both macOS rocksdb jobs; `build-server (rocksdb, 11)` reports 814 tests, 29 failures and 619 errors, dominated by 573 occurrences of `Failed to wait for task 'N' completed` with `Caused by: java.lang.AssertionError: N` from `StandardTaskScheduler.java:532`, while master at the merge base 98477f0 is green. The task-lookup call site at line 808 is already covered by an existing review and is not repeated below.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -974,14 +982,16 @@ protected Iterator<Edge> queryEdgesByIds(Object[] edgeIds,
...
+                        this.queryValidEdgesFromBackend(query);
                 @SuppressWarnings({"unchecked", "rawtypes"})
                 Iterator<Edge> r = (Iterator) it;
                 return r;
             }
 
             query.mustSortByInput(false);
-            Iterator<HugeEdge> it = this.queryEdgesFromBackend(query);
+            Iterator<HugeEdge> it =
+                    this.queryValidEdgesFromBackend(query);
```
bitflicker64: ‼️ `queryEdgesByIds(Object[], boolean)` builds its own query at line 950, `new IdQuery(HugeType.EDGE)`, so it carries the `Query` defaults `showHidden=false` and `showDeleting=false` (Query.java:99-100). Routing it through `queryValidEdgesFromBackend()` here and at line 986 now drops every edge whose label is hidden. Before this PR both call sites went straight to `queryEdgesFromBackend()` and did no such filtering.

Auth relationships use hidden edge labels: `HugeBelong.P.BELONG = Hidden.hide("belong")` (HugeBelong.java:232) and `HugeAccess.P.ACCESS = Hidden.hide("access")` (HugeAccess.java:276). `RelationshipManager` reaches the backend through `this.tx().queryEdges(id)` in `delete(Id)` (line 94), `get(Id)` (107), `exists(Id)` (120) and `queryById()` (161, behind `list(List<Id>)`). So `get` now throws `NotFoundException`, `exists` returns false so `update` fails its check, and `delete` skips its `if (edges.hasNext())` block and returns null with no error and no commit.

`queryEdgesByIds` receives no caller query, so the fix is either to put lines 986 and 994 back on `queryEdgesFromBackend()`, or to set `showHidden(true)` and `showDeleting(true)` on the `IdQuery` at line 950. Note that `CachedGraphTransactionTest.testQueryByIdsFiltersDeletingLabels`, added by this PR, asserts the current behaviour, so it needs to change with the fix.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -377,7 +377,8 @@ protected void prepareDeletions(Map<Id, HugeVertex> removedVertices,
             }
             // Query all edges of the vertex and remove them
             Query query = constructEdgesQuery(v.id(), Directions.BOTH, new Id[0]);
-            Iterator<HugeEdge> vedges = this.queryEdgesFromBackend(query);
+            Iterator<HugeEdge> vedges =
+                    this.queryValidEdgesFromBackend(query);
```
bitflicker64: ⚠️ Line 379 binds to `constructEdgesQuery(Id, Directions, Id...)` (line 1306), which delegates to the private `List<Id>` overload at line 1356. That builds `new ConditionQuery(HugeType.EDGE)` at line 1364 and never touches the display flags, so the query arrives with `showHidden=false` and `showDeleting=false`. Through `queryValidEdgesFromBackend()` the deletion path now skips any adjacent edge whose label is hidden, or whose label status is `DELETING` at that moment.

`prepareDeletions()` is the only place that collects a removed vertex's edges into `removedEdges`; the other writer at line 919 is an explicit `removeEdge()`. A skipped edge therefore keeps its row and its index entries with no owner vertex, and nothing later in this path revisits it. `EntityManager.delete()` removing a `~user` vertex that still has `~belong` or `~access` edges is a concrete instance.

Please leave vertex deletion on the unfiltered `queryEdgesFromBackend()`, or set `showHidden(true)` and `showDeleting(true)` on the query built at line 379.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -1859,7 +1876,8 @@ private void removeLeftIndexIfNeeded(Map<Id, HugeVertex> vertices) {
             return;
         }
         IdQuery idQuery = new IdQuery(HugeType.VERTEX, ids);
-        Iterator<HugeVertex> results = this.queryVerticesFromBackend(idQuery);
+        Iterator<HugeVertex> results =
+                this.queryValidVerticesFromBackend(idQuery);
```
bitflicker64: 🧹 Same wrapper problem, smaller blast radius. `removeLeftIndexIfNeeded()` builds `new IdQuery(HugeType.VERTEX, ids)` at line 1878, which is the `IdQuery(HugeType, Set<Id>)` constructor (IdQuery.java:50) chaining to `IdQuery(HugeType)`, so `showHidden` stays false. The stored version of a hidden-label vertex is now filtered out before `indexTx.updateVertexIndex(existedVertex, true)` can run, so its stale index entries survive the overwrite. `~user`, `~group`, `~role`, `~target` and `~project` (HugeUser.java:261, HugeGroup.java:164, HugeRole.java:186, HugeTarget.java:276, HugeProject.java:256) are all overwritten in place by `EntityManager.save()`.

Filed as minor because the call is gated by `removeLeftIndexOnOverwrite` (line 332), backed by `vertex.remove_left_index_at_overwrite`, which defaults to false (CoreOptions.java:356-362). Whatever fix lands for lines 994 and 381 should cover this site too.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -1963,6 +2015,33 @@ private boolean rightResultFromIndexQuery(Query query, HugeElement elem) {
...
+                    conditionQueryNeedsPostFilter(cq)) {
+                    return true;
+                }
+            }
+            query = query.originQuery();
+        }
+        return false;
+    }
+
+    private static boolean conditionQueryNeedsPostFilter(ConditionQuery query) {
```
bitflicker64, line 2035: 🧹 The `edgeIndexWithLabel` carve-out cannot be reached from `rightResultFromIndexQuery()`, the predicate's other caller. That method already returns true for the same shape earlier: the block at lines 1966-1982 fires when `cq.condition(HugeKeys.LABEL) != null && cq.resultType().isEdge()`, and returns true both when `cq.conditions().size() == 1` and when `cq.optimized() == OptimizedType.INDEX`. So whenever `edgeIndexWithLabel` is true, control never reaches line 1985.

The carve-out only changes the cache decision in `queryNeedsPostFilter()`, but it is spelled as part of a predicate named for post-filtering in general, which invites the next reader to widen it and quietly weaken `rightResultFromIndexQuery()`.

Please move the edge-index-with-label exemption into `queryNeedsPostFilter()` where it is actually used, and leave `conditionQueryNeedsPostFilter()` as the plain `optimized() != OptimizedType.NONE` test that line 1985 needs.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -1870,33 +1882,67 @@ private void removeLeftIndexIfNeeded(Map<Id, HugeVertex> vertices) {
...
             }
             // Process results that query from left index or primary-key
             // Only index query will come here
-            return query.resultType().isVertex() != elem.type().isVertex() ||
-                   rightResultFromIndexQuery(query, elem);
+            boolean matched =
+                    query.resultType().isVertex() != elem.type().isVertex() ||
+                    rightResultFromIndexQuery(query, elem);
+            if (!matched) {
+                warnLeftRecord(elem);
```
bitflicker64, line 1921: 🧹 The `LOG.warn` behind this call cannot fire for a query that leaves `showHidden` at its default, which is narrower than the block above reads.

A left record's label name is `SchemaElement.UNDEF`, `"~undefined"` (SchemaElement.java:57), returned by `HugeVertex.label()` (198-200) and `HugeEdge.label()` (117-119), and `Graph.Hidden.isHidden` is a `"~"` prefix test. So `invalidRecord` at 1912 is true for every left record when `query.showHidden()` is false, and the early `return true` runs before control reaches 1920.

The warning is not lost. On `queryVertices(Query)` and `queryEdges(Query)` it comes from `filterInvalidRecords` at 1890, which is where master emitted it. Line 1921 is reached only by the `showHidden(true)` callers, such as `EntityManager.queryEntity` (158) and `traverseByLabel` (2336). On the internal call sites this PR routes through `filterUnmatchedRecords` with no downstream `filterInvalidRecords` (380, 806, 983, 990, 1809, 1874) it adds nothing, since all six build their own query and leave `showHidden` false (Query.java:99).

Requested change: record that split in the comment at 1907-1911, naming 1890 as the call that covers ordinary queries. As written the block reads as though this line restores left-record logging for index queries in general.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -318,7 +318,8 @@ private boolean needCacheVertex(HugeVertex vertex) {
     @Watched(prefix = "graphcache")
     protected Iterator<HugeVertex> queryVerticesFromBackend(Query query) {
         if (this.enableCacheVertex() &&
-            query.idsSize() > 0 && query.conditionsSize() == 0) {
+            query.idsSize() > 0 && query.conditionsSize() == 0 &&
+            !queryNeedsPostFilter(query)) {
```
bitflicker64, line 322: 🧹 This guard decides before the backend call. The edge path re-checks after `super.queryEdgesFromBackend(query)` at line 424, and its new comment gives the reason: that call can promote `query.optimized()` through origin-query propagation. Nothing records why the vertex path needs no equivalent.

It does not need one, and the reason is worth stating, because the clause is live rather than defensive. `queryNumber` falls back to `this.queryVertices(q)` at GraphTransaction.java:597-598 with `q` being whatever `optimizeQueries` handed its fetcher (561), so `q` arrives here as an `IdQuery` whose origin `ConditionQuery` already carries `PRIMARY_KEY`, set at 1547 before the `IdQuery` is built at 1556, or `INDEX`, set at GraphIndexTransaction.java:402 while `indexQuery` runs. The PR's own `testQueryByPrimaryValuesAndPropsWithCachedVertex` covers the first shape. `optimized()` is therefore already fixed when this line runs, and `queryVerticesByIds` cannot promote it further: `GraphTransaction.query(Query)` (535-540) short-circuits to `super.query(query)` for a non-`ConditionQuery`, so `optimizeQueries` never runs beneath it.

Requested change: add a one-line comment here mirroring the one at 424. The asymmetry currently reads as an oversight.

bitflicker64 (commented):
Blocking: no. Summary: The reordering holds up at this head and the regressions reported against 88ba98c are gone: `filterUnmatchedRecords` now runs upstream of `keepInputOrderIfNeeded`, so each record is tested while `resultsFilter` still belongs to its own flattened sub-query, the `queryValid*FromBackend()` wrappers are off every internal call site, and `queryNeedsPostFilter` is never less conservative than `rightResultFromIndexQuery` on any chain shape I could construct. Two documentation-grade items below.

Evidence:

- `git diff 98477f0 acbee16` and `git diff 88ba98c acbee16` read in full.
- `queryNeedsPostFilter` (GraphTransaction.java:2012-2033) traced against every branch of `rightResultFromIndexQuery` (1946-2010). `OptimizedType` ordinals (ConditionQuery.java:805-811) plus the max-keeping propagation in `ConditionQuery.optimized(...)` (681-696) mean a mixed flatten promotes the parent past the `INDEX` carve-out, so the re-check at CachedGraphTransaction.java:424 declines to cache.
- `InputOrderIterator.fetchBatch` (QueryResults.java:378-419) tolerates ids filtered out upstream, so moving the filter above it does not break order restoration.
- The ramtable short-circuit at CachedGraphTransaction.java:391-394 now returns before `filterUnmatchedRecords`, which on master ran downstream in `queryEdges(Query)`. Benign, because that branch runs before `this.query(...)`, so the caller's `optimized()` is still `NONE` and the old filter was a no-op on those rows; hidden and deleting labels are still handled by `filterInvalidRecords`. Worth knowing it changed, though.
- `HugeEdge.sysprop` (321-347) returns `fatherId()` for `LABEL` and the own id for `SUB_LABEL`, so filtering per `tempQuery` on the `parentElQueryWithSortKeys` path tests correctly.
- `gh pr checks 3182` is green on 23 of 24 checks at this head, the exception being the `codecov/project` threshold.

Findings already raised on this PR are not repeated.

imbajin (approved):
Fix TODOs in next PR

PR #2994, opened 2026-04-13
fix(server): clarify condition resolution semantics for label queries

contrueCT:
## Purpose of the PR

- close #2992

`ConditionQuery.condition()` historically combines several meanings in one API:

- no matching condition
- EQ/IN conditions whose intersection is empty
- one resolved value
- the raw list from a sole `IN` relation
- an exception when several relations still resolve to multiple values

This PR preserves that legacy behavior, adds explicit condition-resolution APIs, and
migrates the high-risk `LABEL` call sites to semantics that match each caller.

### Visual overview

Condition resolution semantics and label-query optimization for PR #2994

The diagram contrasts strict and tolerant single-value resolution and shows why
negative or ambiguous label predicates stay local to preserve complete results.
Eligible positive EQ/IN label predicates can still use label indexes; the diagram
is a summary, not an exhaustive query-plan description.

### Make condition resolution explicit

- `containsCondition(HugeKeys key)` reports any top-level relation for the system key.
  Its `Object` implementation is private; the existing operator-based overload remains public.
- `containsConditionValues(key)` reports whether a top-level `EQ`/`IN` relation exists,
  including an empty `IN` relation.
- `conditionValues(key)` returns the resolved `EQ`/`IN` intersection. Pair it with
  `containsConditionValues(key)` when absence and an empty intersection must differ.
- `conditionValue(key)` returns `null` for an empty result, returns the value for a
  singleton, and rejects a multi-value result.
- `singleConditionValueOrNull(key)` returns the value only for a singleton and returns
  `null` for both empty and multi-value results.
- `condition(key)` remains backward-compatible, including returning the raw list for a
  sole `IN` relation.

### Migrate label-sensitive callers

- Use strict `conditionValue()` semantics where serializers and sort-key paths require
  one resolved label.
- Use `singleConditionValueOrNull()` where an optimization is valid only for exactly one
  resolved label.
- Resolve label intersections when collecting matched indexes, including multi-label and
  conflicting-label queries.
- Apply the same semantics to graph/index transactions, traversers, and HStore.
- Keep RamTable's multi-label adjacency fast path for nonempty valid label candidates;
  flatten them into per-label queries and reject empty or unsupported conditions.

### Preserve correctness for negative-label predicates

A downstream unsafe label predicate, including one after a range, side effect, or
inside a child traversal, keeps candidate-filtering predicates local. This avoids
losing matches from labels without equivalent index coverage.

The fallback handles query controls and SEARCH predicates separately:

[...]

PR #3184, opened 2026-09-02
fix(hstore): don't push sysprop-only range queries down to the store

SebastianGruza:
## Purpose of the PR

- part of #3090 (interim mitigation for the operator-sinking property codec mismatch)

On 1.7.x with HStore, every sort-key **prefix-equality or range** traversal fails, e.g. `g.V('a').outE('flow').has('asset','ETC')` → `Can't construct Cardinality from code 0`, or with a range on the second sort key → `Unsupported data type UNKNOWN`. Minimal REST+Gremlin reproducer: https://gist.github.com/SebastianGruza/616f81e915f00f08c4be3dc447ca77c7 (analysis in https://github.com/apache/hugegraph/issues/3090#issuecomment — see the September comment).

Root cause of the crash path: such a query reaches `HstoreTable.queryByRange()` with **sysprop conditions only** (owner vertex, direction, label, sort values) — all already enforced by the scan key range — but `queryByRange()` pushed the serialized `ConditionQuery` to the store **unconditionally**. The store-side row decoder then tries to parse property values it cannot parse (server writes raw values, store reader expects a self-describing `(cardinality<<6)|dataType` byte) and crashes on the first row in range.

## Main Changes

1. `queryByRange()` now applies the **same `prepareConditionQuery()` guard that `queryByPrefix()` already uses**: the query is pushed down only when user-prop conditions remain. No on-disk format change, no store-side change; the proper versioned sinking codec remains tracked in #3090.
2. Both `prepareConditionQuery()` / `prepareConditionQueryList()` now operate on a **copy** of the origin query instead of mutating it via `resetConditions()` — core still uses the origin query for its own result filtering after the scan returns.

## Verifying these changes

- New unit tests in `HstoreTableTest` (a scan-recording `Session` fake):
  - sysprop-only range query → `scan()` receives `null` query bytes, origin query untouched;
  - range query with a user-prop condition → query bytes are pushed **and** the origin query keeps all its conditions (guards the copy-not-mutate change).
- `mvn -pl hugegraph-server/hugegraph-hstore -am test -Dtest=HstoreTableTest`: 7/7 pass.
- **Live A/B on a local PD + HStore + Server cluster** (single-node, fresh graph, reproducer above):
  - `master` (98477f0): sort-key prefix query fails with `Can't construct Cardinality from code 0`, range on the second sort key fails with `Unsupported data type UNKNOWN` — exactly as reported;
  - this branch: **all 6 reproducer queries return correct results** (3 / 1 / 2 / 2 / 1 / 0 edges), verified with a jar swap in both directions on the same running cluster.

- [ ] Nope
- [x] Logic (the HStore range-scan pushdown path)

**Note on `codecov/patch`:** the red patch check is an artifact — the surefire run of the `hugegraph-hstore` module is not part of the coverage upload, so the report shows 0 % for the changed lines although `HstoreTableTest` (8 tests) covers them. Scale measurement (13 sort-key shapes on 5 M edges, HStore vs RocksDB id sets): https://github.com/SebastianGruza/hugegraph-validation/tree/master/results/pr-3184-sortkeys

### DIFF
--- a/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
+++ b/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
@@ -631,7 +631,6 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
             type |= query.inclusiveEnd() ?
                     Session.SCAN_LTE_END : Session.SCAN_LT_END;
         }
-        ConditionQuery cq;
         Query origin = query.originQuery();
         byte[] position = null;
         byte[] ownerStart = this.ownerByQueryDelegate.apply(query.resultType(),
@@ -648,6 +647,7 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
         if (query.paging() && !query.page().isEmpty()) {
             position = PageState.fromString(query.page()).position();
         }
+        byte[] queryBytes = null;
         if (origin instanceof ConditionQuery &&
             (query.resultType().isEdge() || query.resultType().isVertex())) {
             // Same guard as queryByPrefix(): only push the query down to the
@@ -656,13 +656,11 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
             // label, sort values), which are already enforced by the key
             // range, and the store-side row decoder cannot parse the raw
             // property layout written by the server (see issue #3090).
-            cq = prepareConditionQuery((ConditionQuery) origin);
-            byte[] queryBytes = cq == null ? null : cq.bytes();
-            return session.scan(this.table(), ownerStart,
-                                ownerEnd, start, end, type, queryBytes, position);
+            ConditionQuery cq = prepareConditionQuery((ConditionQuery) origin);
+            queryBytes = cq == null ? null : cq.bytes();
         }
-        return session.scan(this.table(), ownerStart,
-                            ownerEnd, start, end, type, null, position);
+        return session.scan(this.table(), ownerStart, ownerEnd, start, end,
+                            type, queryBytes, position);
     }
 
     static boolean shouldUseOrderedRangeScan(IdRangeQuery query) {
--- a/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
+++ b/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
@@ -568,8 +568,11 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
             }
         }
         if (newConditions.size() > 0) {
-            conditionQuery.resetConditions(newConditions);
-            return conditionQuery;
+            // NOTE: copy before reset, the origin query is still used by core
+            // for result filtering after the backend scan returns
+            ConditionQuery pushdown = conditionQuery.copy();
+            pushdown.resetConditions(newConditions);
+            return pushdown;
         } else {
             return null;
         }
@@ -594,8 +597,10 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
             }
         }
         if (newConditions.size() > 0) {
-            conditionQuery.resetConditions(newConditions);
-            return conditionQuery;
+            // NOTE: copy before reset, see prepareConditionQuery()
+            ConditionQuery pushdown = conditionQuery.copy();
+            pushdown.resetConditions(newConditions);
+            return pushdown;
         } else {
             return null;
         }
@@ -642,16 +647,16 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
         }
         if (origin instanceof ConditionQuery &&
             (query.resultType().isEdge() || query.resultType().isVertex())) {
-            cq = (ConditionQuery) query.originQuery();
-
-            // LOG.debug("query {} with ownerKeyFrom: {}, ownerKeyTo: {}, " +
-            //          "keyFrom: {}, keyTo: {}, " +
-            //          "scanType: {}, conditionQuery: {}",
-            //          this.table(), bytes2String(ownerStart),
-            //          bytes2String(ownerEnd), bytes2String(start),
-            //          bytes2String(end), type, cq.bytes());
+            // Same guard as queryByPrefix(): only push the query down to the
+            // store when user-prop conditions remain. A sort-key prefix/range
+            // query keeps sysprop conditions only (owner vertex, direction,
+            // label, sort values), which are already enforced by the key
+            // range, and the store-side row decoder cannot parse the raw
+            // property layout written by the server (see issue #3090).
+            cq = prepareConditionQuery((ConditionQuery) origin);
+            byte[] queryBytes = cq == null ? null : cq.bytes();
             return session.scan(this.table(), ownerStart,
-                                ownerEnd, start, end, type, cq.bytes(), position);
+                                ownerEnd, start, end, type, queryBytes, position);
         }
         return session.scan(this.table(), ownerStart,
                             ownerEnd, start, end, type, null, position);
--- a/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
+++ b/hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
@@ -569,9 +569,11 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
         }
         if (newConditions.size() > 0) {
             // NOTE: copy before reset, the origin query is still used by core
-            // for result filtering after the backend scan returns
+            // for result filtering after the backend scan returns; drop the
+            // back reference so the serialized payload stays flat
             ConditionQuery pushdown = conditionQuery.copy();
             pushdown.resetConditions(newConditions);
+            pushdown.setOriginQuery(null);
             return pushdown;
         } else {
             return null;
@@ -600,6 +602,7 @@ public class HstoreTable extends BackendTable<Session, BackendEntry> {
             // NOTE: copy before reset, see prepareConditionQuery()
             ConditionQuery pushdown = conditionQuery.copy();
             pushdown.resetConditions(newConditions);
+            pushdown.setOriginQuery(null);
             return pushdown;
         } else {
             return null;
--- a/hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
+++ b/hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.hugegraph.backend.store.hstore;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
@@ -29,6 +30,7 @@ import org.apache.hugegraph.backend.page.PageInfo;
 import org.apache.hugegraph.backend.page.PageState;
 import org.apache.hugegraph.backend.query.Condition;
 import org.apache.hugegraph.backend.query.ConditionQuery;
+import org.apache.hugegraph.backend.query.IdPrefixQuery;
 import org.apache.hugegraph.backend.query.IdRangeQuery;
 import org.apache.hugegraph.backend.query.Query;
 import org.apache.hugegraph.backend.store.BackendEntry;
@@ -177,6 +179,42 @@ public class HstoreTableTest {
         Assert.assertNull(pushed.originQuery());
     }
 
+    @Test
+    public void testPrefixListQueryPushesCopyAndKeepsOrigin() {
+        // prepareConditionQueryList() is reached from queryByPrefixList() and
+        // from the streaming query(Session, Iterator, String); one origin
+        // query is shared by every prefix query of the batch
+        ConditionQuery origin = new ConditionQuery(HugeType.EDGE);
+        origin.eq(HugeKeys.OWNER_VERTEX, IdGenerator.of("v1"));
+        origin.eq(HugeKeys.DIRECTION, Directions.OUT);
+        origin.eq(HugeKeys.LABEL, IdGenerator.of(1L));
+        origin.query(Condition.eq(IdGenerator.of(7L), 100));
+        int before = origin.conditions().size();
+        List<IdPrefixQuery> queries = Arrays.asList(
+                new IdPrefixQuery(origin, IdGenerator.of(keyBytes(1),
+                                                         IdType.STRING)),
+                new IdPrefixQuery(origin, IdGenerator.of(keyBytes(2),
+                                                         IdType.STRING)));
+
+        ScanRecordingSession session = new ScanRecordingSession();
+        List<BackendColumnIterator> iterators = this.newTestTable()
+                .queryByPrefixList(session, queries, "g+oe");
+
+        Assert.assertTrue(session.scanCalled);
+        Assert.assertEquals(2, session.lastOwnerKeys.size());
+        Assert.assertEquals(2, iterators.size());
+        Assert.assertNotNull(session.lastQueryBytes);
+        // the shared origin query keeps every condition, including the
+        // owner vertex that the pushed copy drops
+        Assert.assertEquals(before, origin.conditions().size());
+        Assert.assertNotNull(origin.condition(HugeKeys.OWNER_VERTEX));
+        ConditionQuery pushed = ConditionQuery.fromBytes(session.lastQueryBytes);
+        Assert.assertNull(pushed.condition(HugeKeys.OWNER_VERTEX));
+        Assert.assertNotNull(pushed.condition(HugeKeys.LABEL));
+        Assert.assertFalse(pushed.userpropConditions().isEmpty());
+        Assert.assertNull(pushed.originQuery());
+    }
+
     private HstoreTable newTestTable() {
         HstoreTable table = new HstoreTable("hugegraph", "g+oe");
         table.ownerByQueryDelegate = (type, id) -> new byte[]{0};
@@ -210,6 +248,7 @@ public class HstoreTableTest {
 
         private boolean scanCalled = false;
         private byte[] lastQueryBytes = null;
+        private List<HgOwnerKey> lastOwnerKeys = null;
 
         @Override
         public BackendColumnIterator scan(String table, byte[] ownerKeyFrom,
@@ -219,6 +258,21 @@ public class HstoreTableTest {
             this.scanCalled = true;
             this.lastQueryBytes = query;
             return new TestColumnIterator();
+        }
+
+        @Override
+        public List<BackendColumnIterator> scan(String table,
+                                                List<HgOwnerKey> keys,
+                                                int scanType, long limit,
+                                                byte[] query) {
+            this.scanCalled = true;
+            this.lastQueryBytes = query;
+            this.lastOwnerKeys = keys;
+            List<BackendColumnIterator> iterators = new ArrayList<>();
+            for (int i = 0; i < keys.size(); i++) {
+                iterators.add(new TestColumnIterator());
+            }
+            return iterators;
         }
 
         @Override
@@ -311,14 +365,6 @@ public class HstoreTableTest {
         @Override
         public BackendColumnIterator scan(String table, byte[] ownerKey,
                                           byte[] prefix) {
-            throw new UnsupportedOperationException();
-        }
-
-        @Override
-        public List<BackendColumnIterator> scan(String table,
-                                                List<HgOwnerKey> keys,
-                                                int scanType, long limit,
-                                                byte[] query) {
             throw new UnsupportedOperationException();
         }
 
--- a/hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
+++ b/hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
@@ -169,6 +169,12 @@ public class HstoreTableTest {
         // the pushed-down query is a copy: the origin query keeps all its
         // conditions for core-side filtering after the scan returns
         Assert.assertEquals(before, origin.conditions().size());
+        // pushed payload: user-prop condition survives, owner-vertex is
+        // dropped, and the back reference to the origin query is cleared
+        ConditionQuery pushed = ConditionQuery.fromBytes(session.lastQueryBytes);
+        Assert.assertNull(pushed.condition(HugeKeys.OWNER_VERTEX));
+        Assert.assertFalse(pushed.userpropConditions().isEmpty());
+        Assert.assertNull(pushed.originQuery());
     }
 
     private HstoreTable newTestTable() {
--- a/hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
+++ b/hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
@@ -18,21 +18,29 @@
 package org.apache.hugegraph.backend.store.hstore;
 
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hugegraph.backend.id.Id.IdType;
 import org.apache.hugegraph.backend.id.IdGenerator;
 import org.apache.hugegraph.backend.page.PageInfo;
 import org.apache.hugegraph.backend.page.PageState;
+import org.apache.hugegraph.backend.query.Condition;
+import org.apache.hugegraph.backend.query.ConditionQuery;
 import org.apache.hugegraph.backend.query.IdRangeQuery;
 import org.apache.hugegraph.backend.query.Query;
 import org.apache.hugegraph.backend.store.BackendEntry;
 import org.apache.hugegraph.backend.store.BackendEntry.BackendColumn;
 import org.apache.hugegraph.backend.store.BackendEntry.BackendColumnIterator;
 import org.apache.hugegraph.backend.store.BackendEntryIterator;
+import org.apache.hugegraph.store.HgOwnerKey;
 import org.apache.hugegraph.store.client.util.HgStoreClientConst;
 import org.apache.hugegraph.type.HugeType;
+import org.apache.hugegraph.type.define.Directions;
+import org.apache.hugegraph.type.define.GraphMode;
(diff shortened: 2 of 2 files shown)


hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
```diff
@@ -594,8 +597,10 @@ private ConditionQuery prepareConditionQueryList(ConditionQuery conditionQuery)
             }
         }
         if (newConditions.size() > 0) {
-            conditionQuery.resetConditions(newConditions);
-            return conditionQuery;
+            // NOTE: copy before reset, see prepareConditionQuery()
+            ConditionQuery pushdown = conditionQuery.copy();
```
bitflicker64, line 603: ⚠️ The copy fix lands here, but the entry guard of this method is still `containsLabelOrUserpropRelation()` (line 588), while `prepareConditionQuery()` uses `userpropConditions()` (line 559). That call returns true for a bare `HugeKeys.LABEL` relation (hugegraph-core `backend/query/ConditionQuery.java:249-253`), so a batched edge query with a label and no user property still reaches `resetConditions()` with sysprop-only conditions and is pushed at line 544. Same shape this PR removes from `queryByRange()`. The path is live: `query(Session, List<IdPrefixQuery>, String)` at line 331.

The guard is pre-existing and outside this diff, so not a change request on this PR: worth a follow-up on #3090, or a line in the new comment saying why the list path is different.

Confidence: the guard mismatch is confirmed by reading the two methods. Whether this batch path actually hits the decode failure is not, since I could not reproduce it locally.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
```diff
@@ -568,8 +568,11 @@ private ConditionQuery prepareConditionQuery(ConditionQuery conditionQuery) {
             }
         }
         if (newConditions.size() > 0) {
-            conditionQuery.resetConditions(newConditions);
-            return conditionQuery;
+            // NOTE: copy before reset, the origin query is still used by core
+            // for result filtering after the backend scan returns
+            ConditionQuery pushdown = conditionQuery.copy();
```
bitflicker64, line 574: 🧹 The copy keeps a link back to the query it came from, so the payload grows by one nesting level. `ConditionQuery.copy()` ends with `query.originQuery(this)` (hugegraph-core `backend/query/ConditionQuery.java:577`), `Query.originQuery` is non-transient (`backend/query/Query.java:78`), and `bytes()` serializes with a `Gson` that registers type adapters but no exclusion strategy (`ConditionQuery.java:81-85, 980-983`).

Measured on a build of this head, for an edge query with owner vertex, direction, label and one user-prop condition, using the shape the pipeline produces (the origin is itself a copy from `GraphTransaction.java:1591`, so one level of nesting predates this PR):

- before: 1466 bytes
- after: 2275 bytes
- after, with the back reference cleared: 657 bytes

No behaviour change: store-side consumers read only the top-level query (`hg-store-core/.../business/FilterIterator.java:53-85`, `hg-store-node/.../query/stages/FilterStage.java:36-52`) and nothing in `hugegraph-store` reads `originQuery`.

Requested change: `pushdown.setOriginQuery(null);` after the reset, in both this method and `prepareConditionQueryList()`. `Query.setOriginQuery(Query)` is public (`backend/query/Query.java:142-144`), and the pushdown is discarded right after `bytes()`, so nothing else observes the link.

hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
```diff
@@ -124,6 +132,59 @@ public void testRangeScanBudgetIncludesOneLookaheadRecord() {
...
+        ConditionQuery origin = new ConditionQuery(HugeType.EDGE);
+        origin.eq(HugeKeys.OWNER_VERTEX, IdGenerator.of("v1"));
+        origin.query(Condition.eq(IdGenerator.of(7L), 100));
+        int before = origin.conditions().size();
+
+        ScanRecordingSession session = new ScanRecordingSession();
+        this.newTestTable().queryByRange(session, edgeRangeQuery(origin));
+
+        Assert.assertTrue(session.scanCalled);
+        Assert.assertNotNull(session.lastQueryBytes);
```
bitflicker64, line 170: 🧹 This asserts only that some bytes were pushed. What is worth pinning down is which conditions survive `prepareConditionQuery()`, and that is unchecked: the test still passes if the code pushes the untrimmed origin, or keeps the owner-vertex condition the method drops. The size assertion on line 171 does guard the copy-not-mutate change, so only the payload content is unverified.

Requested change: decode with `ConditionQuery.fromBytes(session.lastQueryBytes)` (hugegraph-core `backend/query/ConditionQuery.java:763-773`) and assert that the user-prop condition survives and `condition(HugeKeys.OWNER_VERTEX)` is null. I checked the round trip works on this head.

Separately, a question rather than a change: is this shape reachable? An `IdRangeQuery` with an edge result type seems to come only from `BinarySerializer.writeQueryEdgeRangeCondition()` (line 717), reached only from the sort-keys branch that calls `resetUserpropConditions()` (`GraphTransaction.java:1602`). If so `prepareConditionQuery()` always returns null here in practice, and this test pins a synthetic shape.

bitflicker64 (commented):
Blocking: no. Summary: The guard is the right shape and I did not find a correctness regression. For sort-key edge queries core already resets user props on the copy it sends down (GraphTransaction.java:1591-1602), the scan key range encodes owner vertex, direction, label, sub-label and sort values (BinarySerializer.java:667-717), and core still filters with cq.test() because the query is marked OptimizedType.SORT_KEYS (GraphTransaction.java:1590, 1933), so dropping the pushdown does not widen the result set. Three notes inline: the sibling guard in prepareConditionQueryList() still lets label-only queries through, the pushdown copy adds a nesting level to the serialized payload, and the new user-prop test does not check what was pushed. Evidence: `mvn -o -pl hugegraph-server/hugegraph-hstore -am test -Dtest=HstoreTableTest` on JDK 11 at 847fab9 gives BUILD SUCCESS with 7/7 passing, matching the PR description (the same command on JDK 17 fails in hugegraph-commons on Lombok, in files this PR does not touch). Payload sizes come from probes compiled against this head. `gh -R apache/hugegraph pr checks 3184` reports no checks on the branch, so the HStore integration lanes are unverified here, and I did not start a cluster, so the store-side decode failure itself is unreproduced.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
SebastianGruza: Done in 1072872 — `setOriginQuery(null)` on the pushdown copy in both prepare methods. Thanks for measuring the payload; the nesting predating this PR (origin already being a copy from `GraphTransaction.java:1591`) was a good catch I had missed entirely.

hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
SebastianGruza: Done in 1072872 — the test now decodes with `ConditionQuery.fromBytes` and asserts the user-prop survives, `OWNER_VERTEX` is gone and `originQuery` is null.

On reachability: agreed — for EDGE sort-key traversals core resets user props before the backend sees the query (that is exactly why suppressing the pushdown is safe), so on current server paths `prepareConditionQuery()` should return null here in practice. The test deliberately pins the method contract rather than a live traversal shape: `queryByRange` also serves VERTEX result types and any future caller, and if core ever stops stripping user props, this is the behavior we want locked. Happy to add a comment in the test saying the shape is synthetic.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
SebastianGruza: Agreed, and thanks for confirming the guard mismatch. Leaving the list path untouched here to keep the diff minimal — will flag `containsLabelOrUserpropRelation()` vs `userpropConditions()` (and the related question of whether label-only pushes can hit the same decode path, plus the cross-module `Id` equality concern for LABEL/SUB_LABEL) as a follow-up under #3090.

bitflicker64 (commented):
Blocking: no. Summary: The guard is sound: the pushed-down query is filter-only on the store side, every sysprop that can reach the edge range path is already encoded in the scan key, and core re-tests the results regardless. The copy-not-mutate change additionally stops the backend mutating a query its caller still owns. Evidence: built this head with JDK 11, `mvn -o -pl hugegraph-server/hugegraph-hstore -am test -Dtest=HstoreTableTest` gives 7/7. Store side is filter-only: `hg-store-node/.../grpc/HgStoreWrapperEx.java` hands the query bytes straight to `FilterIterator.of`, which returns the raw iterator when they are empty (`hg-store-core/.../business/FilterIterator.java:45-47`) and otherwise runs `parseEdge`/`parseVertex` then `query.test()` (lines 60-72). Core re-tests: the sort-keys branch sets `optimized(OptimizedType.SORT_KEYS)` (`GraphTransaction.java:1590`), and `queryEdges(Query)` feeds `filterUnmatchedRecords` (line 1014) into `rightResultFromIndexQuery`, which calls `cq.test(elem)` (line 1933). Key coverage: `BinarySerializer.writeQueryEdgeRangeCondition()` (line 667) writes owner vertex, direction, label and sub-label into the key, and the only server-code producer of an `OTHER_VERTEX` condition (`EdgeExistenceTraverser.java:48`) uses an EQ sort value, so `hasRangeCondition()` (line 660) is false for it and it never reaches that method. Payload shape: `ConditionQuery.copy()` sets `originQuery(this)` and resets `optimizedType`/`resultsFilter` (`ConditionQuery.java:575-585`), which is why `setOriginQuery(null)` is needed before `bytes()` (line 980).

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
```diff
@@ -642,16 +650,16 @@ protected BackendColumnIterator queryByRange(Session session,
...
-            //          bytes2String(ownerEnd), bytes2String(start),
-            //          bytes2String(end), type, cq.bytes());
+            // Same guard as queryByPrefix(): only push the query down to the
+            // store when user-prop conditions remain. A sort-key prefix/range
+            // query keeps sysprop conditions only (owner vertex, direction,
+            // label, sort values), which are already enforced by the key
+            // range, and the store-side row decoder cannot parse the raw
+            // property layout written by the server (see issue #3090).
+            cq = prepareConditionQuery((ConditionQuery) origin);
+            byte[] queryBytes = cq == null ? null : cq.bytes();
```
bitflicker64: 🧹 With the guard in place both exits are the same `session.scan(this.table(), ownerStart, ownerEnd, start, end, type, ..., position)` call, differing only in the query bytes (lines 661-662 vs 664-665).

Requested change: initialise `byte[] queryBytes = null;` before the `if`, assign it inside the branch, and end the method with a single `return session.scan(...)`. While there, the method-scoped `ConditionQuery cq;` at line 634 can become a block-local; that declaration predates this PR, so treat it as optional.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
```diff
@@ -594,8 +599,11 @@ private ConditionQuery prepareConditionQueryList(ConditionQuery conditionQuery)
             }
         }
         if (newConditions.size() > 0) {
-            conditionQuery.resetConditions(newConditions);
-            return conditionQuery;
+            // NOTE: copy before reset, see prepareConditionQuery()
+            ConditionQuery pushdown = conditionQuery.copy();
+            pushdown.resetConditions(newConditions);
+            pushdown.setOriginQuery(null);
```
bitflicker64, line 605: 🧹 The copy-not-mutate change lands in both prepare methods, but the new tests cover only `prepareConditionQuery()`, and only through `queryByRange()`. `prepareConditionQueryList()` has two live call sites, `queryByPrefixList()` (line 542) and the streaming `query(Session, Iterator<IdPrefixQuery>, String)` (line 360), and neither is exercised: `HstoreTableTest` has 7 tests and none reach the prefix or list paths. Its entry guard `containsLabelOrUserpropRelation()` (line 590) admits label-only edge queries, so both call sites do reach line 603.

Requested change: make `ScanRecordingSession.scan(String, List<HgOwnerKey>, int, long, byte[])` record instead of throw, and add a `queryByPrefixList()` case asserting the shared origin query keeps its `OWNER_VERTEX` condition after the scan.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
SebastianGruza: Done in 0ecc10a — `queryBytes` is initialised before the `if`, assigned inside it, and the method ends with a single `session.scan(...)`; `cq` is block-local now.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
SebastianGruza: Done in 0ecc10a — `ScanRecordingSession.scan(String, List<HgOwnerKey>, int, long, byte[])` now records (query bytes, owner keys, one iterator per key) instead of throwing, and `testPrefixListQueryPushesCopyAndKeepsOrigin` drives `prepareConditionQueryList()` through `queryByPrefixList()` with one origin shared by two prefix queries: the origin keeps all of its conditions including `OWNER_VERTEX`, the decoded pushed payload has no `OWNER_VERTEX`, keeps `LABEL` and the user-prop condition, and `originQuery` is null. 8/8 with `mvn -pl hugegraph-server/hugegraph-hstore -am -Dtest=HstoreTableTest test` on Temurin 17.

hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
```diff
@@ -124,6 +134,101 @@ public void testRangeScanBudgetIncludesOneLookaheadRecord() {
...
+        // dropped, and the back reference to the origin query is cleared
+        ConditionQuery pushed = ConditionQuery.fromBytes(session.lastQueryBytes);
+        Assert.assertNull(pushed.condition(HugeKeys.OWNER_VERTEX));
+        Assert.assertFalse(pushed.userpropConditions().isEmpty());
+        Assert.assertNull(pushed.originQuery());
+    }
+
+    @Test
+    public void testPrefixListQueryPushesCopyAndKeepsOrigin() {
+        // prepareConditionQueryList() is reached from queryByPrefixList() and
```
bitflicker64, line 184: 🧹 Worth stating what sits above these two call sites: nothing calls either of them in a live server, so `prepareConditionQueryList()` does not run today. As written, lines 184-186 read as a description of a live path.

`queryByPrefixList` (`HstoreTable.java:526`) has one production caller, `HstoreTable.query(Session, List<IdPrefixQuery>, String)` (line 331), whose only caller is `HstoreStore.query(List<HugeType>, List<IdPrefixQuery>)` (`HstoreStore.java:506`), and that overload is never called: `HstoreStore.java:492` passes a `QueryWrapper implements Iterator<IdPrefixQuery>` and so binds to the `Iterator` overload at line 540. The streaming `query(Session, Iterator<IdPrefixQuery>, String)` (`HstoreTable.java:348`, called from `HstoreStore.java:557`) is reachable only through `BackendStore.query(Iterator<Query>, Function, HugeGraph)` (`BackendStore.java:75-78`), which carries the comment `// TODO: unused now`; nothing in the repository supplies its `queryWriter` argument.

This withdraws the reachability half of my earlier comments (`3916985809`, `3921393173`), which called these call sites live and are where the wording here came from. The guard mismatch itself still stands, but it cannot reach the store row decoder as things are, which is worth recording in the #3090 follow-up.

Requested change, optional and not worth a repush on its own: reword lines 184-186 to say the method is called from those two sites and that neither has a live caller today.

bitflicker64 (commented):
Blocking: no. Summary: Both cleanups asked for on 1072872 landed without changing behaviour and I found no correctness defect at this head; the single note withdraws the reachability half of my own earlier comments.

Evidence: on JDK 11 at this head, `mvn -o -pl hugegraph-server/hugegraph-hstore -am test` is BUILD SUCCESS with 11 tests (`HstoreTableTest` 8/8) and `mvn -o -pl hugegraph-server/hugegraph-hstore checkstyle:check` is clean; `git diff 1072872 0ecc10a` is the `queryBytes` hoist plus the new test, with both former exits of `queryByRange()` collapsing into the same `session.scan(...)` call.

On the second behaviour change, which the title does not advertise: with user props present the range path now pushes a copy stripped of the `OWNER_VERTEX` conditions instead of the untrimmed origin. `onlyOwnerVertex()` (`HstoreTable.java:612`) drops only conditions whose relations are all on `OWNER_VERTEX`, and `writeQueryEdgeRangeCondition()` already pins that vertex in the scan key (`writePartitionedId(HugeType.EDGE, vertex, start)`, `BinarySerializer.java:680`), so the trimmed filter cannot admit a row the key range would not return anyway. Suppressing a sysprop-only pushdown also puts hstore where `RocksDBTable.queryByRange()` (`RocksDBTable.java:258-269`) already sits: it pushes no query at all.

hugegraph-server/hugegraph-hstore/src/test/java/org/apache/hugegraph/backend/store/hstore/HstoreTableTest.java
SebastianGruza: Done in eb5e05a — the comment now says the method is called from `queryByPrefixList()` and the streaming `query(Session, Iterator, String)`, that neither has a live caller in the server today, and that the test pins the method contract. Comment-only change, no code touched.

contrueCT:
@SebastianGruza @imbajin  I think it makes sense to merge #3184 before #2994.

The change here is relatively isolated: it avoids pushing sysprop-only range conditions to HStore when those constraints are already covered by the key range, while user-property residual conditions are still pushed down. Also, copying the `ConditionQuery` before trimming it avoids mutating the query still used by the core layer.

More importantly, #3184 removes an existing HStore failure unrelated to #2994. Without it, the HStore regression for #2994 is partially blocked by #3090, which makes failures harder to attribute.

So I suggest merging #3184 first, then rebasing #2994 on the latest master and rerunning the HStore/RocksDB regression matrix.

#3090 should still remain open since #3184 is only a mitigation for the underlying codec issue.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
```diff
@@ -640,21 +647,20 @@ protected BackendColumnIterator queryByRange(Session session,
...
-        }
-        return session.scan(this.table(), ownerStart,
-                            ownerEnd, start, end, type, null, position);
+            // Same guard as queryByPrefix(): only push the query down to the
+            // store when user-prop conditions remain. A sort-key prefix/range
+            // query keeps sysprop conditions only (owner vertex, direction,
+            // label, sort values), which are already enforced by the key
+            // range, and the store-side row decoder cannot parse the raw
+            // property layout written by the server (see issue #3090).
+            ConditionQuery cq = prepareConditionQuery((ConditionQuery) origin);
```
bitflicker64, line 663: ⚠️ The guard lands here, but the same unguarded pushdown survives in `queryAll()` (lines 448-469), which pushes `((ConditionQuery) query).bytes()` on both exits with no guard at all. Unlike the `prepareConditionQueryList()` path discussed earlier on this PR, that one has a live caller: `queryBy()` (line 405) -> `queryByCond()` (line 707) -> `queryAll()`.

Reachability for a sysprop-only condition query such as `g.V().hasLabel('person')` or `g.E().hasLabel('flow')`: `BinarySerializer.writeQueryEdgePrefixCondition()` (`BinarySerializer.java:721-763`) returns `null` when `OWNER_VERTEX` is absent (the loop breaks on the first key, `count == 0`), and `writeQueryCondition()` (line 765-769) returns the query unchanged for non-index types, so a plain `ConditionQuery` arrives at `queryBy()` with conditions but no id, prefix or range and falls through to `queryAll()`. `HstoreSessionsImpl` then hands those bytes straight to the store (`scan(String, byte[])` at line 582-588, paging overload at 710-729), i.e. the same `FilterIterator` -> `parseEdge`/`parseVertex` decoder this PR is avoiding. The shard scan `queryByRange(Session, Shard, ConditionQuery)` (line 710-728) calls `query.bytes()` unconditionally too.

Worth noting that `prepareConditionQuery()` cannot simply be copied to `queryAll()`: nothing else enforces the label there, since `GraphTransaction.rightResultFromIndexQuery()` returns `true` early for a label-only edge `ConditionQuery` (`GraphTransaction.java:1914-1918`), so dropping the pushdown would return unfiltered rows.

Requested change: record `queryAll()` and the shard overload explicitly in the #3090 follow-up, and add a line to the comment above saying the guard applies to the range path because the key range already covers its sysprops, while the full-scan path still needs the pushdown to filter.

Confidence: the reachability trace is confirmed statically on this head. I could not reproduce the decode failure on a live PD + HStore cluster, so the size of the residual gap is taken from the root cause in the PR description rather than measured.

bitflicker64 (commented):
Blocking: no. Summary: The range-path guard is sound and I found no correctness regression at this head: suppressing the pushdown only removes store-side filtering, every sysprop reaching the edge range path is already encoded in the scan key, and core re-tests the results; the copy-not-mutate change stops the backend stripping conditions from a query its caller still owns. One note: the same unguarded pushdown is still live in `queryAll()`. Evidence: built this head with JDK 17, `mvn -o -pl hugegraph-server/hugegraph-hstore -am test -Dtest=HstoreTableTest` gives 8/8. Store side is filter-only: `FilterIterator.of` returns the raw iterator when the query bytes are empty (`hg-store-core/.../business/FilterIterator.java:44-50`) and otherwise runs `parseEdge`/`parseVertex` then `query.test()` (lines 52-84). Key coverage: `BinarySerializer.writeQueryEdgeRangeCondition()` (line 667-717) writes owner vertex, direction, label and sub-label into the key. Core re-tests: `optimizeQuery()` sets `optimized(OptimizedType.SORT_KEYS)` (`GraphTransaction.java:1589`) and `queryEdges(Query)` feeds `filterUnmatchedRecords` (line 1014) into `rightResultFromIndexQuery`, which calls `cq.test(elem)` (line 1933). `setOriginQuery(null)` is payload-only: the store deserializes into `org.apache.hugegraph.query.ConditionQuery` (hugegraph-struct 1.7.0), whose `test(BaseElement)` reads `ids`, `resultsFilter`, `conditions` and `element2IndexValueMap` and never `originQuery` or `optimizedType` (verified with `javap -c`). manual_review_needed: no check runs are reported for `eb5e05ab` (combined status `pending`, no contexts), and no live PD + HStore cluster was available, so the author's A/B result was not reproduced.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreTable.java
SebastianGruza: Done in 89bb0ec: four lines above the guard now say it applies to the range path only, because the key range already covers its sysprops, while the full-scan path (`queryAll()` and the shard overload of `queryByRange`) still needs the pushdown to filter. `queryAll()` and the shard overload are recorded in the #3090 follow-up: https://github.com/apache/hugegraph/issues/3090#issuecomment-5694243559.

Two things from the lab on the reachability itself, master `36811483`, the same cluster as in the PR description:

- The `g.E().hasLabel('flow')` example does not reach `queryAll()` on hstore. `HstoreFeatures.supportsQueryByLabel()` returns `false`, so for a label-only query `GraphTransaction.optimizeQuery` (line 1655) does not hand the query to the backend; `GraphIndexTransaction.queryByLabel` goes to the label index table and fetches the edges by id. Measured: a label with the label index enabled and 3 edges → 3 rows, also with `valueMap(true)`; a label with the label index disabled → 400 "label index is disabled", raised in core. The same 3 rows through `g.V(x).outE(l).has('asset','ETC')` → `Can't construct Cardinality from code 0`, so the rows do trip the decoder. `g.E()` without `has()` sends a plain `Query` (`HugeGraphStep.makeQuery`), so `queryAll()` pushes `null`.
- Two paths really do push an unguarded `ConditionQuery`: the paged full scan `g.E().has('~page','').limit(2).valueMap(true)` (paging branch of `queryAll()`) and REST `/traversers/edges/scan?page=&page_limit=2` (shard overload). Both returned rows on master, on the same owner's edges on which the range path fails. I did not trace why the decoder is not tripped there, so I leave both open in #3090 rather than call them safe.

SebastianGruza:
@contrueCT thanks, agreed on the order. Once #2994 is rebased on a master that contains this PR, I will run the same HStore/RocksDB matrix (the same queries on both backends, compared as id sets, 170+ shapes) on the new head and post the result under #2994.

@bitflicker64 the `queryAll()` point is handled in 89bb0ec (a comment line) and recorded in #3090 (https://github.com/apache/hugegraph/issues/3090#issuecomment-5694243559); details and the measurement are in the inline thread.

bitflicker64 (approved):
Blocking: no. Summary: The range-path guard is sound. Sort-key edge queries already have their sysprops encoded in the scan key and core still re-tests results with cq.test(), so dropping the sysprop-only pushdown cannot widen results. Building the pushdown from a copy also stops the backend stripping conditions from a query core still uses. Earlier review points are addressed at this head, and the remaining queryAll() and shard-scan gap is tracked in #3090. Evidence: on this head with JDK 11, `mvn -o -pl hugegraph-server/hugegraph-hstore -am test -Dtest=HstoreTableTest` passes 8/8. All required checks and workflow runs are green. codecov/patch is red but not a required check, and the changed lines are covered by HstoreTableTest.

imbajin:
+1, agreed with the scope and merge order. @SebastianGruza @contrueCT @bitflicker64 this PR can go in ahead of #2994. The range guard is appropriately scoped, and the review threads are now resolved.

After merging, please rebase #2994 and post the HStore/RocksDB regression results for the new head there, as planned. Keep #3090 open for the codec and remaining scan-path follow-ups, and update both threads to distinguish what #3184 fixes from what remains outstanding.

imbajin (approved):
Follow-up context and merge order are noted in https://github.com/apache/hugegraph/pull/3184#issuecomment-5701198188.

PR #3132, opened 2026-07-31
feat(helm): add HStore deployment chart

bitflicker64:
## Purpose of the PR

- close #3131

Add an official Helm installation path for the distributed HugeGraph HStore
topology on Kubernetes. The chart packages the PD, Store, and Server startup
contract that otherwise has to be reconstructed by each operator.

All six image-side prerequisites have merged, including
[#3130](https://github.com/apache/hugegraph/pull/3130) (2026-08-15) and
[#3159](https://github.com/apache/hugegraph/pull/3159) (`321ba4da`,
2026-08-29). So have the three image-side fixes this chart's defaults depend
on: [#3185](https://github.com/apache/hugegraph/pull/3185),
[#3187](https://github.com/apache/hugegraph/pull/3187) and
[#3189](https://github.com/apache/hugegraph/pull/3189), between 2026-09-07 and
2026-09-09. The branch now contains all of `master` through `60c8803d`. The PR
stays a draft until the checklist under "Remaining before this leaves draft"
is complete.

**At a glance (2026-09-10).** Head `2bff77a6`, level with `master` (`60c8803d`), which carries apache/hugegraph#3185, #3187 and #3189. Static gates green on this head: `helm lint` on three presets, renders 16/14/16, 63 unit tests. What the merges bought the chart is defaults instead of overrides: `pd.readinessPath` and `store.waitPath` now default to `/v1/ready`, `HG_SERVER_STARTUP_TIMEOUT_S` is derived from the Server startup probe budget, and the two TODO comments in `values.yaml` are gone. Templates, `values.schema.json` and `tests/` stay byte-identical with the `helm-dev` testing branch (chart 0.1.6, tag `helm-dev-20260910` on hugegraph/hugegraph). Two gates keep this in draft, both external: a 1.8.x release for the image pin, and the open review thread on the Store readiness gate. One gap is on me: the fault battery has not been re-run against images built from merged `master`, so the runtime evidence below is still the 2026-09-05 run on the equivalent pull-request tree. Full detail in the folded sections below.

<details>
<summary><b>About the change: what the chart contains</b></summary>

## Main Changes

- Add the first `helm/hugegraph` chart with chart version `0.1.0`:
  - PD and Store StatefulSets with persistent storage and headless Services.
  - Server Deployment and client Service.
  - PD, Store, and optional Server PodDisruptionBudgets.
  - Optional Server HPA and Ingress.
  - A `helm test` connection hook.
- Add `values.schema.json` with render-time validation for replica counts,
[...]

PR #3187, opened 2026-09-02
fix(docker): make the Server startup timeout configurable

### LINKED ISSUE #3186: [Bug] Docker entrypoint hardcodes start-hugegraph.sh -t 120; make the Server startup timeout configurable
### Bug Type (问题类型)

server status (启动/运行异常)

### Before submit

### Environment (环境信息)

- Server Version: master (98477f0f), the `hugegraph/server:latest` image built from it; the line is unchanged since #2952 (0154c064, 2026-03-20)
- Backend: hstore, 3 PD x 3 Store x 3 Server on Kubernetes (Helm chart from #3132)
- OS: 4-node Kind (k8s 1.35) on a 15.7 GB host, and a 4-node kubeadm cluster with 7.6 GB workers
- Data Size: n/a, the failure is at startup

**Expected.** The time a Server container is allowed to spend starting should be something the deployment can set. Under Kubernetes the orchestrator already owns that budget through the startup probe, so the container should not enforce a shorter one of its own, or should at least let the budget be raised.

**Actual.** `hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh` line 220 runs

```bash
```

`-t` sets `SERVER_STARTUP_TIMEOUT_S` in `start-hugegraph.sh` (script default 30), and when it expires the script prints `Starting HugeGraphServer failed` and exits 1, which ends the container. The 120 is a literal in the entrypoint; no environment variable or argument reaches it.

Measured on 2026-09-01: 2 of 6 Server starts across two independent installs terminated this way, each after exactly 24 five-second polls, exit code 1, then recovered on the next restart. The dying container's log both times:

```
Connecting to HugeGraphServer (http://0.0.0.0:8080/graphs).Using G1GC ...
........................Starting HugeGraphServer failed
```

The Helm chart gives the Server a 450-second startup budget (`startupProbe` failureThreshold 90 x periodSeconds 5), sized for `wait-storage.sh` plus JVM start on a loaded node. That budget is unreachable: the process is gone at 120 seconds, 27 percent of it, and no probe setting can extend a process that terminates itself first.

Server startup: the entrypoint's 120s timeout fires long before the chart's 450s probe budget

The cost today is a slow first start, not data: the container restarts and the second attempt usually succeeds because `init-store` has already written its marker. On a slow or contended host it can take several restarts, and every one of them looks like a crash loop to whoever is watching the rollout.

### Proposal

Read the timeout from the environment, defaulting to the current value so nothing changes for anyone who does not set it:

```bash
./bin/start-hugegraph.sh -j "${JAVA_OPTS:-}" -t "${HG_SERVER_STARTUP_TIMEOUT_S:-120}"
```

and document `HG_SERVER_STARTUP_TIMEOUT_S` next to the other `HG_SERVER_*` variables in `docker/README.md`. The Helm chart would then set it from the same value that sizes the startup probe, so the two timers cannot disagree. Happy to send the PR.

### Related

- #3043 covered the container not restarting after a JVM crash; this is the opposite case, the container ending a JVM that is still starting.
- #2952 introduced the `-t 120` line while moving the compose setup to bridge networking.
- The Helm chart in #3132 documents this as a limitation until the timeout is configurable.

bitflicker64:
## Purpose of the PR

- close #3186

`docker-entrypoint.sh` runs `start-hugegraph.sh` with a literal `-t 120`, so a Server that needs longer than 120 seconds to answer on its REST port is ended by its own container. Under an orchestrator the startup budget belongs to the startup probe, and a probe cannot extend a process that terminates itself first: with the 450-second probe budget the Helm chart in #3132 configures, the process is gone at 120 seconds. Measured on 2026-09-01, 2 of 6 Server starts across two independent installs died this way and recovered only on container restart (log excerpts and the measurement setup are in the issue).

Server startup: today the hardcoded 120 s timer ends the container inside the probe 450 s budget; with this PR one value sizes both timers

## Main Changes

- `docker-entrypoint.sh` reads the timeout from `HG_SERVER_STARTUP_TIMEOUT_S`, defaulting to the current 120, so nothing changes for deployments that do not set it. Both Server images share this entrypoint, so one change covers the RocksDB and HStore images.
- The value is validated as a positive whole number before init-store runs. A bad value fails the container immediately with a clear message instead of surfacing later as an arithmetic error inside `wait_for_startup`.
- `docker-entrypoint-test.sh` now records the arguments the `start-hugegraph.sh` stub receives and asserts the default `-t 120`, an explicit `-t 450` override, and rejection of `2m` before the server would have been started.
- Documented as a new section in `hugegraph-server/hugegraph-dist/docker/README.md`. The issue proposed the environment table in `docker/README.md`, but #3149 removed that table, so the Server docker README is the current home for entrypoint variables.

## Verifying these changes

    - `bash hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh` (already wired into `server-ci.yml`) covers the default, an override, and an invalid value.

- [x]  Modify configurations

## Documentation Status

- [x]  `Doc - Done`


hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
```diff
@@ -98,6 +98,19 @@ if [[ -n "${HG_SERVER_AUTH_TOKEN_SECRET:-}" ]]; then
...
 fi
 
+# How long the entrypoint lets the server take to answer on its REST port
+# before it gives up and ends the container. An orchestrator that already
+# owns this budget through a startup probe needs to raise it, otherwise the
+# container terminates a JVM that is still starting and the probe never gets
+# to decide. Validated here so a typo fails before init-store runs, rather
+# than reaching the arithmetic in wait_for_startup.
+SERVER_STARTUP_TIMEOUT_S="${HG_SERVER_STARTUP_TIMEOUT_S:-120}"
+if [[ ! "${SERVER_STARTUP_TIMEOUT_S}" =~ ^[1-9][0-9]*$ ]]; then
```
imbajin: ⚠️ `HG_SERVER_STARTUP_TIMEOUT_S` accepts any positive integer, but `wait_for_startup` computes `stop_s=$((now_s + timeout_s))` (the shared `util.sh` implementation). With the documented input `9223372036854775807`, Bash overflows that sum to a negative value, so the loop is skipped and the container reports a startup timeout immediately instead of honoring the configured budget. Please reject values that cannot be safely added to the current epoch or use overflow-safe/saturating deadline arithmetic, and add a boundary test.

imbajin (commented):
Blocking: no. Summary: The configurable timeout is wired correctly for normal values, but its accepted input range can overflow the startup deadline calculation. Evidence: Bash arithmetic reproduces a negative deadline for 9223372036854775807; all reported GitHub checks are successful.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
```diff
@@ -98,6 +98,19 @@ if [[ -n "${HG_SERVER_AUTH_TOKEN_SECRET:-}" ]]; then
     fi
 fi
 
+# How long the entrypoint lets the server take to answer on its REST port
+# before it gives up and ends the container. An orchestrator that already
+# owns this budget through a startup probe needs to raise it, otherwise the
+# container terminates a JVM that is still starting and the probe never gets
+# to decide. Validated here so a typo fails before init-store runs, rather
+# than reaching the arithmetic in wait_for_startup.
+SERVER_STARTUP_TIMEOUT_S="${HG_SERVER_STARTUP_TIMEOUT_S:-120}"
```
bitflicker64: ⚠️ `:-` treats an explicitly empty value as unset, so `HG_SERVER_STARTUP_TIMEOUT_S=` never reaches the regex below and silently becomes 120. Driving this entrypoint with a stubbed `bin/`: empty exits 0 with `-t 120`, while `0`, `+5` and a whitespace-only value are all rejected. That contradicts the README text added in this PR (lines 161-163: a value that is not a positive whole number "stops the container at startup instead of silently falling back to the default"), and it reinstates the failure this PR exists to fix, since a deployment that believes it set 450 still dies at 120. The repo's own compose files use that interpolation style (`docker/docker-compose.yml:34`), so `HG_SERVER_STARTUP_TIMEOUT_S: ${SOME_VAR:-}` yields an empty value whenever the host variable is unset.

Dropping the colon makes an empty value fail the guard while an unset one still defaults. Verified: unset gives `-t 120`, empty exits 1, 450 gives `-t 450`, and `docker-entrypoint-test.sh` still exits 0.

```suggestion
SERVER_STARTUP_TIMEOUT_S="${HG_SERVER_STARTUP_TIMEOUT_S-120}"
```

If you would rather keep empty meaning "unset", the README sentence needs to say so instead.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
```diff
@@ -232,4 +233,25 @@ rm -f "${TEST_HOME}/docker/init_complete"
...
+start_calls_before_invalid_timeout=$(wc -l < "${TEST_HOME}/docker/start-hugegraph-args")
+if (
+    cd "${TEST_HOME}"
+    HG_SERVER_STARTUP_TIMEOUT_S=2m bash ./docker-entrypoint.sh
+); then
+    echo "invalid startup timeout unexpectedly succeeded" >&2
+    exit 1
+fi
+[[ "$(wc -l < "${TEST_HOME}/docker/start-hugegraph-args")" -eq \
+   "${start_calls_before_invalid_timeout}" ]]
```
bitflicker64: 🧹 This block pins down that the server was not started, but not the ordering stated in the comment above the guard (`Validated here so a typo fails before init-store runs`). I copied the entrypoint, moved the whole validation block from line 101 down to just above the `./bin/start-hugegraph.sh` call so it runs after all three `init-store.sh` call sites, and `docker-entrypoint-test.sh` still exited 0. The invalid run executes `init-store.sh` in that variant, and every `init-store-calls` assertion sits earlier in the file, so nothing observes it.

Snapshotting the init-store count alongside the start count closes the gap. Verified under bash 5.3: exit 0 against this PR, exit 1 against the relocated guard.

```suggestion
start_calls_before_invalid_timeout=$(wc -l < "${TEST_HOME}/docker/start-hugegraph-args")
init_calls_before_invalid_timeout=$(wc -l < "${TEST_HOME}/docker/init-store-calls")
if (
    cd "${TEST_HOME}"
    HG_SERVER_STARTUP_TIMEOUT_S=2m bash ./docker-entrypoint.sh
); then
    echo "invalid startup timeout unexpectedly succeeded" >&2
    exit 1
fi
[[ "$(wc -l < "${TEST_HOME}/docker/start-hugegraph-args")" -eq \
   "${start_calls_before_invalid_timeout}" ]]
[[ "$(wc -l < "${TEST_HOME}/docker/init-store-calls")" -eq \
   "${init_calls_before_invalid_timeout}" ]]
```

hugegraph-server/hugegraph-dist/docker/README.md
```diff
@@ -144,3 +144,20 @@ native `HEALTHCHECK` instructions. `docker ps` shows real health status:
...
+The entrypoint gives the Server a fixed budget to answer on its REST port and
+ends the container when the budget runs out. It is 120 seconds by default. Set
+`HG_SERVER_STARTUP_TIMEOUT_S` to a positive whole number of seconds to change
+it:
+
+```bash
+docker run -itd --name=graph -p 8080:8080 -e HG_SERVER_STARTUP_TIMEOUT_S=450 hugegraph/hugegraph:1.7.0
+```
+
+Raise it on slow or contended hosts, and wherever an orchestrator already owns
```
bitflicker64: 🧹 Section 6 directly above says `docker ps` shows real health status, but the health budget does not move with this variable. For the `docker run` example above, `hugegraph-server/Dockerfile:74` and `hugegraph-server/Dockerfile-hstore:76` set `--start-period=90s --interval=15s --retries=3`, so a container given 450s is marked `unhealthy` about two minutes in while the entrypoint is still legitimately waiting. The compose files replace that with their own (`docker/docker-compose.yml:41-46`: `start_period: 60s`, `interval: 10s`, `retries: 30`, so roughly 360s) and `hubble` gates on it through `depends_on: condition: service_healthy`, so a timeout raised past that budget blocks dependents rather than only mislabeling the container. Please add a sentence pointing at `--health-start-period` and the compose `start_period`.

bitflicker64 (commented):
Blocking: no. Summary: The 120s default is preserved when the variable is unset and the guard runs ahead of every config write and init-store call, but `:-` also lets an explicitly empty value fall back to 120 silently, which the README section added here states does not happen. The new test asserts that a bad value does not start the server without pinning the ordering guarantee the code comment claims. Evidence: ran `bash hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh` at 5cad9dc8217f0aa8139eb68ac1e32c6ab2af0df6 under bash 5.3 (exit 0); drove the entrypoint with a stubbed `bin/` for empty, whitespace, `0`, `+5` and `450`, where empty alone exits 0 with `-t 120`; reran the suite against an entrypoint with the validation block relocated below `init-store.sh`, which also exits 0; healthcheck values read from `hugegraph-server/Dockerfile:74`, `hugegraph-server/Dockerfile-hstore:76` and `docker/docker-compose.yml:41-46`.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
bitflicker64: Fixed in 6ca3df360. Reproduced first: with `now_s=1788438716`, `$((now_s + 9223372036854775807))` gives `-9223372035066337093`, so `[ "$now_s" -lt "$stop_s" ]` is false and the loop never probes.

The accepted range is now 1 to 86400 seconds. The pattern is bounded to five digits (`^[1-9][0-9]{0,4}$`) before the upper-bound comparison runs, so the comparison itself cannot overflow either; bash wraps a 20-digit literal silently rather than erroring, so the regex has to do that part. Boundary tests added for 86400 accepted, 86401 and 9223372036854775807 rejected.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint.sh
bitflicker64: Fixed in 6ca3df360, taking the suggestion. Confirmed the behaviour: `${v:-120}` on an empty value yields `120`, `${v-120}` yields empty and so fails the guard, and an unset variable still yields `120` either way. `docker/docker-compose.yml:34-35` uses the `${VAR:-}` style, so this was reachable.

The README sentence now says an empty value is rejected too, rather than the reverse. Empty and whitespace-only are both in the rejected set in the test.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
bitflicker64: Fixed in 6ca3df360. Your relocated-guard variant is now caught: I reproduced it by moving the block below all three `init-store.sh` call sites, and the suite fails with `startup timeout '' was rejected only after init-store ran`.

One thing worth flagging from checking it. The bare `[[ ]]` assertion style does not fail on bash 3.2, still the `/bin/bash` of macOS: `set -e` there ignores a failing `[[ ]]`, so my first run of your variant passed locally and only failed under bash 5.3. The new assertions use an explicit `|| { echo ...; exit 1; }` so they fail on both. Verified under 3.2.57 and 5.3.15.

hugegraph-server/hugegraph-dist/docker/README.md
bitflicker64: Fixed in 6ca3df360. Section 7 now says the health check keeps a budget of its own that this variable does not move: the images' `--interval=15s --start-period=90s --retries=3` marks a container `unhealthy` around 135 seconds, raised with `--health-start-period`, while the Compose files substitute `start_period: 60s`, `interval: 10s`, `retries: 30`, roughly 360 seconds, which is what `depends_on: condition: service_healthy` gates Hubble on.

hugegraph-server/hugegraph-dist/docker/README.md
```diff
@@ -144,3 +144,30 @@ native `HEALTHCHECK` instructions. `docker ps` shows real health status:
 | `hugegraph/hugegraph-store` | `GET /v1/health` on port 8520 |
 
 The entrypoints supervise the Java process directly — when Java exits, the container exits. If started with a restart policy (the provided compose files use `restart: unless-stopped`), Docker will bring it back automatically. The old cron-based monitor (`-m true`) is for VM/bare-metal deployments only and is not used in Docker images.
+
+## 7. Server Startup Timeout
+
+The entrypoint gives the Server a fixed budget to answer on its REST port and
+ends the container when the budget runs out. It is 120 seconds by default. Set
+`HG_SERVER_STARTUP_TIMEOUT_S` to a whole number of seconds between 1 and 86400
```
imbajin: ⚠️ The new variable is not propagated by any provided Compose topology. `docker/docker-compose.yml`, `docker/docker-compose-hstore.yml`, and `docker/docker-compose-3pd-3store-3server.yml` do not declare `HG_SERVER_STARTUP_TIMEOUT_S` in the Server environment. Reproduction on this head: `HG_SERVER_STARTUP_TIMEOUT_S=450 docker compose -f docker/docker-compose.yml config --format json` emits no such key, so the container still uses the entrypoint default `120` and the healthcheck budget cannot be aligned. Please add the variable to each Server environment/anchor with unset-only default semantics and cover it in the Compose render test.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
```diff
@@ -232,4 +233,60 @@ rm -f "${TEST_HOME}/docker/init_complete"
...
 
+last_start_args() { tail -n 1 "${TEST_HOME}/docker/start-hugegraph-args"; }
+
+[[ "$(last_start_args)" == *"-t 120"* ]]
+
+(
+    cd "${TEST_HOME}"
+    HG_SERVER_STARTUP_TIMEOUT_S=450 bash ./docker-entrypoint.sh
+)
+[[ "$(last_start_args)" == *"-t 450"* ]]
```
imbajin: ⚠️ The positive timeout checks are false-green on the supported macOS Bash 3.2.57. With `set -euo pipefail`, a failing standalone `[[ ... ]]` returns nonzero but the script continues (`set -e; [[ 1 == 2 ]]; echo survived` prints `survived`), so a broken override can still reach PASS. Please use explicit `if ! ...; then ... exit 1; fi` assertions for the default and override cases and the final unset case.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
```diff
@@ -232,4 +233,60 @@ rm -f "${TEST_HOME}/docker/init_complete"
 )
 grep -Fqx -- '-n' "${TEST_HOME}/docker/init-store-password"
 
+last_start_args() { tail -n 1 "${TEST_HOME}/docker/start-hugegraph-args"; }
+
+[[ "$(last_start_args)" == *"-t 120"* ]]
```
imbajin: ⚠️ `last_start_args` records the whole `$*`, and `*"-t 120"*` / `*"-t 450"*` only checks a substring. A launcher receiving `-t 1200` or `-t 4500` would pass these assertions; the stub never invokes the real launcher. Please record `"$@"` as separate arguments or parse the positional arguments and assert the exact `-t` value.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
```diff
@@ -232,4 +233,60 @@ rm -f "${TEST_HOME}/docker/init_complete"
...
+        echo "startup timeout '${invalid_timeout}' was rejected only after" \
+             "init-store ran" >&2
+        exit 1
+    }
+done
+
+# An unset variable still keeps the historical default.
+(
+    cd "${TEST_HOME}"
+    bash ./docker-entrypoint.sh
```
imbajin: ⚠️ This block claims to exercise an unset variable but only starts a child shell; it does not unset an inherited `HG_SERVER_STARTUP_TIMEOUT_S`. Running the test with `export HG_SERVER_STARTUP_TIMEOUT_S=450` makes this call exercise `450` instead of the default `120`, and the bare assertion can still report success on Bash 3.2. Please invoke it with `env -u HG_SERVER_STARTUP_TIMEOUT_S` or unset the variable inside the subshell, while keeping an explicit failing assertion.

imbajin (commented):
Blocking: no. Summary: The configurable timeout is not propagated through the provided Compose topologies, and the new shell tests can report success for incorrect values on Bash 3.2. Evidence: reproduced with Docker Compose config rendering and /bin/bash 3.2.57; all findings are in the changed Docker documentation/test paths.

hugegraph-server/hugegraph-dist/docker/README.md
bitflicker64: Fixed in 0ba30a1e7. Confirmed the reproduction first: `docker compose -f docker/docker-compose.yml config` emitted no such key.

All three topologies now carry `HG_SERVER_STARTUP_TIMEOUT_S: ${HG_SERVER_STARTUP_TIMEOUT_S-120}`, unset-only as you asked, so an absent host value renders 120 while an explicitly empty one reaches the entrypoint and is rejected there instead of silently becoming the default. The HA file gets it through the `x-server-environment` anchor, so all three Servers share it.

`test-compose.sh render` now asserts the default and a 450 override on every Server service, and `compose_auth` strips the variable so the baseline render cannot be coloured by the developer's environment. Verified the assertion fails when the key is removed from a Compose file.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
bitflicker64: Fixed in 0ba30a1e7. The positive cases now use an `assert_start_timeout` helper that exits explicitly and prints the value it saw, so bash 3.2 no longer reaches PASS through a failing `[[ ]]`.

Verified on 3.2.57: changing the entrypoint default to 999 now reports `expected start-hugegraph.sh -t 120, got -t 999` instead of passing. The suite passes on both 3.2.57 and 5.3.15.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
bitflicker64: Fixed in 0ba30a1e7. The stub now writes `printf '%s\n' "$@" > ./docker/start-hugegraph-argv`, one argument per line, and `last_start_timeout` walks that vector to return the exact argument following `-t`. Assertions compare it with `!=` rather than a substring.

Verified with your case: making the entrypoint pass `-t 1200` now reports `expected start-hugegraph.sh -t 120, got -t 1200`, where the old substring check passed. Call counting moved to a separate `start-hugegraph-calls` log so the argv file can be overwritten per call.

hugegraph-server/hugegraph-dist/docker/docker-entrypoint-test.sh
bitflicker64: Fixed in 0ba30a1e7. Both unset cases now run under `env -u HG_SERVER_STARTUP_TIMEOUT_S`, and the assertion is the explicit helper described above rather than a bare `[[ ]]`.

Verified: with `HG_SERVER_STARTUP_TIMEOUT_S=450` exported, the suite still passes, and reverting the `env -u` reports `expected start-hugegraph.sh -t 120, got -t 450` on bash 3.2.

docker/test-compose.sh
```diff
@@ -263,6 +291,23 @@ run_render() {
...
+    render_with_timeout "${RENDER_DIR}/standalone-timeout.json" 450 \
+                        -f "${DOCKER_DIR}/docker-compose.yml"
+    render_with_timeout "${RENDER_DIR}/hstore-timeout.json" 450 \
+                        -f "${DOCKER_DIR}/docker-compose-hstore.yml"
+    render_with_timeout "${RENDER_DIR}/ha-timeout.json" 450 \
+                        -f "${DOCKER_DIR}/docker-compose-3pd-3store-3server.yml"
+    assert_startup_timeout "${RENDER_DIR}/standalone-timeout.json" 450 server
+    assert_startup_timeout "${RENDER_DIR}/hstore-timeout.json" 450 server
+    assert_startup_timeout "${RENDER_DIR}/ha-timeout.json" 450 \
+                           server0 server1 server2
```
bitflicker64, line 319: ⚠️ These assertions cannot see the `-` vs `:-` choice the rest of the change rests on.

Both cases they cover are satisfied by either spelling: unset renders `120` and `450` renders `450` under `${VAR-120}` and `${VAR:-120}` alike. The difference only shows for an explicitly empty host value, the case `docker-entrypoint.sh` was changed to reject and that `docker-compose.yml:37-39` and README section 7 ("an empty value included, stops the container at startup") promise reaches the container.

Verified on this head: rewriting all three Compose files to `${HG_SERVER_STARTUP_TIMEOUT_S:-120}` still gives `Compose render contracts passed`, while the empty host value silently renders `"120"` instead of `""`. With the block below the suite passes as it stands and fails that variant with `expected server HG_SERVER_STARTUP_TIMEOUT_S= in .../standalone-empty.json`.

```suggestion
    assert_startup_timeout "${RENDER_DIR}/ha-timeout.json" 450 \
                           server0 server1 server2
    render_with_timeout "${RENDER_DIR}/standalone-empty.json" "" \
                        -f "${DOCKER_DIR}/docker-compose.yml"
    render_with_timeout "${RENDER_DIR}/hstore-empty.json" "" \
                        -f "${DOCKER_DIR}/docker-compose-hstore.yml"
    render_with_timeout "${RENDER_DIR}/ha-empty.json" "" \
                        -f "${DOCKER_DIR}/docker-compose-3pd-3store-3server.yml"
    assert_startup_timeout "${RENDER_DIR}/standalone-empty.json" "" server
    assert_startup_timeout "${RENDER_DIR}/hstore-empty.json" "" server
    assert_startup_timeout "${RENDER_DIR}/ha-empty.json" "" \
                           server0 server1 server2
```

docker/docker-compose.yml
```diff
@@ -34,6 +34,10 @@ services:
       PASSWORD: ${HUGEGRAPH_ADMIN_PASSWORD:-}
       HG_SERVER_AUTH_TOKEN_SECRET: ${HUGEGRAPH_AUTH_TOKEN_SECRET:-}
       HG_SERVER_MIN_FREE_MEMORY: "0"
+      # Unset-only default: a host value overrides it, an absent one
+      # renders the entrypoint default. ":-" would turn an empty host
+      # value into 120 silently, which the entrypoint rejects on purpose.
+      HG_SERVER_STARTUP_TIMEOUT_S: ${HG_SERVER_STARTUP_TIMEOUT_S-120}
```
bitflicker64, line 40: 🧹 Under Compose the alignment this knob buys only works downward, which is worth one word in the comment at `docker/test-compose.sh:296-297`.

Lowering the container budget to meet the health budget needs no file edit. Raising it past the health budget cannot be done from the host at all: `interval: 10s`, `retries: 30`, `start_period: 60s` (lines 45-50 here, `docker-compose-hstore.yml:100-105`, and the `x-server-common` anchor at `docker-compose-3pd-3store-3server.yml:86-91`) are literals, so the container is `unhealthy` around 360 seconds. `docker/README.md` starts every topology with `up -d --wait` (lines 76, 122, 175) and `hubble` gates on `service_healthy`, so a 450 set through Compose still gives up near 360 seconds. Section 7 says as much for the reader; the test comment reads as alignment in both directions.

Interpolating the Server health budget, say `start_period: ${HG_SERVER_HEALTH_START_PERIOD:-60s}`, would make it symmetric, but those lines are untouched here, so a follow-up issue is the right home. Just narrowing the comment is enough for this PR.

docker/test-compose.sh
```diff
@@ -43,6 +44,33 @@ render() {
     compose_auth "$@" config --format json > "${output}"
 }
 
+# Same render with an explicit host value, to prove the variable reaches the
+# Server environment rather than only defaulting there.
+render_with_timeout() {
+    local output="$1" timeout="$2"
+    shift 2
+    env HG_SERVER_STARTUP_TIMEOUT_S="${timeout}" \
```
bitflicker64: 🧹 `render_with_timeout` re-lists `compose_auth`'s four environment variables instead of reusing it, so the two drift apart on the next variable added.

Lines 52-56 are a verbatim copy of lines 33-37 with `env -u` swapped for a value. Anything added to `compose_auth` has to be added here too or `render` and `render_with_timeout` stop rendering the same stack. Not hypothetical: your own note on this PR says #3189 makes the Compose files require `HG_PD_AUTH_SECRET_KEY`, at which point `render_with_timeout` fails on the two HStore topologies unless it is edited as well.

One definition removes the second place to remember. Verified with exactly this text in place: `bash docker/test-compose.sh render` still prints `Compose render contracts passed`.

```bash
STARTUP_TIMEOUT_ENV=(-u HG_SERVER_STARTUP_TIMEOUT_S)

compose_auth() {
    env "${STARTUP_TIMEOUT_ENV[@]}" \
        HUGEGRAPH_VERSION="${VERSION}" \
        HUBBLE_IMAGE="${RENDER_HUBBLE_IMAGE}" \
        HUGEGRAPH_ADMIN_PASSWORD="${PASSWORD}" \
        HUGEGRAPH_AUTH_TOKEN_SECRET="${SECRET}" \
        docker compose "$@"
}

render_with_timeout() {
    local output="$1" timeout="$2"
    shift 2
    (
        STARTUP_TIMEOUT_ENV=(HG_SERVER_STARTUP_TIMEOUT_S="${timeout}")
        compose_auth "$@" config --format json
    ) > "${output}"
}
```

bitflicker64 (commented):
Blocking: no. Summary: nothing here should hold up the merge; the overflow, empty-value, argv-substring and Bash 3.2 threads are all genuinely closed at this head. One test gap worth fixing: the new Compose render assertions pass just as happily under `:-` as under `-`, so they cannot protect the empty-value semantics the entrypoint and README depend on. Four smaller notes follow. Evidence: at 0ba30a1e, `docker-entrypoint-test.sh` and `docker/test-compose.sh render` both exit 0 locally (Docker Compose v5.1.2, GNU sed on PATH), and all 24 reported checks on this head are successful.

docker/docker-compose-hstore.yml
```diff
@@ -92,6 +92,7 @@ services:
       HG_SERVER_REST_URL: http://server:8080
       HG_SERVER_MIN_FREE_MEMORY: "0"
       HG_SERVER_INIT_STORE_ENABLED: "false"
+      HG_SERVER_STARTUP_TIMEOUT_S: ${HG_SERVER_STARTUP_TIMEOUT_S-120}
```
bitflicker64, line 97: 🧹 The reason for the missing colon lives in only one of the three files.

`docker-compose.yml:37-39` explains it. Here, and at `docker-compose-3pd-3store-3server.yml:71`, a bare `${HG_SERVER_STARTUP_TIMEOUT_S-120}` sits directly above `HG_SERVER_AUTH_TOKEN_SECRET: ${HUGEGRAPH_AUTH_TOKEN_SECRET:-}`, so to a later editor it reads as a dropped colon next to a neighbour that has one. Since the render suite cannot currently catch that edit, the comment is the cheap half of the guard.

```suggestion
      # Unset-only default: ":-" would turn an empty host value into 120
      # silently, which the entrypoint rejects on purpose.
      HG_SERVER_STARTUP_TIMEOUT_S: ${HG_SERVER_STARTUP_TIMEOUT_S-120}
```

The HA anchor at line 71 wants the same two lines at its two-space indent.

hugegraph-server/hugegraph-dist/docker/README.md
```diff
@@ -144,3 +144,35 @@ native `HEALTHCHECK` instructions. `docker ps` shows real health status:
...
+to change it:
+
+```bash
+docker run -itd --name=graph -p 8080:8080 -e HG_SERVER_STARTUP_TIMEOUT_S=450 hugegraph/hugegraph:1.7.0
+```
+
+The Compose topologies under `docker/` pass the variable through to every
+Server with an unset-only default, so `HG_SERVER_STARTUP_TIMEOUT_S=450 docker
+compose up -d` raises the budget without editing a Compose file, while leaving
+it unset renders the same 120.
```
bitflicker64: 🧹 The variable is a Compose knob now, but `docker/README.md`, the file Compose users actually read, never mentions it.

That document walks a user through `.env` (line 26), `HUGEGRAPH_ADMIN_PASSWORD` (42), `HUGEGRAPH_AUTH_TOKEN_SECRET` (42), `HUBBLE_PUBLISH_HOST` (56), `HUGEGRAPH_VERSION` (231) and `HUBBLE_IMAGE` (238), and carries the `up -d --wait` commands this timeout governs. It is not in this PR's diff, so a reader who never opens the Server docker README has no way to learn the knob exists.

Please add a line to `docker/README.md` beside the other host variables, pointing here for the accepted range and the health-check note.

docker/test-compose.sh
bitflicker64: Fixed in f2c6a1ff4, taking the suggestion. Confirmed the blind spot first: rewriting all three Compose files to `${HG_SERVER_STARTUP_TIMEOUT_S:-120}` left the whole suite green, and an empty host value rendered `"120"` there against `""` here.

All three topologies are now rendered with an empty value and asserted to carry one. Verified the assertion fails the `:-` rewrite with exactly `expected server HG_SERVER_STARTUP_TIMEOUT_S= in .../standalone-empty.json`.

docker/docker-compose.yml
bitflicker64: Fixed in f2c6a1ff4. The comment now says the knob aligns the two budgets downward only: lowering the container budget to meet the health budget needs no file edit, raising it past that budget does, because the health values are literals giving every Server about 360 seconds and the documented `up -d --wait` gives up there.

Checked the rendered values rather than the source lines: `interval: 10s`, `retries: 30`, `start_period: 60s` on the Server of all three topologies. Section 7 keeps the reader-facing version, and `docker/README.md` now carries it too. Agreed that interpolating the health budget belongs in a follow-up issue rather than here.

docker/test-compose.sh
bitflicker64: Fixed in f2c6a1ff4, taking the suggestion. `STARTUP_TIMEOUT_ENV` is the only thing the two renders differ by, and `render_with_timeout` rebinds it in a subshell.

Verified the drift it prevents: adding a variable to `compose_auth` alone and asserting on the rendered Server environment now reports `true` for the baseline, the override and the empty render alike. Under the duplicated version the latter two would have missed it, which is the #3189 case you describe.

docker/docker-compose-hstore.yml
bitflicker64: Fixed in f2c6a1ff4. Both files now carry the same two-line note as the standalone file, at their own indent, so the missing colon does not read as a typo beside `HG_SERVER_AUTH_TOKEN_SECRET: ${HUGEGRAPH_AUTH_TOKEN_SECRET:-}`.

hugegraph-server/hugegraph-dist/docker/README.md
bitflicker64: Fixed in f2c6a1ff4. `docker/README.md` gets a "Server startup timeout" section beside the other host variables, with the `up -d` example, the unset and empty semantics, the roughly 360 second health budget this variable does not move, and a link to section 7 for the accepted range and the health check details.

imbajin (approved):

PR #3193, opened 2026-09-05
fix(server): make query result batch boundaries explicit

### LINKED ISSUE #3190: [Improve] Make query-result batch boundaries explicit
## Background

PR #3140 introduced Server-side input-ID ordering for HStore query results. Issue #3180 later exposed an interaction between that ordering path and flattened index queries: InputOrderIterator can probe the next query while detecting the current query boundary, and QueryList can update the shared origin ConditionQuery resultsFilter during that probe.

PR #3182, reviewed at head acbee16753e4bf0aba35e114e23470aab0417ccd, is intended as a bounded correctness hotfix. It applies index-candidate filtering before input-order restoration and protects cache paths that could otherwise bypass that filtering. This fixes the reported failure without redesigning the iterator boundary itself.

A follow-up refactor is needed so correctness no longer depends on the buffering and side effects of ordinary Iterator.hasNext() calls.

### 1. Boundary detection can activate the next query

InputOrderIterator.fetchBatch() currently uses the following sequence:

~~~java
long queryVersion = this.queryResults.queryVersion();
List<Query> queries = this.queryResults.currentQueries();

do {
    results.add(this.origin.next());
} while (this.origin.hasNext() &&
         queryVersion == this.queryResults.queryVersion());
~~~

The origin.hasNext() call is not a read-only boundary check. It can advance FlatMapperIterator or QueryTrackingIterator into the next flattened query. QueryList.IndexQuery may then update the shared origin query's resultsFilter before InputOrderIterator observes that queryVersion changed.

PR #3182 moves filtering into a safer position, but the underlying contract remains implicit: the current element must be filtered and buffered before another hasNext() call can cross the query boundary.

### 2. Input-order restoration is installed from the first active segment

QueryResults.keepInputOrderIfNeeded() decides whether to create InputOrderIterator from currentQueries after the first successful origin.hasNext() call.

This can miss a later segment whose ordering requirement differs from the first segment:

~~~text
segment A: mustSortByInput = false, backend = [1]
segment B: mustSortByInput = true,  input = [3, 2], backend = [2, 3]

expected: [1, 3, 2]
possible current result: [1, 2, 3]
~~~

Input-order requirements must therefore be evaluated per query batch, not once for the flattened result stream.

### 3. Query context is represented by mutable stream-wide state

queryVersion, currentQueries, resultsFilter, input IDs, and mustSortByInput collectively describe the active query segment, but they are stored or observed through different objects. Mapper, filter, flat-map, paging, and cache paths do not share one explicit boundary contract.

This makes it difficult to prove that each result is parsed, TTL-checked, condition-filtered, and ordered with the context that produced it.

### 4. Filtering stages have implicit and asymmetric ownership

PR #3182 separates index-candidate matching from invalid-record filtering, but the ownership of diagnostics and cleanup remains easy to misread.

For ordinary queries, filterUnmatchedRecords() deliberately lets hidden records and records with deleting labels pass through without running rightResultFromIndexQuery(). The downstream filterInvalidRecords() then calls warnLeftRecord() and applies visibility rules. An undefined label is named ~undefined and is therefore hidden, so with the default showHidden=false the later warning inside filterUnmatchedRecords() is unreachable. It is reachable only for callers that enable showHidden.

Internal callers may use filterUnmatchedRecords() without the downstream invalid-record filter. Their access, warning, and left-index cleanup behavior therefore differs from the public query path by design.

The refactor must assign one explicit owner to each operation and preserve their order:

- residual index-condition matching;
- hidden, deleting-label, and undefined-label visibility;
- left-record diagnostics;
- asynchronous left-index cleanup.

A query-boundary change must not silently suppress, duplicate, or move these side effects to internal queries that previously bypassed them.

### 5. Cache eligibility depends on when optimization metadata becomes final

The vertex and edge cache paths currently evaluate queryNeedsPostFilter() at different points.

The vertex path checks before the backend call. This is valid because optimized vertex fallbacks arrive as an IdQuery whose origin ConditionQuery already carries PRIMARY_KEY or INDEX, and querying that IdQuery does not run optimizeQueries again.

The edge path must check both before and after super.queryEdgesFromBackend(query), because backend processing can still promote optimized() through the origin-query chain. Caching the result after that promotion could otherwise bypass required condition filtering.

This timing contract is live behavior, not a defensive implementation detail. The new batch model must either remove the timing asymmetry by carrying finalized cache/filter metadata or represent the two decision points explicitly and test them.

## Goal

Introduce an explicit query-batch or query-segment boundary for query result iteration.

The implementation does not have to use a class named QueryBatch, but it must establish these invariants:

- A batch carries the query metadata required to process its results, including its query or query snapshot, input IDs, input-order requirement, and filter context.
- Each result is processed with the context of the batch that produced it.
- Parse, TTL filtering, condition filtering, and input-order restoration complete at a batch-aware boundary before that context is discarded.
- Checking whether the current batch has more results does not activate the next query or fetch the next backend page.
- Mixed batches can independently choose whether input-order restoration is required.
- Ordinary Iterator.hasNext() buffering behavior is not part of the correctness contract.

## Suggested direction

Prefer an explicit batch abstraction exposed by QueryResults and produced by QueryList/PageEntryIterator. InputOrderIterator should consume one batch directly instead of inferring boundaries by comparing queryVersion after calling origin.hasNext().

A smaller boundary-aware iterator API, such as hasNextInCurrentQuery() plus an explicit operation that advances to the next query, may also be considered. However, the boundary must propagate through QueryTrackingIterator, MapperIterator, FilterIterator, FlatMapperIterator, page iteration, and any other wrapper that can advance the origin. Adding such a method only to InputOrderIterator would leave the same implicit coupling elsewhere.

## Scope

The expected implementation area includes:

- QueryResults and InputOrderIterator
- QueryList and QueryTrackingIterator
- PageEntryIterator and BatchIdHolder paths
- Mapper, filter, and flat-map wrappers that can cross a query boundary
- GraphTransaction processing order
- Cache behavior when a query still requires condition filtering

After the explicit boundary is in place, reassess whether queryVersion/currentQueries and the post-filter cache guards introduced by #3182 can be removed or simplified. Removal is not a requirement unless equivalent correctness and cache behavior are demonstrated.

## Non-goals

- Do not redesign the HStore partition merge, ORDER_BY_KEY protocol, or physical-key cursor introduced by #3140.
- Do not change the public Gremlin, REST, or graph-query result semantics.
- Do not introduce a request-level fallback between old and new query execution paths.
- Do not combine unrelated cache-policy or index-planner changes into this refactor.

## Acceptance criteria

Add focused RED tests before replacing the current boundary behavior.

- [ ] Two flattened queries with different filters retain the correct results for both batches; activating query B cannot cause an element from query A to be evaluated with filter B.
- [ ] Mixed ordering requirements work in both directions, including a first batch with mustSortByInput=false followed by a batch with mustSortByInput=true.
- [ ] Empty intermediate batches do not leak filter or ordering state into adjacent batches.
- [ ] Mapper results that are null or expand one backend entry into multiple elements preserve the originating batch context.
- [ ] Multiple BatchIdHolder batches are processed independently.
- [ ] The final element of a batch still receives TTL and condition filtering before the next batch becomes active.
- [ ] Exhausting or inspecting the current batch does not fetch or prepare the next backend page; tests assert backend/page fetch count, not only consumed element count.
- [ ] Existing #3140 range-index ordering, limit, offset, paging, and cursor behavior remains unchanged.
- [ ] The vertex and edge regressions from #3180/#3182 pass on the applicable memory, RocksDB, and HStore test profiles.
- [ ] Warm-cache PRIMARY_KEY queries, PRIMARY_KEY queries with residual conditions, and vertex/edge INDEX + LABEL queries remain correct.
- [ ] Residual-condition matching, invalid-record visibility, left-record diagnostics, and asynchronous left-index cleanup each have an explicit owner. Ordinary, internal, showHidden, and showDeleting query paths preserve their current warning, visibility, and cleanup behavior without duplicate side effects.
- [ ] Vertex and edge cache eligibility remains correct across query optimization. Tests cover the vertex pre-backend decision and the edge post-backend re-check, or prove that finalized batch metadata makes the asymmetry unnecessary.
- [ ] Close and exception propagation still release the active and remaining iterator/page resources exactly once.

## References

- #3180
- #3140
- #3182
- Filtering and diagnostic ownership: https://github.com/apache/hugegraph/pull/3182#discussion_r3923381651
- Vertex/edge cache decision timing: https://github.com/apache/hugegraph/pull/3182#discussion_r3923381662

I will continue to follow up on this issue.

contrueCT:
## Purpose of the PR

Closes #3190.

`QueryResults` currently detects a query boundary by probing `hasNext()` and then checking whether the active query changed. That probe can activate the next index query or backend page before the current batch finishes processing. Ordering is also selected from the first active segment, so an unsorted batch `[1]` followed by a batch with input IDs `[3, 2]` can incorrectly produce `[1, 2, 3]`.

This change gives each batch its own results and captured processing context. Parsing, TTL checks, residual filtering and input-order restoration finish within that batch before the caller consumes a flattened stream.

## Main Changes

- Introduce `QueryBatch` and `QueryResultContext` to carry the input IDs, ordering decision, visibility flags and effective results filter with the results that produced them.
- Make `QueryResults`, `QueryList` and `PageEntryIterator` compose explicit batches. Remove `queryVersion`/`currentQueries` boundary inference and shared results-filter propagation.
- Apply vertex and edge processing within each batch, preserving public/internal visibility and index-cleanup ownership. Keep native scans on their original single-batch path so backend page tokens and capacity checks remain intact.
- Cache candidates before residual matching and reapply the current batch context on hits. Removing the old post-filter cache exclusion intentionally makes more query shapes cacheable, including indexed edge lookups. Cache capacity limits remain unchanged, but workloads with many distinct predicate values may increase cache-key churn; no cache-performance improvement is claimed. Preserve mixed hit/miss input ordering, the original edge-cache request key, off-heap serialization, TASK/SERVER query types and the RamTable fast path.
- Share the remaining batch cursor between element iteration, mapping and materialization, preserving prefetched results and page metadata. Remove the unused `QueryResults.queries()` diagnostics list; query identity remains in each batch context.
- Preserve the existing holder boundary when a backend page has no raw records. Keep following the cursor when an existing batch is emptied by parsing or TTL filtering, and retain negative edge-cache entries. This avoids scanning stale label-index pages during HStore schema cleanup.
- Bypass the vertex cache for paged backend queries so primary-key optimization retains `PageState`. Give each empty page an independent batch cursor, preventing one consumer from exhausting another consumer's cursor.
- Synchronize reflection filtering with the renamed transaction helpers and the cache overrides. Add lifecycle, edge-cache boundary, and index-cleanup side-effect regressions through the actual query-processing stages.
- Map and filter elements without creating a singleton collection/iterator per element. Skip the TTL wrapper for TTL-capable stores, copy edge-cache group indexes only on writes, and use an empty batch label when the batch key equals the outer cache key. Distinct batch keys remain explicit and retain the existing candidate/batch limits.
- Make batch wrappers close idempotently and preserve the primary exception when cleanup also fails. Native iterators that close themselves still rely on their existing backend close contract.

Explicit query batch boundaries: finish processing A with its own filter and ordering context before activating B.

The flattened consumer can advance after A is exhausted; an iterator operating inside A never probes B to discover A's boundary. HStore partition merging, `ORDER_BY_KEY` and physical-key cursors are outside this change.

Multi-branch edge queries activate branches lazily. Errors raised while preparing a branch (including query optimization and index preparation) or starting its backend read can therefore surface from `hasNext()`/`next()` rather than the initial query call; earlier branches may already have yielded results. Validation reached while flattening the request still runs immediately, and the single-branch path retains immediate branch activation. This avoids preparing sibling branches before the current batch has finished.

## Verifying these changes

- [x] Already covered by existing tests, including `VertexCoreTest`, `EdgeCoreTest` and cache regressions.

Three focused tests first failed on the old implementation: mixed ordering returned `[1, 2, 3]` instead of `[1, 3, 2]`, and the next-query and next-page fetch counters reached 2 when only 1 batch should have been activated. All now pass. Additional regressions fail on the previous PR head for materialization/mapping after `hasNext()` and duplicate/accumulating query diagnostics; the shared cursor fixed those regressions, and the unused diagnostics API has since been removed. The paging regression traverses 10,000 real `PageEntryIterator` pages and checks each batch context, results and page activation count.

Verification on JDK 17 and Maven 3.9.16:

| Scope | Snapshot | Selected | Skipped | Failures / errors |
| --- | --- | ---: | ---: | ---: |
| Diagnostics cleanup: seven focused unit-test classes | `00e365db` | 80 | 0 | 0 / 0 |
| Allocation/cache follow-up: seven focused unit-test classes | `9ad4a27e` | 80 | 0 | 0 / 0 |
| Review cleanup: QueryResultsTest, QueryListTest, GraphTransactionTest | `cf403002` | 37 | 0 | 0 / 0 |
| Seven focused unit-test classes | `05186a0e` | 78 | 0 | 0 / 0 |
| RocksDB: primary-key/paging cases and `QueryListTest` | `17788a4a` | 23 | 0 | 0 / 0 |
| HStore: primary-key paging with cold/warm cache, filtered-empty and missing results | `17788a4a` | 1 | 0 | 0 / 0 |
| Memory: `VertexCoreTest`, `EdgeCoreTest` | `c694bb0a` | 436 | 69 | 0 / 0 |
| RocksDB: `VertexCoreTest`, `EdgeCoreTest` | `c694bb0a` | 436 | 29 | 0 / 0 |
| HStore: search/joint-index, input-order and paging regressions | `c694bb0a` | 26 | 0 | 0 / 0 |

Both new review regressions failed on `c694bb0a`: primary-key paging raised `Invalid PageState 'null'`, and interleaved empty-page consumption exposed the shared cursor. They pass with this follow-up. The primary-key test covers cold and warm caches, residual filtering to an empty result, a missing primary key, and terminal page metadata.

The latest supplement adds 10 tests. Seven isolated fault-injection variants are rejected by the new tests: omitted holder/source closing, a changed cache capacity boundary, a discarded oversized-batch tail, and incorrect cleanup decisions for invisible, mismatching, or matching-with-stale-index records. The unchanged implementation passes all selected diagnostic tests. Reflection-filter entries were checked against 15 actual declared methods across the transaction and cache classes.

Coverage includes raw-empty backend pages versus filtered-empty batches, missing and expired vertices through cache materialization, empty batches/holders, null and expanded mapper results, mixed ordering in both directions, real page-fetch counters and cursor metadata, limits, cross-batch materialization capacity, materialization after partial consumption, page metadata after closing the source, early close and suppressed exceptions, residual filtering on warm vertex caches, and an off-heap edge-cache hit with zero backend reads. The HStore selection also passed on the pre-change baseline using the same isolated PD/Store setup.

At `c694bb0a`, a separate four-method HStore reproduction exercises schema cleanup after creating 10,000 edges per label. Before the empty-page fix, three subsequent tests timed out in setup. With the fix, cleanup completes and the sequence matches baseline `36811483`: three tests pass, while `testQueryOutEdgesOfVertexBySortkeyWithMoreFieldsInPage` fails in its body with `Cardinality from code 0`. This existing HStore serialization error remains unresolved; the 26-test selection above is not a full HStore-suite pass.

<details>
<summary>Reproduction commands</summary>

```bash
mvn test -pl hugegraph-server/hugegraph-test -am -P unit-test \
  -Dtest=QueryResultsTest,QueryListTest,GraphTransactionTest,CachedGraphTransactionTest,IdHolderTest,QueryTest,PageStateTest \
  -DfailIfNoTests=false -Drat.skip=true

# Run once with memory, then with rocksdb
mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,memory \
  -Dtest=VertexCoreTest,EdgeCoreTest -DfailIfNoTests=false -Drat.skip=true

# Requires an isolated, initialized PD/Store and an HStore properties file
mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,hstore \
  -Dbackend=hstore -Dconfig_path=/path/to/hstore-test.properties \
  '-Dtest=VertexCoreTest#testQueryByJointIndexesWithSearch*+testQueryByTextContainsPropertyOrderByMatchedCount*+testQueryByRangeIndexKeeps*+testQueryByPage*+testQueryByMultiLabelInPage,EdgeCoreTest#testQueryEdgeByPage*+testQuery*EdgesOfVertexInPaging' \
  -DfailIfNoTests=false -Drat.skip=true

# Four-method HStore cleanup reproduction (baseline has one serialization error)
mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,hstore \
  -Dbackend=hstore -Dconfig_path=/path/to/hstore-test.properties \
  '-Dtest=EdgeCoreTest#testQueryEdgesWithLimitOnSuperVertexAndFilterProp+testQueryByUnionHasDate+testQueryOutEdgesOfVertexBySortkeyWithMoreFieldsInPage+testUpdateEdgeProperty' \
  -DfailIfNoTests=false -Drat.skip=true

# Latest review follow-up on RocksDB
mvn test -pl hugegraph-server/hugegraph-test -am -P core-test,rocksdb \
  '-Dtest=QueryListTest,VertexCoreTest#testQueryByPrimaryValues*+testQueryByPage*+testQueryByMultiLabelInPage' \
  -DfailIfNoTests=false -Drat.skip=true

# Repeat the new primary-key case with the HStore configuration above:
# -Dtest=VertexCoreTest#testQueryByPrimaryValuesInPageWithVertexCache

mvn editorconfig:format
mvn clean compile -Dmaven.javadoc.skip=true -Drat.skip=true
git diff --check
```

</details>

Formatting and full reactor compilation passed at `9ad4a27e` and again at `00e365db`. The diagnostics cleanup also passed all 80 selected tests, including the revised batch-context and 10,000-page activation checks. The two additional tests cover deferred branch failure/cleanup in the shared flat-map iterator and compact root cache labels coexisting with distinct batches without mutating previously published values. Existing off-heap, lifecycle and cache-limit regressions pass in the 80-test selection. No throughput benchmark or backend-suite rerun was performed for this allocation/cache follow-up. Verification used `-Drat.skip=true`; the complete UnitTestSuite, API and TinkerPop suites were not run. The 78-test unit verification used Java files matching `05186a0e` byte-for-byte. The accessor/comment cleanup at `cf403002` passed formatting, full reactor compilation and the 37 targeted tests listed above. The backend suites remain labeled with the snapshots actually tested and were not rerun for this test/auth-registration supplement.

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API (Gremlin/REST interfaces)
- [x] Other affects (internal query iteration, paging and cache lifecycle)
- [ ] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -316,135 +318,142 @@ private boolean needCacheVertex(HugeVertex vertex) {
...
     @Watched(prefix = "graphcache")
-    protected Iterator<HugeVertex> queryVerticesFromBackend(Query query) {
-        if (this.enableCacheVertex() &&
-            query.idsSize() > 0 && query.conditionsSize() == 0 &&
-            !queryNeedsPostFilter(query)) {
-            return this.queryVerticesByIds((IdQuery) query);
-        } else {
-            return super.queryVerticesFromBackend(query);
+    protected QueryResults<HugeVertex> fetchVertexBatch(Query query) {
+        if (!this.enableCacheVertex() || query.idsSize() == 0 || query.conditionsSize() != 0) {
```
bitflicker64: ‼️ This bypass needs `|| query.paging()`.

Unlike the `queryVerticesFromBackend` override it replaces, this method is the QueryList *fetcher*, so it runs on the inner query. `GraphTransaction.optimizeQuery:1564` rewrites `label + primary-key` into `new IdQuery(query, id)`, and `Query(HugeType, Query originQuery)` (`Query.java:84-109`) copies neither page nor conditions, so that leaf reaches line 322 with `idsSize()==1` and `conditionsSize()==0` even after `QueryList.java:187-188` sets its page.

Line 353 then returns a `QueryResults` whose `metadata` is the batch's `BatchIterator` delegating to a plain `ArrayList` iterator, so `metadata(PageInfo.PAGE)` is null, and `mapBatches` passes that null through `filterExpiredBatches`. `QueryList.java:198` calls `PageInfo.pageState(results.iterator())`, which is `E.checkState(page instanceof PageState, "Invalid PageState '%s'", page)`. So `g.V().hasLabel(<pk label>).has(<pk>, v).has("~page", "")` fails with `Invalid PageState 'null'` wherever `supportsQueryByPage()` is true. The only escape is a cache miss that returns nothing, via the early return at 341-343.

The green paging tests do not contradict this. Index-paged queries go through `IndexQuery.iterator`, which takes its cursor from `pageIds.pageState()` (`QueryList.java:297`) rather than from result metadata, and no `VertexCoreTest` case pairs a primary-key label with its own primary-key property under `~page`. `InMemoryDBStore.supportsQueryByPage()` is false, so the unit tests cannot reach it either.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -316,135 +318,142 @@ private boolean needCacheVertex(HugeVertex vertex) {
...
+        });
+    }
 
-        Iterator<HugeEdge> rs = super.queryEdgesFromBackend(query);
-        if (queryNeedsPostFilter(query)) {
-            // The backend query may promote query.optimized() through origin-
-            // query propagation, so re-check before caching
-            return rs;
+    private void cacheEdgeBatch(Id cacheKey, Id batchKey, List<HugeEdge> candidates) {
+        synchronized (this.edgesCache) {
```
bitflicker64, line 414: 🧹 This monitor is shared by every transaction on the graph.

`edgesCache` comes from `CacheManager` (lines 104-108, 122-135). `edge.cache_type` defaults to `l2` (`CoreOptions.java:559-565`), so it is `levelCache(...)`, which like `cache(...)` returns one instance per name out of the static `INSTANCE.caches` map (`CacheManager.java:37, 59, 136-152`). The name is `edge-<graph>`, so all concurrent `CachedGraphTransaction`s for that graph serialise here, where the previous code did a lock-free `edgesCache.update(cacheKey, edges)`.

It is only reached on a miss (lines 392, 403) and the critical section is bounded, since the group caps at 100 entries (449-450), so the cost is small. Still, striping the lock per `cacheKey`, or just tolerating a lost update from a concurrent sibling batch, would keep the miss path lock-free.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/page/PageEntryIterator.java
```diff
@@ -17,112 +17,77 @@
...
+            QueryBatch.closeAll(previous);
+            if ((this.remaining != Query.NO_LIMIT && this.remaining <= 0L) ||
+                this.pageInfo.offset() >= this.queries.total()) {
+                return null;
+            }
+            long size = this.remaining == Query.NO_LIMIT ? this.pageSize :
+                        Math.min(this.pageSize, this.remaining);
+            QueryList.PageResults<R> page = this.queries.fetchNext(this.pageInfo, size);
+            this.pageBatches = page.results().batches();
+            if (!this.pageBatches.hasNext()) {
```
bitflicker64, line 64: 🧹 This `hasNext()` consumes and closes the static `PageResults.EMPTY`.

`IndexQuery.iterator` returns the shared `PageResults.emptyIterator()` when a holder yields no ids (`QueryList.java:292`; the singleton is declared at 337-339). `page.results().batches()` on it is that singleton's one-shot `BatchIterator`, so this call flips its `fetched` flag and, on the following turn, closes it and clears its `queries` list, process-wide and unsynchronised.

Consequence: the first empty index page in the JVM takes the "one empty batch" path (lines 70-77) and every later one takes the "no batch" path (65-68). Both advance the offset by one and emit nothing, so it is benign today, but the base code guarded against exactly this (`if (this.pageResults != QueryList.PageResults.EMPTY)`, base `PageEntryIterator.java:95-99`). Returning a fresh `PageResults` from `emptyIterator()` would drop the coupling.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/QueryResultContext.java
```diff
@@ -0,0 +1,118 @@
...
+            visibility = current;
+            if (current instanceof ConditionQuery) {
+                ConditionQuery condition = (ConditionQuery) current;
+                if (optimized == OptimizedType.NONE) {
+                    optimized = condition.optimized();
+                }
+                if (filter == null) {
+                    filter = condition.resultsFilter();
+                }
+                if (current.resultType().isGraph()) {
```
bitflicker64, line 64: 🧹 `match` is last-wins here while `optimized` and `filter` above are first-wins.

Lines 56-61 keep the innermost non-`NONE` `optimized()` and the innermost non-null `resultsFilter()`, but line 63 reassigns `match` on every graph-typed `ConditionQuery` in the chain, so `matchQuery()` ends up as the outermost one. `rightResultFromIndexQuery` then runs `cq.test(elem, context.resultsFilter())` (`GraphTransaction.java:1951`) with a filter and an optimized type that need not belong to `cq`. The old code read all three off a single query: `query`, or `query.originQuery()` when that was the `ConditionQuery`.

The chains are also longer now that `QueryList.indexIdQuery` roots the `IdQuery` at `bindQuery` instead of `parent()`. I could not build a chain where the two directions disagree, so this may be deliberate, but a line of comment saying why they differ would save the next reader the same trace.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -853,26 +854,43 @@ public Iterator<Vertex> queryVertices(Query query) {
...
+            // still produce an empty batch, which must keep its page cursor.
+            return results;
         }
-        return vertices;
+        QueryResultContext context = new QueryResultContext(
+                query, this.storeFeatures().supportsQuerySortByInputIds());
+        return results.mapBatches(batch -> new QueryBatch<>(batch.results(), context));
+    }
+
+    protected QueryResults<HugeVertex> fetchVertexBatch(Query query) {
```
bitflicker64, line 878: 🧹 Optional follow-up: `HugeFactoryAuthProxy.registerPrivateActions()` no longer matches this class.

That list hides sensitive members from the Gremlin sandbox (`HugeFactoryAuthProxy.java:239-281`). `filterInvalidRecords`, `filterUnmatchedRecords`, `queryNeedsPostFilter` and `conditionQueryNeedsPostFilter` are still listed but gone at this head, while none of the internals introduced here are registered: `queryVertexBatchesFromBackend`, `backendBatches`, `fetchVertexBatch`, `processBatches`, `queryEdgeBatchesFromBackend`, `queryEdgeBatchesFromBackendInternal`, `queryEdgesFromMemory`, `fetchEdgeBatch`, `filterInvalidRecord`, `filterUnmatchedRecord`, `filterExpiredBatches` (declared at 860, 870, 882, 891, 1075, 1079, 1123, 1132, 1893, 1898, 1984).

That file is outside this diff and the list was already imperfect (`filterExpiredResultFromFromBackend` is a typo matching nothing, and line 96 carries a `TODO: add some test to ensure the effect`), so this is a follow-up rather than something to fix here. Flagging it because the renames are what made it stale.

bitflicker64 (commented):
Blocking: yes. Summary: The batching design holds up. `BatchIterator` close and exception semantics, the shared cursor across `mapBatches`/`flatMap`, the raw-empty-page versus filtered-empty-batch distinction, and rebuilding the batch context on an edge-cache hit all check out. One blocker: moving the QueryList fetcher up from `super::query` to `fetchVertexBatch` puts the vertex cache inside the paging machinery, and the joined cache-plus-backend result it returns carries no page metadata, so a paged primary-key vertex query dies in `PageInfo.pageState`. Please add the `paging()` bypass plus a `VertexCoreTest` case for that shape; the other four comments are optional.

Evidence: the chain is traced through unmodified code and spelled out inline on `CachedGraphTransaction.java:322`, including why the existing green paging tests do not reach it. Static analysis only, not executed against a page-capable backend. CI at c694bb0 is green on all 24 checks.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -316,135 +318,143 @@ private boolean needCacheVertex(HugeVertex vertex) {
...
-            query.bigCapacity() || queryNeedsPostFilter(query)) {
-            // Don't cache all-edge, paging, large, or post-filtered queries
-            return super.queryEdgesFromBackend(query);
+    @Override
+    @Watched(prefix = "graphcache")
+    protected QueryResults<HugeEdge> fetchEdgeBatch(Query query) {
+        QueryResultContext context = new QueryResultContext(query);
+        List<Query> chain = context.queries();
+        Query request = chain.get(chain.size() - 1);
+        if (!this.enableCacheEdge() || request.empty() || request.paging() || request.bigCapacity()) {
```
bitflicker64, line 372: 🧹 Was dropping the `queryNeedsPostFilter` exclusion from this guard intentional?

Base `:399` had it in the same condition, plus a re-check after the backend call at `:424` ("re-check before caching"). Base `queryNeedsPostFilter` returned true for `hasSearchCondition()`, and for `optimized() != NONE` unless the edge query also carried a LABEL condition. So at base `g.E().has(<indexed prop>, v)` and primary-key/sort-key optimized edge queries were never cached, while `g.E().hasLabel(L).has(p, v)` was.

Here only `empty`/`paging`/`bigCapacity` remain. Since `fetchEdgeBatch` is now the `QueryList` fetcher rather than a `queryEdgesFromBackend` override, the index-derived `IdQuery` leaf reaches it with `request` (line 371, the chain root) equal to the user `ConditionQuery`: non-empty, non-paging, default capacity. Line 404 then caches it.

Results stay correct, because `processBatches` re-applies `filterUnmatchedRecord` on hits as well as misses. The cost is key count: `edge.cache_capacity` defaults to 1,000,000 entries (`CoreOptions.java:566-572`) on a per-graph cache, so high-cardinality indexed lookups now mint one guaranteed-miss key per predicate value and evict the adjacency entries the cache exists to serve. `fetchVertexBatch` at `:322` dropped the same guard, which looks harmless there since `verticesCache` is id-keyed.

If the widening is deliberate, a line in the PR body would save the next reader this trace. If not, `context.conditionFilterRequired()` (the context is already built on line 369) is the ready-made predicate, and the base's LABEL carve-out is worth keeping.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/page/QueryList.java
```diff
@@ -344,10 +354,11 @@ public boolean hasNextPage() {
         }
 
         public Query query() {
-            List<Query> queries = this.results.queries();
-            E.checkState(queries.size() == 1,
-                         "Expect query size 1, but got: %s", queries);
-            return queries.get(0);
+            return this.query;
```
bitflicker64: 🧹 `query()` and `get()` have no callers left after this change.

`query()` existed at base, deriving from `this.results.queries()` behind an `E.checkState(queries.size() == 1)`. This commit reworks it onto a new `query` field (line 339) and a third constructor parameter (line 341), with `emptyIterator()` passing `null` for it (line 374).

But the rewritten `PageEntryIterator` is the only `PageResults` consumer in main code (`PageEntryIterator.java:62`), and it reads `results()`, `hasNextPage()`, `page()` and `total()` only. Base `PageEntryIterator:79` called `pageResults.query()`, and `:60/81/97/106` called `pageResults.get()`; both are gone. `git grep` over this head finds no call to either accessor, and `QueryListTest:49-50` uses `.results()`.

Suggest dropping `get()`, `query()`, the field and the constructor parameter, returning `PageResults` to `(results, pageState)`, which also removes the `null` argument in `emptyIterator()`.

bitflicker64 (commented):
Blocking: no. Summary: The batch rewrite holds up at this head. `BatchIterator` close and suppression semantics, prefetch preservation across `backendBatches` into `mapBatches`, `null` mapper returns, filter ordering, root-of-chain visibility flags and paging metadata propagation all check out, and the previous review's `Invalid PageState 'null'` blocker is fixed by the `query.paging()` bypass. Two minor points below: the vertex and edge caches quietly lost their `queryNeedsPostFilter` exclusion, and `PageResults.query()`/`get()` are now dead code. Neither blocks.

Evidence: static reading of the exact-head diff plus the surrounding unchanged code, and a local build of 05186a0e (`hugegraph-struct` plus `hugegraph-server/{hugegraph-core,hugegraph-api,hugegraph-test}`) where the seven selected unit classes give `Tests run: 78, Failures: 0, Errors: 0`, matching the PR body. Backend suites were not re-run here; `InMemoryDBStore.supportsQueryByPage()` is false, so a local unit run cannot reach the paging and cache paths this PR reworks.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/QueryBatch.java
```diff
@@ -0,0 +1,262 @@
...
+
+    public QueryResultContext context() {
+        return this.context;
+    }
+
+    public <T> QueryBatch<T> map(Function<R, T> mapper) {
+        return this.flatMap(value -> {
+            T mapped = mapper.apply(value);
+            return mapped == null ? Collections.emptyIterator() :
+                   Collections.singleton(mapped).iterator();
```
bitflicker64: ⚠️ Each `map` or `filter` stage allocates a set and an iterator per surviving element.

`map` routes through `flatMap` and returns `Collections.singleton(mapped).iterator()` for every value that passes, so each surviving element costs one `SingletonSet` plus one iterator, per stage. `filter` at line 117 is built on `map`, so it pays the same. Rejected elements are free, since they get the shared `Collections.emptyIterator()`.

A plain vertex read stacks four such stages: `.map(this::parseEntry)` and `filterExpiredBatches` in `GraphTransaction.fetchVertexBatch:883`, `filterUnmatchedRecord` in `processBatches:892`, and `filterInvalidRecord` in `queryValidVerticesFromBackend:888`. Edges stack three on top of the one real expansion at `fetchEdgeBatch:1133`. `MapperIterator.fetch()`, which this replaces, loops on a null result in place and allocates nothing.

Requested change: give `map` its own `BatchIterator` that applies the function and loops while the result is null; `filter` can keep delegating to it.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -2009,45 +1981,13 @@ private boolean rightResultFromIndexQuery(Query query, HugeElement elem) {
...
-            Query query, Iterator<T> results) {
-        if (this.store().features().supportsTtl() || query.showExpired()) {
-            return results;
-        }
-        // Filter expired vertices/edges with TTL
-        return new FilterIterator<>(results, elem -> {
-            if (elem.expired()) {
-                DeleteExpiredJob.asyncDeleteExpiredObject(this.graph(), elem);
-                return false;
+    protected <T extends HugeElement> QueryResults<T> filterExpiredBatches(QueryResults<T> batches) {
```
bitflicker64, line 1982: 🧹 The TTL stage is now unconditional.

Base `filterExpiredResultFromBackend` returned the source iterator untouched when `store().features().supportsTtl()` or `query.showExpired()` held, so a TTL-capable store carried no wrapper at all. This version always adds a `filter` stage, which is one of the four counted in the `QueryBatch` comment, and moves the store-feature test into the per-element predicate.

Requested change: read `storeFeatures().supportsTtl()` once and return `batches` unchanged when it is true, leaving `context.showExpired()` and `elem.expired()` in the predicate since those do vary.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -1051,87 +1069,77 @@ public Iterator<Edge> queryEdges(Query query) {
...
-
-            return edgeIterators.reduce(ExtendableIterator::concat)
-                                .orElse(Collections.emptyIterator());
+                return this.queryEdgeBatchesFromBackendInternal(cq);
+            };
+            // Preserve immediate validation for a single query without activating a sibling.
+            if (flattened.size() == 1) {
+                return fetcher.apply(flattened.get(0));
+            }
+            return QueryResults.flatMap(flattened.iterator(), fetcher);
```
bitflicker64, line 1111: 🧹 Multi-branch edge queries lost their eager validation, and only the single-query case says so.

Base built a `Stream<Iterator<HugeEdge>>` and ended with `reduce(ExtendableIterator::concat)`. That is a terminal operation, so every `queryEdgesFromBackendInternal(cq)` ran at call time and any validation failure surfaced from the call itself. `QueryResults.flatMap` here runs each branch only when iteration reaches it, so a failure in the second or later flattened query now surfaces mid-iteration. Line 1109 documents the choice for `flattened.size() == 1` and nothing covers the multi-branch case.

Requested change: say in the PR body which failures are expected to move from call time to iteration time for multi-branch edge queries, or apply the same eager first step there.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -316,135 +318,143 @@ private boolean needCacheVertex(HugeVertex vertex) {
...
-
-        Id cacheKey = new QueryId(query);
-        Object value = this.edgesCache.get(cacheKey);
-        @SuppressWarnings("unchecked")
-        Collection<HugeEdge> edges = (Collection<HugeEdge>) value;
-        if (value != null) {
-            for (HugeEdge edge : edges) {
+        Id cacheKey = new QueryId(request);
+        Id batchKey = new QueryId(query);
+        CachedEdgeQuery group = new CachedEdgeQuery(this.edgesCache.get(cacheKey));
```
bitflicker64, line 379: 🧹 The read path shallow-copies the group's index list for nothing.

Line 377 builds a `CachedEdgeQuery` on every edge fetch that reaches the cache, and the constructor at 427-430 does `new ArrayList<>((List<Object>) cached)`. Nothing on the read path mutates it: `get()` only scans. `cacheEdgeBatch:413` already makes its own copy inside `synchronized (this.edgesCache)`, which is where the copy belongs.

Requested change: scan the cached list directly for the lookup and keep the copy in `cacheEdgeBatch`.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -316,135 +318,143 @@ private boolean needCacheVertex(HugeVertex vertex) {
...
         }
-
-        Id cacheKey = new QueryId(query);
-        Object value = this.edgesCache.get(cacheKey);
-        @SuppressWarnings("unchecked")
-        Collection<HugeEdge> edges = (Collection<HugeEdge>) value;
-        if (value != null) {
-            for (HugeEdge edge : edges) {
+        Id cacheKey = new QueryId(request);
+        Id batchKey = new QueryId(query);
```
bitflicker64, line 376: 🧹 The batch key stored in the value can be larger than the value it labels.

`QueryId.asString()` is `Query.toString()`, and `Query.toString():610` appends `id in <ids>`. `put` stores that string next to the candidates, so an index-derived leaf writes up to `query.batch_size` ids of rendered text (`CoreOptions.java:466-472`, default 1000) beside a candidate list capped at `MAX_CACHE_EDGES_PER_QUERY`, which is 100 (line 63).

In the common adjacency shape the leaf is also the chain root, so `batchKey` and `cacheKey` render the same string and the value holds a verbatim copy of its own cache key.

Requested change: skip storing the label when `batchKey.equals(cacheKey)` and treat a group of one unlabelled entry as that case, which keeps the stored value a nested list and leaves off-heap serialization unchanged.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/tx/GraphTransaction.java
```diff
@@ -45,14 +44,16 @@
...
 import org.apache.hugegraph.backend.query.Aggregate.AggregateFunc;
+import org.apache.hugegraph.backend.query.Aggregate;
 import org.apache.hugegraph.backend.query.Condition;
-import org.apache.hugegraph.backend.query.ConditionQuery;
 import org.apache.hugegraph.backend.query.ConditionQuery.OptimizedType;
+import org.apache.hugegraph.backend.query.ConditionQuery;
 import org.apache.hugegraph.backend.query.ConditionQueryFlatten;
 import org.apache.hugegraph.backend.query.IdQuery;
 import org.apache.hugegraph.backend.query.Query;
+import org.apache.hugegraph.backend.query.QueryBatch;
```
bitflicker64, line 54: 🧹 Four imports are left unused, and two were reordered for no reason. Anchored here because lines 23, 68, 70 and 105 fall outside every hunk in this file.

`java.util.Collections` (23), `org.apache.hugegraph.iterator.FlatMapperIterator` (68), `org.apache.hugegraph.iterator.ListIterator` (70) and `com.google.common.collect.Iterators` (105) each appear exactly once in the file at this head, on their own import line, and all four are used at base. `style/checkstyle.xml:59` enables `UnusedImports`, inheriting the `info` severity set at line 23, so the `validate`-bound execution logs them without failing the build.

Lines 47-48 and 50-51 also moved `Aggregate.AggregateFunc` above `Aggregate` and `ConditionQuery.OptimizedType` above `ConditionQuery`, which is neither alphabetical nor related to the change. `PageEntryIterator.java:23-24` has the same swap.

Requested change: delete the four import lines and restore the original order of the two pairs.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/page/PageEntryIterator.java
```diff
@@ -17,112 +17,77 @@
...
     public PageEntryIterator(QueryList<R> queries, long pageSize) {
         this.queries = queries;
         this.pageSize = pageSize;
-        this.pageInfo = this.parsePageInfo();
-        this.queryResults = new QueryResults<>(this, queries.parent());
-
-        this.pageResults = QueryList.PageResults.emptyIterator();
+        this.pageInfo = PageInfo.fromString(queries.parent().pageWithoutCheck());
+        E.checkState(this.pageInfo.offset() < queries.total(),
+                     "Invalid page offset '%s' exceeds the size of IdHolderList",
```
bitflicker64: 🧹 The offset check dropped the page token from its message.

Base `parsePageInfo` reported `"Invalid page '%s' with an offset '%s' exceeds the size of IdHolderList"` with both the raw page string and the offset. This check keeps only the offset, and it fires exactly when a caller supplies a stale or hand-written `~page` value, which is the case where the token itself is the thing worth seeing in the log.

Requested change: put `queries.parent().pageWithoutCheck()` back in the message alongside the offset.

bitflicker64 (commented):
Blocking: no. Summary: The batch rework holds up at this head, and four of my earlier points are addressed: the vertex cache now has the `paging()` bypass (`CachedGraphTransaction.java:322`), `PageResults.emptyIterator()` returns a fresh instance (`QueryList.java:362-365`), `registerPrivateActions()` matches the renamed members again, and the dead `PageResults.query()`/`get()` accessors are gone. My threads on the shared `edgesCache` monitor (`:412`) and the dropped `queryNeedsPostFilter` exclusion (`:372`) are still open. New here: the iterator plumbing allocates per surviving element at every stage, the TTL stage is now unconditional, multi-branch edge queries lost their eager validation, and the edge cache copies its group index on reads. Evidence: exact-head diff for cf40300 against origin/master, plus `Query.toString():610`, `CoreOptions.QUERY_BATCH_SIZE` default 1000, `style/checkstyle.xml:23,59`. All 24 check runs on cf40300 are green.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/QueryResults.java
```diff
@@ -18,234 +18,362 @@
...
-                return true;
+    public R one() {
+        return one(this.iterator());
+    }
+
+    /**
+     * Source query of the current batch. A known single source is available before
+     * activation; composed streams start empty. Closing clears the diagnostics.
+     */
+    public List<Query> queries() {
```
bitflicker64, line 225: 🧹 `queries()` has no main-code caller left after this change.

At base it had exactly one: `QueryList.PageResults.query()` (base `QueryList.java:347`), which `cf403002` deleted. Ordering now comes from `QueryResultContext`, and base `PageEntryIterator.java:79` (`queryResults.setQuery(pageResults.query())`) is gone. Over this head, `git grep '\.queries()'` finds only `context.queries()` calls plus assertions in `QueryResultsTest.java:161,164,166` and `QueryListTest.java:129,131,135`.

The bookkeeping still runs on the hot path: an `ArrayList` per `QueryResults` (line 51), `queries.clear()` plus `queries.add(...)` on every batch activation (lines 91-92), a clear on close (line 100), and a re-seed in `toList()` (line 250). Same shape as the dead `PageResults.query()`/`get()` pair you removed last round.

Requested change: drop the `queries` field and this accessor, and rework the assertions onto `results.batches()` where the test structure allows. Note that `java.util.Collections` (line 21) is then unused, since lines 51 and 226 are its only uses. If it should stay, name its reader in the Javadoc at lines 221-224, because no main code reads it today.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/cache/CachedGraphTransaction.java
```diff
@@ -316,135 +318,149 @@ private boolean needCacheVertex(HugeVertex vertex) {
...
     @Watched(prefix = "graphcache")
-    protected Iterator<HugeVertex> queryVerticesFromBackend(Query query) {
-        if (this.enableCacheVertex() &&
-            query.idsSize() > 0 && query.conditionsSize() == 0 &&
-            !queryNeedsPostFilter(query)) {
-            return this.queryVerticesByIds((IdQuery) query);
-        } else {
-            return super.queryVerticesFromBackend(query);
+    protected QueryResults<HugeVertex> fetchVertexBatch(Query query) {
+        if (!this.enableCacheVertex() || query.paging() ||
```
bitflicker64, line 322: 🧹 This `paging()` test is on the leaf, while `fetchEdgeBatch` tests the chain root.

Separate from my earlier thread that asked for this guard: now that it exists, the two overrides disagree about what "paged" means. Line 322 tests `query.paging()` on the `QueryList` fetcher's leaf; `fetchEdgeBatch` builds the same context and tests `request.paging()` on `chain.get(chain.size() - 1)`, the chain root (lines 369-372).

The difference is observable. `QueryList.IndexQuery.iterator(int, String, long)` builds its leaf through `indexIdQuery` (`QueryList.java:326-332`) as `new IdQuery(parent().resultType(), bindQuery)`, and `Query(HugeType, Query)` sets `this.page = null` (`Query.java:95`), so that leaf reaches line 322 with `paging() == false`, `idsSize() > 0` and `conditionsSize() == 0`. The vertex cache therefore still runs inside a paged index query, and is correct only for the reason I traced on the earlier thread: that path takes its cursor from `pageIds.pageState()`, not from result metadata.

Requested change: state in a comment that the leaf test is deliberate and that the index-paging leaf is safe because its cursor comes from the `IdHolder`. Matching `fetchEdgeBatch` and testing the root instead would also work, but it bypasses the vertex cache for every paged index query, so it is the more expensive option.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/page/QueryList.java
```diff
@@ -25,15 +25,17 @@
 import org.apache.hugegraph.backend.id.Id;
 import org.apache.hugegraph.backend.page.IdHolder.BatchIdHolder;
 import org.apache.hugegraph.backend.page.IdHolder.FixedIdHolder;
-import org.apache.hugegraph.backend.query.ConditionQuery;
 import org.apache.hugegraph.backend.query.ConditionQuery.OptimizedType;
+import org.apache.hugegraph.backend.query.ConditionQuery;
 import org.apache.hugegraph.backend.query.IdQuery;
 import org.apache.hugegraph.backend.query.Query;
+import org.apache.hugegraph.backend.query.QueryBatch.BatchIterator;
+import org.apache.hugegraph.backend.query.QueryBatch;
```
bitflicker64, line 33: 🧹 These two import pairs kept the ordering that `9ad4a27e` reverted in the other two files.

This PR moved `ConditionQuery` below `ConditionQuery.OptimizedType` (lines 28-29) and added `QueryBatch.BatchIterator` above `QueryBatch` (lines 32-33). `9ad4a27e` restored outer-class-first for exactly these two patterns elsewhere: `Aggregate`/`Aggregate.AggregateFunc` and `ConditionQuery`/`ConditionQuery.OptimizedType` in `GraphTransaction.java`, and `QueryBatch`/`QueryBatch.BatchIterator` in `PageEntryIterator.java`. This file was missed, and nothing flags it because `style/checkstyle.xml:43` has `ImportOrder` commented out.

Requested change: swap both pairs so this file matches the ordering the same commit restored in the other two.

bitflicker64 (commented):
Blocking: no. Summary: The batch rework holds at this head, and `9ad4a27e` addresses all seven points from the `cf40300` round: map allocation, the TTL short-circuit, the branch-validation note plus `testLaterBranchFailureClosesEarlierBranch`, the read-path copy, the batch-key size, the unused imports and reorders in `GraphTransaction.java`, and the offset message. Two caveats. First, only half of the base TTL short-circuit came back: base `filterExpiredResultFromBackend` returned early on `supportsTtl() || query.showExpired()`, while `GraphTransaction.java:1982-1992` returns early on `supportsTtl()` alone and still builds a filter stage to evaluate `context.showExpired()` per element. That is fine given the per-batch context, but the wrapper is not gone for a `showExpired` query on a non-TTL store. Second, the backend suites were not re-run here. Three minor points below; none blocks. My earlier threads on the shared `edgesCache` monitor (`:414`) and the dropped `queryNeedsPostFilter` exclusion (`:372`) are unchanged at this head.

What I checked and found sound: the compact `""` batch label cannot collide, because `Query.toString():588` always starts with `` `Query `` and `IdPrefixQuery`/`IdRangeQuery` build on `super.toString()`, so no `QueryId.asString()` is empty. Off-heap storage of the nested `List<Object>` round-trips: `OffheapCache.ValueType.valueOf` maps `String` to `STRING(DataType.TEXT)` and `serializeList`/`deserializeList` recurse (`OffheapCache.java:281-297`). The `ExtendableIterator(candidates, source)` wrapper at `CachedGraphTransaction.java:409-410` does not strand the backend cursor: `hugegraph-commons/hugegraph-common/src/main/java/org/apache/hugegraph/iterator/ExtendableIterator.java:69-90` closes every iterator still in `itors`, with suppression.

Evidence: static reading of the exact-head diff against `36811483a` plus the surrounding unchanged code. Executed: a local JDK 11 build of `hugegraph-struct` and `hugegraph-server/{hugegraph-core,hugegraph-api,hugegraph-test}` at `9ad4a27e` running the seven selected unit classes gives `Tests run: 80, Failures: 0, Errors: 0`, matching the PR body. `InMemoryDBStore.supportsQueryByPage()` is false (`InMemoryDBStore.java:442-444`), so a local unit run cannot reach the paging paths this PR reworks. All 24 check runs on `9ad4a27e` are green.

bitflicker64 (approved):
Blocking: no. Summary: At 00e365d the batch rework holds, and the three points left from the last round are addressed: the unused `QueryResults.queries()` diagnostics are removed, the leaf-level `paging()` test in `fetchVertexBatch` is explained in a comment, and the `QueryList` import order is restored. All earlier threads are resolved. Batch close and suppression semantics, the shared cursor, raw-empty versus filtered-empty pages, copy-on-write edge-cache groups, per-batch visibility and TTL context, and the sandbox method filter all check out. Evidence: static review of the full diff against 36811483a and the unchanged code around it. The seven focused unit classes (QueryResultsTest, QueryListTest, GraphTransactionTest, CachedGraphTransactionTest, IdHolderTest, QueryTest, PageStateTest) give `Tests run: 80, Failures: 0, Errors: 0` on JDK 11 at 00e365d. All 24 checks at this head are green.

contrueCT:
@imbajin This PR may be ready to merge ?

imbajin (approved):

PR #3194, opened 2026-09-05
perf(docker): reuse runtime layers and scope Maven builds

### LINKED ISSUE #3179: perf(docker): track follow-up image build optimizations
## Background

The shared Buildx Bake graph from apache/hugegraph#3171 and hugegraph/actions#27 removed four repeated Maven reactor builds. A real `latest` publication completed successfully with all four images, both `linux/amd64` and `linux/arm64`, the Compose/Gremlin checks, and the standalone Server smoke test.

Production run: https://github.com/hugegraph/actions/actions/runs/33302286614

The candidate build took 614 seconds. The main critical-path work was:

| Work | Observed duration |
| --- | ---: |
| Full Maven reactor build | about 285s |
| ARM64 runtime package installation under QEMU | about 149–171s |
| Local multi-platform image export | up to about 53s |
| Per-image registry cache export | up to about 100s |

The shared build-cache export took about 237 seconds but overlapped much of the runtime work, so BuildKit vertex durations must not be added directly.

### 1. Put stable runtime dependency layers before source-dependent artifacts

The four runtime stages currently copy Maven artifacts before running `apt-get`. Any source change therefore invalidates package installation as well, which is particularly expensive under ARM64 emulation.

Move stable package installation before `COPY --from=build`. Where a command also edits a copied configuration file, split the stable package installation from the source-dependent configuration edit.

Keep the existing operational tools, including `vim`, `procps`, `curl`, `dumb-init`, and `iproute2` where applicable. Removing commonly used operational tools is explicitly out of scope.

Expected result: reuse the runtime dependency layer on ordinary source changes and reduce total build time by roughly 1–2 minutes, subject to a real benchmark.

### 2. Evaluate a module-scoped Maven build

Determine whether the image build can request only the Server, PD, and Store distribution modules plus their reactor dependencies instead of installing all 38 modules.

Do not skip required assembly inputs or silently weaken validation. Compare the produced distributions and image contents before adopting this change.

### 3. Re-evaluate native ARM builds only if the simpler changes are insufficient

A native ARM runner could remove QEMU runtime work, but it would add runner coordination and manifest assembly complexity. Keep the current single-job Bake flow unless measurements show a clear end-to-end benefit without reintroducing temporary architecture tags or partially published images.

### 4. Measure cache export separately

Registry cache export is a meaningful part of the build step, but removing it can merely shift time to the final image push or the next build. Compare both the candidate-build duration and the complete publish duration; do not optimize one metric by moving the same work elsewhere.

## Acceptance criteria

- Preserve direct Dockerfile builds and the shared `docker/bake.hcl` flow.
- Preserve the current runtime tools and published image behavior.
- Produce the same four final image tags with both amd64 and arm64 variants.
- Keep the existing local multi-platform inspection, Compose/Gremlin checks, and standalone Server smoke test before publication.
- Benchmark on the same runner class with a real registry cache and publish path.
- Report both the candidate-build duration and the complete job duration before and after.
- Prefer the runtime-layer reorder first; pursue the other candidates only when their measured benefit justifies the added complexity.

lokidundun:
### Visual summary

Docker build cache and Maven scope

## Purpose of the PR

- close #3179 

## Main Changes

- Move runtime package installation before `COPY --from=build` in all four Dockerfiles, allowing dependency layers to survive application source changes.
- Keep configuration edits that depend on copied files after the corresponding `COPY`.
- Build the Server, PD, and Store distribution modules with `-pl` and `-am`, reducing the reactor from 38 to 27 modules while retaining required dependencies.
- Preserve existing runtime packages, configurable Maven arguments, and identical shared Maven build stages across all four Dockerfiles.

The existing single-job Bake flow and QEMU-based ARM64 build remain in place.

## Verifying these changes

- [x] Already covered by existing checks: multi-platform image inspection, Compose integration, Gremlin CRUD, and Standalone smoke tests.
- [x] Additional verification: before/after benchmarks and distribution/image content comparisons.

### Runtime dependency layer benchmark

Both variants used the same GitHub-hosted Ubuntu runner class and separate GHCR registry caches. After seeding each cache, identical configuration comments were added to trigger distribution changes and a full Maven rebuild.

| Scenario / metric | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| Initial cache seed: image build | 552 s | 510 s | 42 s (7.6%) |
| Initial cache seed: complete job | 704 s | 669 s | 35 s (5.0%) |
| Source change: image build | 601 s | 461 s | 140 s (23.3%) |
| Source change: complete job | 755 s | 602 s | 153 s (20.3%) |

Before the change, the four ARM64 runtime package installation steps ran under QEMU and took 150.6–160.5 seconds each. After the change, all eight runtime dependency layers across amd64 and arm64 were restored from registry cache in 2.0–9.3 seconds per layer, without rerunning package installation.

Cache export was measured separately:

| Cache export vertex | Before | After |
| --- | ---: | ---: |
| Shared build cache | 205.2 s | 165.7 s |
| PD | 87.8 s | 5.5 s |
| Store | 56.1 s | 4.9 s |
| HStore Server | 83.3 s | 6.6 s |
| Standalone Server | 40.3 s | 8.1 s |

These BuildKit vertices overlap; their durations must not be added together or treated as independent end-to-end savings.

### Maven reactor benchmark

This separate experiment compared the full reactor with the scoped reactor using the same source and existing runtime layer optimization.

| Metric | Full reactor | Scoped reactor | Reduction |
| --- | ---: | ---: | ---: |
| Reactor modules | 38 | 27 | 11 modules |
| Maven-reported build time | 270 s | 176 s | 94 s (34.8%) |
| Distribution build, cache and local export | 424 s | 269 s | 155 s (36.6%) |
| Complete distribution job, including artifact upload | 492 s | 351 s | 141 s (28.7%) |

The distribution-job timings above are not complete image publication timings.

### Correctness checks

- Compared 391 distribution files, including permissions and normalized nested JAR contents.
- Compared all four images on both amd64 and arm64: 780 records per architecture covering application files, installed packages, and image configuration, with no missing, added, or changed records.
- JAR comparisons normalized ZIP timestamps, compression, entry order, and specified build-time metadata; this verifies effective content equivalence rather than byte-identical archives.
- The runtime-layer CI built and published all four images with both architectures, passing local platform inspection, Compose, Gremlin CRUD, Standalone smoke tests, and published manifest checks. Functional checks in this flow ran on amd64.
- A separate native ARM experiment passed functional checks on both architectures and final multi-platform publication. That experimental workflow is not included in this PR.

### Successful CI runs

| Validation | Result | CI run |
| --- | --- | --- |
| Runtime layer benchmark, multi-platform build and publication | All 4 jobs passed | [33950351952](https://github.com/lokidundun/incubator-hugegraph/actions/runs/33950351952) |
| Full/scoped distribution benchmark, content comparison, and native architecture validation | All 6 jobs passed | [33953324289](https://github.com/lokidundun/incubator-hugegraph/actions/runs/33953324289) |
| Full/scoped image content comparison on amd64 and arm64 | Both jobs passed | [33953049148](https://github.com/lokidundun/incubator-hugegraph/actions/runs/33953049148) |

These are fork validation runs for the implementation approach. The Maven experiments enabled the same module selection through `MAVEN_ARGS`; this PR places that selection directly in the Dockerfiles. These runs do not replace CI on the final PR commit.

Measurements are single-run observations using GHCR, not the official Docker Hub publication environment. Savings from separate experiments are not additive.

### Commands for local verification

Run from the repository root using Bash.

Build the selected distributions:

```bash
mvn install \
  -pl hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist \
  -am -e -B -ntp \
  -Dmaven.test.skip=true \
  -Dmaven.javadoc.skip=true
```

Inspect the shared Bake configuration:

```bash
docker buildx bake -f docker/bake.hcl --print
```

Build all four images through the shared Bake flow:

```bash
docker buildx bake -f docker/bake.hcl --progress=plain
```

The default Bake targets include amd64 and arm64. Local multi-platform loading requires a compatible builder and the containerd image store; building ARM64 on an amd64 host also requires emulation.

Check a direct Dockerfile build:

```bash
docker buildx build \
  --platform linux/amd64 \
  -f hugegraph-server/Dockerfile \
  -t hugegraph-standalone:pr-check \
  --load .
```

The commands above are build checks; the linked CI runs provide the integration, content-comparison, and publication validation.

- [ ] Dependencies
- [ ] Modify configurations
- [ ] The public API
- [x] Other affects: Docker build layer caching and Maven reactor selection.
- [ ] Nope

## Documentation Status

- [ ] `Doc - TODO`
- [ ] `Doc - Done`
- [x] `Doc - No Need`

### DIFF
--- a/docker/bake.hcl
+++ b/docker/bake.hcl
@@ -19,6 +19,14 @@ variable "MAVEN_ARGS" {
   default = ""
 }
 
+variable "MAVEN_PROJECTS" {
+  default = "hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist"
+}
+
+variable "RUNTIME_DEPS_EPOCH" {
+  default = "1"
+}
+
 variable "SOURCE_REVISION" {
   default = "local"
 }
@@ -38,8 +46,10 @@ variable "EXPORT_CACHE" {
 target "_common" {
   context = "."
   args = {
-    MAVEN_ARGS     = MAVEN_ARGS
-    SOURCE_REVISION = SOURCE_REVISION
+    MAVEN_ARGS         = MAVEN_ARGS
+    MAVEN_PROJECTS     = MAVEN_PROJECTS
+    RUNTIME_DEPS_EPOCH = RUNTIME_DEPS_EPOCH
+    SOURCE_REVISION    = SOURCE_REVISION
   }
   platforms = [
     "linux/amd64",
--- a/hugegraph-server/Dockerfile
+++ b/hugegraph-server/Dockerfile
@@ -28,14 +28,14 @@ ARG MAVEN_ARGS
 ARG SOURCE_REVISION=local
 
 RUN --mount=type=cache,id=hugegraph-maven-${SOURCE_REVISION},target=/root/.m2,sharing=locked \
-    mvn install $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
+    mvn install -pl hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist \
+        -am $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
     && rm ./hugegraph-server/*.tar.gz ./hugegraph-pd/*.tar.gz ./hugegraph-store/*.tar.gz
 
 # 2nd stage: runtime env
 # Note: ZGC (The Z Garbage Collector) is only supported on ARM-Mac with java > 13
 FROM eclipse-temurin:11-jre-jammy
 
-COPY --from=build /pkg/hugegraph-server/apache-hugegraph-server-*/ /hugegraph-server/
 LABEL maintainer="HugeGraph Docker Maintainers <dev@hugegraph.apache.org>"
 
 # TODO: use g1gc or zgc as default
@@ -46,7 +46,7 @@ ENV JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseContainerSupport -XX:Max
 
 WORKDIR /hugegraph-server/
 
-# 1. Install runtime dependencies and configure server
+# 1. Install runtime dependencies
 # Note: iproute2 provides `ss`, which the bin/util.sh port preflight needs.  The
 # jammy base image ships neither ss nor netstat, so without it the preflight is
 # permanently inconclusive and a duplicate start is no longer caught.  It
@@ -59,8 +59,10 @@ RUN apt-get -q update \
        iproute2 \
        vim \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* \
-    && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=build /pkg/hugegraph-server/apache-hugegraph-server-*/ /hugegraph-server/
+RUN sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
 
 # 2. Init docker script
 COPY hugegraph-server/hugegraph-dist/docker/scripts/remote-connect.groovy ./scripts
--- a/hugegraph-server/Dockerfile
+++ b/hugegraph-server/Dockerfile
@@ -25,10 +25,11 @@ WORKDIR /pkg
 COPY . .
 
 ARG MAVEN_ARGS
+ARG MAVEN_PROJECTS="hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist"
 ARG SOURCE_REVISION=local
 
 RUN --mount=type=cache,id=hugegraph-maven-${SOURCE_REVISION},target=/root/.m2,sharing=locked \
-    mvn install -pl hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist \
+    mvn install -pl "$MAVEN_PROJECTS" \
         -am $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
     && rm ./hugegraph-server/*.tar.gz ./hugegraph-pd/*.tar.gz ./hugegraph-store/*.tar.gz
 
@@ -51,6 +52,7 @@ WORKDIR /hugegraph-server/
 # jammy base image ships neither ss nor netstat, so without it the preflight is
 # permanently inconclusive and a duplicate start is no longer caught.  It
 # replaces lsof, which no shipped script calls any more.
+ARG RUNTIME_DEPS_EPOCH=1
 RUN apt-get -q update \
     && apt-get -q install -y --no-install-recommends --no-install-suggests \
        dumb-init \
--- a/hugegraph-server/Dockerfile-hstore
+++ b/hugegraph-server/Dockerfile-hstore
@@ -28,17 +28,13 @@ ARG MAVEN_ARGS
 ARG SOURCE_REVISION=local
 
 RUN --mount=type=cache,id=hugegraph-maven-${SOURCE_REVISION},target=/root/.m2,sharing=locked \
-    mvn install $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
+    mvn install -pl hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist \
+        -am $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
     && rm ./hugegraph-server/*.tar.gz ./hugegraph-pd/*.tar.gz ./hugegraph-store/*.tar.gz
 
 # 2nd stage: runtime env
 # Note: ZGC (The Z Garbage Collector) is only supported on ARM-Mac with java > 13
 FROM eclipse-temurin:11-jre-jammy
-
-COPY --from=build /pkg/hugegraph-server/apache-hugegraph-server-*/ /hugegraph-server/
-# remove hugegraph.properties and rename hstore.properties.template for default hstore backend
-RUN cd /hugegraph-server/conf/graphs \
-    && rm hugegraph.properties && mv hstore.properties.template hugegraph.properties
 
 LABEL maintainer="HugeGraph Docker Maintainers <dev@hugegraph.apache.org>"
 
@@ -48,7 +44,7 @@ ENV JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseContainerSupport -XX:Max
 
 WORKDIR /hugegraph-server/
 
-# 1. Install runtime dependencies and configure server
+# 1. Install runtime dependencies
 # Note: iproute2 provides `ss`, which the bin/util.sh port preflight needs.  The
 # jammy base image ships neither ss nor netstat, so without it the preflight is
 # permanently inconclusive and a duplicate start is no longer caught.  It
@@ -61,8 +57,13 @@ RUN apt-get -q update \
        iproute2 \
        vim \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* \
-    && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=build /pkg/hugegraph-server/apache-hugegraph-server-*/ /hugegraph-server/
+# remove hugegraph.properties and rename hstore.properties.template for default hstore backend
+RUN cd /hugegraph-server/conf/graphs \
+    && rm hugegraph.properties && mv hstore.properties.template hugegraph.properties
+RUN sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
 
 # 2. Init docker script
 COPY hugegraph-server/hugegraph-dist/docker/scripts/remote-connect.groovy ./scripts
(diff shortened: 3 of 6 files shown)


hugegraph-server/Dockerfile
```diff
@@ -59,8 +59,10 @@ RUN apt-get -q update \
        iproute2 \
        vim \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* \
-    && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=build /pkg/hugegraph-server/apache-hugegraph-server-*/ /hugegraph-server/
```
bitflicker64, line 64: ⚠️ With the apt layer now ahead of this `COPY`, its cache key no longer includes the application source, so the layer is reused until the `eclipse-temurin:11-jre-jammy` digest changes. What is lost is the per-source-change reinstall: rebuilding the same commit already hit cache before, but any source change used to force a fresh `apt-get install`, and now nothing short of a base image update does.

This only bites on a registry-cache flow, and that flow is out of tree: `docker/bake.hcl` gates every `cache-to` behind `EXPORT_CACHE`, which defaults to `false`, and no workflow here runs `bake` to build or push. The PR's own numbers show the effect, with eight runtime layers restored in 2.0-9.3 s "without rerunning package installation".

Could you add a documented way to force a reinstall, for example `ARG RUNTIME_DEPS_EPOCH=1` immediately before the apt block in all four files, bumped when packages need refreshing? `--no-cache-filter` is not an option as things stand, since the runtime stage has no `AS` name. A short note on the cache behaviour in `docker/README.md` would help whoever operates the publish flow.

bitflicker64 (commented):
Blocking: no. Summary: Both mechanisms hold up. The four build stages stay byte-identical, so the shared Bake cache and the existing docker-bake-check guard still work, and the `-pl ... -am` scoping is content-safe because all three assembly descriptors are dependency-driven rather than module-driven. Four non-blocking points: the runtime apt layer now refreshes only when the base image digest changes, the reactor scope is no longer settable through `MAVEN_ARGS`, `.dockerignore` was not narrowed to match the new scope, and Dockerfile-hstore gained an avoidable layer. Evidence: `git diff origin/master...304bea1`; lines 20-33 of all four Dockerfiles hash identically; the server, pd and store assembly descriptors contain no `<moduleSet>`, so `-am` builds exactly the closure they consume; `.github/workflows/cluster-test-ci.yml` still runs an unfiltered full-reactor `mvn clean package` on every pull request, so compile coverage of the 11 dropped modules is retained.

hugegraph-server/Dockerfile
```diff
@@ -28,14 +28,14 @@ ARG MAVEN_ARGS
 ARG SOURCE_REVISION=local
 
 RUN --mount=type=cache,id=hugegraph-maven-${SOURCE_REVISION},target=/root/.m2,sharing=locked \
-    mvn install $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
+    mvn install -pl hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist \
+        -am $MAVEN_ARGS -e -B -ntp -Dmaven.test.skip=true -Dmaven.javadoc.skip=true \
```
bitflicker64, line 32: 🧹 The module list is hardcoded ahead of `$MAVEN_ARGS`, which is the only Maven knob `docker/bake.hcl` exposes. Maven accumulates repeated `-pl` values rather than letting a later one replace an earlier one, so `MAVEN_ARGS` can still add modules or drop them with a `!` prefix, but it can no longer set the scope outright. The PR description notes that the fork experiments "enabled the same module selection through `MAVEN_ARGS`"; that route closes here.

Drift between the four copies is already covered by the `docker-bake-check` job in `.github/workflows/docker-build-ci.yml`, so this is only about overridability.

Suggested change: hoist the list into a build arg beside the existing `ARG MAVEN_ARGS` and pass it through `_common.args` in `docker/bake.hcl`.

```dockerfile
ARG MAVEN_PROJECTS="hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist"
```

with the command becoming `mvn install -pl "$MAVEN_PROJECTS" -am $MAVEN_ARGS ...`. All four files stay byte-identical, so the CI identity check still passes.

hugegraph-server/Dockerfile
```diff
@@ -28,14 +28,14 @@ ARG MAVEN_ARGS
...
     && rm ./hugegraph-server/*.tar.gz ./hugegraph-pd/*.tar.gz ./hugegraph-store/*.tar.gz
```
bitflicker64, line 33: 🧹 The reactor is scoped but the build context is not, so the caching win is smaller than the benchmark suggests. `.dockerignore` at head excludes only build output, archives, IDE and OS files, `.git`, `.github`, `**/*.md` and the compose files, so `hugegraph-test`, `hg-pd-test`, `hg-store-test`, `hugegraph-cluster-test/**`, `hugegraph-example`, `hg-pd-cli`, `hg-store-cli` and `install-dist` all still land in the context and still feed the `COPY . .` cache key on line 25. Editing any of them therefore invalidates this shared build stage and pays for a full scoped Maven run, for modules the build no longer compiles.

A follow-up rather than a change here, but worth recording. One caveat for whoever picks it up: these directories cannot be ignored wholesale, because the root pom lists them in `<modules>` and Maven fails when a listed module's `pom.xml` is absent, so only their `src/` subtrees can be excluded.

hugegraph-server/Dockerfile-hstore
```diff
@@ -61,8 +57,13 @@ RUN apt-get -q update \
...
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* \
-    && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=build /pkg/hugegraph-server/apache-hugegraph-server-*/ /hugegraph-server/
+# remove hugegraph.properties and rename hstore.properties.template for default hstore backend
+RUN cd /hugegraph-server/conf/graphs \
+    && rm hugegraph.properties && mv hstore.properties.template hugegraph.properties
+RUN sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
```
bitflicker64, line 66: 🧹 Both of these `RUN`s edit only the tree copied on line 62, so they are invalidated together and the split just costs an image layer. This is the one file where that layer is avoidable: the plain server Dockerfile also gained a standalone `sed` layer, but it has a single edit with nothing to fold into.

```suggestion
RUN cd /hugegraph-server/conf/graphs \
    && rm hugegraph.properties && mv hstore.properties.template hugegraph.properties \
    && cd /hugegraph-server \
    && sed -i "s/^restserver.url.*$/restserver.url=http:\/\/0.0.0.0:8080/g" ./conf/rest-server.properties
```

The comment on line 63 then covers only part of what the merged `RUN` does, so it is worth widening at the same time.

docker/bake.hcl
```diff
@@ -19,6 +19,14 @@ variable "MAVEN_ARGS" {
   default = ""
 }
 
+variable "MAVEN_PROJECTS" {
+  default = "hugegraph-server/hugegraph-dist,hugegraph-pd/hg-pd-dist,hugegraph-store/hg-store-dist"
```
bitflicker64, line 23: 🧹 This default is now a fifth copy of the module list, and nothing checks it against the other four. A drift here would go unnoticed.

- `docker-bake-check` diffs the `AS build` stage of the four Dockerfiles, so a change to `ARG MAVEN_PROJECTS=...` there has to land in all four. It never compares that ARG with this variable, and its jq assertion on `bake --print` checks targets, platforms and outputs but not `args`.
- The `docker-build` job runs plain `docker build --build-arg SOURCE_REVISION=...`, so CI builds with the Dockerfile default.
- `bake` always passes this value, so the images from `bake.hcl` use this copy.

If someone later adds a module to the Dockerfile list (a new dist, say) and misses this line, every CI check still passes, but bake builds a different reactor from the one CI tested. With a module like that, `rm ./hugegraph-*/*.tar.gz` or a runtime `COPY --from=build` could then fail only in the bake flow.

Requested change: keep one source of truth. Either set `default = null` here, which makes bake leave the arg unset so the Dockerfile `ARG` default applies while `MAVEN_PROJECTS=... docker buildx bake` still overrides it, or add a jq assertion to `docker-bake-check` that `.target.pd.args.MAVEN_PROJECTS` matches the `ARG MAVEN_PROJECTS` default in `hugegraph-pd/Dockerfile`. `RUNTIME_DEPS_EPOCH` on line 27 duplicates its default the same way. A drift there would only cause one extra cache miss, but the same change covers it.

bitflicker64 (commented):
Blocking: no. Summary: All four points from the review on 304bea1 are handled at this head: the epoch ARG, the `MAVEN_PROJECTS` build arg, the merged hstore layer, and the `.dockerignore` follow-up note in the README. The build stages are still byte-identical and the runtime reordering keeps image behaviour the same. One small point is left. The default module list now lives in `docker/bake.hcl` as well as in the four Dockerfiles, and CI only checks the Dockerfile copies. Evidence: `git diff 36811483...230580d` and `git diff 304bea1 230580d`. The `docker-bake-check` job in `.github/workflows/docker-build-ci.yml` diffs only the `AS build` stages, and its jq assertion on `bake --print` never reads `args`. The `docker-build` job runs plain `docker build` with only `SOURCE_REVISION`. The red `hstore` check (`VertexCoreTest.testAddVertexWithTtlAndTtlStartTime`) is the TTL timing flake already noted on #3187. That job runs Java core tests and does not use the files changed here, but it needs a rerun.

imbajin (commented):
No blocking findings at `6bac8f274` after two independent global reviews and one adversarial review. Review score: **9/10**. Bake defaults and explicit overrides pass locally; the scoped Maven reactor validates, all four shared build stages match, and all four Docker image builds pass in CI. The five earlier review threads are addressed and resolved. No additional code changes are needed.

The remaining CI failure was Maven Central returning HTTP 502 while CodeQL downloaded `maven-jar-plugin:3.2.0`; I reran the failed job. Please wait for that rerun before merging. A local image rebuild was unavailable because the Docker daemon was not running; image-build evidence comes from CI.

imbajin (dismissed):
+1. The CodeQL rerun has passed and all current checks are green at 6bac8f274. The independent review remains clean (9/10); ready to merge.

imbajin (approved):

bitflicker64 (approved):

PR #3197, opened 2026-09-09
doc: simplify all AGENTS.md & update memories

bitflicker64:
### Visual summary

Documentation sync
## Purpose of the PR

Agent guidance repeated generic instructions and stale repository details across six files.
Documentation requirements also disagreed between AGENTS.md, the PR template and the contribution guide.

## Main changes

- Reduce all six AGENTS.md files to fewer than 100 lines while keeping component relationships,
  common commands and repository-specific pitfalls inline. Link detailed POM, CI and deployment references.
- Require user-visible feature, configuration and deployment changes to ship with matching documentation.
  In-repository docs belong in the same PR; website changes use a paired hugegraph-doc PR.
  Doc - TODO means pending before merge, and a follow-up issue alone does not satisfy the requirement.
- Correct stale Serena notes for deployment topology, distribution output, compiler targets and test scope.
- Update README navigation without imposing a repository-wide prose reflow.

## Validation

- Independent read-only review of the final diff: no remaining substantive findings.
- All six AGENTS.md files are under 100 lines; their local links resolve.
- git diff --check passed.
- Documentation-only change; no runtime tests executed.

## Documentation Status

- [x] Doc - TODO: synchronize the website contribution guidelines before merging.
- [ ] Doc - Done
- [ ] Doc - No Need

Documentation in this PR: AGENTS.md, module AGENTS.md files, docs/CONTRIBUTING.md,
README.md, the PR template and Serena notes.
Paired hugegraph-doc PR: pending.

Related repositories may adopt the policy separately; their full PR templates are not byte-identical.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -65,3 +65,5 @@ For example:
 - [ ]  `Doc - TODO` <!-- Your PR changes impact docs and you will update later -->
 - [ ]  `Doc - Done` <!-- Related docs have been already added or updated -->
 - [ ]  `Doc - No Need` <!-- Your PR changes don't impact/need docs -->
+
+Related hugegraph-doc PR or issue: <!-- required for Doc - TODO and Doc - Done -->
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,6 +83,18 @@ mvn test -pl hugegraph-store/hg-store-test -am
 
 Before writing new tests, check existing suites under `hugegraph-server/hugegraph-test/`.
 
+## Design philosophy
+
+Match the codebase, do not improve on it.
+
+- Extend existing code before introducing a new abstraction. A new interface, factory
+  or base class needs a reason beyond symmetry.
+- Keep the change minimal: the smallest edit that fully solves the issue. Adjacent
+  cleanups belong in their own PR.
+- One concern per PR. Do not bundle an unrelated fix because you were already in the file.
+- Comments stay short. When something needs real context to explain, link the GitHub
+  issue instead of writing a paragraph in the source.
+
 ## Style & Pre-commit
 
 - Line 120, 4-space indent, LF, UTF-8, **no star imports**
@@ -101,6 +113,10 @@ Before writing new tests, check existing suites under `hugegraph-server/hugegrap
   and `install-dist/scripts/dependency/known-dependencies.txt`.
 - `hugegraph-commons` is shared by every module; `hugegraph-struct` must precede PD/Store;
   server backends depend on `hugegraph-core`.
+- Feature, config and `Dockerfile` changes ship a paired PR in `apache/hugegraph-doc`.
+  Land both together; docs that trail the code go stale without anyone noticing.
+- README: wrap at 120, keep it thin, push detail to hugegraph-doc, and fold the
+  non-urgent parts into `<details>`.
 
 ## Additional context files
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -125,6 +125,12 @@ Note that since GitHub requires submitting code through `username + token` (inst
 
 Go to the web page of GitHub fork repo, there would be a chance to create a Pull Request after pushing to a new branch, click the button "Compare & pull request" to do it. Then edit the description for proposed changes, which can just be copied from the commit message.
 
+If the change touches a feature, a configuration item or a `Dockerfile`, open the matching
+documentation PR in [hugegraph-doc](https://github.com/apache/hugegraph-doc) at the same time
+and link the two together. Both should merge together. Documentation that trails the code
+goes stale quietly, and the gap is only found when a user or a search tool reads the wrong
+thing.
+
 Note: please make sure the email address you used to submit the code is bound to the GitHub account. For how to bind the email address, please refer to https://github.com/settings/emails:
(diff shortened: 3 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -65,3 +65,5 @@ For example:
 - [ ]  `Doc - TODO` <!-- Your PR changes impact docs and you will update later -->
 - [ ]  `Doc - Done` <!-- Related docs have been already added or updated -->
 - [ ]  `Doc - No Need` <!-- Your PR changes don't impact/need docs -->
+
+Related hugegraph-doc PR or issue: <!-- required for Doc - TODO and Doc - Done -->
```
bitflicker64, line 69: ⚠️ Two things work against this field being filled in.

1. Line 63, directly above, still says `DO NOT REMOVE THIS SECTION. CHECK THE PROPER BOX ONLY.` A contributor following that ticks a box and leaves this line empty.
2. `Doc - Done` is defined as "Related docs have been already added or updated", which also covers in-repo docs (`README.md`, module READMEs, `docs/`, `AGENTS.md`). This PR is itself an example. For those, a hugegraph-doc link does not exist, yet the hint marks it required.

The hint is also an HTML comment, so it disappears once the PR is rendered, and reviewers never see that the field was meant to be required.

Suggested change: reword line 63 to cover the field too, for example `CHECK THE PROPER BOX AND FILL IN THE LINK BELOW`, and scope the hint to `required for Doc - TODO; for Doc - Done, when the docs live in hugegraph-doc`.

AGENTS.md
```diff
@@ -101,6 +113,10 @@ Before writing new tests, check existing suites under `hugegraph-server/hugegrap
   and `install-dist/scripts/dependency/known-dependencies.txt`.
 - `hugegraph-commons` is shared by every module; `hugegraph-struct` must precede PD/Store;
   server backends depend on `hugegraph-core`.
+- Feature, config and `Dockerfile` changes ship a paired PR in `apache/hugegraph-doc`.
```
bitflicker64, line 116: 🧹 The rule here and the rule in `docs/CONTRIBUTING.md` are not the same as the one in the template. This line and the CONTRIBUTING paragraph both say the paired hugegraph-doc **PR** lands together with the code. `.github/PULL_REQUEST_TEMPLATE.md` then accepts `hugegraph-doc PR or issue` and keeps `Doc - TODO` ("you will update later"), so an issue link plus deferred docs passes the template. An agent reading this file will hold contributors to a stricter rule than the one they are shown.

Suggested change: choose one rule and write it the same way in all three files. Either allow a tracking issue here and in CONTRIBUTING, or drop `or issue` from the template field.

AGENTS.md
```diff
@@ -101,6 +113,10 @@ Before writing new tests, check existing suites under `hugegraph-server/hugegrap
...
+  Land both together; docs that trail the code go stale without anyone noticing.
+- README: wrap at 120, keep it thin, push detail to hugegraph-doc, and fold the
```
bitflicker64, line 118: 🧹 `wrap at 120` does not describe the current `README.md`. At this head, 47 of its 390 lines are longer than 120 columns. That count covers the badge rows, the module table (L114-116) and the ASCII architecture diagram, none of which can wrap, and also plain prose (L25-27, L34, L175, L204-208). Given as an agent instruction, this invites reflowing the whole README in unrelated PRs, which cuts against the `Keep the change minimal` rule added above. It is also a formatting rule rather than a cross-module dependency, so it fits `Style & Pre-commit` better than this list.

Suggested change: limit it to new or edited prose (`wrap new prose at 120; tables, links and diagrams are exempt`) and move it under `Style & Pre-commit`.

bitflicker64 (commented):
Blocking: no. Summary: The design-philosophy section and the doc-sync paragraph are additive and read well, but the three edited files disagree about what the rule actually is, and the README guidance does not match the README it describes. Separately, open point 2 in the description says the PR template is byte identical in hugegraph-toolchain and hugegraph-computer; it is not (the Documentation Status block is identical, the rest of the file differs in several places), so that follow-up is a hand merge rather than a copy. Evidence: `git show 96a48ad:.github/PULL_REQUEST_TEMPLATE.md`, `AGENTS.md`, `docs/CONTRIBUTING.md` and `README.md` at head; `awk 'length>120' README.md` returns 47 lines, including prose; `diff` of `.github/PULL_REQUEST_TEMPLATE.md` fetched via `gh api` from apache/hugegraph, apache/hugegraph-toolchain and apache/hugegraph-computer default branches.

.github/PULL_REQUEST_TEMPLATE.md
imbajin: Addressed in 899f0722c. The template now visibly asks contributors to select a status and provide the documentation location. In-repository files are accepted; a paired hugegraph-doc PR is needed only when website documentation is affected.

AGENTS.md
imbajin: Addressed in 899f0722c. AGENTS.md, CONTRIBUTING.md and the template now agree: required docs must be ready before merge; in-repository docs ship in the same PR, affected website docs use a coordinated paired PR, and a follow-up issue alone is insufficient. TODO remains a pending state, not permission to defer beyond merge.

AGENTS.md
imbajin: Addressed in 899f0722c. Removed the blanket README wrapping instruction. README remains a concise entry point; only two navigation descriptions were adjusted, without unrelated reflow.

imbajin (approved):

PR #3204, opened 2026-09-10
fix(store-client): enhance thread interrupts in NodeTxExecutor

### LINKED ISSUE #3199: [Bug] A stalled HStore node holds REST workers for minutes: store-client commit retry swallows interrupts
### Bug Type (问题类型)

server status (启动/运行异常)

### Before submit

### Environment (环境信息)

- Server Version: apache/hugegraph master `36811483`; the behaviour is identical in 1.7.0 and 1.5.0 (`HgStoreClientConst.NODE_MAX_RETRYING_TIMES = 10` and `HgStoreClientConfig.GRPC_DEFAULT_TIMEOUT_SECONDS = 100` are unchanged since 1.5.0)
- Backend: hstore, PD + 3 store nodes on 3 VMs (Debian, Temurin 17 for PD/store, Temurin 11 for the server), server with 16 REST workers (8 CPUs), `restserver.request_timeout` at its default 30 s
- Also observed on a separate POC deployment with `grpc.timeout.seconds=60`: `PUT /graph/vertices/batch` → 500 `DEADLINE_EXCEEDED: deadline exceeded after 59.97s ... remote_addr=<store>:8500` in `HgStoreStreamBlockingStub.scanBatchOneShot`, after which the REST server stopped answering any request

**Expected:** when one store node stops answering (GC pause, write stall, hung process), requests that touch its partitions fail after one gRPC deadline or after `restserver.request_timeout`, and the server keeps serving everything else.

**Actual:** every write whose commit lands on a partition of the unreachable store holds a REST worker thread for `11 × grpc.timeout.seconds + 38 s` (about 19 minutes on defaults, 12 minutes with 60 s), and `restserver.request_timeout` does not stop it. At a moderate write rate the worker pool fills within tens of seconds and `LoadDetectFilter` answers 503 to everything, reads and writes to healthy partitions included, and keeps doing so for minutes after the client stops writing.

Cause, `hg-store-client`, `NodeTxExecutor.retryingInvoke()`:

1. The commit is retried up to `NODE_MAX_RETRYING_TIMES = 10` (a constant, not configurable) with a sleep schedule of 1, 1, 1, 2, 3, 4, 5, 6, 7, 8 s, and every attempt is a blocking `HgStoreSessionBlockingStub.batch` call with the `grpc.timeout.seconds` deadline. After a `DEADLINE_EXCEEDED` the next attempt only waits the full deadline again.
2. The `InterruptedException` from `Thread.sleep` is caught, logged as `Failed to sleep`, and the loop continues. The interrupt that `restserver.request_timeout` sends to the Grizzly worker is therefore swallowed, and the REST request time limit is ineffective on exactly this path.

Related, not addressed here: a multi-id read (`HstoreTable.query` → `getWithBatch` → `batchPrefix` → one blocking `scanBatchOneShot` per store) has no retry and no isolation of a single store, so one unreachable node fails the whole read, including the ids that live on healthy stores. That is the path in the POC stack trace above; it does not change the conclusion, because the interrupt does work there.

**Reproduction** (deterministic, no need to wait for a real outage):

```
# 1. on one store node
kill -STOP $(pgrep -f "Dname=HugeGraphStore")
# 2. on the server: one POST /graph/vertices per second for 300 s, each in its own thread,
#    and a probe GET /graph/vertices/"<id>" every 2 s, also after the writes stop
# 3. kill -CONT
```

Measured on the cluster above with `grpc.timeout.seconds=20` so that one run fits in minutes:

| | master `36811483` | with the fix below |
|---|---|---|
| REST unavailable (probe gets 503) | 431 of 503 s | 12 of 300 s |
| REST still dead after the writer stopped | 200 s | 0 s |
| 300 writes | 27 × 201, 243 × 503, 23 × 500 | 131 × 201, 152 × 500 after 20.0 s, 7 × 503 |
| slowest write | 257 s (= 11 × 20 + 38) | 20.1 s |
| server log | `Failed to sleep` ×30, `reached the upper limit` ×30 | `Failed to sleep` 0, `Not retrying after` ×152 |

`kill -CONT` restores the cluster without a restart in both cases. Scripts and full logs: [hugegraph-validation, finding F15](https://github.com/SebastianGruza/hugegraph-validation/blob/master/docs/findings.md#f15).

**Proposed fix** (PR in preparation): in `retryingInvoke` abort the loop when the thread is interrupted (restoring the interrupt flag) and do not retry after `DEADLINE_EXCEEDED` or `CANCELLED`; keep retrying `UNAVAILABLE` and other transport errors as today, since #3130 relies on that. Separately worth considering: moving the attempt count and the schedule into `HgStoreClientConfig`.

Any vertex write whose id maps to a partition of the unreachable store; in the measurement, label `node` with `CUSTOMIZE_STRING` ids and no properties.

`vertexLabel("node").useCustomizeStringId().create()`; the problem does not depend on the schema.

SebastianGruza:
## Purpose

Fixes #3199 by bounding HStore client retry latency and preserving cancellation/interrupt semantics when a store node stops responding.

NodeTxExecutor retry flow

Before this change, a stalled RPC could be retried up to 10 times. With `grpc.timeout.seconds=100`, repeated deadlines and backoff could hold one caller for about 19 minutes. `Thread.sleep` interruptions were swallowed, so REST `request_timeout` and Gremlin `evaluationTimeout` could not release the worker. Under load, one stalled store could exhaust the REST worker pool and cause cascading 503 responses.

## Core changes

- Classify failures across cause chains and suppressed parallel-commit failures:
  - `FATAL`: thread interrupt or `CANCELLED`; stop immediately, restore the interrupt flag, and keep `InterruptedException` as the root cause.
  - `DEADLINE`: `DEADLINE_EXCEEDED` may retry once so refreshed partition leaders can be used; the second deadline in the same call fails, even if other failures occur between them.
  - `RETRYABLE`: `UNAVAILABLE`, `NOT_LEADER`, and other transport failures retain the existing bounded-backoff retry behavior.
- Collect all parallel session failures and aggregate them with suppressed causes, so one racing partition cannot decide the retry path for the whole commit.
- Keep the existing retry count and backoff for retryable transport failures.

## Problem solved

- A stalled call is bounded to at most two RPC deadlines instead of up to 11.
- Request and evaluation interrupts can release the calling thread.
- Store-replacement and replicated-leader recovery remain retryable.
- Parallel-commit retry decisions are deterministic and retain all failure evidence.

## Verification

- `NodeTxExecutorTest`: 11/11 passed on Temurin 17.
- Exact-head command:
  ```bash
  mvn test -pl hugegraph-store/hg-store-test -am -P store-client-test -Dtest=NodeTxExecutorTest -DfailIfNoTests=false
  ```
- No public API or configuration change.

### DIFF
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
@@ -22,13 +22,18 @@ import static org.apache.hugegraph.store.client.util.HgStoreClientConst.NODE_MAX
 import static org.apache.hugegraph.store.client.util.HgStoreClientConst.TX_SESSIONS_MAP_CAPACITY;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Queue;
+import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collector;
@@ -131,35 +136,7 @@ final class NodeTxExecutor {
                 if (!allSuccess.get()) {
                     throw HgStoreClientException.of(msg);
                 }
-                AtomicReference<Throwable> throwable = new AtomicReference<>();
-                Collection<HgStoreSession> sessions = this.sessions.values();
-                sessions.parallelStream().forEach(e -> {
-                    if (e.isTx()) {
-                        try {
-                            e.commit();
-                        } catch (Throwable t) {
-                            throwable.compareAndSet(null, t);
-                            allSuccess.set(false);
-                        }
-                    }
-                });
-                if (!allSuccess.get()) {
-                    if (isTx) {
-                        try {
-                            sessions.stream().forEach(HgStoreSession::rollback);
-                        } catch (Exception e) {
-
-                        }
-                    }
-                    Throwable cause = throwable.get();
-                    if (cause.getCause() != null) {
-                        cause = cause.getCause();
-                    }
-                    if (cause instanceof HgStoreClientException) {
-                        throw (HgStoreClientException) cause;
-                    }
-                    throw HgStoreClientException.of(cause);
-                }
+                this.commitSessions(this.sessions.values());
                 return true;
             });
 
@@ -233,6 +210,52 @@ final class NodeTxExecutor {
     //        }
     //    }
     // };
+
+    /**
+     * Commit every tx session in parallel. When at least one commit fails, roll back (in tx
+     * mode) and throw one exception that carries ALL failures: the first one as the cause, the
+     * others as suppressed. The retry loop looks at all of them, so whether a commit is retried
+     * no longer depends on which partition happened to fail first.
+     */
+    void commitSessions(Collection<HgStoreSession> sessions) {
+        Queue<Throwable> failures = new ConcurrentLinkedQueue<>();
+        sessions.parallelStream().forEach(e -> {
+            if (e.isTx()) {
+                try {
+                    e.commit();
+                } catch (Throwable t) {
+                    failures.add(t);
+                }
+            }
+        });
+        if (failures.isEmpty()) {
+            return;
+        }
+        if (isTx) {
+            try {
+                sessions.stream().forEach(HgStoreSession::rollback);
+            } catch (Exception e) {
+                // keep the commit failure as the reported one
+            }
+        }
+        throw aggregate(failures);
+    }
+
+    static HgStoreClientException aggregate(Collection<Throwable> failures) {
+        Iterator<Throwable> it = failures.iterator();
+        Throwable first = it.next();
+        Throwable cause = first.getCause() != null ? first.getCause() : first;
+        HgStoreClientException result = cause instanceof HgStoreClientException ?
+                                        (HgStoreClientException) cause :
+                                        HgStoreClientException.of(cause);
+        while (it.hasNext()) {
+            Throwable other = it.next();
+            if (other != result && other != cause) {
+                result.addSuppressed(other);
+            }
+        }
+        return result;
+    }
 
     private boolean doAction(HgTriple<String, HgOwnerKey, Object> nodeParams,
                              Function<NodeTkv, Boolean> action) {
@@ -389,17 +412,17 @@ final class NodeTxExecutor {
                                     try {
                                         buffer = supplier.get();
                                     } catch (Throwable t) {
-                                        if (i + 1 > NODE_MAX_RETRYING_TIMES) {
-                                            log.error(maxTryMsg, t);
-                                            throw HgStoreClientException.of(
-                                                    t.getMessage(), t);
-                                        }
                                         if (!isRetryable(t)) {
                                             // A deadline or a cancellation will not
                                             // get better by waiting the full deadline
                                             // again; fail fast and let the caller decide.
                                             log.warn("Not retrying after: {}",
-                                                     t.getMessage());
+                                                     t.getMessage(), t);
+                                            throw HgStoreClientException.of(
+                                                    t.getMessage(), t);
+                                        }
+                                        if (i + 1 > NODE_MAX_RETRYING_TIMES) {
+                                            log.error(maxTryMsg, t);
                                             throw HgStoreClientException.of(
                                                     t.getMessage(), t);
                                         }
@@ -428,11 +451,16 @@ final class NodeTxExecutor {
     /**
      * Retry only failures that a fresh attempt can plausibly fix (a transport error, a
      * leader change). A DEADLINE_EXCEEDED would wait the full deadline again, a CANCELLED
-     * means the caller's thread was interrupted; neither is retried.
+     * means the caller's thread was interrupted; neither is retried, also when it is one of
+     * several failures of a parallel commit.
      */
     static boolean isRetryable(Throwable t) {
+        return isRetryable(t, Collections.newSetFromMap(new IdentityHashMap<>()));
+    }
+
+    private static boolean isRetryable(Throwable t, Set<Throwable> seen) {
         Throwable c = t;
-        while (c != null) {
+        while (c != null && seen.add(c)) {
             if (c instanceof InterruptedException) {
                 return false;
             }
@@ -442,8 +470,12 @@ final class NodeTxExecutor {
                     return false;
                 }
             }
-            if (c.getCause() == c) {
-                break;
+            // A parallel commit reports the other partitions' failures as suppressed;
+            // one non-retryable failure among them makes the whole attempt non-retryable.
+            for (Throwable suppressed : c.getSuppressed()) {
+                if (!isRetryable(suppressed, seen)) {
+                    return false;
+                }
             }
             c = c.getCause();
         }
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
@@ -406,8 +406,12 @@ final class NodeTxExecutor {
                                         // The caller (e.g. a REST worker hitting
                                         // restserver.request_timeout) gave up: stop
                                         // retrying instead of holding its thread.
+                                        // InterruptedException as the root cause: the
+                                        // server's task cancel path recognises it
+                                        // (HugeException.isInterrupted()).
                                         throw HgStoreClientException.of(
-                                                "Interrupted before retry " + i);
+                                                "Interrupted before retry " + i,
+                                                new InterruptedException());
                                     }
                                     T buffer = null;
                                     try {
@@ -453,9 +457,10 @@ final class NodeTxExecutor {
                                             Thread.sleep(sleepTime * 1000L);
                                         } catch (InterruptedException e) {
                                             Thread.currentThread().interrupt();
+                                            e.addSuppressed(t);
                                             throw HgStoreClientException.of(
                                                     "Interrupted while waiting to retry: " +
-                                                    t.getMessage(), t);
+                                                    t.getMessage(), e);
                                         }
                                     }
                                     return buffer;
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
@@ -33,7 +33,6 @@ import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collector;
@@ -65,8 +64,6 @@ final class NodeTxExecutor {
     private static final String maxTryMsg =
             "the number of retries reached the upper limit : " + NODE_MAX_RETRYING_TIMES +
             ",caused by:";
-    private static final String msg =
-            "Not all tx-data delivered to real-node-session successfully.";
 
     static {
         System.setProperty("java.util.concurrent.ForkJoinPool.common.parallelism",
@@ -128,13 +125,9 @@ final class NodeTxExecutor {
                 if (this.entries.isEmpty()) {
                     return true;
                 }
-                AtomicBoolean allSuccess = new AtomicBoolean(true);
                 for (HgPair<HgTriple<String, HgOwnerKey, Object>, Function<NodeTkv, Boolean>> e :
                         this.entries) {
                     doAction(e.getKey(), e.getValue());
-                }
-                if (!allSuccess.get()) {
-                    throw HgStoreClientException.of(msg);
                 }
                 this.commitSessions(this.sessions.values());
                 return true;
@@ -418,6 +411,11 @@ final class NodeTxExecutor {
                                         buffer = supplier.get();
                                     } catch (Throwable t) {
                                         Failure failure = classify(t);
+                                        if (failure != Failure.DEADLINE) {
+                                            // a different failure in between means the
+                                            // next deadline is not "in a row" again
+                                            deadlineRetried[0] = false;
+                                        }
                                         if (failure == Failure.FATAL) {
                                             // The caller's thread was interrupted or the
                                             // call was cancelled: fail fast.
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
@@ -398,6 +398,7 @@ final class NodeTxExecutor {
     }
 
     <T> Optional<T> retryingInvoke(Supplier<T> supplier) {
+        boolean[] deadlineRetried = {false};
         return IntStream.rangeClosed(0, NODE_MAX_RETRYING_TIMES).boxed()
                         .map(
                                 i -> {
@@ -412,14 +413,31 @@ final class NodeTxExecutor {
                                     try {
                                         buffer = supplier.get();
                                     } catch (Throwable t) {
-                                        if (!isRetryable(t)) {
-                                            // A deadline or a cancellation will not
-                                            // get better by waiting the full deadline
-                                            // again; fail fast and let the caller decide.
+                                        Failure failure = classify(t);
+                                        if (failure == Failure.FATAL) {
+                                            // The caller's thread was interrupted or the
+                                            // call was cancelled: fail fast.
                                             log.warn("Not retrying after: {}",
                                                      t.getMessage(), t);
                                             throw HgStoreClientException.of(
                                                     t.getMessage(), t);
+                                        }
+                                        if (failure == Failure.DEADLINE) {
+                                            // One retry: the NOT_WORK notice sent for the
+                                            // failed RPC reloads the partition leaders, so
+                                            // the next attempt can reach a new leader. A
+                                            // second deadline in a row would only wait the
+                                            // full deadline again on the same stalled store.
+                                            if (deadlineRetried[0]) {
+                                                log.warn("Not retrying a second deadline: {}",
+                                                         t.getMessage(), t);
+                                                throw HgStoreClientException.of(
+                                                        t.getMessage(), t);
+                                            }
+                                            deadlineRetried[0] = true;
+                                            log.warn("Deadline exceeded, retrying once in " +
+                                                     "case the partition leader moved: {}",
+                                                     t.getMessage());
                                         }
                                         if (i + 1 > NODE_MAX_RETRYING_TIMES) {
                                             log.error(maxTryMsg, t);
@@ -449,37 +467,51 @@ final class NodeTxExecutor {
     }
 
     /**
-     * Retry only failures that a fresh attempt can plausibly fix (a transport error, a
-     * leader change). A DEADLINE_EXCEEDED would wait the full deadline again, a CANCELLED
-     * means the caller's thread was interrupted; neither is retried, also when it is one of
-     * several failures of a parallel commit.
+     * How a failed attempt is treated by {@link #retryingInvoke}: FATAL is never retried (the
+     * caller's thread was interrupted, or the call was cancelled), DEADLINE is retried exactly
+     * once (the partition leader may have moved after the failed RPC invalidated the partition
+     * cache; a second deadline in a row would only wait the full deadline again on the same
+     * stalled store), everything else (transport errors, NOT_LEADER, store replacement) is
+     * retried up to NODE_MAX_RETRYING_TIMES as before. For a parallel commit the failures of
+     * the other partitions arrive as suppressed exceptions and are classified too; the most
+     * severe class wins.
      */
-    static boolean isRetryable(Throwable t) {
-        return isRetryable(t, Collections.newSetFromMap(new IdentityHashMap<>()));
+    enum Failure {
+        RETRYABLE, DEADLINE, FATAL
     }
 
-    private static boolean isRetryable(Throwable t, Set<Throwable> seen) {
+    static Failure classify(Throwable t) {
+        return classify(t, Collections.newSetFromMap(new IdentityHashMap<>()));
+    }
+
+    private static Failure classify(Throwable t, Set<Throwable> seen) {
+        Failure worst = Failure.RETRYABLE;
         Throwable c = t;
         while (c != null && seen.add(c)) {
             if (c instanceof InterruptedException) {
-                return false;
+                return Failure.FATAL;
             }
             if (c instanceof StatusRuntimeException) {
                 Status.Code code = ((StatusRuntimeException) c).getStatus().getCode();
-                if (code == Status.Code.DEADLINE_EXCEEDED || code == Status.Code.CANCELLED) {
-                    return false;
+                if (code == Status.Code.CANCELLED) {
+                    return Failure.FATAL;
+                }
+                if (code == Status.Code.DEADLINE_EXCEEDED) {
+                    worst = Failure.DEADLINE;
                 }
             }
-            // A parallel commit reports the other partitions' failures as suppressed;
-            // one non-retryable failure among them makes the whole attempt non-retryable.
             for (Throwable suppressed : c.getSuppressed()) {
-                if (!isRetryable(suppressed, seen)) {
-                    return false;
+                Failure f = classify(suppressed, seen);
+                if (f == Failure.FATAL) {
+                    return f;
+                }
+                if (f.compareTo(worst) > 0) {
+                    worst = f;
                 }
             }
             c = c.getCause();
         }
-        return true;
+        return worst;
     }
 
     private boolean isValid(Object obj) {
--- a/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
+++ b/hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
@@ -46,6 +46,8 @@ import org.apache.hugegraph.store.client.util.HgStoreClientConst;
 import org.apache.hugegraph.store.term.HgPair;
 import org.apache.hugegraph.store.term.HgTriple;
 
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import lombok.extern.slf4j.Slf4j;
 
 /**
@@ -376,30 +378,42 @@ final class NodeTxExecutor {
         return IntStream.rangeClosed(0, NODE_MAX_RETRYING_TIMES).boxed()
                         .map(
                                 i -> {
+                                    if (Thread.currentThread().isInterrupted()) {
+                                        // The caller (e.g. a REST worker hitting
+                                        // restserver.request_timeout) gave up: stop
+                                        // retrying instead of holding its thread.
+                                        throw HgStoreClientException.of(
+                                                "Interrupted before retry " + i);
+                                    }
                                     T buffer = null;
                                     try {
                                         buffer = supplier.get();
                                     } catch (Throwable t) {
-                                        if (i + 1 <= NODE_MAX_RETRYING_TIMES) {
-                                            try {
-                                                int sleepTime;
-                                                // The first three times try once every second
-                                                if (i < 3) {
-                                                    sleepTime = 1;
-                                                } else {
-                                                    // Subsequent incremental
-                                                    sleepTime = i - 1;
-                                                }
-                                                log.info("Waiting {} seconds " +
-                                                         "for the next try.",
-                                                         sleepTime);
-                                                Thread.sleep(sleepTime * 1000L);
-                                            } catch (InterruptedException e) {
-                                                log.error("Failed to sleep", e);
-                                            }
-                                        } else {
+                                        if (i + 1 > NODE_MAX_RETRYING_TIMES) {
                                             log.error(maxTryMsg, t);
                                             throw HgStoreClientException.of(
+                                                    t.getMessage(), t);
+                                        }
+                                        if (!isRetryable(t)) {
+                                            // A deadline or a cancellation will not
+                                            // get better by waiting the full deadline
+                                            // again; fail fast and let the caller decide.
+                                            log.warn("Not retrying after: {}",
+                                                     t.getMessage());
+                                            throw HgStoreClientException.of(
+                                                    t.getMessage(), t);
+                                        }
+                                        // The first three times try once every second,
+                                        // subsequent incremental
+                                        int sleepTime = i < 3 ? 1 : i - 1;
+                                        log.info("Waiting {} seconds for the next try.",
+                                                 sleepTime);
+                                        try {
+                                            Thread.sleep(sleepTime * 1000L);
+                                        } catch (InterruptedException e) {
+                                            Thread.currentThread().interrupt();
+                                            throw HgStoreClientException.of(
+                                                    "Interrupted while waiting to retry: " +
                                                     t.getMessage(), t);
                                         }
                                     }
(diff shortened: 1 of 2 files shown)


hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -376,32 +378,44 @@ <T> Optional<T> retryingInvoke(Supplier<T> supplier) {
...
-                                            } catch (InterruptedException e) {
-                                                log.error("Failed to sleep", e);
-                                            }
-                                        } else {
+                                        if (i + 1 > NODE_MAX_RETRYING_TIMES) {
                                             log.error(maxTryMsg, t);
                                             throw HgStoreClientException.of(
                                                     t.getMessage(), t);
                                         }
+                                        if (!isRetryable(t)) {
```
bitflicker64: ⚠️ On a commit that spans more than one partition, whether this gate fires is decided by a race, so the fix is not guaranteed to engage on the very workload the issue describes.

Evidence:

- `doCommit()` commits every session in parallel and keeps only the **first** throwable: `sessions.parallelStream().forEach(...)` with `throwable.compareAndSet(null, t)` (`NodeTxExecutor.java:136-145`). Which `ForkJoinPool` task wins is nondeterministic.
- That single throwable is the only one rethrown (`:154-161`), so it is the only input `isRetryable()` ever sees for the whole commit — the other nodes' failures are discarded, not even suppressed.
- So for a commit touching a stalled node (`DEADLINE_EXCEEDED`) and a node being replaced (`UNAVAILABLE`, the #3130 recovery this PR is careful to preserve): if the `UNAVAILABLE` lands first, the commit is retried up to 11 times and each attempt still blocks on the stalled partition for the full `grpc.timeout.seconds` (`AbstractGrpcClient.java:128`). That is `11 × grpc.timeout.seconds` of held caller thread — the stall from #3199, unchanged.
- Nothing pins this: all four new tests use a supplier with a single failure mode, and `testTransientFailureIsStillRetried` / `testDeadlineExceededIsNotRetried` each exercise one node's worth of behaviour.

Requested change: collect every failure instead of the first (a `ConcurrentLinkedQueue<Throwable>` or `Collections.synchronizedList`) and make the decision deterministic — retry only when *every* captured failure is retryable, attaching the rest as suppressed on the thrown `HgStoreClientException` — and add a test for a two-session commit where one session fails `DEADLINE_EXCEEDED` and the other `UNAVAILABLE`, asserting exactly one attempt.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -376,32 +378,44 @@ <T> Optional<T> retryingInvoke(Supplier<T> supplier) {
...
+                                            // A deadline or a cancellation will not
+                                            // get better by waiting the full deadline
+                                            // again; fail fast and let the caller decide.
+                                            log.warn("Not retrying after: {}",
```
bitflicker64, line 420: 🧹 This log drops the throwable, on what this change makes the *common* failure path.

Evidence:

- `log.warn("Not retrying after: {}", t.getMessage())` passes only the message, so nothing reaches the logger's throwable slot.
- The branch four lines up keeps it: `log.error(maxTryMsg, t)` (`:393`). Before this PR that was the only place a store failure was logged with its cause chain; after it, a stalled node takes this branch on the first attempt and never reaches `:393`, so the nested `StatusRuntimeException` and its stack trace are no longer in the server log at all.

Requested change: `log.warn("Not retrying after: {}", t.getMessage(), t);` — slf4j appends a trailing `Throwable` argument as the cause.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -376,32 +378,44 @@ <T> Optional<T> retryingInvoke(Supplier<T> supplier) {
...
-                                                log.info("Waiting {} seconds " +
-                                                         "for the next try.",
-                                                         sleepTime);
-                                                Thread.sleep(sleepTime * 1000L);
-                                            } catch (InterruptedException e) {
-                                                log.error("Failed to sleep", e);
-                                            }
-                                        } else {
+                                        if (i + 1 > NODE_MAX_RETRYING_TIMES) {
                                             log.error(maxTryMsg, t);
```
bitflicker64: 🧹 Ordering the attempt-budget check before the retryability check makes the last attempt log the wrong reason.

Evidence:

- At `i == NODE_MAX_RETRYING_TIMES` the loop takes `:392-396` first, so a `DEADLINE_EXCEEDED` on the final attempt is reported as `"the number of retries reached the upper limit : 10"` (`maxTryMsg`, `:58-60`) even though the new gate at `:397` is what should have described it.
- This is reachable in the scenario the PR targets: attempts 0-9 fail `UNAVAILABLE` while a store is being replaced (still retried by design), then attempt 10 hits the deadline on the new node.

Requested change: move the `if (!isRetryable(t))` block above the `i + 1 > NODE_MAX_RETRYING_TIMES` block so each exit logs the reason that actually applied.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutorTest.java
```diff
@@ -115,4 +123,67 @@ public void testParallelReplacementUsesOneCurrentSession() throws Exception {
...
+        NodeTxExecutor executor = NodeTxExecutor.graphOf("graph", null);
+        AtomicInteger attempts = new AtomicInteger();
+        try {
+            assertThrows(HgStoreClientException.class, () ->
+                    executor.retryingInvoke(() -> {
+                        attempts.incrementAndGet();
+                        Thread.currentThread().interrupt();
+                        throw new RuntimeException("simulated transport failure");
+                    }));
+            assertEquals(1, attempts.get());
```
bitflicker64, line 195: 🧹 The new pre-attempt interrupt guard has no coverage; this test exercises only the sleep handler.

Evidence:

- The supplier increments `attempts`, *then* sets the flag, then throws (`:163-165`). So the `i == 0` guard at `NodeTxExecutor.java:381-387` runs on a clean thread, `isRetryable(new RuntimeException(...))` is true, and the abort comes from the `catch (InterruptedException)` around `Thread.sleep` at `NodeTxExecutor.java:411-418`.
- This assertion proves it: `assertEquals(1, attempts.get())`. Had the guard fired, the supplier would never have run and `attempts` would be `0`.
- The untested branch is the one that matters for a caller whose `restserver.request_timeout` expired *between* store calls rather than during one — the case the guard's own comment cites.

Requested change: add a case that sets the flag before the call, e.g. `Thread.currentThread().interrupt(); assertThrows(HgStoreClientException.class, () -> executor.retryingInvoke(() -> { attempts.incrementAndGet(); return "ok"; })); assertEquals(0, attempts.get());`, with the same `finally { Thread.interrupted(); }` guard this test already uses.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutorTest.java
```diff
@@ -115,4 +123,67 @@ public void testParallelReplacementUsesOneCurrentSession() throws Exception {
...
     }
+
+    @Test
+    public void testIsRetryableClassifiesFailures() {
+        assertFalse(NodeTxExecutor.isRetryable(Status.DEADLINE_EXCEEDED.asRuntimeException()));
+        assertFalse(NodeTxExecutor.isRetryable(Status.CANCELLED.asRuntimeException()));
+        assertFalse(NodeTxExecutor.isRetryable(new InterruptedException("interrupted")));
+        // The status is usually wrapped by the time it reaches the retry loop
+        assertFalse(NodeTxExecutor.isRetryable(HgStoreClientException.of(
+                "commit failed", new RuntimeException(Status.DEADLINE_EXCEEDED.asRuntimeException()))));
```
bitflicker64: 🧹 Two style nits in the new test code.

Evidence:

- This line is 104 characters, over the project's `LineLength` max of 100 (`style/checkstyle.xml:29-32`). It is the only added line in the diff that exceeds it.
- `import io.grpc.StatusRuntimeException` (`:43`) is unused: the tests only call `Status.<CODE>.asRuntimeException()` and never name the type. `UnusedImports` is enabled at `style/checkstyle.xml:59`.
- Neither breaks the build — the checkstyle plugin is bound only in `hugegraph-server/pom.xml:295` and `hugegraph-commons/pom.xml:147`, not in the store modules — so this is purely to keep the new code inside the shared style.

Requested change: wrap the argument list onto a second line and drop the unused import.

bitflicker64 (commented):
Blocking: no. Summary: the fix holds at `4350ef99` — the exception the retry loop sees really does carry the gRPC status (`NotifyingExecutor.invoke()` rethrows `err(t)` = `HgStoreClientException.of(msg, t)` at `NotifyingExecutor.java:74-82`, `GrpcStoreNodeSessionImpl.commit()` wraps it in `RuntimeException(t)` at `:139-141`, and `doCommit()` unwraps one level at `NodeTxExecutor.java:154-161`), and no recovery path is lost by dropping `DEADLINE_EXCEEDED` from the retry set, because the node-invalidation notice that would make a retry resolve a different leader only fires for `UNAVAILABLE` (`NotifyingExecutor.java:244-255`) while leader moves arrive in-band as `PARTITION_FAULT_TYPE_NOT_LEADER` on an `HgStoreClientException` with no gRPC cause (`:150-171`), which `isRetryable()` still retries. Interrupt-flag hygiene is safe for pooled callers: Grizzly clears the flag at the top of every task (`AbstractThreadPool.Worker.doWork()`, grizzly-framework 3.0.1 line 526), as does `ThreadPoolExecutor.runWorker`. Five comments below, none blocking; the important one is that `doCommit()` keeps only the first of several parallel commit failures, so on a mixed-failure commit whether the new fail-fast engages is decided by a race. Evidence: static review of the exact-head diff (2 files, +128/-18) against merge-base `36811483`; suite wiring checked (`ClientSuiteTest` lists `NodeTxExecutorTest.class`, `hg-store-test/pom.xml:227-237` includes `**/ClientSuiteTest.java`, run by `pd-store-ci.yml:286`). No CI signal at this head — all seven workflows are `action_required` awaiting maintainer approval, so `gh api repos/apache/hugegraph/commits/4350ef99.../status` is `pending` and `check-runs` is empty. The new tests could not be executed locally: `mvn test -pl hugegraph-store/hg-store-test -am -P store-client-test` fails in the untouched `hugegraph-commons/hugegraph-common` module with lombok annotations unprocessed, so the author's 6/6 result is unverified here.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Done in 8adf522. `commitSessions()` now collects every session failure in a `ConcurrentLinkedQueue` and throws one `HgStoreClientException` (first failure unwrapped one level as before, the others attached as suppressed); `isRetryable()` walks the suppressed failures as well as the cause chain (identity-set guarded against cycles), so one `DEADLINE_EXCEEDED` among several failures makes the attempt non-retryable regardless of which partition reported first. Tests: `testMixedCommitFailuresAreNotRetried` (two sessions, `UNAVAILABLE` + `DEADLINE_EXCEEDED` thrown fresh on every call, exactly one attempt, both sessions rolled back, one suppressed) and `testAllRetryableCommitFailuresAreRetried`.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Done in 8adf522 — the throwable is passed as the trailing argument.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Done in 8adf522 — `isRetryable()` is checked first, the attempt budget second.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutorTest.java
SebastianGruza: Done in 8adf522 — `testInterruptBeforeCallSkipsTheAttempt` sets the flag before the call and asserts `attempts == 0` with the flag still set.

hugegraph-store/hg-store-test/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutorTest.java
SebastianGruza: Done in 8adf522 — line wrapped, unused import removed; no added line exceeds 100 characters now.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -411,6 +448,40 @@ <T> Optional<T> retryingInvoke(Supplier<T> supplier) {
...
+
+    private static boolean isRetryable(Throwable t, Set<Throwable> seen) {
+        Throwable c = t;
+        while (c != null && seen.add(c)) {
+            if (c instanceof InterruptedException) {
+                return false;
+            }
+            if (c instanceof StatusRuntimeException) {
+                Status.Code code = ((StatusRuntimeException) c).getStatus().getCode();
+                if (code == Status.Code.DEADLINE_EXCEEDED || code == Status.Code.CANCELLED) {
```
bitflicker64: ⚠️ Not retrying `DEADLINE_EXCEEDED` also drops a leader-failover recovery. The Javadoc's reasoning (a retry waits the full deadline again) holds when the retry goes to the same store, which is not the case once the partition leader has moved. This corrects my earlier summary, which only looked at node eviction.

Evidence:

- Every RPC error sends a `NOT_WORK` notice (`NotifyingExecutor.java:117-123`, `:244-255`). `HgStoreNodePartitionerImpl.notice()` then calls `pdClient.invalidPartitionCache()` (`HgStoreNodePartitionerImpl.java:190-195`), which reloads shard-group leaders from PD right away (`ClientCache.java:220-230`, `:178-186`).
- The retried `doCommit()` supplier re-runs `doAction()` for every entry (`NodeTxExecutor.java:132-135`), so it routes against the reloaded leaders (`NodeTxSessionProxy.java:731-734`, `:862-866`).
- A new raft leader is elected after the 3 s election timeout (`HgStoreEngineOptions.java:91`) and pushes its shard group to PD (`PartitionEngine.java:605-681`), well inside the 100 s default `grpc.timeout.seconds` (`HgStoreClientConfig.java:30`).

So with replicated partitions, attempt 1 used to reach the new leader; now the call fails after attempt 0. Interrupted REST workers still stop after one deadline either way, but callers with no timeout that interrupts them (async jobs are only interrupted on cancel) lose the recovery. Was the SIGSTOP cluster at the shipped `default-shard-count: 1` (`hugegraph-pd/hg-pd-dist/src/assembly/static/conf/application.yml:95`)? With 3 replicas the before run should have recovered on attempt 1.

Requested change: allow at most one `DEADLINE_EXCEEDED` retry per `retryingInvoke()` call (keep `CANCELLED` and `InterruptedException` non-retryable), update the Javadoc to match, and make `testDeadlineExceededIsNotRetried` expect two attempts.

bitflicker64 (commented):
Blocking: no. Summary: the earlier follow-ups are in at 8adf522 and the interrupt handling looks right; one inline comment on `DEADLINE_EXCEEDED` no longer riding through a partition leader failover, which corrects what my earlier summary said about that. Evidence: static trace through `NotifyingExecutor`, `HgStoreNodePartitionerImpl`, `ClientCache` and `PartitionEngine` at the exact head; `mvn test -pl hugegraph-store/hg-store-test -am -P store-client-test -Dtest=NodeTxExecutorTest` passes locally on JDK 17 (9 tests, 0 failures). No CI signal at this head: all seven workflows are `action_required`, waiting for maintainer approval.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Agreed, and yes — the SIGSTOP cluster runs with the shipped `default-shard-count: 1`, so it never had a leader to fail over to; with replicas the first retry would indeed have reached the new leader. Done in b588aa8: `retryingInvoke()` now classifies a failure as FATAL (interrupt, `CANCELLED`: never retried), DEADLINE (retried exactly once per call, then failed on a second deadline in a row) or RETRYABLE (unchanged); suppressed failures of a parallel commit are classified as well and the most severe class wins. Javadoc updated. Tests: `testDeadlineExceededIsRetriedExactlyOnce` expects two attempts, `testDeadlineThenNewLeaderSucceeds` covers the failover case, the mixed-commit test now expects two attempts, `testClassifyFailures` replaces the old `isRetryable` test. 10/10 locally on JDK 17.

bitflicker64 (commented):
Blocking: no. Summary: At b588aa88 the retry classification, the commit failure aggregation and the interrupt handling look correct, and every earlier inline point is addressed. One minor issue: two of the three new interrupt exits drop the InterruptedException, and the server's task-cancel path looks for it. Evidence: static review of the full exact-head diff (2 files, +323/-47) against merge-base 60c8803d, traced through NotifyingExecutor, GrpcStoreNodeSessionImpl, HugeException, HugeTask and DistributedTaskScheduler; grpc-stub 1.39.0 ClientCalls bytecode checked for the interrupt cancel cause. Tests were not run locally. No CI signal at this head: all seven workflows are action_required, waiting for maintainer approval.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -373,35 +398,65 @@ boolean ifAnyTrue(Supplier<Stream<HgPair<HgStoreNode, NodeTkv>>> nodeStreamSuppl
...
     <T> Optional<T> retryingInvoke(Supplier<T> supplier) {
+        boolean[] deadlineRetried = {false};
         return IntStream.rangeClosed(0, NODE_MAX_RETRYING_TIMES).boxed()
                         .map(
                                 i -> {
+                                    if (Thread.currentThread().isInterrupted()) {
+                                        // The caller (e.g. a REST worker hitting
+                                        // restserver.request_timeout) gave up: stop
+                                        // retrying instead of holding its thread.
+                                        throw HgStoreClientException.of(
```
bitflicker64, line 409: 🧹 Two of the new interrupt exits drop the `InterruptedException`, so the server cannot tell them apart from a store failure.

Evidence:

- `:409-410` throws `HgStoreClientException.of("Interrupted before retry " + i)` with no cause. `:456-458` uses the store failure `t` as the cause and discards `e`. Neither has an `InterruptedException` as its root cause.
- The FATAL path keeps one. A blocking stub on an interrupted thread fails with `CANCELLED`, and the `InterruptedException` is its cause: grpc-stub 1.39.0 `ClientCalls.blockingUnaryCall` calls `call.cancel("Thread interrupted", e)`.
- `HugeException.isInterrupted()` only checks the root cause (`HugeException.java:56-61`). `HugeTask.fail()` relies on it so that a cancelled task is not recorded as failed (`HugeTask.java:351-355`). Take a task that is cancelled while its thread is between store calls or in the retry sleep. It now logs a WARN with a stack trace. If the worker reaches `fail()` before `cancel()` sets CANCELLED (`:323` interrupts, `:336` sets the status), the task is stored as FAILED and `cancel()` returns false, so `DistributedTaskScheduler.cancel()` does not save CANCELLED (`:317-321`). Before this change the loop swallowed the interrupt and carried on, so this path did not exist.

Requested change: make `InterruptedException` the root cause on both exits. For example, use `HgStoreClientException.of("Interrupted before retry " + i, new InterruptedException())`. In the sleep handler, use `HgStoreClientException.of("Interrupted while waiting to retry: " + t.getMessage(), e)` and attach `t` with `addSuppressed`. Then extend `testInterruptStopsRetrying` and `testInterruptBeforeCallSkipsTheAttempt` to assert the root cause.

SebastianGruza: Done in 172b2d0 — both exits now have an `InterruptedException` as the root cause (`HgStoreClientException.of("Interrupted before retry " + i, new InterruptedException())`; in the sleep handler the `InterruptedException` is the cause and the store failure `t` is attached to it as suppressed). `testInterruptStopsRetrying` and `testInterruptBeforeCallSkipsTheAttempt` assert the root cause, the first one also the suppressed store failure. 10/10 locally.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -373,35 +398,70 @@ boolean ifAnyTrue(Supplier<Stream<HgPair<HgStoreNode, NodeTkv>>> nodeStreamSuppl
...
+                                            // full deadline again on the same stalled store.
+                                            if (deadlineRetried[0]) {
+                                                log.warn("Not retrying a second deadline: {}",
+                                                         t.getMessage(), t);
+                                                throw HgStoreClientException.of(
+                                                        t.getMessage(), t);
                                             }
-                                        } else {
+                                            deadlineRetried[0] = true;
+                                            log.warn("Deadline exceeded, retrying once in " +
```
bitflicker64, line 439: ⚠️ The PR description still describes the pre-`b588aa88` behaviour, so the numbers reviewers and the squash commit message will carry are wrong for this head.

Evidence:

- The body says "Do not retry a failure whose cause chain carries `Status.Code.DEADLINE_EXCEEDED` … (`isRetryable()`)". At this head a deadline **is** retried once (this block, plus the guard four lines down), and `isRetryable()` no longer exists: `git grep -n isRetryable 172b2d0 -- hugegraph-store` returns nothing — it is `classify()` / `Failure` now.
- The before/after table's `slowest write | 20.1 s` was measured with `grpc.timeout.seconds=20` under the never-retry behaviour, i.e. one deadline. With the single retry the worst case is two deadlines plus the 1 s backoff: about 41 s in that setup, and about 201 s on the shipped default `grpc.timeout.seconds=100`. That is still a large improvement over `11 × timeout + 38 s`, but it is twice what the table claims.

Requested change: update the "Main Changes" bullet to describe the retry-once classification and rename `isRetryable()` to `classify()`, and either re-run the SIGSTOP measurement at `172b2d0` or annotate the table as taken at `920bbbdf`.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -129,35 +136,7 @@ void doCommit() {
...
-                    Throwable cause = throwable.get();
-                    if (cause.getCause() != null) {
-                        cause = cause.getCause();
-                    }
-                    if (cause instanceof HgStoreClientException) {
-                        throw (HgStoreClientException) cause;
-                    }
-                    throw HgStoreClientException.of(cause);
-                }
+                this.commitSessions(this.sessions.values());
```
bitflicker64, line 132: 🧹 Extracting `commitSessions()` removed the only writer of `allSuccess`, leaving a dead variable and an unreachable branch.

Evidence:

- `AtomicBoolean allSuccess = new AtomicBoolean(true)` (`:131`) is now only ever read. The `allSuccess.set(false)` that used to live in the inline parallel-commit block moved into `commitSessions()`, which reports through the thrown exception instead.
- So `if (!allSuccess.get()) { throw HgStoreClientException.of(msg); }` (`:136-138`) can never fire, and `msg` is referenced only from that unreachable throw.

Requested change: delete `:131` and `:136-138` (and the now-unused `msg` field, plus the `java.util.concurrent.atomic.AtomicBoolean` import, if nothing else uses them). If the gate was meant to catch a `doAction()` that returned `false`, wire the loop's return value into it instead of leaving it dead.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -373,35 +398,70 @@ boolean ifAnyTrue(Supplier<Stream<HgPair<HgStoreNode, NodeTkv>>> nodeStreamSuppl
...
+                                            throw HgStoreClientException.of(
+                                                    t.getMessage(), t);
+                                        }
+                                        if (failure == Failure.DEADLINE) {
+                                            // One retry: the NOT_WORK notice sent for the
+                                            // failed RPC reloads the partition leaders, so
+                                            // the next attempt can reach a new leader. A
+                                            // second deadline in a row would only wait the
+                                            // full deadline again on the same stalled store.
+                                            if (deadlineRetried[0]) {
```
bitflicker64: 🧹 The comment and the javadoc say "a second deadline **in a row**", but the flag is never reset, so the code means "a second deadline anywhere in this invocation".

Evidence:

- `boolean[] deadlineRetried = {false}` (`:401`) is set once at `:441` and never cleared, including after an attempt that fails with something else or that succeeds partially.
- Concretely: attempt 0 `DEADLINE_EXCEEDED` (retried), attempts 1-9 `UNAVAILABLE` (retried, ~38 s of backoff, partition leaders reloaded several times over), attempt 10 `DEADLINE_EXCEEDED` — the guard fires and the call aborts, even though these two deadlines are not consecutive and the stated rationale ("the `NOT_WORK` notice … reloads the partition leaders") applies just as much the second time.
- The same wording is in the `Failure` javadoc at `:478`.

Requested change: pick one. Either reset `deadlineRetried[0] = false` whenever an attempt fails with a non-`DEADLINE` classification, which makes the code match "in a row"; or reword `:433` and `:478` to "a second deadline in this invocation" so the budget is honestly described as one deadline retry per call.

bitflicker64 (commented):
Blocking: no. Summary: at `172b2d0` the retry classification, the parallel-commit failure aggregation and both interrupt exits are correct, and every inline point from the earlier reviews is addressed; three new minor points remain, the main one being that the PR description still describes the pre-`b588aa88` behaviour. Evidence: `HugeException.isInterrupted()` (`hugegraph-core/.../HugeException.java:56-61`) tests `rootCause()`, so putting an `InterruptedException` at the bottom of both exits is what `HugeTask.fail()` (`task/HugeTask.java:351`) needs; `HgStoreClientException` uses the plain `RuntimeException` constructors (`type/HgStoreClientException.java:24-46`) so suppression is enabled and `aggregate()`/`classify()` really see the other partitions' failures; `GrpcStoreNodeSessionImpl.commit()` -> `NotifyingExecutor.invoke()` -> `err(t)` keeps the `StatusRuntimeException` in the cause chain; `grep -rn 'StatusException' hugegraph-store/` is empty, so matching only `StatusRuntimeException` is not a live gap; no added line exceeds the 100-char `LineLength` max (`style/checkstyle.xml:29-32`). Not verified: the new tests were not executed here (`mvn -pl hugegraph-store/hg-store-test -am -P store-client-test -Dtest=NodeTxExecutorTest` was not run) and there is no CI signal at this head (combined status `pending`, zero check runs).

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Done in 35e0a6d — `allSuccess`, the unreachable throw and the unused `msg` constant are gone (the `doAction()` return value was never wired into that gate before either, so nothing changes behaviourally).

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Done in 35e0a6d — I kept the "in a row" semantics and made the code match: the budget flag is reset whenever an attempt fails with a non-DEADLINE classification, so deadline → UNAVAILABLE → deadline retries the second deadline too (`testDeadlineBudgetResetsAfterAnotherFailure`, 4 attempts). 11/11 locally.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Done — the description now describes `classify()` / `Failure` with the single deadline retry (and the reset after a different failure from 35e0a6d), and I re-ran the SIGSTOP measurement at `35e0a6d` instead of annotating: with one deadline retry the REST outage under the 1 request/s load is 128 of 300 s (in 20 short windows) against 12 s for the never-retry head and 431 of 503 s for master, the slowest write is 45 s (two deadlines plus backoff) and the server is still healthy 0 s after the writer stops. The table in the body has all three columns and states that the old column is superseded.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
```diff
@@ -373,35 +391,75 @@ boolean ifAnyTrue(Supplier<Stream<HgPair<HgStoreNode, NodeTkv>>> nodeStreamSuppl
...
-                                                         "for the next try.",
-                                                         sleepTime);
-                                                Thread.sleep(sleepTime * 1000L);
-                                            } catch (InterruptedException e) {
-                                                log.error("Failed to sleep", e);
+                                        Failure failure = classify(t);
+                                        if (failure != Failure.DEADLINE) {
+                                            // a different failure in between means the
+                                            // next deadline is not "in a row" again
+                                            deadlineRetried[0] = false;
```
bitflicker64: ⚠️ Resetting the budget here lets one call block on up to six full deadlines, not two.

Evidence:

- The loop makes 11 attempts (`rangeClosed(0, NODE_MAX_RETRYING_TIMES)`). With D, X, D, X, ... D, where X is any fast retryable failure, every D gets retried: 6 × `grpc.timeout.seconds` (600 s on defaults) plus about 38 s of backoff.
- That pattern is reachable. The DEADLINE sends NOT_WORK without the node (`NotifyingExecutor.java:249-254`), which reloads partition leaders. While raft is still electing, or PD still lists the stalled store, the next attempt can fail quickly with NOT_LEADER, a partition fault or UNAVAILABLE. That resets this flag, and the attempt after it waits on the stalled store again.
- The description gives the bound as 2 × deadline + backoff. That holds only for back-to-back deadlines. The request-timeout interrupt still frees REST workers, but callers that are never interrupted (Gremlin jobs without `evaluationTimeout`, background tasks) can be held for about 10 minutes again.

Could you bound deadlines per `retryingInvoke` call instead, e.g. count DEADLINE failures and stop at the second one even if something else failed in between? The javadoc on `Failure` and `testDeadlineBudgetResetsAfterAnotherFailure` would need updating to match.

bitflicker64 (commented):
Blocking: no. Summary: at 35e0a6d the interrupt handling, the parallel-commit failure aggregation and the FATAL/DEADLINE/RETRYABLE classification look correct; one important point on the new deadline budget reset, which lets a single call wait on up to six full deadlines. Separately, the PR title still says "stop retrying after DEADLINE_EXCEEDED", but a deadline is now retried once, and the title becomes the squash commit subject. Evidence: static review of the full exact-head diff against merge-base 60c8803d, traced through NotifyingExecutor.handleErr()/notifyErrConsumer() and HgStoreClientConst.NODE_MAX_RETRYING_TIMES; checked whether store shutdown could surface CANCELLED (lognet grpc-spring-boot-starter 4.5.5 only calls graceful shutdown(), so no concrete path to FATAL there). No CI checks reported at this head.

hugegraph-store/hg-store-client/src/main/java/org/apache/hugegraph/store/client/NodeTxExecutor.java
SebastianGruza: Agreed, the reset reopened the long wait through D, X, D, X. Done in 5ffb221 — the budget is a per-call counter now: the second `DEADLINE_EXCEEDED` ends the call whatever failed in between, so one call blocks on at most two deadlines. `testDeadlineBudgetResetsAfterAnotherFailure` became `testSecondDeadlineFailsEvenWithAnotherFailureInBetween` (D, UNAVAILABLE, D → three attempts, then the deadline is thrown); the `Failure` javadoc and the PR description say "at most once per call" instead of "in a row". Title changed to "bound DEADLINE_EXCEEDED retries to one per call" as you suggested. `NodeTxExecutorTest` 11/11 on JDK 17.

bitflicker64 (approved):
Blocking: no. Summary: At 5ffb221 a deadline is retried at most once per call, interrupts stop the retry loop with an InterruptedException at the root cause, and every parallel commit failure is taken into account when deciding whether to retry. Every earlier inline thread is addressed, and I found no new issues. Evidence: static review of the full diff against merge-base 60c8803d, traced through NotifyingExecutor, GrpcStoreNodeSessionImpl.commit(), HgStoreClientException (plain RuntimeException constructors, so suppressed failures are kept) and HugeException.isInterrupted() / HugeTask.fail(). CI at this head: all 7 workflows and 24 check runs pass. The store job ran ClientSuiteTest, which includes NodeTxExecutorTest: 43 tests, 0 failures, 0 errors.

imbajin (approved):

PR #3209, opened 2026-09-16
feat(core): add DECIMAL (BigDecimal) property data type

### LINKED ISSUE #3206: [Feature] Add a DECIMAL (BigDecimal) property data type for exact amounts
### Feature Description (功能描述)

Add an arbitrary-precision decimal property type: `DataType.DECIMAL` backed by `java.math.BigDecimal`, with REST `data_type: DECIMAL` and `PropertyKey.Builder.asDecimal()`.

### Motivation

The numeric property types today are `BYTE/INT/LONG/FLOAT/DOUBLE`. Values that do not fit a `long` and must not be rounded (token balances in wei go up to 2^256 - 1, 78 digits; money amounts in general) can only be stored as `TEXT`. That loses the one place where the server itself does arithmetic: `update_strategies` in `PUT /graph/{vertices,edges}/batch` (`SUM` / `BIGGER` / `SMALLER`). `UpdateStrategy` already computes in `BigDecimal`, but the result goes back to the property's type, so with `DOUBLE` a `SUM` of `10^18 + 1` is `10^18`; `TEXT` fails the strategy's `Number` type check. For accumulating balances during an import this blocks the use case.

### Proposal (points to confirm)

1. **Type code `12`** in `DataType` (server and the `hugegraph-struct` copy), name `"decimal"`. Please confirm the code is free.
2. **No sort key, no index, no OLAP range.** `isNumber()` stays `false`; the schema builders reject these with an explicit message. Reason: there is no fixed-width byte-order-preserving encoding for a decimal, and faking one through `LongEncoding` would be lossy. `SUM/MAX/MIN` aggregate types on the property key are allowed as for numbers.
3. **Encoding** in `BytesBuffer`: `vint(len)` + unscaled two's-complement bytes + `vint(scale)`. Exact for any precision, 33 bytes for a uint256, scale preserved; existing encodings untouched.
4. **JSON: always a string** on output (`toPlainString()`), a string or a number literal accepted on input. A JSON number is a `double` to most clients; a string is the only lossless representation. In the batch update a fraction has to be sent as a string, because the request's `properties` map is parsed by Jackson before any schema is known (a `0.000000000000000001` literal becomes a `double`); integral literals are exact.

Also in the same change: exact comparison in `ConditionQuery` when one side is a `BigDecimal` (instead of through `doubleValue()`), a string variant in the store-side row decoder, and one fix in `BatchAPI.updateExistElement`: the JSON value is normalised through the property key *before* the strategy runs, on both paths. Today the strategy receives the raw JSON value, which only works for the types Jackson happens to produce; a decimal (or a date) sent as a string fails the type check.

### Status

Implementation and tests are ready on the branch `feat/decimal-datatype` in my fork (one commit on top of `master` `60c8803`, 25 files): unit tests (`DataTypeTest`, `BytesBufferTest`, `JsonUtilTest`), core (`PropertyKey/IndexLabel/EdgeLabel/VertexCoreTest`), API (`VertexApiTest`: `SUM` on 2^256-2 + 1, two entries of one vertex within one request, `BIGGER`) and struct. Run locally with the CI scripts: unit, core on rocksdb and memory, api on rocksdb, all green. I will open the PR tomorrow to leave time for comments on the points above; the client/loader in hugegraph-toolchain will be a separate change.

If any of the four points looks wrong, I would rather hear it here than in the code review.

SebastianGruza:
## Purpose of the PR

- close #3206

The numeric property types today are `BYTE/INT/LONG/FLOAT/DOUBLE`. Values that do not fit a `long` and must not be rounded (token balances in wei, up to 2^256 - 1; money amounts in general) can only be stored as `TEXT`, which loses the one place where the server itself does arithmetic: `update_strategies` in `PUT /graph/{vertices,edges}/batch` (`SUM` / `BIGGER` / `SMALLER`). `UpdateStrategy` already computes in `BigDecimal`, but the result goes back to the property's type: with `DOUBLE` a `SUM` of `10^18 + 1` is `10^18`, and `TEXT` fails the strategy's `Number` type check. This PR adds an exact decimal type so that accumulating balances during an import works. The design points were posted in #3206 on 2026-09-14; no objections so far.

## Main Changes

- **`DataType.DECIMAL(12, "decimal", BigDecimal.class)`** in the server enum and in the `hugegraph-struct` copy, with `isDecimal()` and `valueToDecimal()` (exact for `BigDecimal`, `BigInteger` and integral Java numbers; `Float`/`Double` through their shortest decimal representation; decimal strings). `PropertyKey.Builder.asDecimal()`, REST `data_type: DECIMAL`.
- **No primary key, no sort key, no index, no OLAP write type with an index**: `isNumber()` stays `false` on purpose; `VertexLabelBuilder`, `EdgeLabelBuilder`, `IndexLabelBuilder` and `PropertyKeyBuilder` reject these with an explicit message. A primary key would go through `LongEncoding`/`NumericUtil`, which collapses fractions into a `double` and overflows a `long` on uint256 (review round 3). There is no fixed-width byte-order-preserving encoding for a decimal, and faking one through `LongEncoding` would be lossy. `SUM/MAX/MIN` aggregate types on the property key are allowed, as for numbers.
- **Encoding** in `BytesBuffer` (server core and struct): `vint(len)` + unscaled two's-complement bytes + `vint(scale)`. Exact for any precision, scale preserved, 33 bytes for a uint256. Existing encodings untouched; `OffheapCache` gets the new value type appended at the end of its enum.
- **JSON**: always a plain string on output (`toPlainString()`, `HugeGraphSONModule`); a string or a number literal accepted on input. A JSON number is a `double` to most clients, so a string is the only lossless representation.
- **`ConditionQuery`** compares exactly when one side is a `BigDecimal` (instead of through `doubleValue()`); the store-side row decoder (`GraphStoreIterator`) maps a decimal to a string variant.
- **`BatchAPI.updateExistElement`**: the JSON value is normalised through the property key *before* the `update_strategies` strategy runs, on both paths (two entries of one id within a request; request vs stored element). Found by the new API test: the strategy used to receive the raw JSON value, which only worked for the types Jackson happens to produce, so a decimal (or a date) sent as a string failed the type check.

- **Review round 1 (9d5eaab)**: the `hugegraph-struct` copy converts DECIMAL values too (`DataType.valueToDecimal()`, a decimal branch in `PropertyKey.convSingleValue()`, `Builder.asDecimal()`), so a decimal default value reloaded from JSON as a string normalises on the store side as well; `BigDecimalSerializer.serializeWithType()` for the typed GraphSON v2/v3 mappers of gremlin-server, keeping TinkerPop's `gx:BigDecimal` type id and carrying the plain string in `@value`; `OLAP_SECONDARY`/`OLAP_RANGE` rejected for a DECIMAL key in `PropertyKeyBuilder.checkOlap()` and the no-index guard repeated in `IndexLabelBuilder.build()`, which the OLAP path reaches without `checkFields()`.

Known limit, documented in the issue: in the batch update a **fraction has to be sent as a string**, because the request's `properties` map is parsed by Jackson before any schema is known (`0.000000000000000001` becomes a `double` literal); integral literals are exact. `hugegraph-client` / loader / Hubble will get the type in a separate toolchain change.

## Verifying these changes

- [ ] Trivial rebase, no need to test
- [x] Unit tests / core tests / API tests added and passing locally (JDK 11, through the CI scripts):
  - `unit/core/DataTypeTest`: predicates, `valueToDecimal` for uint256 max, wei scale, integral and binary numbers, invalid strings
  - `unit/serializer/BytesBufferTest`: exact byte layout for `-1.5`, `0`, uint256 max; scale round trip; decimal lists
  - `unit/util/JsonUtilTest`: string on output, string or number on input
  - `core/PropertyKeyCoreTest`: create, value normalisation, `calcSum()`, list cardinality
  - `core/IndexLabelCoreTest`: secondary / range / shard / unique on a decimal all rejected
  - `core/EdgeLabelCoreTest`: decimal sort key rejected, decimal edge property fine
  - `core/VertexLabelCoreTest.testAddVertexLabelWithDecimalPrimaryKey`: a single and a composite decimal primary key rejected, a decimal next to a text primary key fine
  - `core/VertexCoreTest`: uint256 and 18-fraction-digit values through commit and reload, exact `has()` vs the neighbouring value, `gt/lt/gte`, update, invalid values
  - `api/VertexApiTest`: `PUT /graph/vertices/batch` with `SUM`: `2^256-2` + `1` (number literal), then two entries of one vertex in one request (`"0.000000000000000000"`, `"0.000000000000000001"`), then `BIGGER`; response and `GET` carry the exact string
  - struct `PropertyKeyTest`: groovy schema string, struct `BytesBuffer` round trip; decimal conversion from string, integral, `BigInteger` and a default value (single and list)
  - `unit/core/DataTypeTest.testValueToDecimalBounds` and struct `PropertyKeyTest`: `1E+999999999`, `1E-999999999`, `1E+129`, 129 digits rejected; 128 digits, `1E+128`, `1E-128`, uint256 with 18 fraction digits accepted
  - `unit/serializer/HugeGraphSONModuleTest`: a `BigDecimal` through `GraphSONMessageSerializerV1d0/V2d0/V3d0` with `HugeGraphIoRegistry`, `gx:BigDecimal` with the plain string in `@value`, round trip for `1.5`, `1E-18`, uint256 max
  - `core/PropertyKeyCoreTest.testAddOlapPropertyKeyWithDecimalType`: `OLAP_SECONDARY`/`OLAP_RANGE` on a decimal rejected, no `*olap_by_rank` index, `OLAP_COMMON` allowed
  - Results: struct 4/4; `unit-test` 687/688 (`SecurityManagerTest.testFile` fails identically on plain master on a non-English locale, unrelated); `core-test` on rocksdb and memory for the four touched classes green (383 tests on rocksdb); `api-test,rocksdb` 162 tests, 0 failures.
- [x] End-to-end suite for the review paths (`cluster/decimal_e2e.py`, 37 checks: schema with a decimal default value, no-index rule, OLAP write types, REST with batch `SUM` and the default value, Gremlin through the REST proxy and straight on gremlin-server with GraphSON v1/v2/v3), run on hstore and rocksdb against dists from `a28554e` and `9d5eaab`: before 24 pass / 9 fail (the `OLAP_SECONDARY` gap and all 8 typed GraphSON answers), after 33 pass / 0 fail on both backends; logs in `results/decimal/e2e/`.
- [x] Cluster check on HStore (PD + 3 stores) and RocksDB: 10 000 vertices with `balance/hi/lo DECIMAL`, 5 rounds of `PUT /graph/vertices/batch` with `update_strategies: {balance: SUM, hi: BIGGER, lo: SMALLER}`, random increments up to 2^255 with 18 fraction digits, 30 % negative, batches of 500, 4 writer threads, 50 accounts per round appearing twice in one request; 2 000 edges with a decimal `amount` behind an INT sort key. Every value read back and compared exactly with a Python `Decimal` oracle: 50 250 upserts, 0 errors, **0 / 10 000 mismatches** on both backends; decimal sort key / range index rejected with the intended message. Script and logs: `cluster/decimal_sum_bench.py` and `results/decimal/` in https://github.com/SebastianGruza/hugegraph-validation.
- [ ] Docs: the property-key data type list in the `hugegraph-doc` repository needs a `DECIMAL` row; I will open that PR once the type is in.

## Note on compatibility

A graph that already contains a DECIMAL property key cannot be opened by a server built without this change (`No enum constant DataType.DECIMAL` at startup), as with any new data type; worth one line in the release notes.

The `BigDecimal` serializer is registered through `registerCommonSerializers()`, so it applies to every `java.math.BigDecimal` a response carries, not only to DECIMAL property values: a Groovy decimal literal such as `g.inject(1.5)` or `2 * 1.1` came back as the JSON number `1.5` before this PR and comes back as the string `"1.5"` now (GraphSON v1 and the REST `/gremlin` proxy), and as `{"@type":"gx:BigDecimal","@value":"1.5"}` instead of `@value: 1.5` on GraphSON v2/v3. Deliberate, because a JSON number is a `double` to most clients; a `BigDecimal` does not record where it came from, so the serializer cannot keep numbers for non-property values. For the release notes as well.

Values are bounded to 128 significant digits and an absolute scale of 128 (`DataType.DECIMAL_MAX_PRECISION` / `DECIMAL_MAX_SCALE`), checked in `valueToDecimal()` on every input path including the `SUM` result of a batch update; `1E+999999999` is rejected with `Decimal value out of bounds`.

## Release notes

- New property data type `DECIMAL` (`java.math.BigDecimal`, `PropertyKey.Builder.asDecimal()`, REST `data_type: DECIMAL`) for exact amounts, e.g. token balances up to 2^256 - 1 with 18 fraction digits; usable with the `SUM`/`BIGGER`/`SMALLER` update strategies of `PUT /graph/{vertices,edges}/batch`. Values are bounded to 128 significant digits and an absolute scale of 128. A decimal cannot be a vertex primary key, a sort key, an index field of any type or an OLAP write type with an index.
- Every `java.math.BigDecimal` in a REST or Gremlin response is now serialised as a JSON string (`"1.5"`) instead of a number: on GraphSON v1 and the REST `/gremlin` proxy the value itself, on GraphSON v2/v3 `{"@type":"gx:BigDecimal","@value":"1.5"}`. This also applies to Groovy decimal literals such as `g.inject(1.5)`, which came back as the number `1.5` before. In batch updates a fraction has to be sent as a string.
- A graph that contains a `DECIMAL` property key cannot be opened by a server built without this change (`No enum constant DataType.DECIMAL` at startup).

## Note on public API

New enum constant `DataType.DECIMAL` (code 12), new builder method `PropertyKey.Builder.asDecimal()`, new REST value `data_type: DECIMAL`. No change to existing types, encodings or endpoints; graphs without decimal properties are unaffected.


hugegraph-struct/src/main/java/org/apache/hugegraph/struct/schema/PropertyKey.java
```diff
@@ -424,6 +424,9 @@ public String convert2Groovy(boolean attachIdFlag) {
             case UUID:
                 builder.append(".asUUID()");
                 break;
+            case DECIMAL:
```
imbajin, line 434: ⚠️ Blocking: yes. Summary: This makes struct schema generation emit an API that the struct builder does not expose. Evidence: struct PropertyKey.Builder has no asDecimal(), and convSingleValue has no DECIMAL conversion/valueToDecimal; after schema/default JSON values arrive as strings, validValueOrThrow("1.5") returns null/throws, unlike server PropertyKey. Requested change: add the struct conversion and builder method, plus string/numeric/default-value tests.

imbajin (commented):
Blocking issues remain in the struct-side DECIMAL schema integration; targeted exact-head tests passed, but this review is not an approval.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/io/HugeGraphSONModule.java
```diff
@@ -956,4 +962,36 @@ public Blob deserialize(JsonParser jsonParser,
             return Blob.wrap(bytes);
         }
     }
+
+    private static class BigDecimalSerializer extends StdSerializer<BigDecimal> {
```
bitflicker64, line 965: ‼️ `BigDecimalSerializer` only overrides `serialize()`, but this module is registered into the typed GraphSON mappers used by gremlin-server (`GraphSONMessageSerializerV2d0`/`V3d0` in `gremlin-server.yaml`, and V3d0 also answers `application/json`). Jackson then calls `serializeWithType()`, which `StdSerializer` does not implement. I checked this at a28554e by building a `ResponseMessage` with `new BigDecimal("1.5")` and serializing it through each serializer configured with `ioRegistries: [HugeGraphIoRegistry]`. V1d0 returns `"1.5"`. V2d0 and V3d0 both fail with `InvalidDefinitionException: Type id handling not implemented for type java.math.BigDecimal (by serializer of type ...HugeGraphSONModule$BigDecimalSerializer)`. Without the registry the same serializers emit `{"@type":"gx:BigDecimal","@value":1.5}`. So `g.V().values('balance')` on a DECIMAL key fails over GraphSON v2/v3, and so does any existing script that returns a BigDecimal, such as a Groovy decimal literal (`g.inject(1.5)`). That used to work. Requested change: implement `serializeWithType` (for example via `typeSer.writeTypePrefix/writeTypeSuffix`, as the other typed serializers in this module do), or limit the string serializer to `JsonUtil` and leave the TinkerPop `gx:BigDecimal` handling alone. Add a test that serializes a BigDecimal through GraphSON v2 and v3 with `HugeGraphIoRegistry`.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/builder/IndexLabelBuilder.java
```diff
@@ -472,6 +472,9 @@ private void checkFields(Set<Id> propertyIds) {
             E.checkArgument(pkey.aggregateType().isIndexable(),
                             "The aggregate type %s is not indexable",
                             pkey.aggregateType());
+            E.checkArgument(!pkey.dataType().isDecimal(),
```
bitflicker64, line 481: ⚠️ The decimal guard is in `checkFields()`, which only runs on the user-facing `create()` path. OLAP property keys build their index through `SchemaTransaction.createIndexLabelForOlapPk()`, which calls `IndexLabelBuilder.build()` directly and skips `checkFields()`. `PropertyKeyBuilder.checkOlap()` also rejects only `OLAP_RANGE` for non-numeric types. On RocksDB at a28554e, `schema.propertyKey("rank").asDecimal().writeType(WriteType.OLAP_SECONDARY).create()` succeeds and creates index label `*olap_by_rank type=SECONDARY`. That contradicts the rule this PR states (no index of any type on a decimal). The secondary index key is also built from `value.toString()` (`SplicingIdGenerator.concatValues`), and for BigDecimal that output depends on scale and can use exponent notation (`1E+21`), so equal numbers can map to different index keys. Requested change: reject `OLAP_SECONDARY` (and any OLAP write type that builds an index) for `DataType.DECIMAL` in `PropertyKeyBuilder.checkOlap()`, or move the decimal check into `build()`, and add a core test for it.

hugegraph-struct/src/main/java/org/apache/hugegraph/struct/schema/PropertyKey.java
bitflicker64, line 434: ⚠️ The struct copy now knows `DataType.DECIMAL`, but struct `PropertyKey.convSingleValue()` has no decimal branch (only number/date/uuid/blob), and struct `DataType` has no `valueToDecimal()`. For a DECIMAL key, a `String` or `Long` value falls through to `checkDataType()` and returns null. One concrete case is `defaultValue()`: userdata is reloaded from JSON, so a decimal default arrives as a string, and `validValueOrThrow(raw)` then throws. The server-side `PropertyKey` in this PR converts these values correctly. Requested change: port `valueToDecimal()` into the struct `DataType` and add the decimal branch to struct `convSingleValue()`, with tests for string, integral and default-value input.

bitflicker64 (commented):
Blocking: yes. Summary: The new BigDecimal serializer breaks typed GraphSON (v2/v3) for every BigDecimal Gremlin result, and an OLAP_SECONDARY decimal key still gets a secondary index despite the new no-index rule. The struct-side conversion gap is still open. Evidence: GraphSONMessageSerializerV2d0/V3d0 configured with HugeGraphIoRegistry at a28554e fail with "Type id handling not implemented for type java.math.BigDecimal" (same serializers without the registry emit gx:BigDecimal); a RocksDB probe created propertyKey("rank").asDecimal().writeType(OLAP_SECONDARY) and index label *olap_by_rank type=SECONDARY. Latest-head CI is green.

hugegraph-struct/src/main/java/org/apache/hugegraph/struct/schema/PropertyKey.java
SebastianGruza: Done in 9d5eaab: `valueToDecimal()` ported into the struct `DataType` unchanged from the server copy, a decimal branch in struct `PropertyKey.convSingleValue()`, and `asDecimal()` on the struct `PropertyKey.Builder`. Tests in struct `PropertyKeyTest`: conversion from a string, `Long`/`Integer`, `BigInteger`, exponent notation and uint256 max, `"1,5"` and a `Date` rejected; a default value from userdata as a string, as an integral literal and as a list under LIST cardinality, all coming back as `BigDecimal`. struct 5/5 on JDK 11.

SebastianGruza: Done in 9d5eaab, same change as in the thread above: `valueToDecimal()` in the struct `DataType`, the decimal branch in `convSingleValue()`, `asDecimal()` on the builder, tests for string, integral, `BigInteger` and default-value input (single and list).

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/io/HugeGraphSONModule.java
SebastianGruza: Done in 9d5eaab, and thanks for checking this through the real serializers, I had only tested `JsonUtil`. `BigDecimalSerializer` now has `serializeWithType()` in the same shape as `IdSerializer` in this module: `typeSer.typeId(value, VALUE_STRING)`, prefix, `serialize()`, suffix. While at it I removed the `BigDecimal` entry from the module's `TYPE_DEFINITIONS`: with it the type id came out as `hugegraph:BigDecimal`, which no client knows; without it the id stays `gx:BigDecimal` from `GraphSONXModule`, and our serializer still wins the lookup because the registry is added later.

Result: V1 gives `"1.5"`, V2 and V3 give `{"@type":"gx:BigDecimal","@value":"1.5"}`. The string in `@value` is deliberate: a number there is decoded as a double by the JS/Python clients, and this type exists to avoid exactly that; Jackson's default `BigDecimal` deserializer and `gx:BigDecimal` in gremlin-python both accept a string. If you would rather keep a number in `@value` for compatibility with the previous `gx:BigDecimal` output, it is a one-line change, but then it should be said explicitly in the type's description.

Test: new `unit/serializer/HugeGraphSONModuleTest` (in `UnitTestSuite`) builds a `ResponseMessage` with a `BigDecimal`, runs it through `GraphSONMessageSerializerV1d0/V2d0/V3d0` configured with `ioRegistries: [HugeGraphIoRegistry]`, checks the type prefix and the string in `@value`, and deserializes the response back to an equal `BigDecimal` for `1.5`, `1E-18` and uint256 max. 3/3 on JDK 11.

End to end on the lab, dists from both heads, hstore and rocksdb (`cluster/decimal_e2e.py`, group R6 in `results/decimal/e2e/` of https://github.com/SebastianGruza/hugegraph-validation): before, every one of the 8 queries through gremlin-server with `Accept` v2.0 and v3.0 (`values()` on uint256 max, `g.inject(1.5)`, `values()` of a default value, `sum()`) → `500 Type id handling not implemented`; after, all 8 return `gx:BigDecimal` with the exact value, `sum()` exact to the 18th fraction digit. `/gremlin` through the REST proxy (`application/json`, untyped) worked on both heads.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/schema/builder/IndexLabelBuilder.java
SebastianGruza: Done in 9d5eaab, both things you asked for: `PropertyKeyBuilder.checkOlap()` rejects for DECIMAL every OLAP write type that builds an index (everything but `OLAP_COMMON`) with "decimal keys can't be indexed", and `IndexLabelBuilder.build()` carries the same guard as `checkFields()`, so the rule also holds on the `createIndexLabelForOlapPk()` path and for any future caller of `build()`. The `toString()`/scale point about the secondary index key becomes moot, since such an index can no longer exist.

Test: `PropertyKeyCoreTest.testAddOlapPropertyKeyWithDecimalType` behind `Assume supportsOlapProperties`: `OLAP_SECONDARY` and `OLAP_RANGE` on a decimal key → `NotAllowException`, `*olap_by_rank` does not exist, `OLAP_COMMON` (no index) still passes. On rocksdb: `PropertyKeyCoreTest` 25/25, `IndexLabelCoreTest` 45/45. End to end through REST on the lab (group R3 in `results/decimal/e2e/`): before, `write_type: OLAP_SECONDARY` on a decimal key → 202 and the key is created; after → 400 "decimal keys can't be indexed", `OLAP_RANGE` 400, `OLAP_COMMON` 202, on hstore and rocksdb.

SebastianGruza:
Thanks for the reviews. All three blocking points are handled in 9d5eaab: the DECIMAL conversion in the struct copy (method, branch in `convSingleValue`, `asDecimal()` on the builder), `serializeWithType` for GraphSON v2/v3 with the `gx:BigDecimal` type id kept, and `OLAP_SECONDARY`/`OLAP_RANGE` rejected for decimals plus a guard in `IndexLabelBuilder.build()`. Four new tests (struct ×2, unit GraphSON, core OLAP), details in the inline threads. Locally on JDK 11: struct 5/5, unit 47/47 for the touched classes, core on rocksdb 384 tests in the four touched classes, 0 failures.

On top of that, end to end on live servers (hstore on PD + 3 stores, rocksdb), dists built from `a28554e` and `9d5eaab`: 37 checks (schema, no-index rule, OLAP, REST with batch `SUM` and a default value, Gremlin through the REST proxy and straight on gremlin-server in GraphSON v2/v3), before 24 pass / 9 fail, after 33 pass / 0 fail on both backends; script `cluster/decimal_e2e.py`, logs `results/decimal/e2e/`, write-up in `docs/decimal-datatype.md` in https://github.com/SebastianGruza/hugegraph-validation.

One thing for the release notes rather than for this PR: a graph that already contains a DECIMAL property key cannot be opened by a server without this change (`No enum constant DataType.DECIMAL` at startup). PR description updated with the three points under "Main Changes" and "Verifying".

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/type/define/DataType.java
```diff
@@ -143,6 +156,42 @@ public <V> Number valueToNumber(V value) {
...
+        if (value instanceof Byte || value instanceof Short ||
+            value instanceof Integer || value instanceof Long) {
+            return BigDecimal.valueOf(((Number) value).longValue());
+        }
+        if (!(value instanceof Number) && !(value instanceof String)) {
+            return null;
+        }
+        String text = value.toString().trim();
+        try {
+            return new BigDecimal(text);
```
bitflicker64: ⚠️ Important. Nothing bounds the exponent here. `"1E+999999999"` parses to unscaled 1 with scale -999999999, and `BytesBuffer` stores it in a few bytes. Every output path then calls `toPlainString()`: `BigDecimalSerializer.serialize()` in HugeGraphSONModule.java:974 (REST responses, including the create response, and GraphSON v1/v2/v3) and GraphStoreIterator.java:261 on the store side.

On JDK 17, `toPlainString()` on that value throws `OutOfMemoryError` at `-Xmx512m` (the minimum heap in `hugegraph-server.sh`). At `-Xmx4g` it returns a 1,000,000,000 character string in about 1.3 s, before Jackson copies it into the response. So any user who can write a vertex can store an 11-character value that costs gigabytes on every read. `"1E-999999999"` does the same.

Requested change: reject values whose `scale()` or `precision()` is above a documented limit (one that still fits uint256 with 18 fraction digits) in `valueToDecimal()`, here and in the struct copy, and add a test with `"1E+999999999"`. The `validValueOrThrow(value)` that `BatchAPI` already runs after the strategy will then also cover a `SUM` result that crosses the limit.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/io/HugeGraphSONModule.java
```diff
@@ -183,6 +184,10 @@ public static void registerCommonSerializers(SimpleModule module) {
 
         module.addSerializer(Blob.class, new BlobSerializer());
         module.addDeserializer(Blob.class, new BlobDeserializer());
+
+        // Decimals travel as strings: JSON numbers are doubles to most clients
+        module.addSerializer(BigDecimal.class, new BigDecimalSerializer());
```
bitflicker64, line 189: 🧹 Minor. This goes through `registerCommonSerializers()`, which `HugeGraphIoRegistry` registers for every GraphSON version and `JsonUtil` also uses, so it applies to every `java.math.BigDecimal`, not only DECIMAL property values. Groovy decimal literals are already `BigDecimal`: `g.inject(1.5)` or `2 * 1.1` through `/gremlin` returned `1.5` before this PR and returns `"1.5"` now (the new `HugeGraphSONModuleTest` asserts the V1 string). The PR description says existing endpoints don't change.

Requested change: add the V1 number-to-string change, and the string in `gx:BigDecimal` `@value`, to the compatibility note and release notes. Or keep numeric output for BigDecimal values that don't come from a DECIMAL property.

bitflicker64 (commented):
Blocking: yes. Summary: a DECIMAL value with a huge exponent, such as "1E+999999999", is accepted and stored in a few bytes, and every read then expands it through toPlainString() into a billion-character string, which exhausts server memory. The BigDecimal serializer is also registered globally, so existing BigDecimal Gremlin results change from JSON numbers to strings. Evidence: static read of DataType.valueToDecimal, HugeGraphSONModule.BigDecimalSerializer and GraphStoreIterator at 9d5eaab; a JDK 17 probe where toPlainString() on that value throws OutOfMemoryError at -Xmx512m and returns 1,000,000,000 chars at -Xmx4g. Latest-head CI is green.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/type/define/DataType.java
SebastianGruza: Done in bae56ca, good catch. The bound: at most 128 significant digits and an absolute scale of at most 128 (`DataType.DECIMAL_MAX_PRECISION` / `DECIMAL_MAX_SCALE`, the same pair in the struct copy), checked by `checkDecimalBounds()` at the end of `valueToDecimal()`. uint256 with 18 fraction digits is 96 digits, so more than 30 digits of headroom remain, and the longest possible `toPlainString()` is about 256 characters. The message reads `Decimal value out of bounds: precision 1, scale -999999999 (at most 128 significant digits and a scale of at most 128 in either direction)`.

One thing your comment touched on that I had not seen: `PropertyKey.convValue()` returned the value untouched when its type already matched, so a ready-made `BigDecimal` (a Gremlin literal in `addV().property()`, the `SUM` result in `BatchAPI`) never reached `valueToDecimal()` at all. The struct test caught it: `validValueOrThrow(new BigDecimal("1E+999999999"))` did not throw. Both copies of `convValue()` now skip the short-circuit for decimals, so `validValueOrThrow` after the strategy really does cover the `SUM` result, as you wrote.

Tests: `DataTypeTest.testValueToDecimalBounds` (`1E+999999999`, `1E-999999999`, `1E+129` and 129 digits rejected; 128 digits, `1E+128`, `1E-128` and uint256 with 18 fraction digits accepted; the same bound for a ready-made `BigDecimal`), struct `PropertyKeyTest` (the same inputs through `validValueOrThrow`), `PropertyKeyCoreTest` (a string and a `BigDecimal` through `validValue`). End to end on the lab (rocksdb, `cluster/decimal_e2e.py`, `results/decimal/e2e/after2-rocksdb.log` in hugegraph-validation): create with `1E+999999999` → 400 "out of bounds", create with `1E+128` and with 128 nines → 201, batch `SUM` of 128 nines + 1 (a 129-digit result) → 400 "out of bounds"; 37 PASS, 0 FAIL, 4 N-A.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/io/HugeGraphSONModule.java
SebastianGruza: Right, the PR description was inaccurate there. Added to "Note on compatibility" and for the release notes: through `registerCommonSerializers()` the serializer applies to every `java.math.BigDecimal` in a response, so a Groovy literal (`g.inject(1.5)`, `2 * 1.1`) used to come back as the number `1.5` and now comes back as `"1.5"` in V1 and through the REST `/gremlin` proxy, and as `{"@type":"gx:BigDecimal","@value":"1.5"}` instead of a number in `@value` on V2/V3. The "numbers for BigDecimals that are not DECIMAL property values" variant cannot be done in the serializer, since a `BigDecimal` carries no record of where it came from; it would need a wrapper type on property values, which I think is worse than one plain rule, "a BigDecimal is always a string". If the maintainers prefer backward compatibility, the change is one line (`writeNumber` instead of `writeString` in `@value`), but then the JS/Python clients get a double.

SebastianGruza:
Round 2 in bae56ca: DECIMAL values are bounded to 128 significant digits and an absolute scale of 128, checked in `valueToDecimal()` in both copies and, after removing the short-circuit in `PropertyKey.convValue()`, also for ready-made `BigDecimal`s (a Gremlin literal, the `SUM` result). The compatibility note now says that every `BigDecimal` in a response comes out as a string (V1, REST proxy) or as `gx:BigDecimal` with the string in `@value` (V2/V3). Locally on JDK 11: struct 5/5, unit 48/48 for the touched classes, core on rocksdb 339 tests in three classes with no failures, api on rocksdb green. E2E (`cluster/decimal_e2e.py`, now 41 checks with three new ones for the bound) on rocksdb with a dist from this head: 37 PASS, 0 FAIL, 4 N-A, log `results/decimal/e2e/after2-rocksdb.log` in hugegraph-validation.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/type/define/DataType.java
```diff
@@ -43,7 +45,14 @@ public enum DataType implements SerialEnum {
...
     DATE(10, "date", Date.class),
-    UUID(11, "uuid", UUID.class);
+    UUID(11, "uuid", UUID.class),
+    /*
+     * Arbitrary-precision decimal (java.math.BigDecimal). Stored exactly; not a
+     * "number" in the isNumber() sense because it has no fixed-width, sortable
+     * encoding, so it can't be a sort key, a range/secondary index field or an
+     * OLAP range property.
+     */
+    DECIMAL(12, "decimal", BigDecimal.class);
```
imbajin, line 55: ‼️ Blocking: yes. `DECIMAL` is still accepted as a vertex primary key: `VertexLabelBuilder.checkPrimaryKeys()` only checks that the key belongs to `properties`, and this PR adds no type guard. Core primary-id generation then passes the normalized `BigDecimal` through `ConditionQuery.concatValues()`/`LongEncoding`; `NumericUtil.numberToSortableLong()` converts non-integral values through `double` and throws on large integral values. A local probe showed `1.000000000000000001` and `1.000000000000000002` map to the same sortable long, while the uint256 value advertised by this PR throws `ArithmeticException: Overflow`. The REST batch path also hashes the raw string before core normalization. Please reject DECIMAL primary keys or add a lossless canonical primary-key encoding, with tests for fractional precision and uint256 values.

imbajin (commented):
Blocking: yes. Summary: DECIMAL primary-key IDs can collide or fail for valid high-precision values. Evidence: current-head static tracing shows VertexLabelBuilder accepts DECIMAL primary keys, while ConditionQuery/LongEncoding still route the normalized BigDecimal through NumericUtil; a local probe collapsed distinct 18-decimal values to one sortable long and raised ArithmeticException for a uint256 value.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/type/define/DataType.java
SebastianGruza: Done in 834ef89, thanks for tracing that path. I reject a decimal as a primary key rather than adding a canonical encoding: it is consistent with this PR's rule (no sort key, no index, no OLAP write type with an index), and a lossless id would also have to settle scale normalisation (are `1.0` and `1` the same key or different ones?) and touch `SplicingIdGenerator`, which I would rather not do in this PR. The guard sits in `VertexLabelBuilder.checkPrimaryKeys()`, next to the existing check that the key belongs to the properties, in the same shape as the decimal sort-key rejection in `EdgeLabelBuilder`: `The primary key 'balance' of vertex label 'account' can't be a decimal property`. It covers a composite key that has a decimal among its fields too. The REST batch path you mention (hashing the raw string before core normalisation) no longer matters, since such a label cannot exist.

Test: `VertexLabelCoreTest.testAddVertexLabelWithDecimalPrimaryKey`: a single decimal primary key and a composite `(name, balance)` key are rejected with that message, the label does not exist afterwards, and a decimal as a plain property next to a text primary key passes. `VertexLabelCoreTest` 53/53 on rocksdb, 123/123 together with `PropertyKeyCoreTest` and `EdgeLabelCoreTest`. End to end: a new R2 check in `cluster/decimal_e2e.py` (`POST /schema/vertexlabels` with `primary_keys: [balance]` → 400 `The primary key 'd2g_balance' of vertex label 'd2g_badpk' can't be a decimal property`); on rocksdb with a dist from this head 38 PASS, 0 FAIL, 4 N-A, log `results/decimal/e2e/after3-rocksdb.log` in https://github.com/SebastianGruza/hugegraph-validation. The limits table in the PR description and in the docs now lists the primary key as well.

SebastianGruza:
Round 3 in 834ef89: DECIMAL is rejected as a vertex primary key (single and composite) in `VertexLabelBuilder.checkPrimaryKeys()`, with a core test and an E2E check; details in the inline thread. The full list of what this type cannot be is now in one place in the description: primary key, sort key, index field of any type, OLAP write type with an index. Locally on JDK 11: core on rocksdb 123/123 in the three schema test classes; E2E on rocksdb 38 PASS, 0 FAIL, 4 N-A.

bitflicker64 (approved):
Blocking: no. Summary: DECIMAL is stored exactly on every backend path, bounded, excluded from primary keys, sort keys and indexes, and the earlier review points (struct conversion, typed GraphSON, OLAP index bypass, exponent bound, primary-key ids) are all fixed. Please add the string output for every BigDecimal in Gremlin/REST responses to the release notes. Evidence: read core/struct BytesBuffer, PropertyKey, the schema builders, ConditionQuery/Condition and BatchAPI at 834ef89; checked the negative-scale vint round trip; ran a local memory-backend core probe (eq/neq/within/without against a stored 1.50 with 1.5, all scale-insensitive) plus testAddVertexWithPropertyValueOfDecimal; all latest-head CI checks pass.

SebastianGruza:
Thanks for the review and for checking the scale-insensitive comparisons, that was not covered by my tests. Added a "Release notes" section to the description with the three lines that matter for a release: the new type and its limits, every `BigDecimal` in a REST or Gremlin response now being a string (`"1.5"` on v1 and the REST proxy, `gx:BigDecimal` with the string in `@value` on v2/v3, Groovy literals included), and the older-server compatibility note. No code change, no new push; the head stays 834ef89.

bitflicker64 (approved):

SebastianGruza:
Companion PR on the toolchain side: apache/hugegraph-toolchain#771 (`DataType.DECIMAL` and `asDecimal()` in hugegraph-client, `BigDecimal` sent as a plain string, loader/spark conversion, Hubble Groovy export). Its API test is `Assume`-guarded, so it stays green against a server without DECIMAL and switches on once the toolchain CI pins a server commit that includes this PR. Verified end to end against a server built from this branch: client unit 77/77, `DecimalPropertyApiTest` 6/6, loader unit 13/13.

SebastianGruza:
New commit 0146849b, prompted by the review of the toolchain companion apache/hugegraph-toolchain#771: REST now reads JSON fractions as `BigDecimal` (`ObjectMapperResolver` registered next to `JacksonFeature`, `USE_BIG_DECIMAL_FOR_FLOATS`), so a DECIMAL value no longer has to be a string: the literal `12345678901234567890.123456789012345678` is stored exactly, a DOUBLE key narrows it as before, an INT key still rejects a fraction with the same message. `JsonUtil.castNumber` converts a BigDecimal to double instead of asserting on the JSON type. Tests: `testDecimalJsonNumberLiteralIsExact` plus a number literal as the increment in `testBatchUpdateDecimalWithSumStrategy`; `VertexApiTest` from the api-test profile on rocksdb 6/6 against this build. Blast radius checked: nothing in `hugegraph-api` assumes `Double` from JSON, the only place in core was `castNumber`.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/server/ObjectMapperResolver.java
```diff
@@ -0,0 +1,48 @@
...
+ * any Number and narrows it to the key's type as before.
+ */
+@Provider
+public class ObjectMapperResolver implements ContextResolver<ObjectMapper> {
+
+    private final ObjectMapper mapper;
+
+    public ObjectMapperResolver() {
+        this.mapper = new ObjectMapper();
+        this.mapper.enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS);
```
bitflicker64, line 41: ⚠️ Enabling `USE_BIG_DECIMAL_FOR_FLOATS` on the shared Jersey mapper changes every untyped `Object` in every REST body, not only vertex and edge properties. Combined with the `BigDecimalSerializer` this PR adds to `registerCommonSerializers()` (which `JsonUtil` uses), any fraction that the server re-encodes with `JsonUtil.toJson()` now comes back as a JSON string, and existing code that expects a `Number` rejects it.

Two concrete paths:

1. Algorithm and computer jobs. `AlgorithmAPI.post()` and `ComputerAPI.post()` take `Map<String, Object> parameters` and store `JsonUtil.toJson(input)` as the task input. `AlgorithmJob.execute()` / `ComputerJob` read it back with `JsonUtil.fromJson(input, Map.class)`, and `ParameterUtil.parameterDouble()` checks `value instanceof Number`. So `POST /graphs/{graph}/jobs/algorithm/page_rank {"alpha": 0.85}` passes the synchronous `AlgorithmJob.check()` (the value is still a `BigDecimal` there) and then the task fails with `Expect double value for parameter 'alpha': '0.85'`. The same applies to `precision` in `AbstractCommAlgorithm`/`AbstractComputer` and `alpha` in `PageRankComputer`.
2. Schema userdata. `JsonPropertyKey.userdata` (and the vertex/edge/index label equivalents) is a `Map<String, Object>`, persisted by `BinarySerializer`/`TextSerializer` through `JsonUtil.toJson(schema.userdata())`. A DOUBLE key created with `"userdata": {"~default_value": 1.5}` is stored as `"1.5"`. After a reload from the backend (restart, schema cache miss, another server node), `PropertyKey.defaultValue()` calls `validValueOrThrow("1.5")`, `DataType.valueToNumber()` returns null for a `String`, and `HugeElement` line 102 throws `Invalid property value '1.5'` when filling defaults. Before this commit the value was a `Double` and round-tripped as a number.

I reproduced the round trip with plain Jackson 2.9: a mapper with `USE_BIG_DECIMAL_FOR_FLOATS` reads `{"alpha":0.85}` as `BigDecimal`; a mapper with the same `BigDecimal` to `writeString(toPlainString())` serializer writes the task input as `{"parameters":{"alpha":"0.85",...}}`; reading it back gives `java.lang.String`, `instanceof Number == false`. CI is green because no API test sends a fractional algorithm parameter or a fractional default value.

Requested change: scope exact-decimal parsing to property values instead of flipping the global mapper, for example a content deserializer on `JsonElement.properties` (or converting `BigDecimal` back to `Double` for DOUBLE/FLOAT keys only where the key is known), and drop the global `ObjectMapperResolver`. If the global switch is kept, the job and userdata paths need explicit normalisation plus API tests: an algorithm job with a fractional parameter, and a DOUBLE property key with a fractional `~default_value` read after the schema is reloaded.

bitflicker64 (commented):
Blocking: yes. Summary: The DECIMAL type itself (encoding, bounds, no-index and no-primary-key rules, struct conversion, typed GraphSON) looks correct and the earlier review points are resolved, but the new global `ObjectMapperResolver` in 0146849 turns every JSON fraction in any REST body into a `BigDecimal`, which `JsonUtil` then re-encodes as a string: fractional algorithm/computer job parameters fail at execution and a fractional `~default_value` on a DOUBLE key breaks after a schema reload. Evidence: static trace at 0146849bcfbe2d9430789dfa34bab681bba7d010 of `ObjectMapperResolver`, `HugeGraphSONModule.registerCommonSerializers()`, `JsonUtil`, `AlgorithmAPI`/`ComputerAPI`, `AlgorithmJob.execute()`, `ParameterUtil.parameterDouble()`, `BinarySerializer`/`TextSerializer` userdata, `PropertyKey.defaultValue()` and `DataType.valueToNumber()`; a standalone Jackson 2.9 probe showing `{"alpha":0.85}` round-trips to the String `"0.85"`; all 22 latest-head checks pass.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/type/define/DataType.java
```diff
@@ -143,6 +156,65 @@ public <V> Number valueToNumber(V value) {
...
+     * Convert a value to BigDecimal: BigDecimal as is, any other Number and a
+     * decimal string through their exact decimal representation. Float and
+     * Double go through Number.toString(), i.e. the shortest string that
+     * round-trips the binary value, so a client that already holds a lossy
+     * double gets that double, exactly.
+     *
+     * @return the BigDecimal, or null if the value is not a Number or String
+     * @throws IllegalArgumentException if the string is not a decimal number
+     */
+    /*
```
bitflicker64, line 169: 🧹 The `valueToDecimal()` Javadoc (lines 159-168) now sits above the bounds comment and the `DECIMAL_MAX_PRECISION` / `DECIMAL_MAX_SCALE` constants, so the Javadoc tool attaches it to `DECIMAL_MAX_PRECISION`, and `valueToDecimal()` at line 179 has no documentation. The struct copy of `DataType` has the constants before the Javadoc and does not have this problem.

Requested change: move the two constants and their `/* Bounds ... */` comment above the `/** Convert a value to BigDecimal ... */` block, so the Javadoc directly precedes `valueToDecimal()`.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/backend/query/ConditionQuery.java
```diff
@@ -947,6 +947,11 @@ private static boolean numberEquals(Object number1, Object number2) {
             // Otherwise convert to BigDecimal to make two numbers comparable
             Number n1 = NumericUtil.convertToNumber(number1);
             Number n2 = NumericUtil.convertToNumber(number2);
+            if (n1 instanceof BigDecimal || n2 instanceof BigDecimal) {
```
imbajin, line 950: ‼️ This exact-comparison change does not preserve DECIMAL conditions sent to HStore. ConditionQuery.bytes() serializes BigDecimal as an untyped JSON number, and QueryAdapter only writes a value class for Java primitive wrappers; the Store's ConditionQuery.fromBytes() therefore reads the filter value as Double. HStore's FilterIterator then compares against the BigDecimal property using this rounded value, so equality and range queries for high-precision DECIMAL values can miss or misclassify matches. Preserve BigDecimal type and precision in the Store query wire format and its struct-side filter path.

imbajin (commented):
Blocking: yes. Summary: HStore DECIMAL filters lose precision across the Store query wire format, and unbounded GraphSON BigDecimals can expand into enormous strings; the DECIMAL scale bound also has an Integer.MIN_VALUE bypass. Evidence: static trace through ConditionQuery/QueryAdapter/FilterIterator, HugeGraphSONModule, and both DataType copies; 22 exact-head checks pass.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/io/HugeGraphSONModule.java
```diff
@@ -956,4 +961,56 @@ public Blob deserialize(JsonParser jsonParser,
...
+    private static class BigDecimalSerializer extends StdSerializer<BigDecimal> {
+
+        public BigDecimalSerializer() {
+            super(BigDecimal.class);
+        }
+
+        @Override
+        public void serialize(BigDecimal decimal, JsonGenerator jsonGenerator,
+                              SerializerProvider provider) throws IOException {
+            jsonGenerator.writeString(decimal.toPlainString());
```
imbajin, line 974: ⚠️ The new GraphSON deserializer accepts BigDecimal values that bypass the property bounds, while this serializer expands every BigDecimal with toPlainString(). A Gremlin result such as new BigDecimal("1E+999999999") can therefore request about a billion output characters and exhaust a response worker's memory. Apply a safe bound to generic GraphSON BigDecimals or use an exact string representation that does not expand the exponent.

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/type/define/DataType.java
```diff
@@ -143,6 +156,65 @@ public <V> Number valueToNumber(V value) {
...
+            } catch (NumberFormatException e) {
+                throw new IllegalArgumentException(String.format(
+                        "Can't read '%s' as decimal", value));
+            }
+        }
+        return checkDecimalBounds(decimal);
+    }
+
+    public static BigDecimal checkDecimalBounds(BigDecimal decimal) {
+        int scale = Math.abs(decimal.scale());
```
imbajin, line 206: ⚠️ This scale check is bypassed when decimal.scale() is Integer.MIN_VALUE because Math.abs(Integer.MIN_VALUE) overflows and remains negative; a precision-1 BigDecimal then passes the ±128 bound. PropertyKey deliberately routes ready-made BigDecimal values through this check, and BytesBuffer persists the scale, so a later toPlainString response can trigger an enormous allocation. Compare scale directly with -DECIMAL_MAX_SCALE and DECIMAL_MAX_SCALE in both the server and struct copies.

PR #3210, opened 2026-09-16
fix(server): wait for the stores at startup instead of exiting on a cold start

### LINKED ISSUE #3203: [Bug][HStore] Server exits 1 on cold start when Stores miss the 38s partition-lookup retry ceiling
### Bug Type (问题类型)

server status (启动/运行异常)

### Before submit

- [x] I have confirmed and searched that there are no similar problems in the historical issues and documents.

### Environment (环境信息)

- Server Version: `hugegraph/server:latest` published image, built 2026-09-08, `hg-store-client-1.7.0.jar`. Also seen on images built from master `60c8803d` and from `6ec19838`.
- Backend: HStore, 3 PD + 3 Store + 3 Server, replica count 3
- OS: 12 CPUs, 15 G RAM, Ubuntu, Kubernetes v1.37.0 (kind), Docker 29.5.3
- Data Size: none, this happens on a completely empty first install

**Expected.** On a cold start of a distributed cluster, a Server that finds the Stores have
not yet registered with PD waits for them and then starts.

**Actual.** The Server asks PD for partition information, gets
`error code = 105, The number of active stores is less then 3`, retries **10 times over 38
seconds**, hits a hard retry ceiling, and the JVM **exits 1**. Under Kubernetes the container
is restarted and usually succeeds on the second attempt, so the cluster ends up healthy and
the only lasting trace is `RESTARTS: 1`. Outside an orchestrator that restarts it, the Server
stays down.

This is a startup race, not a timeout. It is not related to the entrypoint startup budget
from #3186: the log contains no `The operation timed out(...)` line, and the value was 450 s
here.

The Server's retry window is a fixed 38 s from T+42 to T+80; Stores registering inside it start the Server, after it the Server exits 1

Case B is the reproduction detailed below. Case A stands for the successful path, which is
what the other two Servers on that same install did; its T+64 s is illustrative, while the
fixed timings (T+42, T+80, T+85) are measured.

#### Timeline from one reproduction

| Time | Event |
|---|---|
| 09:44:28 | Server containers start |
| 09:44:38 to 09:44:44 | PD pods start, 10 to 16 s **after** the Server |
| 09:44:39 to 09:44:45 | Store pods start |
| 09:45:11 | Server JVM finishes booting (~42 s) and requests partition information |
| 09:45:11 to 09:45:49 | `error code = 105` x10, backoff 1,1,1,2,3,4,5,6,7,8 s = **38 s** |
| 09:45:49 | `NodeTxExecutor - the number of retries reached the upper limit : 10` |
| 09:45:54 | container **exits 1**, total lifetime 85 s |

#### Log and stack trace

```
2026-09-10 09:45:11 [main] [ERROR] o.a.h.s.c.HgStoreNodePartitionerImpl - An error occurred
  while getting partition information :PD request error, error code = 105,
  msg = The number of active stores is less then 3
2026-09-10 09:45:11 [main] [INFO]  o.a.h.s.c.NodeTxExecutor - Waiting 1 seconds for the next try.
... 10 attempts, backoff growing to 8 s ...
2026-09-10 09:45:49 [main] [ERROR] o.a.h.s.c.NodeTxExecutor - the number of retries reached
  the upper limit : 10,caused by:
java.lang.RuntimeException: PD request error, error code = 105,
  msg = The number of active stores is less then 3
	at org.apache.hugegraph.store.client.HgStoreNodePartitionerImpl.partition(HgStoreNodePartitionerImpl.java:87) ~[hg-store-client-1.7.0.jar:1.7.0]
	at org.apache.hugegraph.store.client.NodeTxSessionProxy.doPartition(NodeTxSessionProxy.java:865) ~[hg-store-client-1.7.0.jar:1.7.0]
	at org.apache.hugegraph.store.client.NodeTxSessionProxy.toNodeTkvList(NodeTxSessionProxy.java:778) ~[hg-store-client-1.7.0.jar:1.7.0]
	at org.apache.hugegraph.store.client.NodeTxSessionProxy.getNodeStream(NodeTxSessionProxy.java:921) ~[hg-store-client-1.7.0.jar:1.7.0]
	at org.apache.hugegraph.store.client.NodeTxSessionProxy.lambda$createTable$20(NodeTxSessionProxy.java:321) ~[hg-store-client-1.7.0.jar:1.7.0]
	at org.apache.hugegraph.store.client.NodeTxExecutor.lambda$isAllTrue$12(NodeTxExecutor.java:344) ~[hg-store-client-1.7.0.jar:1.7.0]
	at org.apache.hugegraph.store.client.NodeTxExecutor.lambda$retryingInvoke$15(NodeTxExecutor.java:381) ~[hg-store-client-1.7.0.jar:1.7.0]
```

The restarted container hits the same error **9 more times** and then succeeds, because by
then the Stores have registered.

#### Why it is intermittent

It is a race between how long the Stores take to register with PD and a fixed 38 second
budget that begins about 42 seconds after the Server container starts. Rates observed across
four campaigns on two different hosts:

| Run | Host | Fresh Server starts | Failed |
|---|---|---|---|
| 2026-09-05 | A | 3 | 1 |
| 2026-09-10 run 1 | B | 3 | 0 |
| 2026-09-10 run 2 | B | 3 | 1 |
| 2026-09-10 F1 run | B | 3 | 1 |
| **total** | | **12** | **3** |

So **3 of 12 fresh Server starts** failed, and **3 of the 4 installs** produced at least one
crashed Server. The per-install figure is the one an operator meets.

Earlier campaigns saw a similar symptom that is deliberately **not** counted above. On
2026-08-29, two of six Server starts exited 1 on install, but each ran about 2 min 45 s
before dying, which fits the 120 s entrypoint self-kill of #3186 rather than this 38 s retry
ceiling, and the logs were lost so neither can be attributed. #3187 has since fixed that
one. If those two turn out to be this bug as well the rate is higher, not lower.

*(Corrected after first posting. The first two rows originally read 6/2 and 10/0. The 6/2
was carried over from #3186, which is the entrypoint's 120 s self-kill and a different
failure, and the 10 was a pod count rather than a count of Server starts. Every campaign
deployed exactly 3 Servers.)*

**It does not reproduce on pod replacement.** 66 Server replacements against an already
running cluster produced zero failures, because a replacement Server asks PD and gets a
straight answer in about 16 seconds. Only a cold cluster has the window. Anyone trying to
reproduce this must recreate the whole cluster, not restart a Server.

### Steps to reproduce

1. Deploy 3 PD + 3 Store + 3 Server on Kubernetes, all at once, with no ordering between them.
2. Watch the Server pods on the very first install.
3. Repeat the full install a few times. Three of the four installs observed so far showed
   at least one Server with `RESTARTS: 1` and `lastState.terminated.exitCode: 1`.

### A note on diagnosing this

The message an operator actually sees is only:

```
Connecting to HugeGraphServer (http://0.0.0.0:8080/graphs)...........Starting HugeGraphServer failed
See /hugegraph-server/logs/hugegraph-server.log for HugeGraphServer log output.
```

That file is inside the container and the container is gone, so `kubectl logs --previous`
returns the line above and nothing else. The stack trace in this report was only obtainable
by mounting a volume at `/hugegraph-server/logs` before the first boot, so the crashed run's
log survived the restart. **Logging the fatal startup cause to stdout would make this class
of failure diagnosable without that trick.**

### Suggested direction

These are suggestions rather than a diagnosis of the right fix:

1. The retry ceiling is `HgStoreClientConst.NODE_MAX_RETRYING_TIMES = 10`
   (`hugegraph-store/hg-store-client/.../util/HgStoreClientConst.java:48`), used only by
   `NodeTxExecutor` at lines 376 and 383. Nothing reads it from configuration or the
   environment, so 38 seconds is the whole budget an operator gets. A cold cluster bootstrap
   can reasonably take longer than that. Either make the budget configurable, or use a longer
   one on the bootstrap path where "stores not registered yet" is an expected transient
   rather than a fault.
2. Treat `error code = 105` during startup as "not ready yet" rather than as a retryable
   failure with a short ceiling, since it is precisely the condition that resolves on its own.
3. Log the terminal cause to stdout, per the note above.

Orchestration can also side-step this by not starting the Server until PD reports the
expected number of active stores. I will carry that as a gate in the Helm chart in #3131 /
apache/hugegraph#3132, but it is a workaround for the retry ceiling, not a fix for it.

### Related

- #3124, same component family (`HgStoreNodePartitionerImpl`, `NodeTxExecutor`) but a
  different trigger: stale DNS after Store pod replacement, not a cold-start race.
- #3123, the storage wait binding to the first PD that answers.
- #3186, the entrypoint startup timeout. Explicitly **not** the cause here.
- #3131, the Helm chart, where the orchestration-side gate would live.

Not applicable, the cluster is empty.

Not applicable, the failure happens before any schema exists.

SebastianGruza:
## Purpose of the PR

- close #3203

With `usePD=true` the first hstore graph opened at startup (a local `conf/graphs` graph, or the system graph created on the first boot) needs `pd.initial-store-count` active stores, and the store client gives up after a fixed 10 retries (about 38 s). Stores that register later than that make the server exit 1 on a cold start; under Kubernetes the container is restarted and usually succeeds the second time, without an orchestrator the server stays down.

## Before → after

HStore cold start before and after the readiness gate

| Scenario | Before | After |
|---|---|---|
| First boot, Stores register late | The first HStore graph exhausts the Store client's fixed retry window and Server exits. | Before opening any graph, wait for PD to report `Cluster_OK`, bounded by `pd.stores_wait_timeout`. |
| Restart of an initialized cluster | Open existing graphs. | Existing partitions bypass the readiness wait, including when a Store is down. |
| PD cannot be reached | Client RPC timeouts can delay startup. | Each probe has a deadline bounded by the remaining wait budget; timeout reports PD's last message. |

## Main Changes

- Add `pd.stores_wait_timeout` in seconds: default `300`; `0` disables the gate and preserves the previous behavior. Applies when `usePD=true`.
- Run the gate in `GraphManager` before local graphs and PD metadata are loaded. An existing partition lets startup continue; an empty cluster waits for PD's own `Cluster_OK`, without deriving an active-Store threshold from replica count.
- Use short-lived gRPC channels with the existing PD credentials and inbound message size setting. The two RPCs in a probe share one deadline of `min(remaining budget, 5 s)`; channels close after the wait and no watchers are created.
- Keep the generic `NodeTxExecutor` retry behavior unchanged. This is a bounded bootstrap wait, not a guarantee that all later graph operations succeed.

## Verifying these changes

- [x] Follow-up validation on JDK 11: `GraphManagerStoresWaitTest` **5/5 passed**, including a real gRPC partition response over 4 MiB. The new regression test fails with `RESOURCE_EXHAUSTED` before the inbound-limit fix and passes afterward.
- [x] `mvn editorconfig:format`, `git diff --check`, and `mvn clean compile -Dmaven.javadoc.skip=true` passed (all 38 reactor modules).

- [x] Original `unit/core/GraphManagerStoresWaitTest` coverage (added to `UnitTestSuite`), 4/4 on JDK 11: an initialised cluster is not waited for (one probe, zero wait); a cold start waits until PD reports OK, with one unreachable answer retried; a black-holed probe gets deadlines that shrink with the budget and the total stays under timeout + one poll; the timeout message keeps PD's last message and names the option.
- [x] Author-reported reproduction on 3 VMs (1 PD + 3 Store + 1 Server from a tarball, `usePD=true`; same setup as in my comment on #3203, script `cluster/repro_coldstart.sh`, logs in `results/issue-3203/fix/` of https://github.com/SebastianGruza/hugegraph-validation):

| run | result |
|---|---|
| master `1a15e762`, cold start, stores +5 s / +71 s after PD, first boot | 18 × `error code = 105`, backoff 1,1,1,2,3,4,5,6,7,8, `upper limit : 10`, **exit 1 after 38 s** |
| this PR, same cold start | 14 progress lines `Waiting for the PD cluster: Cluster_Not_Ready: The number of active stores is 1, less than pd.initial-store-count:3`, then `PD cluster ready after 70s: PD reports Cluster_OK`, **0 × 105, exit 0 after 81 s**, REST 200 |
| this PR, restart with the node2 store killed (`kill -9`, port closed), cluster initialised | `PD cluster ready after 0s: cluster already has 24 partition(s)`, **exit 0 after 8 s**, REST 200 (PD still listed the store as Up 200 s later, so the skip must not depend on PD's store view) |
| this PR, `pd.peers=192.168.80.250:8686` (black hole), `pd.stores_wait_timeout=20` | `Waiting for the PD cluster: 192.168.80.250:8686: UNAVAILABLE (16s left … 10s … 4s)`, `Timed out after 20s waiting for the PD cluster to be ready (192.168.80.250:8686: UNAVAILABLE); start the stores first or raise pd.stores_wait_timeout`, **exit 1 after 31 s** (10 s JVM boot + 20 s budget) |

## Notes

- The wait sits in the `GraphManager` constructor, not in `createSysGraphIfNeed()`: with a local `conf/graphs` hstore graph that local graph is opened before `loadMetaFromPD()` and already trips the retry ceiling (a first version that waited only before the system graph did nothing on the VMs).
- `Cluster_OK` alone would not do for restarts: `StoreNodeService.checkStoreStatus()` reports `Cluster_Not_Ready` whenever fewer than `pd.initial-store-count` stores are active, so with three stores and `initial-store-count: 3` a restart with one store down would wait too. Hence "partitions exist" as the first criterion and `Cluster_OK` only for an empty cluster.
- The `[wait-storage]` gate in the image entrypoint from #3132 stays the right thing for containers; this change covers the tarball and any start without an orchestrator.

The terminal-log forwarding suggestion in #3203 is outside this change; container logging and orchestration gates remain separate concerns.


hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -482,6 +495,82 @@ private HugeGraph createSysGraphIfNeed() {
...
+    private void waitForActiveStores() {
+        int timeout = this.conf.get(ServerOptions.PD_STORES_WAIT_TIMEOUT);
+        if (timeout <= 0) {
+            return;
+        }
+        PDConfig pdConfig = PDConfig.of(this.pdPeers);
+        pdConfig.setAuthority(PdMetaDriver.PDAuthConfig.service(),
+                              PdMetaDriver.PDAuthConfig.token());
+        // same short-lived client as limitStorage(); PDClient has no close()
+        PDClient pdClient = PDClient.create(pdConfig);
```
imbajin: ⚠️ Blocking: no. `PDClient.create()` is not a one-shot client: the first `getPDConfig()` call enters `newBlockingStub()`, which invokes `startWatch()` and opens the watch streams. This local client is discarded after the wait, while `PDClient` exposes no shutdown (`closeStub()` even notes that the managed channel is not closed), so every `usePD` startup leaves an unused watcher set for the process lifetime. Please use a closable one-shot stub or retain and close this client from the GraphManager lifecycle.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -482,6 +495,82 @@ private HugeGraph createSysGraphIfNeed() {
...
+        try {
+            Metapb.PDConfig pd = pdClient.getPDConfig();
```
imbajin: ⚠️ Blocking: no. `pd.stores_wait_timeout` is not an upper bound for this startup wait. `PDConfig.of(this.pdPeers)` keeps the PD client's default 60-second gRPC deadline, so `getPDConfig()` here and each `getActiveStores()` call inside `waitForStores()` can block for up to 60s; the supplier is invoked before the deadline check as well. With `pd.stores_wait_timeout=20` and a black-holed PD, startup can stall well past 20s. Please propagate a per-call deadline bounded by the remaining wait budget (or otherwise time-bound the RPC) and add a delayed/unreachable-PD test.

imbajin (commented):
Blocking: no. Summary: The startup wait needs a hard deadline for each PD RPC and a lifecycle-safe client. Evidence: current-head static tracing shows the new calls use PDClient's default 60-second gRPC deadline, and the temporary client starts watchers when its first blocking stub is created.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -482,6 +495,82 @@ private HugeGraph createSysGraphIfNeed() {
...
+            int required = pd.getMinStoreCount() > 0 ? pd.getMinStoreCount() :
```
bitflicker64: ⚠️ Important. With the `PDConfig` PD serves, `min_store_count` is always 0 (`ConfigService` only sets partition and shard count), so `required` is `shard_count` on every cluster, and this runs on every start, not just the first boot.

That turns a normal restart into a failure: 3 stores, `default-shard-count: 3`, one store down for maintenance. Each raft group still has 2 of 3 replicas, PD's `checkStoreStatus()` reports `Cluster_OK` (it compares against `pd.initial-store-count` and a per-group majority), and before this PR the server opened its graphs. Now `getActiveStores()` returns 2, the loop waits 300 s and throws. The count is also not what PD enforces on first boot: `allocShards()` checks `pd.initial-store-count` and caps shards at `min(shard_count, stores)`, so a cluster with fewer stores than `shard_count` never starts, and one with `initial-store-count` above `shard_count` still hits the original retry ceiling.

Requested change: gate the wait on PD's own readiness instead of a derived count, for example poll `pdClient.getClusterStats()` until the state is `Cluster_OK` (which already encodes `initial-store-count` and the majority check), or at least skip the wait once shard groups exist. Add a test for the restart-with-a-store-down case.

bitflicker64 (commented):
Blocking: no. Summary: the wait requires shard_count active stores on every start with usePD=true, not only on a cold start, so restarting the server while one of three stores is down now blocks for pd.stores_wait_timeout and exits, although PD reports the cluster as usable. Evidence: static read of GraphManager.waitForActiveStores at 0084e77, PD ConfigService (served PDConfig has no min_store_count), StoreNodeService.allocShards/checkStoreStatus. Latest-head CI is green.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
SebastianGruza: Done in 3d64daa. Instead of a `PDClient` there is now a `PdReadinessProbe`: one plaintext gRPC channel per PD peer, a raw `PDGrpc.newBlockingStub` with the same `Authentication` interceptor that `AbstractClient.setBlockingParams()` adds (credentials from `PDConfig.setAuthority()`, so without PD authentication no interceptor is attached), no watchers, closed with `shutdownNow()` in a try-with-resources once the wait is over. PD forwards requests to its leader on the server side, so any peer answers; the probe rotates through the peers so one dead PD does not eat every poll.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
SebastianGruza: Done in 3d64daa. Every probe call gets `withDeadlineAfter(min(remaining budget, 5 s poll))`, and the budget is checked before the call, not after. `testBlackholedPdStaysWithinTheBudget`: a probe that hangs for its whole deadline on every call receives shrinking deadlines (all ≤ 1 s at a 1 s poll) and the whole wait ends below timeout + one poll. On the lab with `pd.peers=192.168.80.250:8686` (a black hole) and `pd.stores_wait_timeout=20`: `Waiting for the PD cluster: 192.168.80.250:8686: UNAVAILABLE (16s left … 10s … 4s)`, then `Timed out after 20s waiting for the PD cluster to be ready (192.168.80.250:8686: UNAVAILABLE); start the stores first or raise pd.stores_wait_timeout`, exit 1 after 31 s (10 s of JVM boot plus the 20 s budget). Log: `results/issue-3203/fix/after-v2-unreachable-pd-timeout20.log` in https://github.com/SebastianGruza/hugegraph-validation.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
SebastianGruza: Done in 3d64daa, thanks, that case was a real regression. The probe first asks `queryPartitions` with an empty query: if the cluster has any partition it is ready and nothing is waited for, whatever PD thinks of the stores at that moment. Only a cluster without partitions (first boot) waits, and it waits for `getClusterStats() == Cluster_OK`, which is exactly what `allocShards()` needs, with no count derived on the server side. One correction to your description: `checkStoreStatus()` also sets `Cluster_Not_Ready` when fewer than `pd.initial-store-count` stores are active (`StoreNodeService.java:831-835`), so with three stores and `initial-store-count: 3` `Cluster_OK` on its own would block a restart with one store down as well; hence "partitions exist" as the first criterion and `Cluster_OK` only for an empty cluster.

Measured on the lab (1 PD + 3 Store + 1 Server from a tarball, `usePD=true`): a cold start with stores at +5 s / +71 s → 14 progress lines carrying PD's message (`The number of active stores is 1, less than pd.initial-store-count:3`), `PD cluster ready after 70s: PD reports Cluster_OK`, 0 × 105, exit 0 after 81 s, REST 200; a restart with the node2 store killed with `kill -9` (port 8500 closed) → `PD cluster ready after 0s: cluster already has 24 partition(s)`, exit 0 after 8 s, REST 200; PD still listed that store as Up 200 s later, which only confirms that the skip must not depend on PD's view of the stores. `testInitialisedClusterIsNotWaitedFor` covers the case (one probe, zero wait). Logs: `results/issue-3203/fix/after-v2-*.log`.

SebastianGruza:
Round 1 in 3d64daa: a one-shot gRPC probe (a channel per PD peer, no watchers, closed after the wait) replaces the temporary `PDClient`, every call carries a deadline bounded by the remaining budget, a cluster that already has partitions is not waited for at all, and a cluster without partitions waits for PD's own `Cluster_OK` instead of a store count derived on the server. Four unit tests (initialised cluster, cold start, black-holed PD within budget, timeout keeping PD's last message), 4/4 on JDK 11. Three lab runs in the table in the description: cold start with 70 s of waiting and exit 0, restart with a killed store in 8 s and exit 0, unreachable PD exiting 1 after the 20 s budget. Description updated.

imbajin (approved):
+1, the current scope is appropriate for #3203. The bounded wait before opening any graph addresses the cold-start failure while preserving existing-cluster startup and the generic Store retry behavior. The earlier review concerns and the probe response-size regression are addressed.

A non-blocking follow-up would be a lightweight PD bootstrap-status query, avoiding the full partition-list response, with probing encapsulated in a closable PD client helper. That broader API/client change does not need to expand this fix.

Validation: JDK 11 full reactor compilation and 5/5 focused tests passed, including a real gRPC response over 4 MiB that failed before the fix. CI for this head is still running; merge after it passes.

PR #3220, opened 2026-09-18
fix(server): bind the meta cluster name before any graph is opened

### LINKED ISSUE #3219: [Bug] Server upgraded from 1.7.0 with usePD=true reads an empty schema: meta cluster prefix 'hg-test' vs hardcoded 'hg'
### Bug Type (问题类型)

data / logic error

### Before submit

### Environment (环境信息)

- Server Version: master `4f5202b8` (PD and store on master as well; the same happens with PD and stores left on 1.7.0)
- Backend: HStore, 1 PD + 3 stores, `usePD=true`
- Data written by the official `apache-hugegraph-incubating-1.7.0` release

After swapping only the jars (`lib/`) to master on a 1.7.0 installation (config and data directories untouched) the server lists the graph and sees the raw data, but not the schema:

```
GET /graphs/hugegraph/schema/propertykeys   -> {"propertykeys": []}
g.V().count()                                -> 29
g.V().hasLabel('person').count()             -> Undefined vertex label: 'person'
```

On the same data 1.7.0 returns 13 property keys and 20 `person` vertices. With `usePD=false` the same swap works (HStore and RocksDB, every schema and data read identical to 1.7.0).

### Root cause

The meta keys in PD are prefixed with `HUGEGRAPH/<cluster>/…`, and `MetaManager.connect()` binds the cluster name once per process: a later call with a different name is silently ignored. Two places supply the name: `GraphManager` from the `cluster` option of `rest-server.properties` (default `hg-test`), and `StandardHugeGraph` with the literal `"hg"`. Whoever connects first wins:

- **1.7.0**: `HugeRestServer.start()` runs first; the `GraphManager` constructor with `usePD=true` calls `loadMetaFromPD()` → `initMetaManager()` → `connect("hg-test")`. The schema is written under `HUGEGRAPH/hg-test/GRAPHSPACE/DEFAULT/hugegraph/SCHEMA/…`.
- **master**: #3008 moved `HugeGremlinServer.prepare()` in front of `HugeRestServer.start()`. `prepare()` puts every file from `conf/graphs` into `settings.graphs`, TinkerPop opens the `StandardHugeGraph`, and it calls `connect("hg")` first. The configured `cluster` is ignored from then on; the server reads `HUGEGRAPH/hg/…`, which only holds the system schema it created itself at startup.

Key listing through `KvClient.scanPrefix("HUGEGRAPH/")` after the swap: 106 keys under `hg-test/…/hugegraph` (written by 1.7.0), 36 under `hg/…/hugegraph` (created by master at startup). Removing `hugegraph.properties` from `conf/graphs` does not help (the graph is not registered in the meta, the server then has no graph at all), and setting `cluster=hg-test` does not either (the literal wins regardless of the option).

### How to reproduce

1. 1.7.0, HStore, `usePD=true` and `pd.peers` in `rest-server.properties`; create a schema and some data.
2. Stop, replace the server `lib/` with master (PD and stores may stay on 1.7.0 or move to master, same result), start without `init-store`.
3. `GET /schema/propertykeys` → empty list, `g.V().hasLabel(...)` → Undefined vertex label.

Scripts and logs from the lab run: https://github.com/SebastianGruza/hugegraph-validation/tree/master/results/upgrade-170-to-master. A fix is in the linked PR.

_No response_

_No response_

SebastianGruza:
## Purpose of the PR

Closes #3219. A server upgraded from 1.7.0 with `usePD=true` reads an empty schema, because `StandardHugeGraph` connects the `MetaManager` under the literal `hg` before `GraphManager` gets to use the configured `cluster` (`hg-test`); since #3008 `HugeGremlinServer.prepare()` opens the graphs from `conf/graphs` before the REST server starts, so the graph-level connect wins. The meta keys are prefixed with the cluster name and `MetaManager.connect()` binds it once per process. Mechanism and key listing in the issue.

## Main Changes

- `HugeGraphServer`: with `usePD=true`, connect the `MetaManager` under `ServerOptions.CLUSTER` before `HugeGremlinServer.prepare()`, through the new static `GraphManager.connectMetaManager(conf)`, which `initMetaManager()` uses as well (same CA handling, one place).
- `StandardHugeGraph`: connect only when nothing connected yet (`isReady()`), and take the cluster name from the new `CoreOptions.PD_CLUSTER` (`pd.cluster`) instead of the literal. The default stays `hg`, so `usePD=false` deployments keep reading their existing keys; with `usePD=true` the server's `cluster` wins, as in 1.7.0. Startup order no longer matters.
- `MetaManager`: `connect()` logs a WARN when it ignores a different cluster name; new `ensureCluster(expected)` throws `IllegalStateException` naming both clusters and the prefix the keys live under. The server calls it right after its own connect, so a conflict fails the startup instead of yielding an empty schema.

## Verifying these changes

- [x] Make sure there is no significant performance impact
- [x] Add new unit test(s): `MetaManagerClusterTest` (6 tests, registered in `UnitTestSuite`): the first connect wins and later names are ignored; `ensureCluster` fails with both names and requires a connection; `GraphManager.connectMetaManager` detects an earlier graph-level connect under `hg` and is idempotent under the configured name; the `pd.cluster` default is `hg`. Mocked `MetaDriver`, no PD needed.
- [x] Lab, 1 PD + 3 stores, data written by the official 1.7.0 release with `usePD=true`: master without this change → 0 property keys, `Undefined vertex label`; master with this change → 13 property keys, 3 vertex labels, 3 edge labels, 4 index labels, 29 V / 63 E, 20/20 REST and Gremlin reads identical to the 1.7.0 baseline, no cluster warning in the log. With `usePD=false` the swap worked before and works after (HStore and RocksDB). Scripts and logs: https://github.com/SebastianGruza/hugegraph-validation/tree/master/results/upgrade-170-to-master

- [ ] Nope
- [x] Dependencies (add/update license info) — no
- [ ] Modify configurations
- [x] Introduce new configurations: `pd.cluster` in `CoreOptions`
- [ ] The public API
- [ ] Other affected components or behavior

Docs: the description of `cluster` should say that with `usePD=true` it is the prefix of the meta keys in PD and must stay the same across upgrades.

## Notes

Out of scope, for a separate thread: a small migration tool for the prefix (`hg-test` → `hg` or back, about 100 keys per graph through `KvClient`), and a startup diagnostic listing the existing prefixes when the configured one is empty.


hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
```diff
@@ -263,9 +263,13 @@ public StandardHugeGraph(HugeConfig config) {
             throw new HugeException(message);
         }
 
-        if (isHstore()) {
+        if (isHstore() && !MetaManager.instance().isReady()) {
```
bitflicker64: 🧹 Minor. With this guard, the WARN added to `MetaManager.connect()` never fires for a graph, so a conflicting `pd.cluster` is dropped without any message.

`pd.cluster` is a per-graph option, but the prefix is bound once per process. With `usePD=false` and two hstore graphs in `conf/graphs`, say one with `pd.cluster=a` and one with `pd.cluster=b`, the second graph sees `isReady() == true`, skips `connect()`, and reads and writes under `HUGEGRAPH/a/` without a log line. That is the same silent empty schema this PR fixes, just triggered by the new option. Before this change the literal `hg` made a conflict impossible.

Could you add an `else` branch that warns (or fails) when `config.containsKey(CoreOptions.PD_CLUSTER.name())` and the value differs from `MetaManager.instance().cluster()`? The `containsKey` check keeps it quiet for usePD=true graphs that leave the default. Please also say in the `PD_CLUSTER` description that the value is process-wide and the first hstore graph opened wins.

hugegraph-server/hugegraph-dist/src/main/java/org/apache/hugegraph/dist/HugeGraphServer.java
```diff
@@ -68,6 +69,15 @@ public HugeGraphServer(String gremlinServerConf, String restServerConf)
                     ServiceConstant.SERVICE_NAME,
                     ServiceConstant.AUTHORITY);
 
+            // Bind the meta cluster name ('cluster' in rest-server.properties)
+            // before any graph is opened: prepare() below opens every graph
+            // in conf/graphs, and an hstore graph would otherwise connect the
+            // MetaManager first under its own 'pd.cluster' (default 'hg'),
+            // hiding the meta written under the configured cluster
+            if (restServerConfig.get(ServerOptions.USE_PD)) {
+                GraphManager.connectMetaManager(restServerConfig);
```
bitflicker64, line 78: 🧹 Minor. This moves the prefix back to `hg-test` for every usePD=true server that has been running a master build since #3008 (merged 2026-07-10). Those servers wrote their schema, graph spaces and users under `HUGEGRAPH/hg/`. After this change they come up with an empty schema, and `ensureCluster` passes, so the log gives no hint. `docker/docker-compose-hstore.yml:92` sets `HG_SERVER_USE_PD: "true"` and no `HG_SERVER_CLUSTER`, so anyone on the compose setup from master gets the `hg-test` default (`ServerOptions.CLUSTER`).

Restoring 1.7.0 behaviour is the right call, and the migration tool is fine as a follow-up. Could you log the bound prefix at INFO here (for example `Meta cluster bound to '{}' (keys under HUGEGRAPH/{}/)`) and add a line to the upgrade notes for snapshot users? Then an empty schema after upgrading shows its cause in the first lines of the log.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/MetaManagerClusterTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Before
+    public void setup() throws Exception {
+        this.originalDriver = swapField("metaDriver", null);
+        this.originalCluster = swapField("cluster", null);
+    }
+
+    @After
+    public void teardown() throws Exception {
+        swapField("metaDriver", this.originalDriver);
```
bitflicker64: 🧹 Minor. The teardown restores `metaDriver` and `cluster`, but every `connect()` in these tests also runs `initManagers()`, which replaces the ten sub-managers (`authMetaManager` through `lockMetaManager`) with ones bound to the Mockito driver and the test cluster. After this class runs, the `MetaManager` singleton reports `isReady() == false` but its managers point at a mock. Any later test in `UnitTestSuite` that reaches a manager without connecting gets a silent mock instead of the NPE it would get today.

`GraphManagerAdminInitTest` in the same package already restores each manager field it swaps. Please snapshot and restore the sub-manager fields too, for example by iterating the non-static declared fields of `MetaManager` in `setup()`/`teardown()`, so the suite leaves the singleton as it found it.

bitflicker64 (commented):
Blocking: no. Summary: binding the server's `cluster` in `HugeGraphServer` before `HugeGremlinServer.prepare()` fixes #3219 the right way, and `usePD=false` keeps its `hg` prefix; three small follow-ups remain: a conflicting per-graph `pd.cluster` is now dropped without the new warning, usePD=true deployments running a master build since #3008 flip from `HUGEGRAPH/hg/` back to `HUGEGRAPH/hg-test/` with nothing in the log naming the prefix, and the new test leaves the singleton's sub-managers bound to a mock driver. Evidence: exact-head diff at cc3018f14daf80225121a86559706c2111039f4b against merge base 4f5202b8; reads of `MetaManager.connect/initManagers`, `GraphManager` constructor and `loadMetaFromPD()`, `StandardHugeGraph` constructor, `ServerOptions.CLUSTER` (default `hg-test`), `docker/docker-compose-hstore.yml:92`; CI at this head: all workflows green, `UnitTestSuite` 745 run / 0 failures in build-server (memory) with the new tests, hstore job (usePD=true via start-server.sh) green; `codecov/project` reports failure (-0.02%).

hugegraph-server/hugegraph-core/src/main/java/org/apache/hugegraph/StandardHugeGraph.java
SebastianGruza: Done in 68b616a2. Added the `else` branch: when the MetaManager is already bound and the graph sets `pd.cluster` explicitly (`config.containsKey(...)`) to another value, a WARN names both clusters and the prefix the keys live under, and the graph's value is ignored. I kept a WARN rather than a throw: with `usePD=true` the server already enforced its own `cluster` through `ensureCluster()`, and with `usePD=false` failing the whole process on the second file in `conf/graphs` seemed worse than a clear log line. The `PD_CLUSTER` description now says the value is process-wide: the server's `cluster` or the first hstore graph opened wins, a later different value is ignored with a warning. Checked on the lab with a second graph carrying `pd.cluster=other`:

```
[WARN] o.a.h.StandardHugeGraph - Graph 'hugegraph2' sets pd.cluster='other' but the meta cluster is already bound to 'hg-test' (keys under HUGEGRAPH/hg-test/); the graph's value is ignored
```

hugegraph-server/hugegraph-dist/src/main/java/org/apache/hugegraph/dist/HugeGraphServer.java
SebastianGruza: Done in 68b616a2. After `connectMetaManager()` the server logs `Meta cluster bound to 'hg-test' (keys under HUGEGRAPH/hg-test/)` at INFO, in the first lines of the startup (line 5 of the log on the lab). I added a paragraph to `hugegraph-store/docs/operations-guide.md`, step "Upgrade Server Nodes": 1.7.0 bound `cluster` (`hg-test`), master builds between #3008 and this PR bound the literal `hg`, from this PR on `cluster` again; whoever sees an empty schema after an upgrade compares the log line with the prefix their data lives under and sets `cluster` in `rest-server.properties`. I did not find a dedicated upgrade-notes file in the repo, so if there is a better place I'll move it. The prefix migration tool stays a follow-up, as agreed.

hugegraph-server/hugegraph-test/src/main/java/org/apache/hugegraph/unit/core/MetaManagerClusterTest.java
SebastianGruza: Done in 68b616a2, thanks, I had missed that. `setup()` now snapshots every non-static instance field of `MetaManager` through reflection and `teardown()` restores them by reference, so the ten sub-managers return to what they were before the class, not just `metaDriver` and `cluster`. A new `testTeardownRestoresTheSubManagers` checks it on `authMetaManager`: a different reference after `connect()`, the original one after `teardown()`. 7/7.

SebastianGruza:
Round 1 in 68b616a2: a WARN for a graph whose `pd.cluster` differs from the bound cluster (the value is process-wide, said so in the option), an INFO line with the bound prefix at startup plus a paragraph in the operations guide about `hg-test`/`hg` after an upgrade, and the test now restores the whole singleton state. On the lab with 1.7.0 data and `usePD=true`: the INFO line is in the log, a second graph with `pd.cluster=other` gets the WARN and the schema stays visible (13 property keys). `MetaManagerClusterTest` 7/7.

bitflicker64 (approved):
Checked the exact head 68b616a2 against master 4f5202b8. Binding `cluster` in `HugeGraphServer` before `HugeGremlinServer.prepare()` restores the 1.7.0 meta prefix for usePD=true with the default `graph.load_from_local_config=false`, usePD=false keeps `hg` through the `pd.cluster` default, and `ensureCluster()` turns a prefix mismatch into a startup error instead of an empty schema. The three points from the earlier round (warning for a conflicting per-graph `pd.cluster`, the startup INFO line plus the upgrade note, restoring the whole `MetaManager` singleton in the test) are in this head. CI is green, including `UnitTestSuite` (746 run, 0 failures) and the hstore job.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
```diff
@@ -727,19 +727,38 @@ public void destroy() {
...
             clientCa = conf.get(ServerOptions.META_CLIENT_CA);
             clientKey = conf.get(ServerOptions.META_CLIENT_KEY);
-            this.ca = new K8sDriver.CA(ca, clientCa, clientKey);
         }
-        this.metaManager.connect(this.cluster, MetaManager.MetaDriverType.PD,
-                                 ca, clientCa, clientKey, endpoints);
+        MetaManager manager = MetaManager.instance();
+        manager.connect(cluster, MetaManager.MetaDriverType.PD,
+                        ca, clientCa, clientKey, endpoints);
+        manager.ensureCluster(cluster);
```
imbajin, line 767: ‼️ Critical. In a `usePD=false` HStore startup, `HugeGraphServer` skips the pre-bind at its `usePD` guard, while `HugeGremlinServer.prepare()` opens the graph before `HugeRestServer.start()`; the graph then binds `MetaManager` through the `pd.cluster` fallback (default `hg`). `GraphManager.initMetaManager()` subsequently calls this method unconditionally, and the default server `cluster` is `hg-test`, so `ensureCluster(cluster)` throws instead of preserving the graph-level binding. This regresses the usePD=false compatibility path described by the PR. Please gate this check to the explicit usePD=true server bind, or otherwise use the already-bound graph cluster for usePD=false, and add a startup test covering that path.

imbajin (commented):
Blocking: yes. Summary: The usePD=false HStore startup path still fails when the graph binds MetaManager under its pd.cluster before GraphManager initializes it. Evidence: exact-head source shows HugeGraphServer pre-binds only when usePD=true, HugeGremlinServer.prepare() runs before HugeRestServer.start(), and ServerOptions.CLUSTER defaults to hg-test while CoreOptions.PD_CLUSTER defaults to hg; CI/status failures: none.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/core/GraphManager.java
SebastianGruza: Thanks for tracing this path; I checked it two ways and would like to be precise, because in the described form it does not occur. `ensureCluster()` is called only from `connectMetaManager()`, which has two callers: `HugeGraphServer` (explicitly behind the `usePD` guard) and `GraphManager.initMetaManager()`. `initMetaManager()` has a single caller, `loadMetaFromPD()` (line 376), and `loadMetaFromPD()` runs only in the constructor under `if (PDExist)` (lines 271-274), i.e. only with `usePD=true`. With `usePD=false` the server never reaches `ensureCluster` at all; `StandardHugeGraph` binds `pd.cluster` and that is the end of it, exactly as in 1.7.0.

Measured on the lab (1 PD + 3 stores, data written by the official 1.7.0 release, a build of this head), with `usePD` and `pd.peers` removed from `rest-server.properties`: with the default `pd.cluster` the server starts, the graph opens, 0 property keys (the data lives under `hg-test`, which is expected), and zero `ensureCluster`/`IllegalStateException` lines in the log; with `pd.cluster=hg-test` in the graph file the server starts and sees all 13 keys, log equally clean. With `usePD=true` restored: 13 keys and `Meta cluster bound to 'hg-test'`. Log: `results/upgrade-170-to-master/logs/usepd-false-round2.txt` in the validation repo.

I added the gate anyway in 8e914e6f, since it is cheap and protects against a future caller: `connectMetaManager()` returns without binding or checking when `usePD=false`, with a comment that the graph-level binding is then the only one and no server-side check may apply to it. Two tests: `testUsePdFalseLeavesTheGraphLevelBindingAlone` (a graph bound `hg`, calling the helper with `usePD=false` does not throw and the cluster stays `hg`) and `testGraphManagerStartupWithUsePdFalseKeepsTheGraphBinding` (a `GraphManager` constructed with `usePD=false` after such a binding completes without an exception, cluster still `hg`). `MetaManagerClusterTest` 9/9.

SebastianGruza:
Round 2 in 8e914e6f, for the note at `GraphManager.java:761`: the `usePD=false` path does not reach `ensureCluster` (`initMetaManager()` only runs from `loadMetaFromPD()`, i.e. under `usePD=true`), which I confirmed on the lab with a build of this head: with `usePD=false` the server starts with the default `pd.cluster` and with `pd.cluster=hg-test` (0 and 13 keys, as the prefixes imply) and no cluster-related line in the log. `connectMetaManager()` nevertheless got an explicit `usePD` gate and two tests covering a start with `usePD=false` after a graph bound `hg`; 9/9. Details and the log in the inline thread.

bitflicker64 (approved):
Checked the exact head 8e914e6f against 68b616a2. The change is the `usePD` gate at the top of `GraphManager.connectMetaManager()`, two new tests, and a `serverConfig()` helper that the existing server-side tests now use. On imbajin's usePD=false point at 68b616a2, I agree with the author's reply: `initMetaManager()` is reached only from `loadMetaFromPD()` under `if (PDExist)` (GraphManager.java:271-274), so that head did not reach `ensureCluster()` with usePD=false. The gate now makes this explicit for future callers, and `testUsePdFalseLeavesTheGraphLevelBindingAlone` fails without it. CI is green at this head, including `UnitTestSuite` (748 run, 0 failures, up from 746) and the hstore job.

imbajin (approved):

PR #3218, opened 2026-09-18
feat(helm): add HStore deployment chart

bitflicker64:
## Purpose of the PR

- close #3131

Add an official Helm installation path for the distributed HugeGraph HStore topology on Kubernetes. The chart packages the PD, Store and Server startup contract that otherwise has to be reconstructed by each operator, with authentication on by default and an optional Hubble UI.

This PR succeeds #3132 (closed, history and earlier review threads preserved there). Its head is the org branch `hugegraph:feat/hstore-helm-chart-3132`, the same branch as hugegraph/hugegraph#221, so this PR and the org-side review stay on one branch and cannot drift. Review happens on hugegraph/hugegraph#221, where the GitHub review apps can run; ASF repository settings block their permissions here.

## Main Changes

- `helm/hugegraph/`: chart 0.1.0 with PD and Store StatefulSets, a Server Deployment, an optional Hubble Deployment, three values presets (`values.yaml`, `values-single.yaml`, `values-cluster.yaml`), a full `values.schema.json`, and 66 helm-unittest cases.
- Encodes the distributed startup contract: Server skips `init-store` and registers per-Pod with PD, the Store waits for a PD quorum on `/v1/ready`, PD readiness is quorum-aware while startup and liveness stay on `/v1/health`, and `HG_SERVER_STARTUP_TIMEOUT_S` is derived from the startup probe budget.
- One shared PD REST secret (`pd.auth`) consumed by PD, the Server storage wait and Hubble; chart-managed admin and JWT Secrets with bring-your-own `existingSecret` paths.
- `.github/workflows/helm-chart-ci.yml`: lint, renders on all presets, fail-path guards, kubeconform, a legacy `--reuse-values` fixture render, and packaging.
- README covering install, upgrade rolls, disaster recovery, troubleshooting and Limitations.

## Verifying these changes

- [ ] Trivial rework / code cleanup without any test coverage. (No Need)
- [ ] Already covered by existing tests, such as *(please modify tests here)*.
- [x] Need tests and can be verified as follows:
    - Helm Chart CI: lint, preset renders, 15 fail-path guards, kubeconform (1.23.0 and current), legacy-values render, packaging.
[...]

PR #3221, opened 2026-09-18
feat(server): add a storage-aware GET /readiness endpoint

### LINKED ISSUE #3212: [Feature] Server needs a storage-aware readiness signal (readiness stays green with zero Stores)
## Feature Description (功能描述)

The Server has no endpoint a Kubernetes readiness probe can use that reflects whether it can actually serve graph traffic. `/versions` (what the #3132 chart uses today) answers 200 as long as the REST layer is up, even when the cluster has no Store at all.

Measured independently by @SebastianGruza on 2026-09-16 (apache/hugegraph#3132, comment of 19:31 CEST; scripts in SebastianGruza/hugegraph-validation, `docs/helm-chart-faults.md` scenario 7): with all Stores scaled to 0 on a running cluster, the Servers stayed `Running` and `ready=true` with 0 restarts for 150+ seconds, `/versions` answered 200 throughout, while every `GET /graph/vertices/<id>` ended in a 500 after the 30 s request bound (#3204). A Kubernetes Service therefore keeps routing traffic to Servers whose every real request fails.

Today the readiness probe answers 200 from /versions while every graph request fails on 0 Stores; the proposed storage-aware readiness turns the pod NotReady so the Service stops routing to it

This is the Server-side twin of #3183, which was fixed for PD by #3185's unauthenticated `/v1/ready`.

Proposal: an unauthenticated, cheap readiness endpoint on the Server (for example `/apis/readiness` or a flag on `/versions`) that returns 503 unless:

1. the Server's PD client currently has a usable PD connection, and
2. at least one Store is registered and Up from that Server's view (the graph list from PD plus one cheap Store call is the shape suggested in the #3132 thread).

It must not require Basic auth: a Kubernetes `httpGet` probe cannot carry a credential without leaking it into the pod spec, which is exactly why #3185 made `/v1/ready` unauthenticated and why the auth interceptor already excludes `/actuator/**`, `/v1/health` and `/v1/ready` on PD.

Chart side once this exists: `server.readinessPath` flips from `/versions` to the new path the same way `pd.readinessPath` moved to `/v1/ready`, and a Server with no working storage drops out of the Service instead of serving 500s.

### Visual summary

Storage-aware readiness

SebastianGruza:
## Purpose of the PR

Closes #3212. A Kubernetes readiness probe on `/versions` keeps a Server in the Service while every graph request fails, because `/versions` answers 200 as long as the REST layer is up, even with no Store in the cluster (measured under #3132: `ready=true` with 0 Stores for 150+ s while every `GET /graph/vertices/<id>` ended in a 500 after the 30 s request bound). Design discussed in the issue.

## Main Changes

`GET /readiness` answers 200 while this Server can serve graph traffic and 503 otherwise, unauthenticated like `/versions` (whitelisted in `AuthenticationFilter` and `PathFilter`, so an `httpGet` probe needs neither a credential nor a graphspace prefix), with a JSON body that carries no addresses:

```
{"ready":true,"storage":"hstore","reason":"ok","active_stores":3,"answered_store":6386607161404282741,
 "pd_reachable":true,"pd_checked_age_ms":312,"stores_age_ms":312,"store_millis":3,"cached":false}
```

Ready means, from this Server's own view: at least one Store from the last Store list PD answered with answers a direct, local, read-only gRPC call (`HgStoreState.getScanState`, a read of the node's own scan-pool stats that never touches raft). The Store list is refreshed from PD in the background on every probe and never waited for once a list is known, so PD being down, slow or restarting only flips the reported `pd_reachable`, never the readiness, as long as a Store answers. The pings run in parallel and the first answer wins, so a Store whose pod just left never eats the budget of the healthy ones.

- `hugegraph-hstore`: `HstoreStorageProbe` (pure logic over a Store lister and a Store pinger, one shared time budget through an executor so a hung PD or Store yields "did not answer within N ms" instead of a hung probe; `KnownStores` holds the last PD answer) and a `storage_readiness` meta handler in `HstoreStore`.
- `hugegraph-api`: `ReadinessAPI` and `StorageReadiness` (first hstore graph is probed as the internal admin through `HugeGraph.metadata(null, "storage_readiness", timeout)`, result cached for `readiness.cache_ttl`; Servers with no hstore graph answer 200 with `storage=embedded`); the api module gains no dependency.
- `ServerOptions`: `readiness.timeout` (default 1000 ms) and `readiness.cache_ttl` (default 2000 ms).

## Verifying these changes

- [x] Add new unit test(s): `HstoreStorageProbeTest` (15: first answer wins, a hung Store does not hide an answering one, known Stores keep the Server ready while PD is down, a hung PD does not delay a probe with known Stores, the PD answer updates the known list, budget bounding, no addresses in the body) and `StorageReadinessTest` (6, including the two filter whitelists).
- [x] E2E on a three-node k3s with the chart from #3218 / hugegraph/hugegraph#221, 1 Hz sampling of `/readiness`, `/versions`, the pod Ready condition and the Service endpoints on every Server pod, load through the Service: **7 PASS** (baseline; Stores→0 pulls every Server out of the Service in ~30 s and back in ~20 s; one Store deleted, no flap; **rolling restart of the Stores under load, zero readiness transitions**; PD→0 stays 200 with `pd_reachable=false`; rolling restart of PD, zero transitions; SIGSTOP the busiest Store, zero transitions). Scripts, per-scenario JSON and the two superseded probe designs (an `existsTable` ping flapped on a roll; probing PD on every request tracked PD instead of storage): https://github.com/SebastianGruza/hugegraph-validation/tree/master/results/issue-3212

- [x] Introduce new configurations: `readiness.timeout`, `readiness.cache_ttl` in `ServerOptions`
- New unauthenticated endpoint `GET /readiness`; no data-format change.

## Chart side

A companion change adds `server.readinessPath` (default `/versions`) to the #3132 chart, mirroring `pd.readinessPath`; set it to `/readiness` on an image that serves it. Diff and measurements are in the validation repo above and will go to hugegraph/hugegraph#221.


hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,348 @@
...
+            Throwable cause = e.getCause() != null ? e.getCause() : e;
+            throw cause instanceof Exception ? (Exception) cause : new RuntimeException(cause);
+        }
+    }
+
+    private static long elapsed(long since) {
+        return System.currentTimeMillis() - since;
+    }
+
+    private static String message(Throwable e) {
```
bitflicker64: ⚠️ `/readiness` is unauthenticated and the PR says the body carries no addresses, but `reason` embeds raw exception messages. Two paths:

- No Store list known yet and the PD client has no live stub (never connected, or reset by `closeStub` after a watch error) with peers refusing fast: `await()` unwraps the `PDException` from `newBlockingStub()`, whose message is `"PD unreachable, pd.peers=" + config.getServerHost()` (PDClient.java:142-143), so every PD peer is listed.
- Every known Store fails and one failed name resolution: `UNAVAILABLE: Unable to resolve host <host>` (grpc-core 1.47 `DnsNameResolver`), if Stores register by DNS name.

`StorageReadiness.check` also appends `e.getMessage()`; `testMapCarriesNoAddresses` covers only the ready path.

Could `reason` use a fixed category (`pd unreachable`, the gRPC status code or exception class) and log the full message instead?

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,348 @@
...
+
+    public static Result probe(KnownStores known, StoreLister lister, StorePinger pinger,
+                               long timeoutMs, ExecutorService executor) {
+        E.checkArgument(timeoutMs > 0, "The probe timeout must be > 0, but got %s", timeoutMs);
+        long deadline = System.currentTimeMillis() + timeoutMs;
+
+        // Refresh the store list from PD in the background; whatever PD
+        // answers lands in `known` for this or the next probe
+        CompletableFuture<List<Metapb.Store>> refresh = new CompletableFuture<>();
+        executor.execute(() -> {
```
bitflicker64: ⚠️ Every cache-missed probe submits a new PD refresh here, even while the previous one is still running, and `EXECUTOR` is an unbounded cached pool. `getActiveStores()` is bounded only by the PD client deadline, 60 s by default (`PDConfig.grpcTimeOut`; `HstoreSessionsImpl.initStoreNode` does not override it).

If PD hangs instead of refusing connections, the probe still returns in milliseconds from the known Stores, so each probe parks one more thread on PD: up to about 30 at the default 2 s TTL, and one per request with `readiness.cache_ttl=0`, which the option allows, on an endpoint anyone can call. Without a live stub they also queue on the `synchronized` `PDClient.newBlockingStub()` alongside graph traffic.

Could the refresh be single-flight, e.g. keep the in-flight `CompletableFuture` and skip submitting until it completes?

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/ReadinessAPI.java
```diff
@@ -0,0 +1,66 @@
...
+import jakarta.inject.Singleton;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.Response;
+
+/**
+ * Storage-aware readiness for Kubernetes and load balancers: 200 while this
+ * server can serve graph traffic, 503 while PD or every Store is unreachable
```
bitflicker64: 🧹 This Javadoc says 503 while PD is unreachable. So do the `StorageReadiness` Javadoc ("PD answers and one active Store answers"), the `HstoreStorageProbe` class Javadoc ("a hung PD or Store turns into not ready") and the `readiness.timeout` description. The probe deliberately stays ready with PD down once a Store list is known (`testKnownStoresKeepTheServerReadyWhilePdIsDown`). Both option descriptions also say one Store call per probe, while it pings every known Store in parallel. Please reword these to match: PD only matters until the first Store list is known, and the Store cost is one call per known Store.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,348 @@
...
+        if (pd == null) {
+            return new Result(false, "pd client not initialised", 0, null,
+                              false, -1L, -1L, 0L).toMap();
+        }
+        return probe(KNOWN, pd::getActiveStores, HstoreStorageProbe::pingScanState,
+                     timeoutMs, EXECUTOR).toMap();
+    }
+
+    private static void pingScanState(Metapb.Store store, long timeoutMs) {
+        ManagedChannel channel = CHANNELS.computeIfAbsent(store.getAddress(), address -> {
```
bitflicker64, line 270: 🧹 `CHANNELS` keeps one plaintext channel per Store address and nothing ever removes or shuts one down. When a Store is replaced or comes back under a different address, the old channel stays allocated for the life of the process. Could channels whose address is no longer in the latest PD list be shut down after a successful refresh in `probe`?

bitflicker64 (commented):
Blocking: no. Summary: the probe design (last known Store list, parallel pings, one shared budget) holds up and is well tested, but the unauthenticated endpoint's body can leak PD peer and Store host names through exception text, and a hung PD makes every cache-missed probe park another thread on an unbounded pool. Two doc and resource nits besides. Evidence: static read of the d0f4a25 diff against PDClient (getStub/newBlockingStub, PDConfig.grpcTimeOut=60000), HstoreSessionsImpl.initStoreNode and grpc-core 1.47 DnsNameResolver; CI green on this head apart from codecov, and HstoreStorageProbeTest (15) passes in the hstore job log.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,348 @@
...
+ * slow, restarting or down does not change the readiness of a server whose
+ * Stores still answer. Every wait is bounded by one shared time budget, so a
+ * hung PD or Store turns into "not ready" instead of a hung probe. The result
+ * carries no addresses, since it is served without authentication.
+ */
+public final class HstoreStorageProbe {
+
+    public static final String META_STORAGE_READINESS = "storage_readiness";
+
+    private static final ExecutorService EXECUTOR = Executors.newCachedThreadPool(
```
bitflicker64, line 73: ⚠️ This ten-line anonymous `ThreadFactory` with its own `AtomicInteger` is the thing `BasicThreadFactory.Builder` exists for — and this repo already uses it exactly this way, four times, in `hugegraph-common`'s `ExecutorUtil`. commons-lang3 is already imported elsewhere in this module (`HstoreSessionsImpl`, `HstoreTable`), so no new dependency.

Requested change:

```java
private static final ExecutorService EXECUTOR = Executors.newCachedThreadPool(
        new BasicThreadFactory.Builder().namingPattern("storage-readiness-%d")
                                        .daemon(true).build());
```

Same names, same daemon flag, eight lines and one import (`java.util.concurrent.ThreadFactory` + `AtomicInteger`) gone.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    /**
+     * Probe through the process-wide PD client and this probe's own plaintext
+     * channels to the stores (the store gRPC server takes no credentials).
+     *
+     * @param graphName the store-side graph name, kept for the meta handler
+     * @param timeoutMs the whole budget for PD plus stores
+     */
+    public static Map<String, Object> probe(String graphName, long timeoutMs) {
```
bitflicker64, line 237: 🧹 `graphName` is never read in this method — the javadoc even says "kept for the meta handler", which is scaffolding for a use that doesn't exist yet. The call site in `HstoreStore` builds `this.namespace + "/" + this.store` just to feed it.

Requested change: drop the parameter here and the concat at the call site. Re-add it the day something reads it — `pd.getActiveStores(graphName)` exists if a per-graph store list is ever wanted.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,348 @@
...
+            }
+        }
+
+        public void pdFailed() {
+            this.pdOk = false;
+            this.pdAt = System.currentTimeMillis();
+        }
+    }
+
+    public static final class Result {
```
bitflicker64, line 173: 🧹 The only production consumer of `Result` is `probe(String, long)` calling `.toMap()` on it immediately; outside this file the class appears only in `HstoreStorageProbeTest`. So these ~55 lines — eight final fields, an eight-positional-arg constructor (`new Result(false, "...", 0, null, false, -1L, -1L, 0L)` is hard to read at the four construction sites), five getters — exist to make test assertions prettier.

Requested change: have `probe(...)` build the `LinkedHashMap` directly (one small private `result(boolean ready, String reason, ...)` helper covers the four return points) and let the tests assert on map entries, the way `testMapCarriesNoAddresses` already does. Deletes the class and the `.toMap()` hops.

bitflicker64 (commented):
Blocking: no. Summary: the design is measured, justified and tested — the simplifications left are local: the probe hand-rolls a ThreadFactory that commons-lang3's `BasicThreadFactory` (the repo's own `ExecutorUtil` pattern) does in one expression, carries a `graphName` parameter nothing reads, and wraps its result in a 55-line `Result` holder whose only production consumer is `toMap()`. Evidence: reviewed the full diff at head d0f4a2548d8953a37cf4c1a44645d0457b423dc5 (10 files, +965/-0) with the head checked out; `grep -n graphName HstoreStorageProbe.java` shows the parameter declared at line 196 and never read; `grep -rn HstoreStorageProbe.Result` outside the file matches only the test; `grep -rn BasicThreadFactory` shows four uses in hugegraph-common's `ExecutorUtil` and commons-lang3 already imported in the hstore module. The endpoint itself, `KnownStores`, the background PD refresh, the first-answer-wins loop and the api/hstore split via the meta handler are all justified in the PR description (two superseded designs measured, seven E2E scenarios) and are not raised.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
SebastianGruza: Done in c9e70b14, thanks, both paths were real: with no Store list yet the `PDException` from `newBlockingStub()` carried the whole `pd.peers`, and with every Store down gRPC carried `Unable to resolve host …` (it was in my own Stores→0 samples, I had not connected the dots). `reason` now carries a fixed category only: the gRPC status code for `StatusRuntimeException`, `pd unreachable` for `PDException`, the class name otherwise (`HstoreStorageProbe.category()`); the full messages go to the log (WARN for PD, DEBUG for Store pings). `StorageReadiness.check()` likewise: class name only, message to the log. `testReasonCarriesNoPdPeersNorStoreHosts` feeds a `PDException` with `pd.peers=pd-0.internal:8686,…` and an `UNAVAILABLE: Unable to resolve host store-0…svc` and asserts neither map contains `internal`, `svc` or `8686`; `StorageReadinessTest` checks the same for a probe exception. On k3s with this build the reasons read `none of 3 known store(s) answered: a store failed: UNAVAILABLE; a store failed: UNAVAILABLE; a store failed: UNAVAILABLE`, and a grep of every sampled body from `stores-zero` and `pd-zero` for `.svc`, `8686` and `hugegraph-store-` finds 0 occurrences.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
SebastianGruza: Done in c9e70b14. The refresh is single-flight: `KnownStores` keeps an `AtomicReference` to the in-flight `CompletableFuture`, `refresh()` returns the running one while it is not `isDone()` and only starts a new one after it finished (CAS on the reference, so two racing probes also share one). A hung PD therefore parks one thread per process regardless of `readiness.cache_ttl`, including 0. `testRefreshIsSingleFlight`: a lister sleeping 3 s, five probes in a row → one call, all five ready from the known list; after it completes the next probe starts the second. The queueing on the `synchronized` `newBlockingStub()` next to graph traffic is thereby bounded to that one thread as well.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/ReadinessAPI.java
SebastianGruza: Done in c9e70b14. The Javadoc of `ReadinessAPI`, `StorageReadiness` and `HstoreStorageProbe` and the descriptions of `readiness.timeout` and `readiness.cache_ttl` now say what the code does: PD only matters until the first Store list is known, afterwards it is refreshed in the background; the cost of a probe is one cheap call to every known Store in parallel, first answer wins; 503 only when none answers within the budget (or, before the first list, when PD does not answer).

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
SebastianGruza: Done in c9e70b14. After every successful listing from PD, `pruneChannels()` shuts down (`shutdownNow`) and removes the channels of addresses no longer listed; a failed listing touches nothing. Wired into the lister lambda of `probe(graphName, timeoutMs)`, so the pure `probe(...)` stays stateless. `testChannelsOfReplacedStoresAreShutDown` uses a small `ManagedChannel` subclass (the hstore module has no Mockito): the channel of a listed address stays open, the replaced one is shut down and removed, a `null` listing removes nothing.

SebastianGruza:
Round 1 in c9e70b14: `reason` carries a category only (gRPC status code, `pd unreachable`, class name) with the full messages in the log; the PD refresh is single-flight; channels of replaced Stores are shut down after a successful listing; Javadoc and option descriptions match the probe. Tests 18 + 6. On k3s with this build `stores-zero` and `pd-zero` PASS, and the sampled bodies contain 0 occurrences of host names or the PD port; results in `results/issue-3212/round1-3212g` of the validation repo.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
SebastianGruza: Done in 213fd9c0, exactly as proposed: `new BasicThreadFactory.Builder().namingPattern("storage-readiness-%d").daemon(true).build()`, like `ExecutorUtil`. The anonymous factory, the `AtomicInteger` and two imports are gone; same thread names, same daemon flag.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
SebastianGruza: Done in 213fd9c0. `probe(long timeoutMs)` without the parameter, and the `namespace + "/" + store` concat in `HstoreStore` is gone with it. If a per-graph store list is ever wanted, `pd.getActiveStores(graphName)` comes back together with the parameter.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
SebastianGruza: Done in 213fd9c0. `Result` is gone: `probe(...)` returns the `Map<String, Object>` built by one private `result(ready, reason, activeStores, answeredStore, pdReachable, pdAgeMs, storesAgeMs, storeMillis)` helper at the four return points, and `StorageReadiness.check()` gets the map directly (no more `toMap()` hops). The tests assert on map entries the way `testMapCarriesNoAddresses` already did; 18/18.

SebastianGruza:
Round 2 in 213fd9c0: `BasicThreadFactory` instead of the hand-rolled factory, the `graphName` parameter and its concat removed, the body built directly as a map through one helper instead of the `Result` holder; tests on map entries, 18/18; net -48 lines. Smoke on k3s with this build: the three Server pods answer 200 with the same body as before the refactor, no readiness errors in the log.

bitflicker64 (commented):
Blocking: no. Summary: the earlier rounds are addressed at 213fd9c and the probe logic holds up. The main gap is that `/readiness` is not exempt from `LoadDetectFilter` the way `/versions` is, so a busy Server fails the probe; separately, an empty PD answer closes the Store channels the probe keeps pinging. Evidence: static read of the full diff at 213fd9c0ffe459df2f2bfa4b564dc2521cd601e9 against `LoadDetectFilter.WHITE_API_LIST`, `ServerOptions.MAX_WORKER_THREADS` (default `2 * CPUS`), `KnownStores.update` and `pruneChannels`; `gh pr checks` green on this head apart from the two codecov statuses.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AuthenticationFilter.java
```diff
@@ -77,6 +77,7 @@ public class AuthenticationFilter implements ContainerRequestFilter, ContainerRe
     private static final AntPathMatcher MATCHER = new AntPathMatcher();
     private static final Set<String> FIXED_WHITE_API_SET = ImmutableSet.of(
             "versions",
+            "readiness",
```
bitflicker64, line 80: ⚠️ `readiness` is added to the auth and path whitelists but not to `LoadDetectFilter.WHITE_API_LIST` (`""`, `apis`, `metrics`, `versions`). So `/readiness` goes through the worker-load and free-memory checks, before the result cache is reached, and gets a 503 once `max_worker_threads - 1` other requests are in flight. `restserver.max_worker_threads` defaults to `2 * CPUS`, so on a 2-CPU pod the probe fails while 3 other requests run.

Under sustained load Kubernetes then drops busy Servers from the Service and shifts their traffic to the rest, so a deployment that moves its readiness probe from `/versions` to `/readiness` can lose every endpoint during a spike while storage is fine.

Could `readiness` go into `WHITE_API_LIST` next to `versions` (`LoadReleaseFilter` reads the same list, so the counter stays balanced), with a case like `testFilter_WhiteListPathIgnored`? If shedding load through readiness is intended, please say so in the `ReadinessAPI` Javadoc.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,362 @@
...
+            List<Metapb.Store> stores = pd.getActiveStores();
+            pruneChannels(CHANNELS, stores);
+            return stores;
+        }, HstoreStorageProbe::pingScanState, timeoutMs, EXECUTOR);
+    }
+
+    /** Shut down the channels of addresses PD no longer lists (replaced Stores). */
+    static void pruneChannels(Map<String, ManagedChannel> channels,
+                              List<Metapb.Store> stores) {
+        if (stores == null) {
```
bitflicker64: 🧹 Follow-up to the channel pruning from round 1. An empty PD answer shuts down every channel here, while `KnownStores.update` ignores an empty answer and keeps the old list (`testEmptyPdAnswerIsNotReadyAndKeepsTheOldList`). As long as PD answers with no active Store, each refresh closes the channels to the Stores that the pings still use, and the next probe opens new ones. `probe()` starts the refresh and pings right away, so a ping in flight when `shutdownNow()` runs fails with UNAVAILABLE, and the old-list rule that keeps the Server ready can still turn a probe into a 503.

Could this return early on an empty list as well (`if (stores == null || stores.isEmpty())`), the same rule `update` applies, and could `testChannelsOfReplacedStoresAreShutDown` check that an empty listing keeps the channels?

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/filter/AuthenticationFilter.java
SebastianGruza: Done in 4a4aa1cc, thanks, that would have been a nasty interaction: a readiness probe shed under load would pull exactly the busiest Servers out of the Service while the storage is healthy. `readiness` is in `LoadDetectFilter.WHITE_API_LIST` next to `versions`; `LoadReleaseFilter` reads the same list, so the `workLoad` counter stays balanced. `testFilter_ReadinessIgnoredLikeVersions` in `LoadDetectFilterTest`: with a 2-thread limit and one request in flight the filter lets `/readiness` through without touching the counter and without a log entry, in the same shape as `testFilter_WhiteListPathIgnored`. Readiness is not meant to shed load; the `ReadinessAPI` Javadoc says it answers from the storage state.

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
SebastianGruza: Done in 4a4aa1cc. `pruneChannels()` returns on an empty list as well (`stores == null || stores.isEmpty()`), the same rule `KnownStores.update()` applies: as long as the pings use the last known list, their channels stay open. `testChannelsOfReplacedStoresAreShutDown` now also checks that an empty listing neither shuts down nor removes the channel of a known Store. 18/18.

SebastianGruza:
Round 3 in 4a4aa1cc: `/readiness` on the `LoadDetectFilter` whitelist (a busy Server still answers its probe from the storage state), an empty PD answer no longer closes the channels, plus `ReadinessApiTest` in the API suite (with and without credentials, on every backend of the suite) so the endpoint is measured by jacoco in the api-test job. Tests: `LoadDetectFilterTest` 11/11, `HstoreStorageProbeTest` 18/18, `StorageReadinessTest` 6/6, `ReadinessApiTest` 2/2 against a server built from this head on rocksdb.

bitflicker64 (approved):
Blocking: no. Summary: both findings from the last round are fixed at 4a4aa1c. `readiness` is in `LoadDetectFilter.WHITE_API_LIST`, and `LoadReleaseFilter` skips it through the same `isWhiteAPI`, so the work-load counter stays balanced. `pruneChannels` now returns on an empty PD answer, the same rule `KnownStores.update` applies. The fixes from the first two rounds are still in place, and I found nothing new in the rest of the diff. Evidence: read the full diff at 4a4aa1cc1177c68631f282f7b86e140718abb669 (14 files, +1204/-1) and the 213fd9c..4a4aa1c delta, plus the call sites in `LoadReleaseFilter`, `GraphManager.graphs()`, `HugeGraphAuthProxy.runAsAdmin` and `HstoreStore.registerMetaHandlers`. `gh pr checks` is green on this head, codecov included. The hstore job log shows `HstoreStorageProbeTest` 18/18, and its API suite grew from 155 to 157 tests (the two in `ReadinessApiTest`) with no failures.

imbajin (commented):
Blocking: yes. Summary: The readiness signal can stay green while graph-serving operations are unavailable, and concurrent probes can wait beyond their configured timeout when caching is disabled. Evidence: static trace of HstoreStorageProbe.getScanState() and StorageReadiness.check().

hugegraph-server/hugegraph-hstore/src/main/java/org/apache/hugegraph/backend/store/hstore/HstoreStorageProbe.java
```diff
@@ -0,0 +1,367 @@
...
+        });
+    }
+
+    private static void pingScanState(Metapb.Store store, long timeoutMs) {
+        ManagedChannel channel = CHANNELS.computeIfAbsent(store.getAddress(), address -> {
+            return ManagedChannelBuilder.forTarget(address).usePlaintext().build();
+        });
+        HgStoreStateGrpc.newBlockingStub(channel)
+                        .withDeadlineAfter(timeoutMs, TimeUnit.MILLISECONDS)
+                        .getScanState(SubStateReq.getDefaultInstance());
```
imbajin, line 228: ⚠️ A successful getScanState response is treated as proof that this Server can serve graph traffic, but this RPC only reads local scan-pool statistics and never checks Raft or graph-data access. A Store whose status RPC remains healthy while its Raft or partition path is unavailable will still make /readiness return 200, so Kubernetes can keep routing graph requests to an unusable instance. Please validate the graph-serving path, or use a Store health signal that covers it, before reporting ready.

hugegraph-server/hugegraph-api/src/main/java/org/apache/hugegraph/api/profile/StorageReadiness.java
```diff
@@ -0,0 +1,130 @@
...
+                holder.add(body);
+                return;
+            }
+            holder.add(check(t -> graph.metadata(null, STORAGE_READINESS_META, t),
+                             timeoutMs, cacheTtlMs));
+        });
+        return holder.get(0);
+    }
+
+    public static synchronized Map<String, Object> check(Probe probe, long timeoutMs,
```
imbajin, line 82: ⚠️ This synchronized method holds the class monitor while the Store probe performs network I/O for up to readiness.timeout. With the supported readiness.cache_ttl=0 setting, concurrent requests each trigger another probe and queue behind the previous call, so request latency can exceed the configured timeout by multiple probes and occupy REST workers. Please share an in-flight probe or bound each caller's wait without holding the monitor during network I/O.

